PR #35, opened 2020-05-03
Adding default .editorconfig

gnovaro:
Hello community, my little contribution. This allows defining a coding standard according to the independent project of the editor or IDE

https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,35 @@
+
+
+# EditorConfig is awesome: https://EditorConfig.org
+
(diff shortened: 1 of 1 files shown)


rathboma:
Thanks for doing this! I think it can be simplified to maybe look more like this one? https://github.com/Kong/insomnia/blob/develop/.editorconfig

Also we should make sure there is an explicit entry for `.vue` files

UtechtDustin:
Can be closed (#282 is merged).

Issue #52, opened 2020-05-07
Postgres: Display views and tables from multiple schemas


CptMonac:
It looks like by default, the tables & views displayed are only those in the public schema

rathboma:
Yes this is correct until I figure out a way to add the a schema dropdown to the UI.

Will work on this soon. I consider this a bug.

dwelch2344:
We tend to work across a ton of schemas, so I'd prefer not to have to switch schemas each time I want to see something in another context. Perhaps a collapsible tree component (eventually being able to specify "all schemas" or check only those you're interested in at the connection level)?

dwelch2344:
Took a quick stab at following the `TableListItem` pattern and wrapping it in a `SchemaContainer` type component. Right now it pulls all tables in all schemas, but being able to specify those schemas in the connection would probably be good to have with this kind of approach. 

Video linked below. Thoughts? 

https://drive.google.com/file/d/1cAYel_bj7h3LD1yw3NuhzOD3OMjenyvt/view?usp=sharing

ifokeev:
I'm trying to connect to not default schema in my `postgres` database and see `no tables` at all.

<img width="1312" alt="Screenshot 2020-05-07 at 10 51 50" src="https://user-images.githubusercontent.com/2017148/81268556-c24f9400-9050-11ea-8b77-ccf80b4c426d.png">

rathboma:
@dwelch2344 That actually looks pretty awesome. Want to create a PR and we can iterate on it?

rathboma:
Another option is to previx the schema on the table name, but I think that's a little ugly, eg `public.foo`, and `contacts.bar`

ghstng:
@rathboma @dwelch2344 I'm good with it -- PR and I can do some styling cleanup

dwelch2344:
@rathboma ew yeah I think that'd be a huge turn off to anyone who leverages schemas for namespacing / isolation. 

IntelliJ / DataGrip have really good tooling for RDBMS, and they handle it similarly with a tree component like this – was where I got the idea ;) I don't love that they mash everything (including multiple database connections) into it, but it does work well when you have multiple data sources for dev/prod and such. Screenshot below for chewing on 

<img width="370" alt="Screen Shot 2020-05-07 at 8 27 23 AM" src="https://user-images.githubusercontent.com/160142/81306591-e43f1b80-903c-11ea-942c-beae31cd7363.png">

[...]

PR #73, opened 2020-05-07
Schema awareness in tables

dwelch2344:
Looking to address #52 by showing multiple schemas with tables grouped under them. Super rough cut, but visually shows pretty well what I'm thinking. 

I haven't dug too deep on the existing logic, but it doesn't look like any tables outside of the `public` schema (default for Postgres, which I'm using in my example) have their column information. Guessing it has something to do with the concept of the "current table" I saw referenced in a few places. 

Would love some insight / feedback :)


src/store/index.js
```diff
@@ -30,6 +30,9 @@ const store = new Vuex.Store({
     pinned(state) {
       const result = state.pinStore[state.database]
       return _.isNil(result) ? [] : result
+    },
+    schemaTables(state){
+      return _.chain(state.tables).groupBy('schema').value()
```
dwelch2344, line 35: Not sure what the right name here would be. Don't love `schemaTables`, but hey - naming things is hard :D

src/store/index.js
```diff
@@ -125,7 +128,7 @@ const store = new Vuex.Store({
       // Ideally here we would run all queries in parallel
       // however running through an SSH tunnel doesn't work
       // it only supports one query at a time.
-      const onlyTables = await context.state.connection.listTables()
+      const onlyTables = await context.state.connection.listTables({schema: null})
```
dwelch2344, line 131: This is the big one. Where do we store and expose which schemata we want to view? IMO shouldn't be done before making a connection, so maybe as a top level component on the Table list? 

So first time you connect you see: 
```
> Schemas (3/3) 
  > public 
    > bar_table
    > foo_view
  > x_schema
    > x_table
  > y_schema
    > y_table
```

But then an action on `Schemas` would allow you to filter what is pulled

rathboma: so several things (eg the auto-complete prompt) require having all table and column data on connect. So that is a design limitation. I think pulling all schemas & tables at once is fine.

I'll test your PR later to see what undocumented assumptions it breaks :-)

rathboma: I dislike the datagrip and dbeaver approach to having a 'schema' folder, of course they're schemas, what else are they?

I think it probably needs to show the schemas collapsed initally, even if it pulls all the table information up-front

dwelch2344: Yeah definitely agree on collapsed. Hadn't thought about the connection problem – been a while since I strayed from postgres ;) 

Yeah, the schema folder is super annoying, though IntelliJ only does it if you enable additional object types / have a wide context. They also allow you to specify a regex-like filter like the following (notice the tree / UI just updates the "Schema Pattern" field)

Screen Shot 2020-05-07 at 9 46 16 AM

Maybe we offer just a box like that and leave it to the (likely technical) users to hone as needed? Could default to `*`, or even a DBMS-specific default like `public.*` or `[^postgres].*` for postgres?

rathboma:
@dwelch2344 I added some changes to deal with db's without schemas, and added a loading screen (if there are a lot of tables).

@ghstng some UI things that I made look terrible:

- There's now a 'loading table x/y' state in the table list view
- The schema icon - can you review and see if it is ok with you?

rathboma:
I think functionality wise it is good to go now

PR #75, opened 2020-05-07
Use os.homedir to find user directory

MasterOdin:
[`os.homedir()`](https://nodejs.org/api/os.html#os_os_homedir) has been in node since 2.3 and is probably better to use than a hand-rolled function.

The builtin function has two more methods to find a user's home directory if the environment variable does not exist.


rathboma:
Thanks! Although I don't think this code is really used anymore (legacy from prior project)

PR #82, opened 2020-05-07
Postgres - double quote identifiers

### LINKED ISSUE #74: Postgres connection should respect identifier "case"
Hey, LOVE this project.  I have been missing postico since I moved to linux, and this looks very promising.

I noticed that postgres connections have trouble when identifiers are not set up all lower cased.  

For example, we have a column named "companyName" (our sin, not yours), and the default is for postgres to cast that to "companyname"

Would you be open to a solution for postgres users such as wrapping table names and column names in double quotes before executing the queries?  Maybe there is a more correct solution than that.

There are a couple of areas where it might be useful.

1) Query Editor, the CodeMirror autocomplete window recognizes the table names as camelCase, but when I select, doesn't wrap in double quotes.
2) Table View, the column names in the filter dropdown appear in camelCase, but I can't filter on companyName by hitting the "search" button.
3) Tables & Views, hitting the "Open in a new tab" button on a table with a camelCased name will error.

hillmanov:
Fixes #74 

Makes sure that postgres table and column names are surrounded in
double-quotes. This is needed when the table or column name contains
upper case letters.

Also, it adds some additional logic to the autocomplete functionality to
automatically surround postgres table and column names with
double-quotes after the complete event has fired.

In order for the autocomplete to show properly for tables that have
double quotes, it was necessary to add both an unquoted and quoted
variant of the table name to this hintOption object. This does clutter
the dialog a bit - it will show both:

tableNameWithQuotes
"tableNameWithQuotes"

without this, however, you will only get autocomplete when typing the
table name - you will not get column name autocompletion hints off that table
name without the quoted variant present as well.

### DIFF
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -123,12 +123,18 @@
         result[ctrlOrCmd('l')] = this.selectEditor
         return result
       },
+      connectionType() {
+        return this.connection.connectionType;
+      },
       hintOptions() {
         const result = {}
         this.tables.forEach(table => {
           const cleanColumns = table.columns.map(col => {
             return col.columnName
           })
+          if (this.connectionType === 'postgresql' && /[A-Z]/.test(table.name)) {
+            result[`"${table.name}"`] = cleanColumns
+          } 
           result[table.name] = cleanColumns
         })
         return { tables: result }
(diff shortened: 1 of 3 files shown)


rathboma:
Thanks! Let me play around with it. Not sure about the UX with the autocomplete and wonder if we can just replace the table name and not autocomplete when folks start quoting themselves.

Oh, I can probably just set it up to autocomplete after the quote char maybe. Let me play with it.

rathboma:
I checked this out, seems awesome. Thanks for the great work!

Issue #32, opened 2020-04-30
Bugged text when adding an Private key


MrGKanev:
Hi,
When I tried to add a private key file every time i see the text like boxes. (https://imgur.com/8jK1lSC)
**OS**:
DISTRIB_ID=ManjaroLinux
DISTRIB_RELEASE=20.0
DISTRIB_CODENAME=Lysia
DISTRIB_DESCRIPTION="Manjaro Linux"
**Using**:
snaps
**version of the snap**: 
1.1.3

rathboma:
Wow, that's weird. What you should see is your native OS file-picker.

Looks like a problem with the snap not being able to access native fonts? I'll spin up a Manjaro VM and test it there. This works fine on Ubuntu VMs, so probably just a missing package dependency I need to specify

MrGKanev:
@rathboma if you need help with testing i can help . :)

rathboma:
Thanks! I'll post to the snapcraft forums and try to replicate the issue. I'll ping this thread when I have a fix!

rathboma:
@MrGKanev can you start beekeeper from the terminal and see if there are any errors in the shell? 

`$> beekeeper-studio`

I'd be looking for a missing font error or something like that.

MrGKanev:
Hi @rathboma, personally I didn't see anything connected with it. Here is the output I saw so you can take another look at it. (https://pastebin.com/FEtad1wG)

MrGKanev:
Small update: Updated to version 1.4.0 The problem still occurs. (gave the snap SSH access)

rathboma:
Potential workaround - 

```
$ sudo rm -f /var/cache/fontconfig/*
$ rm -f ~/.cache/fontconfig
```

https://forum.snapcraft.io/t/snap-store-fonts-on-arch-linux-are-merely-empty-rectangles/15373/9

harveydobson:
I have this same issue, 

> @MrGKanev can you start beekeeper from the terminal and see if there are any errors in the shell?
> 
> `$> beekeeper-studio`
> 
> I'd be looking for a missing font error or something like that.


Here's the output

```
15:57:39.815                     › initializing background

(beekeeper-studio-bin:50167): Gtk-WARNING **: 15:57:39.828: Theme parsing error: gtk.css:1413:23: 'font-feature-settings' is not a valid property name

[...]

PR #92, opened 2020-05-08
Use native filepicker dialog instead of input file

danilopolani:
With this PR when choosing a private key for SSH tunneling a native dialog will be used instead of the `input file` one.

**Added**:

- Native dialog filepicker for SSH private key
- Input group CSS
- When choosing a private key, it will open by default `.ssh` folder (if it does not exists, it will fallback on the home folder, at least on unix)

**Fixes**:

- Bug on my PC (Ubuntu 17.04) returning error "Access denied" when navigating to `.ssh` folder
- Probably #32 too

Read more: https://www.electronjs.org/docs/api/dialog

### DIFF
--- a/src/assets/styles/app/_layout.scss
+++ b/src/assets/styles/app/_layout.scss
@@ -37,6 +37,10 @@ i {
     -webkit-user-select: none;
             user-select: none;
 }
+.clickable,
+.cursor-pointer {
+  cursor: pointer;
+}
 
 // Material-icons
 .material-icons {
@@ -149,6 +153,11 @@ i {
   width: 100%;
   max-width: 250px;
 }
+.btn-small {
+  font-size: 10px;
+  padding: 0 .7rem;
+  min-width: 0;
+}
 .btn-block {
   display: flex;
   justify-content: center;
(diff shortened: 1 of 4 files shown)


rathboma:
@danilopolani are you using the `snap` by any chance?

rathboma:
I still like the idea of this change separate from the permissions issue. I'll do some testing and figure out how to encapsulate this in a component.

danilopolani:
Yeah, I'm using the snap version. I thought about creating a component from this, I can work on it tomorrow if you want (I guess it should applied to Sqlite DB selection too)

rathboma:
So your change won't fix anything for the snap version -- snaps are only allowed (by default) access to NON hidden folders in the home directory.

I'm curious if your change will fix the garbled fonts in Manjaro though.

I'll need to figure out the limitations of the Mac client separately. Like Snaps I also had to define some permissions up front, so I don't think simply switching to the electron-level picker will do it.

Please do componentize the picker, I think using the electron-picker is much better as the web file picker has some broad limitations that the electron one does not have.

danilopolani:
Ehy @rathboma , I updated the code creating an isolated component `FilePicker` to be re-used.

A few notes:

1. I added as props the common settings we need right now (`showHiddenFiles` and `defaultPath`), but we can extend/override other values as well with the `config` object prop.
2. Currently it does not support multiple files selection, because it would need some change in the UI too, but I think we will never need multiple files in the project.

I replaced the SQLite input file too with this, let me know if it's fine or if you need some more changes!

rathboma:
Awesome! Thanks so much. I'll review in the next few days

Issue #57, opened 2020-05-07
JSONB columns display as [object Object]


k3rni:
![image](https://user-images.githubusercontent.com/200996/81264702-24ed6380-9042-11ea-89f3-b7792b8b77b9.png)

here, the `data` column is JSONB with a toplevel array of objects.

Same for query output.

```
$ snap list beekeeper-studio
Name              Version  Rev  Tracking       Publisher         Notes
beekeeper-studio  1.2.0    18   latest/stable  matthew-rathbone  -

rathboma:
lol, what's the problem that looks soo good!

rathboma:
This is fixed, see release 1.3 💪

k3rni:
installed 1.3 with `snap refresh --edge`, fixed!

PR #93, opened 2020-05-08
JSON and JSONB for PSQL

rathboma:
Being able to display complex data types (like json and jsonb). This is **basic** functionality that just turns everything into a sensible string representation.

Before this is done:
- [x] Mutator mapping for table tabs
- [x] Mutator mapping for query tabs

Long term we need to standardize the DB interfaces a little, so that things like return types are all the same


rathboma:
Fixes #57

PR #98, opened 2020-05-08
Order columns using ordinal_position

MasterOdin:
The ordering of columns is not correct if the rows in `information_schema.COLUMNS` are not in proper order (which is in no way guaranteed).

The note in [mysql docs](https://dev.mysql.com/doc/refman/5.6/en/columns-table.html) for pulling information about of `information_schema.COLUMNS`:

> The position of the column within the table. ORDINAL_POSITION is necessary because you might want to say ORDER BY ORDINAL_POSITION. Unlike SHOW COLUMNS, SELECT from the COLUMNS table does not have automatic ordering.

Not having the order by gives me the following result:
Screenshot 2020-05-08 18 44 37

where I would expect `id` to be shown before `column1` and it's not.

### DIFF
--- a/src/lib/db/clients/mysql.js
+++ b/src/lib/db/clients/mysql.js
@@ -106,6 +106,7 @@ export async function listTableColumns(conn, database, table) {
     FROM information_schema.columns
     WHERE table_schema = database()
     AND table_name = ?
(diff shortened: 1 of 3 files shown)


rathboma:
Superstar

PR #99, opened 2020-05-09
Prompt for parameter placeholder values

hillmanov:
[Feature Demo](https://giphy.com/gifs/QvdLEGkLz5poRrTznY)

gif

If you type in a query like this:

```
SELECT * FROM project WHERE id = :id;
```

You will now get a prompt to provide a value for `:id` before the query
is executed.

This works for queries that use this syntax as well:

```
SELECT * FROM project WHERE id = $1 AND state = $2;
```

The following syntax is NOT supported:

```
SELECT * FROM project WHERE id = ? AND state = ?;
```

This would be possible, but I didnt' want to deal with array indexes at
the time. The current implentation is just an easy object lookup, which
makes all the loops and prompts clean and easy.

I have never used vue before, so if this implementation needs some work,
I am happy to do things differently. Thanks for the great project!

### DIFF
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -61,7 +61,7 @@
 
     <!-- Parameter modal -->
     <modal class="vue-dialog beekeeper-modal" name="parameters-modal" @closed="selectEditor" height="auto" :scrollable="true">
-      <form @submit.prevent="deparameterizeQuery">
+      <form @submit.prevent="deparameterizeQueryResolve">
         <div class="dialog-content">
           <div class="dialog-c-title">Provide parameter values</div>
           <div class="modal-form">
@@ -117,7 +117,6 @@
         saveError: null,
         lastWord: null,
 
-        queryParameterPlaceholders: null,
         queryParameterValues: {},
         deparameterizeQueryResolve: null,
       }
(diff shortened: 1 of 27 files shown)


rathboma:
Whhhaaatt, this is amazing. I'll review over the next couple of days

rathboma:
I think this is an amazing feature. @ghstng and I will need to chat about where the UI for the parameters goes.

If @ghstng agrees I think the modal works for now. Long term if you're iterating on a query you probably want a quicker turnaround cycle than the modal popping up. I'm thinking maybe a right-side sidebar, just in the query editor area.

I'm going to add some feedback to the PR around to structure the Vue code, but functionally I think it's great!

hillmanov:
@rathboma Sounds good. Thanks! 

I got the idea for the modal from DBeaver - that is how they prompt for them, but it definitely doesn't have to be done that way. Happy to help if you want it!

src/components/TabQueryEditor.vue
```diff
@@ -197,11 +225,29 @@
...
+          query = query.replace(new RegExp(`(\\W|^)${this.escapeRegExp(param)}(\\W|$)`), `$1${this.queryParameterValues[param]}$2`)
+        });
+        return this.deparameterizeQueryResolve(query);
+      },
       async submitQuery() {
         this.running = true
         try {
+          let query = this.editor.getValue();
+
+          this.queryParameterPlaceholders = this.getQueryParameterPlaceholders()
```
rathboma, line 243: My feedback is mostly around doing things the 'vue way'.  Probably we need a hook into the codemirror API to inflate the `query` text into a data property when it changes.

If you feel like you want to take this on, then please feel free, otherwise I'll do it later this week. Broadly what I think needs to happen is:

1. Add a codemirror callback when the editor area changes and update a `query` property from `data`
2. Add computed properties for `queryParameterPlaceHolders`, and `deparameterizedQuery`
3. Move `escapeRegExp` to `data`

rathboma: Then you can make the `action` on `run` dynamic, so if queryParams > 0 'showmodal' else 'submit'

hillmanov:
@rathboma Sorry for the delay! Thanks for the feedback - I'll make these changes in the next couple of days.

hillmanov:
@rathboma alright, I think I have it cleaned up a bit. 

Things got a little weird with my local repo, and I did a couple of force pushes to my fork. Sorry if that makes the PR odd or hard to review on your end. 

I did have one question, however - I am hoping you can help me understand one of your requests better. It seemed unintuitive to me to put `escapeRegExp` in `data` - why not keep it in `methods` since it isn't related component state, nor does it modify it? Again, I am not very familiar with vue and where to put certain things, but it seems like `methods` was a more natural place for it. I am happy to put it in `data` if that is the better place though, or even in a different utility file if even that would be preferable. 

Thanks!

rathboma:
@hillmanov Actually honestly it should probably just be a constant above the export if we're not using it in the template (we're not).

`const foo = bar`

I'll do another review of your code this weekend. Thanks so much for contributing! I'll make sure to prioritize this.

rathboma:
I messed up your code with another PR, so I fixed it :-)

Issue #89, opened 2020-05-08
SSL - allow users to provide a certificate or CA for SSL connections


rathboma:
Ideal behavior:

Provide an automatically drop-down for common CAs:
- Amazon RDS
- Azure
- GCP
- Digital Ocean

Allow users to provide CA / Cert+key for other connections

rathboma:
If someone wants to test https://github.com/beekeeper-studio/beekeeper-studio/pull/109 that'd be great. I don't have a db set up with a real certificate yet.

Gys:
Sorry, did not work, I get 'Connection lost: The server closed the connection.' after about 10 seconds.

I tried this (MacOS) for https://github.com/beekeeper-studio/beekeeper-studio/pull/109/checks?check_run_id=667232724

Exactly the same settings work in TablePlus.

rathboma:
Thanks for testing! I'll do some work on this and spin up my own digital ocean instance

danilopolani:
DigitalOcean if I remember only needs the SSL flag to be enabled, without providing any cert file, maybe we should investigate on that direction

mholt:
Most CockroachDB installations also recommend / require TLS certificate authentication.

Once this feature lands I will probably be able to switch to using Beekeeper, looks good!

trevorrjohn:
This can be closed as well @rathboma

PR #109, opened 2020-05-12
Add SSL CA, Cert and Key files

danilopolani:
Some notes:

1. I did not have tested this. On the paper it should work smoothly according to the docs, but I don't have any database with SSL enabled to use.
2. I copied the design of the `SSH` panel, I mean it was cool.
3. I don't know if it's better to have field like `full row, col col` or `full row, full row, full row` (see screenshots). I honestly prefer the `full row, col col` layout but up to you :)

Full row, col col
Full row, full row, full row


rathboma:
wow, this is awesome! I'll try to get a database set up to test this with :-D

rathboma:
Someone tested this and I don't think it works quite right, so I'll need to do some more in-depth testing, but to do that I'll need a 'real' database set up with real SSL certs.

iketiunn:
Hi, since the binaries from Github Action is expired, I tested it using @danilopolani's source with my production GCP CloudSQL. And it works great.

OS: macOS 10.15.6
Source: https://github.com/danilopolani/beekeeper-studio/tree/feature/add-ssl-ca-file
Build: using  `yarn electron:serve`, remember to delete the outdated `yarn.lock`
Database: GCP CloudSQL: `PostgreSQL 12.3 on x86_64-pc-linux-gnu, compiled by Debian clang version 10.0.1 , 64-bit`

Screen Shot 2020-09-13 at 12 45 24 PM
Screen Shot 2020-09-13 at 12 46 25 PM

rathboma:
@iketiunn does the production build work if you check the 'use ssl' box without providing a certificate?

iketiunn:
> @iketiunn does the production build work if you check the 'use ssl' box without providing a certificate?

I test on the free postgres hosting: https://www.elephantsql.com/docs/faq.html#Does_ElephantSQL_support_TLS_SSL and it is fine.

Wondering what's the error message(with `pg` connector) showed up in https://github.com/beekeeper-studio/beekeeper-studio/issues/89#issuecomment-629380749 .

mholt:
I would like to renew interest in this feature. How can I help?

trevorrjohn:
> @iketiunn does the production build work if you check the 'use ssl' box without providing a certificate?

I have fixed this issue in https://github.com/beekeeper-studio/beekeeper-studio/pull/376

trevorrjohn:
This can be closed now @rathboma

PR #123, opened 2020-05-15
Add ability to select and copy a cell value

danilopolani:
Unfortunately Tabulator does not support cell-selection, but only row selection. With this PR you can select a single cell and copy its value with Ctrl+C (or Cmd+C) without entering the "edit mode", like the default behaviour of Tabulator editable cells.  

Until it support row editing I think this could work.  

**Before**:

before

**After**:

after


rathboma:
This is awesome! I'll get this reviewed and merged :-)

Issue #63, opened 2020-05-07
[Enhancement] Pick color for each connection to quickly distinguish them


danilopolani:
It would be nice to be able to pick a color (from a fixed selection is fine too, like Sequel Pro) that will be shown as background color in the saved connection to quickly distinguish them.

rathboma:
💯 this would be useful too, agreed.

mlebrun:
Almost have a pull request for this

PR #131, opened 2020-05-18
Connections: add optional color labels for sidebar list

mlebrun:
Add an initial implementation for color labels (https://github.com/beekeeper-studio/beekeeper-studio/issues/63) for Saved Connections. Definitely willing to accept any feedback or comments! Just wanted to take a stab at it 😊 

- adds a default Radio form button for potential future use styled similarly to the checkbox (but without a check)
- re-uses provided colors in theme files

The only weirdness is it looks like it's saving your selected label color but it's really more of a "preview" and will only actually save when you click the connections Save button. It should probably not show the label anymore when you click off of that connection to show that you're selection didn't save.

I also didn't see a contributing guide or any tests folder so like I said, I'm definitely willing to accept any comments or changes because I know I could be wildly off track or something!! Here's a preview:

ezgif-4-04ee3e593e33


src/assets/styles/app/_layout.scss
```diff
@@ -360,6 +360,33 @@ input[type="checkbox"] {
...
+
+  position: relative;
+  display: inline-flex;
+	-webkit-appearance: none;
+	background-color: transparent;
+  box-shadow: inset 0 0 0 2px rgba($theme-base, 0.2);
+  height: $radio-size;
+  line-height: 1;
+  width: $radio-size;
+	padding: 0;
```
mlebrun, line 376: there's some weird whitespace here, will fix that

src/assets/styles/app/_layout.scss
```diff
@@ -360,6 +360,33 @@ input[type="checkbox"] {
   }
 }
 
+// Radio
+// ---------------------------------
+input[type="radio"] {
+  $radio-size:     16px;
+
+  position: relative;
```
mlebrun, line 368: very similar to the checkbox code but wanted the radio buttons to be pretty "on-brand". There's probably a better way to do this 🤷‍♂️

rathboma:
Wow, amazing PR! Thanks for putting this together. @ghstng and I will review and get back to you with any changes. Very excited for this.

src/components/ConnectionInterface.vue
```diff
@@ -57,6 +57,19 @@
                     <span>Save Passwords</span>
                     <i class="material-icons" v-tooltip="'Passwords are encrypted when saved'">help_outlined</i>
                   </label>
+
+                  <!-- Label Color Selection -->
+                  <div class="row">
```
rathboma, line 62: I've been thinking about the `ConnectionLabel` component.

I think for the form it would be easier to maintain if this were all just part of a `ColorPicker` component that renders all the color radios and has its own `v-model`. I think having to individually specify the colors here, all bound to the same variable makes it easy to forget one down the line if this needs to go in another place.

Then in the connection sidebar, I don't think the color needs a component, but rather a simple `span` (or whatever element) with the connection color, I think in the end it is a little weird using an `input` in the sidebar.

src/migration/20200517.js
```diff
@@ -0,0 +1,7 @@
+
+export default {
```
rathboma, line 2: Amazing! Thanks for figuring this out!

src/entity/saved_connection.js
```diff
@@ -77,6 +77,13 @@ export class SavedConnection extends DbConnectionBase {
   name
 
 
+  @Column({
```
rathboma, line 80: Also amazing, thanks for figuring this out!

src/components/ConnectionInterface.vue
mlebrun: Definitely makes sense. I thought it was a little awkward as well using a radio in the sidebar, this sounds much better!

rathboma:
@mlebrun just checking in to see if you have time to come back to this? Would love to merge it!

mlebrun:
Hm, I'll try and fix it up in time but had an offer accepted on a house so I'm right in the thick of it 😅 - I'll try my best but I won't be surprised if it's more likely in ready by next release. But like I said, I'll see what I can do 👍

mlebrun:
Not sure when you're drafting/publishing the release, but I've got most the changes in a branch, just need to fix one style thing! @rathboma

rathboma:
Awesome! Probably no release until the weekend!

mlebrun:
Updated!

rathboma:
@mlebrun ++

PR #144, opened 2020-05-24
fix sqlserver client breaking on CS collation

### LINKED ISSUE #139: SQL Server 2005 - no tables found
I'm trying to open a SQL Server 2005 database.
Connection and database listing is fine, but after a few seconds "finding tables" no tables are shown (_There are no tables in *TABLENAME*_).
Other db management apps correctly list all the 100+ tables.

Not sure if Beekeper is meant to be compatible with a 2005 db.

MasterOdin:
Ported from sqlectron/sqlectron-core#94

Fixes #139 

This PR then makes it so that the capitalization of these tables works for CS and CI collations. When running SQLServer under a CS collation (e.g. `Latin1_General_CS_AS`), it enforces a capitalized version of the INFORMATION_SCHEMA tables (e.g. `INFORMATION_SCHEMA.VIEWS`). When using a CI collation (e..g `Latin1_General_CI_AS`), capitalization does not matter. The other thing to note here is that under a CS collation, `INFORMATION_SCHEMA.COLUMNS.IS_NULLABLE` will return `NO` versus `No` under CI.

Testing this can be done by using docker and using the command:

```bash
docker run --rm -d --name sqlserver -p 1433:1433 -e "ACCEPT_EULA=Y" -e "SA_PASSWORD=Password12" -e "MSSQL_COLLATION=Latin1_General_CS_AS" mcr.microsoft.com/mssql/server:2017-latest
```

### DIFF
--- a/src/lib/db/clients/sqlserver.js
+++ b/src/lib/db/clients/sqlserver.js
@@ -187,7 +187,7 @@ export async function listTables(conn, filter) {
     SELECT
       table_schema,
       table_name
-    FROM information_schema.tables
+    FROM INFORMATION_SCHEMA.TABLES
     WHERE table_type NOT LIKE '%VIEW%'
     ${schemaFilter ? `AND ${schemaFilter}` : ''}
     ORDER BY table_schema, table_name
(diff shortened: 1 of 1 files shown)


rathboma:
Amazing! Thank you!

Issue #152, opened 2020-05-27
Feature Request: Run only selected statement(s)


manngo:
Is it possible to have either an option or an alternative keyboard short cut to run only the current statement (under the cursor) or the hightlighted statements?

rathboma:
Currently it is not a feature, but yes agreed that would be very useful!

rathboma:
This is fixed in master, thanks to @vuongggggg

PR #161, opened 2020-05-29
Add run buttons for selection text and current cursor query

vuongggggg:
Hi,

I'm interesting with this tool and would like to add the new features. 
- Run Selection (Ctrl-Alt-Enter): Run the selection query. If no selection, then the button is disabled as default and can do nothing when using shortcut keys.
- Run Current (Shift-Ctrl-Enter): Run on the current query under cursor.

image

I just done and test manually on Windows and didn't test shortcut keys on macOS yet. Please help if you can. Also, please take a look on this PR and give me your opinions if can. Thanks so much! 😃 

References: 
https://github.com/beekeeper-studio/beekeeper-studio/issues/152

### DIFF
--- a/src/assets/styles/app/query-editor.scss
+++ b/src/assets/styles/app/query-editor.scss
@@ -48,7 +48,11 @@
         padding: 0 $gutter-w;
         font-size: 0.85rem;
         min-width: 48px;
-      }
+
+        &.run-btn {
+          min-width: 100px;
+        }
+      } 
     }
   }
   .bottom-panel {
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -6,9 +6,9 @@
       <div class="toolbar text-right">
         <div class="actions btn-group" ref="actions">
           <a @click.prevent="triggerSave" class="btn btn-flat">Save</a>
+
           <a href="" v-tooltip="'(shift + ctrl + enter)'" @click.prevent="submitCurrentQuery" class="btn btn-primary btn-disabled">Run Current</a>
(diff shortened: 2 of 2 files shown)


rathboma:
Wow, thanks so much for submitting a PR! I'll do a review over the weekend

rathboma:
I took a quick look, and looks fantastic. Like how straightforward your implementation is!

However, one thing we want to account for is the possibility of there being a semi-colon inside a string, like 'some; value', that shouldn't split the query.

I think there's probably a regex to do this (dangerous last words I know).
Googling I've found https://gostash.it/en/stashes/1621-split-sql-text-into-multiple-queries, but would need to test it.

vuongggggg:
Hi, thanks for the reply. I got you at the concern on running selected query as well.

Just did a search and have tried a bit with regex. How do you think about these regex below?
1. Define the starting tokens (from stackoverflow)
https://regex101.com/r/dt9XTK/2 
2. Ignore meaning of tokens.
https://regex101.com/r/dh8Vuj/1

I prefer the second. Sorry if I bother you in weekend. 😨

rathboma:
Agreed the second one is way better. Wow great research!

FOSS is what weekends were made for :-)

vuongggggg:
Hi @rathboma, could you please help me to take a look on new change?

rathboma:
Yep! I'll review this weekend! Thanks!

rathboma:
@vuongggggg I added a couple of tweaks. I think this looks good.

I'm going to take a quick look again tomorrow, but I'm pretty happy with it and will probably merge it then -- I might move some of this logic into another file (like sql_utils.js or something like that).

I think a nice addition would be highlighting the 'current' query using the codemirror 'markText' api, like adding a class with bg color, but I don't think that is needed for this PR.

vuongggggg:
Thanks for the tweaks!

Issue #113, opened 2020-05-13
UI Suggestion: Close tabs via middle mouse click


Dragory:
This is the behaviour in e.g. all browsers and many editors such as VS Code (at least on Windows). I can look into submitting a PR for it as well.

rathboma:
yeeees. Good idea.

rathboma:
should be as easy as adding a `@click.middle` to `CoreTabHeader.vue` if someone wants an easy contribution

PR #164, opened 2020-05-30
Close tab via the middle mouse button.

strotgen:
Adds an event handler in `CoreTabHeader.vue` to close a tab when the user clicks it with the middle mouse button. 

Solves  #113


rathboma:
Yes! You are a legend.

PR #165, opened 2020-05-30
fix for postgresql driver for redshift

### LINKED ISSUE #150: Redshift Connection throwing Error
When trying to connect to Redshift database, I get an error:
**"column el.typarray does not exist"** and the connection does not succeed.

Same error happens when trying to connecting using Potgres connector. I suspect it's gotta do with some version issue of Postgres. 

https://github.com/dbcli/pgcli/issues/316

MasterOdin:
Amazon Redshift is based on postgresql 8.0, and pg_catalog.pg_type.typarray was added in 8.3, meaning that the query would cause an exception to be thrown due to el.typarray not existing when used for Redshift. This change gives us a fallback in the case of an exception on the first query to use one that will work under pg 8.0. Like the original query, this is based on the sql that is executed when using \dT command within psql, with the addition of the clause on n.snpname which was added in later versions of psql for \dT.

Fixes #150

### DIFF
--- a/src/lib/db/clients/postgresql.js
+++ b/src/lib/db/clients/postgresql.js
@@ -19,7 +19,7 @@ let dataTypes = {}
 
 /**
  * Do not convert DATE types to JS date.
- * It gnores of applying a wrong timezone to the date.
+ * It ignores of applying a wrong timezone to the date.
  * TODO: do not convert as well these same types with array (types 1115, 1182, 1185)
  */
 pg.types.setTypeParser(1082, 'text', (val) => val); // date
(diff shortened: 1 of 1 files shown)


rathboma:
Nice! Do you think it would be worth doing a version check rather than a try/catch? See for example `listMaterializedViews`?

this does a `isPostgres` check, and a `server_version_num` check

MasterOdin:
Yes, probably. Looking at that function, it'll also cause an error to be thrown under Redshift since it does not support `server_version_num` or `server_version` commands. So, you could fix things for redshift by doing something like:

```js
const { version } = (await driverExecuteQuery(conn, {query: "select version()"})).rows[0];
const serverVersionNum = parseInt(version.split(" ")[1].split(".").map(s => s.padStart(2, "0")).join("").padEnd(6, "0"), 10);
```

and then use that to condition on specific versions. This is also on the idea that I _believe_ that Redshift returns something like "PostgreSQL 8.0.2 on..." like regular postgresql instances, but as I do not have access to Redshift, I'm not totally sure.

rathboma:
Yeah I like that. This solves for the Redshift case, and also for the <= 8.3 psql case. This can just go at the top of the file I think.

MasterOdin:
@rathboma done

rathboma:
legend

PR #171, opened 2020-05-31
Query highlighting

rathboma:
- Refactored new query breakout code to make it more 'vue friendly'
- Added highlighting of queries
- Added a unit test for query identification

cc @vuongggggg

@ghstng I just made it blue for the timebeing. Leaving the exact stying up to you -- but we definitely want there to be a style difference between text selection and this current query highlight
query-selection


vuongggggg:
I just read the code and have a manual test here. :D 
Just concern about detect the sentence for highlighting by position instead of a give string. 🧀

rathboma:
So when did it fail? When you deleted a character?


Hmm, did you do it differently in your version? How would you change it? I'm using the search cursor from codemirror, but maybe not based on the right event

vuongggggg:
In the previous image, sorry because I didn't explain enough. Yes, I just assumed getSearchCursor(words) couldn't give us the right position of current query in editor in case we have many similar queries.

My version is check by regex (!!) and store the cursor index into data array, then always based on index (I calculated the fromPos and toPos of string) to get the range for highlighting.

One thing I also concern on my version is the performance for checking in the huge content (not sure if this case in practice).

From `master`
```javascript
const cursor = this.editor.getSearchCursor(this.currentlySelectedQuery)
        if (cursor.findNext()) {
          return {
            from: cursor.from(),
            to: cursor.to()
          }
        }
```

Here is mine
```javascript
//...
 let fromPos, toPos = 0, currentQuery
    for (let i = 0; i < queries.length; i++) {
      toPos += i == 0 ? queries[i].length : queries[i].length + 1
      if (toPos > cursorIndex) {
        fromPos = toPos - (queries[i].length - queries[i].search(/\S/))
        currentQuery = queries[i]
        break
      }
    } 

return {
      fromPos,
      toPos,
      query: currentQuery,
}
```

Here is the gifs for showing the behavior on two branches.
Branch `master`
demo-highlight-from-master-branch

Branch `feature/highlight-current-query`
demo-highlight-from-personal-branch

Because my native lang is not English, so sorry about any confused, just let me know pls. 😄

rathboma:
Yes you are correct! I did miss that :-). You did it better in that way!

Although I do think there are benefits to using the searchcursor from codemirror itself, but yes it does need to account for substrings. I'll push a fix later tonight.

Thanks again for your work on this. I'm super excited to ship this feature!

rathboma:
@vuongggggg see #181

PR #181, opened 2020-06-03
Highlight fixes and docs

rathboma:
Fixes for query highlighting and docs to go with it


vuongggggg:
I tried, it looks very awesome to me. So interesting to read code and see how it can be implemented in another way. I just have a small question on #171, please check. 😃

vuongggggg:
Sorry, my question here https://github.com/beekeeper-studio/beekeeper-studio/pull/171/files#diff-395bad58b51eb11cfe052a3cf69abca5R222

rathboma:
Ah, I don't see a question -- do you have a pending review you need to submit? (See the top right of the screen)

Screen Shot 2020-06-03 at 12 54 31 PM

rathboma:
Is your question around why I'm doing this?

I was thinking it wasn't worth highlighting the text if there was only a single query.

vuongggggg:
Ah yes, you got my point at the Pending question and also have clearly answer for it! 💡 
Thanks, then this PR is perfect to me. Sorry again, I'm not familiar with Github PR.

PR #187, opened 2020-06-05
Use lsb_release to determine proper ubuntu version for install docs

MasterOdin:
Unless you only do actually provide the single apt list for dingo, it's better to use `lsb_release` to fill in the distro name dynamically so that way you do not have to keep updating your site. From the [man page](https://linux.die.net/man/1/lsb_release), the `-c` gives us the distro codename and `-s` prints it out in short form (as otherwise it would print `Codename:    bionic`).


rathboma:
Hey! So I named it really badly. I should have named it `release` or `public` or something not affiliated with an Ubuntu release.

It's the same deb download for all Ubuntu versions 16.04+.

Issue #162, opened 2020-05-30
Ability to create schemas.


AqilCont:
I would really like a graphical interface for creating new schemas on the MySQL editor. It would add some more simplicity to the workflow.

rathboma:
While I think this definitely helps accessibility, this is certainly not something we have the bandwidth to add right now, nor can I see this being a core feature for a while.

I think instead we're probably better served providing templates for table creation and letting folks view the DDLs for existing tables.

I'm going to close the issue to keep the issue tracker clean -- this doesn't mean I think your issue is unimportant, but we just aren't capable of building this right now.

PR #190, opened 2020-06-06
ddl generator: add entity relationship diagram

dineug:
#162 Suggest a simple ERD editor for that issue

beekeeper-studio-erd


lion-man44:
Awesome!
I'm looking forward to merge.

Issue #178, opened 2020-06-01
make new connection configurable by URL


mariusa:
Most database providers give only the Connection URL by default, eg
```
postgres://analyst:98kd8a9@ec2-123-73-145-214.compute-1.amazonaws.com:6212/dee932clc3mg8h
```
https://devcenter.heroku.com/articles/heroku-postgresql-credentials

Note this is one string, no need to specify the separate fields.
When creating a new connection, would you please implement support for this type of URL as an alternative to filling in all the fields?

Thanks!

rathboma:
Yes! This always annoys me when I'm using heroku, cant believe I overlooked it.

strotgen:
Hey, I would like to help with this if that's okay.

rathboma:
:+1: that'd be amazing

strotgen:
How were you thinking about doing it? 

Visually: TablePlus has an [Import from URL](https://docs.tableplus.com/gui-tools/manage-connections#importing-connection-url) button on the same screen that allows you to select the database driver. That would be one option. Or maybe adding a button to the left of _Test_ and _Connect_ buttons?

For the URI parsing, I had thought of a simple method that assigns directly to `config`.

rathboma:
I think a import link in the top right on the connection card, even when the DB type drop-down is there would work, then that is parsed and fills in host/port/type/etc ready for the user to test/connect/etc as normal

Also, if someone pastes a valid DB url into the host field it should auto detect it and do the same

mariusa:
Please see https://github.com/beekeeper-studio/beekeeper-studio/pull/201/commits/263b0f5070e484d7b41135d9c0dc95f5801c268f
for a bug & solution

mariusa:
Also, please check Enable SSL too when importing (eg heroku string)

strotgen:
Thanks for spotting that bug, I'll fix it asap.

As for the SSL option, there is already some code for doing that. It just checks for sslmode=require or sslmode=prefer in the params. I could add the other [sslmode options](https://www.cockroachlabs.com/docs/stable/connection-parameters.html#secure-connections-with-urls), if that's the problem.

If it isn't working for you, could you please provide me an example of how heroku strings are formatted.

mariusa:
Thanks!

Heroku URL:
```
postgres://analyst:98kd8a9@ec2-123-73-145-214.compute-1.amazonaws.com:6212/dee932clc3mg8h
```

[...]

PR #201, opened 2020-06-13
Import Connection URL

strotgen:
I have added the functionality described in #178. Adds an import button at the top right for adding connections strings directly. Additionally, host field detects when a connection URL is added and automatically fills the necessary fields.    

I think the (+) button a the top would look better without the `btn` classes, but I am not sure.

Please let me know what you think and if I should make any changes!

### DIFF
--- a/src/components/connection/ImportUrlForm.vue
+++ b/src/components/connection/ImportUrlForm.vue
@@ -32,26 +32,31 @@
     watch: {
       config: {
         handler: function(config) {
-          this.parseUrl(config.host)
+          if (config.host.includes('://')){
+            this.parseUrl(config.host)
+          }
         },
         deep: true
       }
     },
     methods: {
       parseUrl(connectionUrl) {
-        if (!connectionUrl.includes('://')){
-          return;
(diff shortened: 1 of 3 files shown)


rathboma:
Awesome! Thanks for adding this! We'll take a look

src/components/connection/ImportUrlForm.vue
```diff
@@ -0,0 +1,54 @@
...
+          this.connectionUrl = 'https://' + this.connectionUrl
+        }
+        let url = new URL(this.connectionUrl);
+        this.config.connectionType = this.protocol(url.protocol)
+        // set https as the protocol so URL interface can
+        // correctly parse all necessary information
+        url.protocol = 'https'
+        this.config.host = url.hostname
+        this.config.port = url.port
+        this.config.defaultDatabase = url.pathname.slice(1, -1)
```
mariusa: There is a bug here. slice(1, -1) removes the last character too.
Use url.pathname.substr(1) to remove only the 1st ('/')

PR #205, opened 2020-06-17
fix text selection

### LINKED ISSUE #203: Cannot copy text/highlight text in Beekeeper latest update
Cannot copy any Text (CTRL+C). I'm using Linux PopOS.
Why the text-highlighting disabled?

rathboma:
Fixes #203 

Also makes it a little easier to select text.

I think this is important enough to push out a point update

c81ef740-65ed-468c-b49d-8a60a0028de6

### DIFF
--- a/src/assets/styles/app/vendor/tabulator.scss
+++ b/src/assets/styles/app/vendor/tabulator.scss
@@ -92,6 +92,7 @@ $errorColor: $brand-danger;
       background: transparent;
       width: 100%;
       color: $text-dark;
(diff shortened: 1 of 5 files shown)


p0n1:
I really like the demo gif. It's a lot more convenient than it used to be.

rathboma:
Thanks! In the future I'd like to make multi-cell selection easier too (so like it pastes as CSV rather than squishing them together, but that requires much deeper modifications to Tabulator

PR #213, opened 2020-06-21
Add Execute time to the status bar

### LINKED ISSUE #38: Execution time in results
SQL queries sometimes are slow... for example if you don't use indexes. Showing query time in seconds is a nice to have feature.

KasperFranz:
This is my first PR on this project. This pull request resolves #38. 

With this change, I have added the execute time to the status bar.

To make the execution time a bit more readable I have added HumanizeDuration.js (Thanks @rathboma)

There are a few things that should be considered as follow-ups (if the DB APIs allows us to do this) to use the real execution type instead of the time it takes to init the promise and the rest of the magic around the queries :).

### DIFF
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -368,14 +368,14 @@
           const query = this.deparameterizedQuery
           this.$modal.hide('parameters-modal')
 
-          const runningQuery = this.connection.query(query);
-          const queryStartTime = +new Date();
-          const results = await runningQuery.execute();
-          const queryEndTime = +new Date();
-          const queryExecutionTime = queryEndTime - queryStartTime;
+          const runningQuery = this.connection.query(query)
+          const queryStartTime = +new Date()
+          const results = await runningQuery.execute()
+          const queryEndTime = +new Date()
+          const queryExecutionTime = queryEndTime - queryStartTime
           let totalRows = 0
           results.forEach(result => {
-            result.executeTime = queryExecutionTime;
+            result.executeTime = queryExecutionTime
             result.rowCount = result.rowCount || 0
 
             // TODO (matthew): remove truncation logic somewhere sensible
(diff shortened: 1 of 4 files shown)


rathboma:
Nice! Let me poke around to see if the DB API supports execution time

src/components/TabQueryEditor.vue
```diff
@@ -202,6 +203,13 @@
         const rows = this.result.affectedRows || 0
         return `${rows} ${Pluralize('row', rows)} affected`
       },
+      executeTimeText() {
+        if (!this.result) {
+          return null
+        }
+        const executeTime = this.result.executeTime || 0
+        return `${executeTime} ms`
```
rathboma, line 211: We could use this simple package to humanize the time - https://github.com/EvanHahn/HumanizeDuration.js

src/components/TabQueryEditor.vue
```diff
@@ -360,10 +368,14 @@
...
 
-          const runningQuery = this.connection.query(query);
+          const runningQuery = this.connection.query(query)
+          const queryStartTime = +new Date()
           const results = await runningQuery.execute()
+          const queryEndTime = +new Date()
+          const queryExecutionTime = queryEndTime - queryStartTime
           let totalRows = 0
           results.forEach(result => {
+            result.executeTime = queryExecutionTime
```
rathboma, line 378: Seeing as we're not measuring this in the back-end. I'd not put this on `result`, but instead create a new top level entry in `data` - `executeTime`. It's also not per-query, so this makes more sense to me.

Then in the computed it would just be `executeTime` rather than `result.executeTime`

src/components/TabQueryEditor.vue
KasperFranz: Good point, I have addressed that now

src/components/TabQueryEditor.vue
KasperFranz: updated now

KasperFranz:
@rathboma  is there any way to trigger the build process? 
it seems to be timing out in yarn

rathboma:
I'll figure that out! I'll take a look at this tonight. Thanks!

PR #244, opened 2020-07-12
Try for a better autoscaling text

gnovaro:


rathboma:
@ghstng thoughts?

gnovaro:
I think can do better for example https://andy-carter.com/blog/using-scalable-css-units-for-font-sizes

ghstng:
pulled, tks @gnovaro

Issue #253, opened 2020-07-22
My data no longer appears in 1.6.9 - Ubuntu 20.04


martindufresne:
Hello, this morning I downloaded the latest version of the application (1.6.9), but since then, when I connect, my data no longer appears to me. However, version 1.5.3 worked really well. I am on Ubuntu 20.04 and downloaded the appimage. But I also tested with the snap and the DEB and each package has the same result. Here is a screenshot.

![Screenshot from 2020-07-22 08-48-32](https://user-images.githubusercontent.com/2438127/88178733-d7bc6f00-cbf8-11ea-908b-1a861138c9dc.png)

louis77:
Which database engine do you try to connect to?

martindufresne:
I am trying to connect to MySQL databases.

louis77:
Do you have SSL enabled?

martindufresne:
No, it is not activated. And if I reinstall version 1.5.3, everything comes back. So I tell myself that there is something in the latest version that is causing the bug. Could there be a conflict with configuration files since I tested the Snap and DEB version as well?

rathboma:
Hmm, that's concerning. I'll take a look today for you

martindufresne:
thank you so much!

rathboma:
@martindufresne few questions to help me debug:
- What version of MYSQL?
- Can you give me the table schema (columns and types)? Even just the types is helpful

martindufresne:
I am on MySQL 8.0.20. I have attached a screenshot of my table.I have indeed errors in the console. I attach a screenshot. Which is weird, I don't have this problem in version 5.3.
![Screenshot from 2020-07-22 14-20-32](https://user-images.githubusercontent.com/2438127/88213405-b8d4d180-cc26-11ea-9c65-7fc0693cd2b9.png)
![Screenshot from 2020-07-22 14-22-45](https://user-images.githubusercontent.com/2438127/88213529-dbff8100-cc26-11ea-97dc-5f17ff4ad3e0.png)

rathboma:
Ok, one more thing -- can you right click the console click save-as, save the .log file and share it here?

martindufresne:
Of course! It's sent! thanks again

rathboma:
Thanks! Hmm, unfortunately that doesn't show too much. I think one of the tables in the database has a schema that is breaking Beekeeper Studio somehow. I don't think its the orders table, but some other table.

Are you using any mysql plugins at all?

[...]

PR #255, opened 2020-07-22
Add missing table_name alias in listTableColumns for mysql 8

MasterOdin:
MySQL 8 has made it so that columns returned from querying `information_schema` tables are returned as capitalized, regardless of the capitalization you use in your query. Adding the alias here is necessary (like everywhere else in the client) to normalize it to lowercase for both MySQL 5 and 8.

Maybe will resolve problem in #253?

### DIFF
--- a/src/lib/db/clients/mysql.js
+++ b/src/lib/db/clients/mysql.js
@@ -104,7 +104,7 @@ export async function listRoutines(conn) {
 export async function listTableColumns(conn, database, table) {
   const clause = table ? `AND table_name = ?` : ''
   const sql = `
-    SELECT table_name, column_name AS 'column_name', column_type AS 'data_type'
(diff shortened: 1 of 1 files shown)


rathboma:
You are my hero. I would never have found that.

Interestingly, I have a mysql8 install and it works fine without this change, I don't know why.

MasterOdin:
I suspect there's various flags, as well as depends on the minor version. In [cakephp/phinx](https://github.com/cakephp/phinx), things broke when upgrading from 8.0.21 as before that one of the information_schema tables would return based on case used in query, and after it would return always upper case. The 8.x release has been messy.

martindufresne:
It works! Yeah! You're on fire ... Nice work and really, a big, big thank you for your help.

PR #265, opened 2020-08-01
Add modal to create a database

### LINKED ISSUE #224: Create Database Popup window
A simple dialog add database name and choose collation in a drop-down

UtechtDustin:
First of all thanks for that awesome project, im searching a good database manager since some years.
The "best" i found was dbeaver but its very ugly and ui/ux is bad.

This PR is not done yet, i first wanted to ask if it was going in the right direction.
I never used electrone/vue before, so i have to learn it :-)
Are you intresested that i continue develop on that PR or do you think it not good enough right now ?

(The `create database` query dont use prepared statements because it would work with that, if i continue developing on this PR i will fix that)

The modal already works fine, but there is still some work todo.
beekeeper

This PR would fix #224

### DIFF
--- a/src/components/connection/AddDatabaseForm.vue
+++ b/src/components/connection/AddDatabaseForm.vue
@@ -1,0 +1,73 @@
+<template>
+  <div class="save-connection expand">
+    <h3 class="dialog-c-title">Add database</h3>
+
+    <div v-if="error" class="alert alert-danger">
+      {{error}}
+    </div>
+
+    <div class="form-group">
+      <input class="form-control" v-model="databaseName" @keydown.enter.prevent.stop="save" type="text" placeholder="Database Name">
+    </div>
+
+    <div class="form-group">
+      <select v-model="selectedCharset">
+        <option v-for="charset in charSets" v-bind:key="charset" :selected="charset === defaultCharSet" :value="charset">{{charset}}</option>
+      </select>
+    </div>
+
+    <div class="form-group">
+      <select v-model="selectedCollation">
+        <option v-for="collation in collations" v-bind:key="collation" :value="collation">{{collation}}</option>
+      </select>
+    </div>
+
+    <div class="save-actions">
+      <button class="btn btn-flat" @click.prevent="$emit('cancel')">Cancel</button>
+      <button class="btn btn-primary save" @click.prevent="save">Add</button>
+    </div>
+  </div>
+</template>
+<script>
+export default {
+  props: ['connection'],
+  data() {
+    return {
+      charSets: [],
+      collations: [],
+      defaultCharSet: null,
+      databaseName: null,
+      selectedCharset: null,
+      selectedCollation: null,
+      error: null
+    }
+  },
+  async mounted(){
+    this.charSets = await this.connection.listCharsets()
+    this.defaultCharSet = await this.connection.getDefaultCharSet()
+    this.selectedCharset = this.defaultCharSet
+    console.log(this.selectedCharset)
+    await this.updateCollations()
+  },
+  methods: {
+    async updateCollations() {
+      this.collations = await this.connection.listCollations(this.selectedCharset)
+      this.selectedCollation = this.collations[0]
+    },
+    async save() {
+      try {
+        await this.connection.createDatabase(this.databaseName, this.selectedCharset, this.selectedCollation)
+        this.$noty.success('The database was created')
+        this.$emit('databaseCreated', this.databaseName)
+      } catch (err) {
+        this.error = `The database could not be created: ${err.message}"`
+      }
+    }
+  },
+  watch: {
+    selectedCharset() {
+      this.updateCollations()
+    }
+  }
+}
+</script>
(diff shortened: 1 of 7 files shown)


src/components/connection/AddDatabaseForm.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <div class="save-connection expand">
+    <h3 class="dialog-c-title">Add database</h3>
+
+    <div v-if="error" class="alert alert-danger">
+      {{error}}
+    </div>
+
+    <div class="form-group">
+      <input class="form-control" v-model="databaseName" @keydown.enter.prevent.stop="save" type="text" placeholder="Database Name">
```
rathboma, line 10: So this keydown does work, but the right way to do this is to have a `<form>` tag with a `submit` event.

Then the `cancel` button has `type=button`, and you don't need any keydown on the input. For a good example see `ImportButton.vue`

src/components/connection/AddDatabaseForm.vue
```diff
@@ -0,0 +1,71 @@
...
+  </div>
+</template>
+<script>
+export default {
+  props: ['connection'],
+  data() {
+    return {
+      charSets: [],
+      collations: [],
+      defaultCharSet: 'utf8mb4',
```
rathboma, line 38: Is this default supported across all database types?

We should get this value from the connection too, like `await connection.getDefaultCharset`
Probably we also need a `connection.supportsDatabaseCreation` or something

For example: Sqlite3 cannot create new databases, becuase a file === database.

src/components/connection/AddDatabaseForm.vue
```diff
@@ -0,0 +1,71 @@
...
+      databaseName: null,
+      selectedCharset: null,
+      selectedCollation: null,
+      error: null
+    }
+  },
+  async mounted(){
```
rathboma, line 45: this whole component looks :+1:

src/components/sidebar/core/DatabaseDropdown.vue
```diff
@@ -11,14 +11,24 @@
       <a class="refresh" @click.prevent="refreshDatabases" :title="'Refresh Databases'">
         <i class="material-icons">refresh</i>
       </a>
+      <a class="refresh" @click.prevent="$modal.show('config-add-database')" :title="'Add Database'">
```
rathboma, line 14: I think this is in the right place, we'll probably tidy these buttons up after merge, but looks great so far.

src/components/sidebar/core/DatabaseDropdown.vue
```diff
@@ -30,7 +40,11 @@
     methods: {
       async refreshDatabases() {
         this.dbs = await this.connection.listDatabases()
-      }
+      },
+      async databaseCreated() {
+        this.$modal.hide('config-add-database')
+        await this.refreshDatabases()
```
rathboma, line 46: I'd also suggest switching to the new database after creation. `this.currentDatabase = newDatabase` and pass that through into the created callback

src/lib/db/clients/mysql.js
```diff
@@ -292,6 +295,41 @@ export async function listDatabases(conn, filter) {
...
+  ];
+
+  const { data } = await driverExecuteQuery(conn, { query: sql, params });
+
+  return data
+      .map((row) => row.Collation);
+}
+
+export async function createDatabase(conn, databaseName, charset, collation) {
+  const sql = 'create database ' + databaseName + ' character set ' + charset + ' collate ' + collation;
```
rathboma, line 321: you should able to do it with backticks, then use `wrapIdentifier`, see for example truncateAllTables

src/components/connection/AddDatabaseForm.vue
UtechtDustin: This is currently hardcoded (and should be get from the database) as i first wanted some feedback for my other changes (as this is my first vue/electron project)

rathboma: Yep! I figured.

Your Vue code is great! Thanks for working so hard to make the changes good.

UtechtDustin: done

src/components/sidebar/core/DatabaseDropdown.vue
UtechtDustin: done

src/lib/db/clients/mysql.js
UtechtDustin: done

UtechtDustin:
@rathboma 
Maybe you could help me with that.

It works fine for mysql/mariadb now.
But i never used sqllite/mssql/etc. so i dont know if it is possible to define charset AND collation on all databasetypes (and how i get a list of them).
Do you maybe know that ?

rathboma:
Yes I can definitely help with that! I'll take a look over the next week. Thanks for starting the work

jatinhariani:
Hey. I could certainly use this feature and would like to help get this into master. I'm familiar with Vue and what @rathboma is trying to get you to do here. Would you like my help with this @UtechtDustin ?

UtechtDustin:
@jatinhariani Sure, what information do you need ?

rathboma:
So the way this needs to work:
1. Each database client (src/lib/db) needs a way to indicate the collations and encodings it supports
2. Dropdown should fetch those and populate the form with them (already done)
3. If they don't support one or the other, make those values null

UtechtDustin:
?

PR #268, opened 2020-08-01
Added ability to duplicate existing connections

### LINKED ISSUE #259: Allow to duplicate/clone a saved connection to edit a copy and keep most of the parameters
Not much more to say than the title: it would be a nice addition to be able to copy/duplicate saved or recent connections to have another one with similar settings but not exactly the same, eg. keeps the SSH settings.

UtechtDustin:
This PR will add a new option to duplicate a existing connection.
Fix #259

beekeeper gif

### DIFF
--- a/src/components/ConnectionInterface.vue
+++ b/src/components/ConnectionInterface.vue
@@ -2,7 +2,7 @@
   <div class="interface connection-interface">
     <div class="interface-wrap row">
       <sidebar class="connection-sidebar" ref="sidebar">
-        <connection-sidebar :defaultConfig="defaultConfig" :selectedConfig="config" @remove="remove" @edit="edit" @connect="handleConnect"></connection-sidebar>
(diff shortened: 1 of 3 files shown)


src/components/ConnectionInterface.vue
```diff
@@ -140,6 +140,19 @@
         }
         this.$noty.success(`${config.name} deleted`)
       },
+      async duplicate(config) {
+        const duplicateConfig = Object.assign( Object.create( Object.getPrototypeOf(config)), config);
```
rathboma, line 144: Nice, please add a comment here about what this does (clone the ES6 class), because I think this is not obvious
```

File: src/components/ConnectionInterface.vue [hunk 2]
```diff
@@ -140,6 +140,19 @@
...
+        duplicateConfig.id = null
+        duplicateConfig.name += ' (copy)'
```
rathboma, line 146: Due to our text trimming, can we do this the other way around?

"Copy of Foo Bar"

PR #274, opened 2020-08-07
Added the ability to use multiple filters

UtechtDustin:
This PR added the ability to use multiple filters

Peek 2020-08-07 09-06


rathboma:
Damn, this is awesome. Will take a look tonight. @ghstng and I can figure out how to handle the UX of this. Might also clash with some changes we've been preparing

src/components/tableview/TableTable.vue
```diff
@@ -143,22 +154,42 @@ export default {
...
-      if (this.filter.type && this.filter.field) {
-        if (this.filter.value) {
-          this.tabulator.setFilter(
-            this.filter.field,
-            this.filter.type,
-            this.filter.value
-          );
-        } else {
-          this.tabulator.clearFilter();
+      const filterArray = JSON.parse(JSON.stringify(this.filter));
```
rathboma, line 180: why do you do this?

UtechtDustin: `this.filter` is a observer, as i already told this is my first vue application so maybe there is a better method to get the object from a observer.
I googled that and this was the solution i found :-)

rathboma: so I don't think you need to worry about converting filter to another data type.

I think you can just pass it as-is `this.tabulator.setFilter(this.filter)`

What I would do is make a computed property called something like `checkedFilter` or `resolvedFilter`, and in there do the validity check.

Then we can use that computed field to do a couple of things:
1. in the Filter UI we can use it to disable the 'go' button if filters are not valid
2. We can just pass that through to tabulator in this method without any checking

rathboma: once this one change is made, I'll merge it, but @ghstng and I need to iterate on the UI a little. We have a multi-filter UI in mind.

UtechtDustin: Thats fine for me, i use a computed function now :-)

rathboma:
I'm going to merge this into another branch so we can work on styling. After chatting we want to give this feature proper attention, which means reworking the filtering UI to be a bit more cohesive when there are multiple filters.

UtechtDustin:
No problem, im not a ui expert :-)

PR #275, opened 2020-08-07
Feature/context menu

### LINKED ISSUE #260: SQL Formatter
Would it be possible to add an sql formatter?

### LINKED ISSUE #270: Context menu options on query window or query tab click
- **Context menu options on right click to the query tab**
Pin one or more query windows 
Closing options like (close, close all, close all but this, close all but pinned)
Option to open file containing folder

- **Context menu options on right click to the query editor window**
Allow to format the SQL queries

UtechtDustin:
This PR adds a context-menu component and would Fix #270 and would Fix #260.
Already implemented context-menus:
- Tabs
  - close
  - close other
  - close all
  - duplicate a query (tab) 
- Sql editor
  - format sql

The formatting also have a shortkey `meta/strg+alt+f`.

Tabs:
Peek 2020-08-07 10-36

Editor:
Peek 2020-08-07 10-35

### DIFF
--- a/src/components/CoreTabs.vue
+++ b/src/components/CoreTabs.vue
@@ -6,6 +6,7 @@
           v-for="tab in tabItems"
           :key="tab.id"
           :tab="tab"
+          :tabsCount="tabItems.length"
           :selected="activeTab === tab"
           @click="click"
           @close="close"
--- a/src/components/CoreTabs.vue
+++ b/src/components/CoreTabs.vue
@@ -178,10 +178,23 @@
         }
       },
       duplicate(tab) {
-        let duplicatedTab = Object.assign( Object.create( Object.getPrototypeOf(tab)), tab)
-        duplicatedTab.id = uuidv4()
-        duplicatedTab.title = "Query #" + this.newTabId
+        let duplicatedTab = {
+            id: uuidv4(),
+            type: tab.type,
+            connection: tab.connection,
+            unsavedChanges: true
+        }
 
+        if(tab.type === 'query') {
+          const query = new FavoriteQuery()
+          query.text = tab.query.text
+
+          duplicatedTab['title'] = "Query #" + this.newTabId
+          duplicatedTab['unsavedChanges'] = true
+          duplicatedTab['query'] = query
+        } else if(tab.type === 'table') {
+          duplicatedTab['table'] = tab.table
+        }
         this.addTab(duplicatedTab)
       }
     },
--- a/src/components/CoreTabs.vue
+++ b/src/components/CoreTabs.vue
@@ -11,6 +11,7 @@
           @close="close"
           @closeAll="closeAll"
           @closeOther="closeOther"
+          @duplicate="duplicate"
           ></core-tab-header>
       </ul>
       <span class="actions">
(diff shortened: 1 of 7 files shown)


ghstng:
siiick @rathboma

rathboma:
Man, you're on fire! Thanks for all the PRs. Reviewing them all soon

UtechtDustin:
No Problem, i searched for a long time for a good (looking) database tool for linux.
So why should i not help to create one :-)

src/components/CoreTabs.vue
```diff
@@ -152,20 +154,36 @@ import { mapGetters } from 'vuex'
...
+      },
+      closeOther(tab) {
+        this.tabItems = [tab]
+        this.activeTab = tab;
+        if (tab.query && tab.query.id) {
+          tab.query.reload()
+        }
+      },
+      duplicate(tab) {
+        let duplicatedTab = Object.assign( Object.create( Object.getPrototypeOf(tab)), tab)
```
rathboma, line 181: So this cloning doesn't work here because table tabs have a `table` or a `query` embedded in them which is the same table object as comes from the store.

So think we have to manually set id/title/querytext/etc

Some things this needs to do:
1. Manually set query text on a new query object so that saved queries can be duplicated (only one copy of a saved query should be open at a time)
2. Don't clone, but just use the same `table` object for new table tabs, so that we maintain referential integrity with the store

Other questions:

src/components/TabQueryEditor.vue
```diff
@@ -254,6 +262,7 @@
       keymap() {
         const result = {}
         result[this.ctrlOrCmd('l')] = this.selectEditor
+        result[this.ctrlOrCmd('alt+f')] = this.formatSql
```
rathboma, line 265: So we probably need to use the common/menu code to provide the context menu functionality (because we'll want to add menu-bar items to do all these things too, and that will also handle the keyboard shortcuts).

I'll need to think about how to do that.
I think we'll just need to import the menu entries as needed, then use them for the context menu, same way we do for the client-side menu. It might be fiddly to do though.

src/components/CoreTabs.vue
UtechtDustin: I commit a new version of the duplicate function, which builds a new object instead of cloning the old one.
Please check if that version is ok.

src/components/TabQueryEditor.vue
UtechtDustin: I also tried to implement that one, but i guess im to new for that.  
If i tried to add new menu entries (or for testing remove the whole menu object) nothing changed on the menu (also after a restart).

Feel free to move that part of the code :-)

rathboma:
Sorry I haven't merged this yet -- I want to spend a little time figuring out how to integrate with the app menu, and also maybe use x-menu for the menu UI to keep it consistent with buttons

rathboma:
I forked your change and merging it now. Thanks a bunch! This rocks.

UtechtDustin:
Good work ! :-)

rathboma:
@UtechtDustin would you be interesting in doing some freelance consulting work on Beekeeper? If you'd like to talk about it, send me an email - matthew [at] beekeeperstudio [dot] io

Issue #39, opened 2020-05-03
Editing rows by double click


AqilCont:
A very loved feature from our MySQL Workbench. Clicking on a row in the retrieved data from a query replaces the text present there with a text input, where you can input the new value. Would appreciate this a lot, and would make me happily switch over.

rathboma:
:100:. We're working towards this feature step by step. Stay tuned!

First step is the #33 PR which allows viewing table data, from there I can work towards implementing the editing interface.

UtechtDustin:
@rathboma can be closed i guess, as editing is possible with  #276.

rathboma:
Yep, this works! Got a few small changes I need to make, but then a release will be ready

PR #276, opened 2020-08-07
Foreign key lookups & Table Editing

### LINKED ISSUE #118: enhancement: Foreign Key relation with a link like PhpMyAdmin
Is a nice to have. When a click in a relational link open the another table data view.

rathboma:
Fixes #118 #39 

SUPER EARLY PREVIEW:
early-preview-editing

- usual caveat - this will change a lot before release

Currently working for SQLite

- Edit your table data! Buffers changes, then you can commit whenever you're ready
- Click a foreign key to view that item in the other table (in a new tab, love having tabs)

TODO:
- [x] MySQL integration
- [x] PSQL integration
- [x] SQL Server Integration
- [x] Testing on Redshift (should be disabled?)
- [ ] Testing on MariaDB
- [x] Testing on Mysql 8
- [ ] Test editing on multi-line values (maybe need to use `textarea` vs input)
- [x] Testing JSON values in psql


UtechtDustin:
Looks greate for a first version, i have two ideas which would be great:
- Align the pagination to the center of the bottom bar, without `commit` and `discard` Button it looks positioned incorrectly.
- If there is foreign key it would be create to show the data of that column as tooltip or something similar while (long) hover. If you want to check multiple row it dont have to open multiple tabs for that.

src/lib/db/clients/sqlite.js
```diff
@@ -107,6 +109,48 @@ export function query(conn, queryText) {
...
 
+export async function updateValues(conn, updates) {
+  const updateCommands = updates.map(update => {
+    return {
+      query: `UPDATE ${update.table} SET ${update.column} = ? WHERE ${update.pkColumn} = ?`,
+      params: [update.value, update.primaryKey]
+    }
+  })
+
+  const commands = [{ query: 'BEGIN'}, ...updateCommands, {query: 'COMMIT'}];
```
UtechtDustin, line 120: Does this makes sense ?
As we use `begin`, then the `update queries` and instantly `commit` the `rollback` on the  `catch` part will have no effect or not ?

I added the `commit` to the end in #277

rathboma: So I guess we should commit a the end, after the select queries, so it all happens in one go. I'll make that change.

src/lib/db/clients/mysql.js
```diff
@@ -277,7 +277,7 @@ export function query(conn, queryText) {
 export async function updateValues(conn, updates) {
   const updateCommands = updates.map(update => {
     return {
-      query: `UPDATE \`${update.table}\` SET \`${update.column}\` = ? WHERE \`${update.pkColumn}\` = ?`,
+      query: `UPDATE ${wrapIdentifier(update.table)} SET ${wrapIdentifier(update.column)} = ? WHERE ${wrapIdentifier(update.pkColumn)} = ?`,
```
UtechtDustin, line 280: I think we need the backticks if someone names a table with (reserved terms) `update`, `table`, etc.

rathboma: I've included the library `knex` in the project and want to use that for all the update queries

rathboma: I'll move mysql and sqlite to using that sql generator over the next few days. I just pushed sqlserver and pg support using it. Much simplifies the code. I hate manually building sql.

Oh, the `wrapIdentifier` adds the backticks fyi.

UtechtDustin: Oh good to know :+1:

rathboma:
I think this is actually ready to merge (!)

linaspasv:
> * If there is foreign key it would be great to show the data of that column as tooltip or something similar while (long) hover. If you want to check multiple foreign key rows, you dont have to open multiple tabs for that.

This could be a huge QoL improvement. I hope this is implemented at some point.

PR #277, opened 2020-08-07
Added editing support for mysql/mariadb

UtechtDustin:
ref #276

### DIFF
--- a/src/lib/db/clients/mysql.js
+++ b/src/lib/db/clients/mysql.js
@@ -39,8 +39,10 @@ export default async function (server, database) {
     listTableIndexes: (db, table) => listTableIndexes(conn, db, table),
     listSchemas: () => listSchemas(conn),
     getTableReferences: (table) => getTableReferences(conn, table),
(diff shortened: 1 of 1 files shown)


src/lib/db/clients/mysql.js
```diff
@@ -178,6 +180,13 @@ export async function getTableReferences(conn, table) {
   return data.map((row) => row.referenced_table_name);
 }
 
+export async function getPrimaryKey(conn, database, table) {
+  logger().debug('finding foreign key for', database, table)
+  const sql = `SHOW KEYS FROM ${table} WHERE Key_name = 'PRIMARY'`
+  const { data } = await driverExecuteQuery(conn, { query: sql })
+  return data[0] ? data[0].Column_name : null
```
rathboma, line 187: @MasterOdin Is this one of those things that will be capitalized in mysql8?

UtechtDustin: Yes its starts with a capitalized `C`

PR #278, opened 2020-08-08
Fixed foreign keys for mysql/mariadb & a few optimizations

UtechtDustin:
This PR fixed fetching the foreign keys (which mean with mysql you can open the referenced row in a new tab).

It also fixed the issue that the "open referenced row in new tab icom" will be displayed if the cell value is null.
It also will hide the pagination if there are no other pages.

### DIFF
--- a/src/lib/db/clients/mysql.js
+++ b/src/lib/db/clients/mysql.js
@@ -196,7 +196,7 @@ export async function getTableKeys(conn, database, table) {
     FROM information_schema.key_column_usage
     WHERE table_schema = database()
     AND table_name = ?
-    AND ((referenced_table_name IS NOT NULL) OR constraint_name LIKE '%PRIMARY%')
+    AND referenced_table_name IS NOT NULL
   `;
 
   const params = [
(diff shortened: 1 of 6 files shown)


src/lib/db/clients/mysql.js
```diff
@@ -190,9 +190,9 @@ export async function getPrimaryKey(conn, database, table) {
 export async function getTableKeys(conn, database, table) {
   const sql = `
     SELECT constraint_name as 'constraint_name', column_name as 'column_name', referenced_table_name as 'referenced_table_name',
-      CASE WHEN (referenced_table_name IS NOT NULL) THEN 'FOREIGN'
-      ELSE constraint_name
-      END as key_type
+      IF(referenced_table_name IS NOT NULL, 'FOREIGN', constraint_name) as key_type,
+      REFERENCED_TABLE_NAME as referenced_table,
+      REFERENCED_COLUMN_NAME as referenced_column
```
rathboma, line 195: I think actually this needs a tweak to change the `where` -- we need to remove the primary key from the filter.

line 199 - remove `or constraint_name like '%PRIMARY%'`

UtechtDustin: while developing i was confused why the primary are alsi needed, but i didnt know why so leaved it.
I will remove it :+1:

UtechtDustin: done

PR #282, opened 2020-08-10
Added Editorconfig and fixed resulting errors

UtechtDustin:
As #35 will not be updated, i adjusted the editorconfig of that PR and fixed the resulting errors.

I removed the `*.py` and `makefile` part because there are no make or .py files on this project.
I also moved
```
indent_style = space
indent_size = 2
```
to the global scope to reduce the complexibility of the editorconfig file, this also should fix your (@rathboma) comment
```
Also we should make sure there is an explicit entry for .vue files
```
Or do we need some other options for vue files ?


rathboma:
don't think we need specific ones actually

PR #286, opened 2020-08-14
Dont show both menus, fixed wrong conditions

### LINKED ISSUE #262: BUG: There're two menu bars and two title bars after changing menu style to native
image

Beekeeper Version: 1.6.11
OS: Windows 10 1903

UtechtDustin:
This PR adjust the `v-if` condition for the titlebar to Fix #262

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,7 +1,7 @@
 <template>
 <div class="style-wrapper">
     <div class="beekeeper-studio-wrapper">
-      <titlebar v-if="$config.isWindows || $config.isMac || menuStyle === 'client'"></titlebar>
+      <titlebar v-if="($config.isWindows || $config.isMac) && menuStyle === 'client'"></titlebar>
       <connection-interface v-if="!connection"></connection-interface>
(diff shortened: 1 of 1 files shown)


UtechtDustin:
@rathboma your fix (https://github.com/beekeeper-studio/beekeeper-studio/commit/727702e1119c108757310e2257a30b28e2e5939e)
But then we also have to hide the `Menu style` menu item for `macos`.

rathboma:
@UtechtDustin I think that should be hidden by the code in `Titlebar.vue`.

rathboma:
This whole logic section is a total mess tbh

rathboma:
Really App.vue should do two things:
1. Decide whether to display titlebar
2. Decide whether to display app menus

Then pass the second one into the titlebar class

`:displayAppMenu="displayAppMenu"`

But not super urgent so haven't done it

UtechtDustin:
> @UtechtDustin I think that should be hidden by the code in `Titlebar.vue`.

Your right, didnt saw that one.
As you already wrote, its quite a mess right now :D

PR #288, opened 2020-08-14
Added an option to copy the connection string

UtechtDustin:
As we have a option to import a connection string we also should have a option to copy that connection string (also usefull for e.g. pdo (PHP) ).

The `hr` will cause a horizontal scrollbar, im not sure whats the best way to add a seperator with `x-menuitem`.


src/components/sidebar/connection/ConnectionListItem.vue
```diff
@@ -101,6 +110,14 @@ export default {
     },
     duplicate() {
       this.$emit('duplicate', this.config)
+    },
+    copyUrl() {
```
rathboma, line 114: If you make this an `async` function you can use `await this.$copyText` then just try/catch as usual

src/components/sidebar/connection/ConnectionListItem.vue
```diff
@@ -101,6 +110,14 @@ export default {
...
+      this.close()
+      this.$copyText(this.config.fullConnectionString)
```
rathboma, line 116: Should this include the password if it is saved?

rathboma:
HR is the best way to do it. We can adjust styles later

src/components/sidebar/connection/ConnectionListItem.vue
UtechtDustin: No, the same getter will be used for the `title` attributes of the connection list item.

src/components/sidebar/connection/ConnectionListItem.vue
UtechtDustin: Good to know, thanks!
I changed that.

UtechtDustin:
> HR is the best way to do it. We can adjust styles later

Okay, i was not sure if there is maybe already a element/component for that.

rathboma:
So good.

Vue tip: in templates you can just use 'connectionType' without the 'this'.

PR #289, opened 2020-08-14
Use the x-mssql mode for mssql

### LINKED ISSUE #199: Backslash as escape character throws off SQL syntax highlighting
It appears that backslash is treated as an escape character for the purposes of the SQL syntax highlighter, so it treats all code following a backslash encased in single-quotes ('\\') as one long string.

image

Notepad++ has an option in its language settings called "Treat backslash as escape character for SQL" which, when disabled, resolves this issue. Perhaps a similar solution could be implemented here?

### LINKED ISSUE #208: Auto completion when using backticks around tables names
I would love the auto completion of column names also when adding backticks around the table name. Maybe it is not always needed but it is something I've got accustomed to. E.g. typing

```
SELECT * FROM `users` WHERE `users`.`
```

would then auto complete the columns.

### LINKED ISSUE #209: Auto completion of columns without table prefix
I would love if column auto completion would also work when not prefixing with the table name. This saves some typing when doing single table queries. Also, not having this makes me feel I have to tell the software how to be smart :) E.g. when typing `SELECT * FROM users WHERE ` it would auto complete column names from `users`.

I understand this might be tricky, especially when using joins. I don't know if it helps from a UX pov, but I would be fine if it auto completes columns from multiple tables in such case.

UtechtDustin:
This PR will change the mode of the codemirror editor.
Codemirror has internal a variable called `backslashStringEscapes` which will decide if a backslash will escape or not.

For the current used mode `x-sql` the  `backslashStringEscapes` is not set which means its true.

Fix #199
Bildschirmfoto von 2020-08-14 22-34-51

Update: This PR will now use also the correct type for `mysql` and `postgresql`.
Because of the correct mode the following two Issues are fixed:

Fix #209 
Bildschirmfoto von 2020-08-14 22-31-05

Fix #208
Bildschirmfoto von 2020-08-14 22-30-25

### DIFF
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -469,9 +469,14 @@
           "Cmd+Shift+F": this.formatSql
         }
 
+        const modes = {
+          'mysql': 'text/x-mysql',
+          'postgresql': 'text/x-pgsql',
+          'sqlserver': 'text/x-mssql',
+        };
         this.editor = CodeMirror.fromTextArea($editor, {
           lineNumbers: true,
-          mode: this.connection.connectionType === 'sqlserver' ? 'text/x-mssql' : "text/x-sql",
+          mode: this.connection.connectionType in modes ? modes[this.connection.connectionType] : "text/x-sql",
           theme: 'monokai',
           extraKeys: {"Ctrl-Space": "autocomplete", "Cmd-Space": "autocomplete"},
           hint: CodeMirror.hint.sql,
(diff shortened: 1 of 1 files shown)


UtechtDustin:
Update: This PR will now use also the correct type for `mysql` and `postgresql.
Because of the correct mode the following two Issues are fixed:

#209 
Bildschirmfoto von 2020-08-14 22-31-05

#208
Bildschirmfoto von 2020-08-14 22-30-25

rathboma:
Awesome!

PR #293, opened 2020-08-15
Use the default postgres database if no default database is given

### LINKED ISSUE #292: Request: Allow not to set a connection's default database
When configuring a Postgresql connection and leaving out the Default Database field, the app tries connect to a database with a name of the configured username, giving a red error message:

"Database >>_User's name_<< doesn't exist"

For a PSQL connection I set the field to `postgres`, but I'm not very fond of this solution, as I don't need that database being open on connection. 

Maybe my workflow is too specific, but as I work with multiple databases and switch between them anyway, it'd be nice to connect to the server and have the left-hand pane empty, without any database loaded.

UtechtDustin:
Currently the underlaying library for postrgres tries to connect with the database username if no default database is given.
It makes more sense to connect to the default `postgres` database if no default database is given.

FIx #292

### DIFF
--- a/src/lib/db/server.js
+++ b/src/lib/db/server.js
@@ -51,7 +51,7 @@ export function createServer(serverConfig) {
         return server.db[dbName];
       }
 
-      if(dbName === null && server.config.client === 'postgresql') {
+      if(dbName === null && serverConfig.client === 'postgresql') {
         dbName = 'postgres'
       }
 
--- a/src/lib/db/server.js
+++ b/src/lib/db/server.js
@@ -51,6 +51,10 @@ export function createServer(serverConfig) {
         return server.db[dbName];
       }
 
+      if(dbName === null) {
+        dbName = 'postgres'
+      }
+
       const database = {
         database: dbName,
         connection: null,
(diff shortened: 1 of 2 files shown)


rathboma:
This should be scoped to just postgres databases I think. Easy way to do that?

rathboma:
Maybe instead this could just be a default when you switch to 'psql' in the connection editor? Rather than a 'magic' setting

UtechtDustin:
Oh shit, that was my mistake.
At first i had it in a other file (only for psql), but that was buggy so i moved it to the `server` file an didnt noticed that this is for all connections.
I agree that implementation makes no sense, i will try to find a better implementation for that later.

UtechtDustin:
Found a better solution :-)

src/lib/db/server.js
```diff
@@ -51,6 +51,10 @@ export function createServer(serverConfig) {
         return server.db[dbName];
       }
 
+      if(dbName === null && server.config.client === 'postgresql') {
```
rathboma, line 54: I think you need `serverConfig.client`?

UtechtDustin: Both possible, i changed it to `serverConfig` now.

PR #294, opened 2020-08-17
Fixed formatting shortcut and added shortcut for toggling comments

### LINKED ISSUE #129: Keyboard shortcut: comment line
It would be nice if we could use cmd + / to comment queries.

UtechtDustin:
This PR will add add a shortcut (`Ctrl+/`) and will Fix #129.
Also this PR will fix the shortcut for formatting code.

Peek 2020-08-17 23-13

### DIFF
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -105,7 +105,7 @@
   import _ from 'lodash'
   import 'codemirror/addon/search/searchcursor'
   import CodeMirror from 'codemirror'
-  require('codemirror/addon/comment/comment')
+  import 'codemirror/addon/comment/comment'
   import Split from 'split.js'
   import { mapState } from 'vuex'
 
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -105,13 +105,14 @@
   import _ from 'lodash'
   import 'codemirror/addon/search/searchcursor'
   import CodeMirror from 'codemirror'
+  require('codemirror/addon/comment/comment')
   import Split from 'split.js'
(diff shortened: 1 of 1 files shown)


UtechtDustin:
@rathboma how i can trigger the checks to run again ?
There was a problem with the internet connection.

rathboma:
I have to do it, don't worry that much though, usually I just ignore those types of failures

src/components/TabQueryEditor.vue
```diff
@@ -105,13 +105,14 @@
   import _ from 'lodash'
   import 'codemirror/addon/search/searchcursor'
   import CodeMirror from 'codemirror'
+  require('codemirror/addon/comment/comment')
```
rathboma, line 108: `import 'codemirror/addon/comment/comment'`

src/components/TabQueryEditor.vue
```diff
@@ -465,8 +475,10 @@
           "Cmd-Enter": this.submitTabQuery,
           "Ctrl-S": this.triggerSave,
           "Cmd-S": this.triggerSave,
-          "Ctrl+Shift+F": this.formatSql,
-          "Cmd+Shift+F": this.formatSql
+          "Shift-Ctrl-F": this.formatSql,
```
rathboma, line 478: doh, thanks

src/components/TabQueryEditor.vue
UtechtDustin: Sorry, its the first project with `import/require`, i tried stuff like:
```
import Comment From '...'
```
Which dont worked but, so i thought require is the right way.
Thanks for the info :+1: 

Btw. i changed that.

PR #296, opened 2020-08-18
Added canel button and show the row count of a table

### LINKED ISSUE #54: HEX id are not shown correctly
HEX fields are not shown properly.

### LINKED ISSUE #128: Add "abort query" button and hot keys combination
Hi,
I'm not sure this feature exists already so I decided to create a feature request.
Sometimes you may write query like "update users set balance=1000000 where is_active" and then just forget to specify, for example, id or something like this. So this query adds  1000000 money to all users and sometimes (haha) this is not what you want to gonna happen. 
In this kind of situation I always using "Ctrl + C" keys combination. And it could be awesome to have this combination in Beekeeper Studio.
Thanks!

### LINKED ISSUE #151: Feature Request: Add row conter to interface
I think it would be nice to know how many rows a table has without running a separate query.
My suggestion is to put it at the bottom of the window next to the pagination buttons.

### LINKED ISSUE #185: Ability to cancel a running query
The query interface provides `cancel` and `discard`. So if someone runs a really long query by mistake they should be able to cancel it.

- Add information to the 'running' state to indicate if the state is `running query` or `loading data into table`
- If state is `running query` provide a `Cancel` button that calls `runningQuery.cancel` to stop query execution
- Add some sort of error, like `Query Execution Cancelled`

### LINKED ISSUE #295: Clicking "END" remove focus from editing field
First, I was looking for a good sql editor for a long time and today I found this amazing project. Thanks for all your great working.

As I tested de editor, I missed the feature to edit fields but right after that i found that it was already merged. So I downloaded the master branch to test and its working really well, but when I'm editing a field and click 'END' the editor remove the focus from the field instead of moving the cursor to the end of the text.
Is this behavior expected?

UtechtDustin:
This PR:

-  Adds a button to cancel queries (Fix #185 and Fix #128)
-  Adds the row count of tables (Fix #151)
-  Fix hex values (Fix #54)
-  Allow to use `END` and `POS1` to jump within editing input (Fix #295)

Peek 2020-08-19 16-17

Bildschirmfoto von 2020-08-18 18-48-47

Peek 2020-08-18 18-52

### DIFF
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -47,6 +47,8 @@
       <query-editor-status-bar
         v-model="selectedResult"
         :results="results"
+        :running="running"
+        @cancelQuery="cancelQuery"
         @download="download"
         @clipboard="clipboard"
         :executeTime="executeTime"
@@ -285,6 +287,11 @@
       }
     },
     methods: {
+      cancelQuery() {
+        this.running = false
+        this.runningQuery.cancel()
+        this.runningQuery = null
+      },
       download(format) {
         this.$refs.table.download(format)
       },
@@ -358,9 +365,9 @@
           const query = this.deparameterizedQuery
           this.$modal.hide('parameters-modal')
 
-          const runningQuery = this.connection.query(query)
+          this.runningQuery = this.connection.query(query)
           const queryStartTime = +new Date()
-          const results = await runningQuery.execute()
+          const results = await this.runningQuery.execute()
           const queryEndTime = +new Date()
           this.executeTime = queryEndTime - queryStartTime
           let totalRows = 0
@@ -378,7 +385,9 @@
           this.results = results
           this.$store.dispatch('logQuery', { text: query, rowCount: totalRows})
         } catch (ex) {
-          this.error = ex
+          if(this.running) {
+            this.error = ex
+          }
         } finally {
           this.running = false
         }
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -379,6 +379,19 @@
           results.forEach(result => {
             result.rowCount = result.rowCount || 0
 
+            let rows = []
+            result.rows.forEach(row => {
+              const tmpRow = {}
+              for (let [key, value] of Object.entries(row)) {
+                if(value instanceof Buffer) {
+                  value = value.toString()
+                }
+                tmpRow[key] = value
+              }
+              rows.push(tmpRow)
+            })
+            result.rows = rows
+
             // TODO (matthew): remove truncation logic somewhere sensible
             totalRows += result.rowCount
             if (result.rowCount > this.$config.maxResults) {
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -379,19 +379,6 @@
           results.forEach(result => {
             result.rowCount = result.rowCount || 0
 
-            let rows = []
-            result.rows.forEach(row => {
-              const tmpRow = {}
-              for (let [key, value] of Object.entries(row)) {
-                if(value instanceof Buffer) {
-                  value = value.toString()
-                }
-                tmpRow[key] = value
-              }
-              rows.push(tmpRow)
-            })
-            result.rows = rows
-
             // TODO (matthew): remove truncation logic somewhere sensible
             totalRows += result.rowCount
             if (result.rowCount > this.$config.maxResults) {
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -41,6 +41,7 @@
       <result-table ref="table" v-else-if="rowCount > 0" :tableHeight="tableHeight" :result="result" :query='query'></result-table>
       <div class="message" v-else-if="result"><div class="alert alert-info"><i class="material-icons">info</i><span>Query Executed Successfully. No Results</span></div></div>
       <div class="message" v-else-if="error"><div class="alert alert-danger"><i class="material-icons">warning</i><span>{{error}}</span></div></div>
+      <div class="message" v-else-if="info"><div class="alert alert-info"><i class="material-icons">warning</i><span>{{info}}</span></div></div>
       <div v-else><!-- No Data --></div>
       <span class="expand" v-if="!result"></span>
       <!-- STATUS BAR -->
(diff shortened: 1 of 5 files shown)


rathboma:
Love this.

Regarding cancel button replacement: You like it better in statusbar or underneath the 'Query Running' text? My one hesitation with the status bar is that if you click as it disappears you'll end up downloading the results, lol. Not sure if that's a good enough reason to change it?

src/components/TabQueryEditor.vue
```diff
@@ -357,15 +369,28 @@
...
           this.executeTime = queryEndTime - queryStartTime
           let totalRows = 0
           results.forEach(result => {
             result.rowCount = result.rowCount || 0
 
+            let rows = []
+            result.rows.forEach(row => {
+              const tmpRow = {}
+              for (let [key, value] of Object.entries(row)) {
+                if(value instanceof Buffer) {
```
rathboma, line 385: See data_mutators.js and add an entry there!

rathboma: If I'm not using that in query results yet, I should be

src/components/TabQueryEditor.vue
```diff
@@ -466,7 +493,9 @@
           "Ctrl-S": this.triggerSave,
           "Cmd-S": this.triggerSave,
           "Ctrl+Shift+F": this.formatSql,
-          "Cmd+Shift+F": this.formatSql
+          "Cmd+Shift+F": this.formatSql,
+          "Ctrl-C": this.cancelQuery,
+          "Cmd-C": this.cancelQuery
```
rathboma, line 498: We need to add a check here to see if the query is running before capturing this keycode because ctrl-c otherwise needs to be `copy`.
Also we don't want cmd-c to be cancel, that would stay ctrl-c (mac is confusing)

src/components/TabQueryEditor.vue
UtechtDustin: I will try it again, im sure i tried it and the tabulator casted the value to a string.
A buffer casting to a string results to a json (see the screenshot in the issue).

i will report if i can do that with a mutator.

UtechtDustin:
> Love this.
> 
> Regarding cancel button replacement: You like it better in statusbar or underneath the 'Query Running' text? My one hesitation with the status bar is that if you click as it disappears you'll end up downloading the results, lol. Not sure if that's a good enough reason to change it?

I also was unsure if it is a good place for the button, but underneath the `Query Running` text it looks not that good.

src/components/TabQueryEditor.vue
UtechtDustin: Oh your right, i just followed the issue suggestion and thought `ctrl+c` makes sense because for terminals its the same shortcut.

What would be a good shortcut for that ?

rathboma:
Just leave it in the statusbar -- if it's annoying in practice we can always move it. @ghstng can move it when he's back from vaca if he wants.

src/components/TabQueryEditor.vue
rathboma: I think ctrl-c is fine, this keybinding can be dynamic!

if (this.running) result['Ctrl+C'] = this.cancelQuery

rathboma: That's the nice thing about vue computed properties :-)

src/components/TabQueryEditor.vue
UtechtDustin: Works and is committed.

src/components/TabQueryEditor.vue
UtechtDustin: Im not sure if i understand what do you mean.

Do you mean a computed property which returns an `object` (with the keymap) if a query is running and if not, it returns an empty object ?
That object will be merged with `runQueryKeyMap` ?

src/mixins/data_mutators.js
```diff
@@ -11,6 +11,7 @@ export default {
 
     genericMutator(value) {
       // if (_.isNil(value)) return NULL
+      if(value instanceof Buffer) return value.toString()
```
rathboma, line 14: or `_.isBuffer(value)`?

UtechtDustin: If you wish i can use that, but it would create a unnecessary overhead because its would check the same thing.

rathboma: doesn't matter to me either way honestly

rathboma:
Oh, one question -- does this `hex` fix also work with table editing? Or does the string conversion break it?

UtechtDustin:
> Oh, one question -- does this `hex` fix also work with table editing? Or does the string conversion break it?
The string conversion inside `node-mysql2` will break it.
I also found nothing how you can safely save hex data using `node-msql2`.

If we put the variable directly inside of the update sql without the params array it works fine, but that would not be the best solution i guess.

src/components/TabQueryEditor.vue
UtechtDustin: I tried the following which doesnt work, because the computed function will be called once not multiple times
```
const runQueryKeyMap = Object.assign({
          "Shift-Ctrl-Enter": this.submitCurrentQuery,
          "Shift-Cmd-Enter": this.submitCurrentQuery,
          "Ctrl-Enter": this.submitTabQuery,
          "Cmd-Enter": this.submitTabQuery,
          "Ctrl-S": this.triggerSave,
          "Cmd-S": this.triggerSave,
          "Shift-Ctrl-F": this.formatSql,
          "Shift-Cmd-F": this.formatSql,
          "Ctrl-/": this.toggleComment,
          "Cmd-/": this.toggleComment
        }, this.cancelQueryKeyMap)
```
and as computed prop
```
cancelQueryKeyMap() {
        if (!this.running) {
          return {};
        }

        return {
          "Ctrl-C": this.cancelQuery,
          "Cmd-C": this.cancelQuery
        }
      }
```

UtechtDustin:
Fix #295

Peek 2020-08-19 16-17

src/components/tableview/TableTable.vue
```diff
@@ -286,7 +286,11 @@ export default {
       lastUpdated: null,
       // callbacks
       ajaxRequestFunc: this.dataFetch,
-      index: this.primaryKey
+      index: this.primaryKey,
+      keybindings: {
```
rathboma, line 290: I just submitted a PR to tabulator to stop this behavior during editing, see https://github.com/olifolkerd/tabulator/pull/2946, it's also an issue with pageup, pagedown.

I'm fine with this patch until that PR is resolved.

UtechtDustin: :+1:

PR #299, opened 2020-08-20
Add select to bool type at edit mode

geovannimp:
Before this PR columns with type bool in edit mode used a textfield that was allowing to any value be placed.

With this PR columns with type bool in edit mode will use a select with only `true` and `false` options.

Peek 19-08-2020 22-14

### DIFF
--- a/src/components/tableview/TableTable.vue
+++ b/src/components/tableview/TableTable.vue
@@ -184,9 +184,10 @@ export default {
         // currently it doesn't fetch the right result if you update the PK
         // because it uses the PK to fetch the result.
         const editable = this.editable && column.columnName !== this.primaryKey
-        const useTextarea = column.dataType === 'text'
-        const editorType = useTextarea ? 'textarea' : 'input'
         const slimDataType = this.slimDataType(column.dataType)
+        const editorType = this.editorType(column.dataType)
+        const useVerticalNavigation = editorType === 'textarea'
+
         const formatter = () => {
           return `<span class="tabletable-title">${column.columnName} <span class="badge">${slimDataType}</span></span>`
         }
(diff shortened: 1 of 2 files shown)


rathboma:
oh looks like there's a conflict -- want to resolve it, then I'll merge!

geovannimp:
I did rebase with master and made some improvements at the code.

geovannimp:
One question, how could we handle nullable fields?

src/components/tableview/TableTable.vue
```diff
@@ -318,6 +323,13 @@ export default {
       }
       return null
     },
+    editorType(dt) {
```
UtechtDustin, line 326: maybe its just me, but i think `dataType` would be better then `dt`.

src/components/tableview/TableTable.vue
```diff
@@ -318,6 +323,13 @@ export default {
...
+      switch (dt) {
+        case 'text': return 'textarea'
```
UtechtDustin, line 328: I locally have a branch for that featureset (select for bool, textarea for text and datepicker for date(time)).
But i was stuck on the textarea, does it works for you ? (screen/video/gif would be great)
For me the size was buggy.

src/components/tableview/TableTable.vue
geovannimp: Agree, just followed `slimDataType` method style

src/components/tableview/TableTable.vue
geovannimp: I didn't tested textarea as it was already implemented, just added support to boolean with select

UtechtDustin: Oh okay, i check your changes a second time.
I thought the `editorType` function will return the type which will be used for `tabulator`.

rathboma: @UtechtDustin I fixed textarea to work properly

src/components/tableview/TableTable.vue
```diff
@@ -210,8 +211,12 @@ export default {
           cellEditCancelled: cell => cell.getRow().normalizeHeight(),
           formatter: (cell) => _.isNil(cell.getValue()) ? '(NULL)' : cell.getValue(),
           editorParams: {
-            verticalNavigation: useTextarea ? 'editor' : undefined,
+            verticalNavigation: useVerticalNavigation ? 'editor' : undefined,
             search: true,
+            ...(column.dataType === 'bool'
+              ? { values: ['true', 'false'] }
```
rathboma, line 217: should these have quotes around them, or just be `[true, false]`?

rathboma: We probably need an entry in `data_mutators.js` for booleans to not turn them into strings too

geovannimp: I think tabulator accept only strings as values. Btw it's working.

PR #303, opened 2020-08-26
Tabulator redraw doesn't seem to work

rathboma:
1. Run a query
2. Scroll the result table down a bit
3. Change tab
4. Change back
EXPECTED: you can still see the result table

ACTUAL: Result table is blank.

Workaround: sort a column, or re-run the query (column resize doesn't work)

After scroll, the `tabulator-table` style is:

```css
padding-top: 954px; padding-bottom: 14448px;
```
Change to another tab, the style is:
```css
padding-top: 954px; padding-bottom: 14448px; min-width: 414px; min-height: 1px; visibility: hidden;
```

Switching back to the tab it becomes:
```css
padding-top: 2178.68px; padding-bottom: 13149.3px; min-height: 1px; visibility: hidden;
```

Notice the padding-top getting really large. Hmm

### DIFF
--- a/src/components/CoreTabs.vue
+++ b/src/components/CoreTabs.vue
@@ -27,8 +27,8 @@
         :key="tab.id"
         :class="{show: (activeTab === tab), active: (activeTab === tab)}"
       >
-        <QueryEditor v-if="tab.type === 'query'" :active="activeTab == tab" :tab="tab" :tabId="tab.id" :connection="connection"></QueryEditor>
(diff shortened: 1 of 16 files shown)


rathboma:
I think I may have fixed it by disabling redraw

src/components/CoreTabs.vue
```diff
@@ -56,6 +56,9 @@
         activeItem: 0,
         newTabId: 1
       }
+    },
+    watch: {
+
```
UtechtDustin, line 61: empty watch ? :D

rathboma: it's a feature!

Issue #301, opened 2020-08-22
Typescript at source code


geovannimp:
Any reason to don't use typescript at source code?
If typescript is desired I could try to make a PR migrating at least the core to TS.

geovannimp:
I understand the fear after CoffeeScript, but I will leave here a few things to consider:
- Typescript is a superset of js, so every javascript code is compatible with typescript
- Every Typescript feature follows ecmascript standard, so any lang feature is only released after been approved to enter in the next js release.
- Think like typescript was babel + types
- Many big projects like VSCode use typescript or are migrating the source code at some level

I did some refactor already to understand the code and see how hard would be to migrate, so I will leave here sobre examples of before/after:

**Components**
![Sem título](https://user-images.githubusercontent.com/4943174/90986006-293c7e80-e556-11ea-9269-4a33b3b11a17.png)

**Database entity (with some error I found with typescript)**
![Sem título2](https://user-images.githubusercontent.com/4943174/90986218-7e2cc480-e557-11ea-9156-3e1f3eeac795.png)

If you want we could keep talking about it here or another place :)

PR #304, opened 2020-08-27
Move core to typescript

geovannimp:
Issue: https://github.com/beekeeper-studio/beekeeper-studio/issues/301

When I open the PR I had already started refactoring, as it was fun, here you have the complete refactor.

Based in @rathboma answer, I didn't touched any .vue file and I'm totaly open to modify the PR to remove any unwanted refactored file.

I placed some TODOs with "bugs" I found when doing the refactor, I think this already prove typescript benefits.

There is a lot more that could be refactored, but the PR would be much bigger and have a lot more conflicts.

Please, let me know what you guys think about and sorry for the PR size.


rathboma:
Hey!

Wow, this must have been a lot of work.

Not 100% sure what I think about moving so much to typescript. I certainly see some benefits, especially for stuff in 'lib', but I do worry about the accessibility of the project to new contributors.

Let me noodle on it a few days.

geovannimp:
No problem, take the time you want.

I did all refactor trying to keep closer to JS as possible, only adding types. So accessibility shouldn't be too hard.

It take some time, but was cool to understand the application architecture. This is something that I think typescript make more accessible because before the refactor I had no idea what type of "variable" was present at store state `connectionConfigs` for example.

rathboma:
@geovannimp Ok. So I just had to solve a bug that would have been caught with typescript (broken auto-updates).

So... I was going to say 'not now', but dammit lets do it. I think it could add some stability and will let me make the db libraries a lot more class-based. 

Questions:
- Can you resolve the last couple of merge conflicts?
- Are there any other things need to be done before this can for real be merged?

geovannimp:
- Can you resolve the last couple of merge conflicts?
	Of course, I will try to do it today or tomorrow.

- Are there any other things need to be done before this can for real be merged?
	- I tested some features and they are all working, but some test by the maintainer would make me more confident.
	- Could you take a look at the TODOs I placed at the code? Those are bugs I found that can be at master or something that I refactored wrong.

geovannimp:
@rathboma Rebase is done

rathboma:
cool, I'll look through the TODOS over the next few days

src/lib/menu/ClientMenuActionHandler.ts
```diff
@@ -0,0 +1,67 @@
+import { ipcRenderer } from 'electron'
+import _ from 'lodash'
+import AppEvent from '../../common/AppEvent'
+
+function send(name: string, arg?: any) {
+  ipcRenderer.send(AppEvent.menuClick, name, arg)
+}
+
+export interface IClientMenuActionHandler {
```
rathboma, line 9: Can you combine this with an interface shared by this and the NativeMenuActionHandler? The interface can go in common/

src/main.ts
```diff
@@ -36,17 +38,19 @@ import VueClipboard from 'vue-clipboard2'
     log.info("starting logging")
     tls.DEFAULT_MIN_VERSION = "TLSv1"
     TimeAgo.addLocale(en)
+    // @ts-ignore
+    // TODO: Check cause I dint find documentation
```
rathboma, line 42: this is fine

src/shims-tsx.d.ts
```diff
@@ -0,0 +1,13 @@
+import Vue, { VNode } from 'vue'
+
+declare global {
```
rathboma, line 3: What is this for?

src/store/index.ts
```diff
@@ -0,0 +1,346 @@
...
+          })
+
+          const materialized = await context.state.connection.listMaterializedViews({schema: null})
+          materialized.forEach(v => v.entityType = 'materialized-view')
+          const tables = onlyTables.concat(views).concat(materialized)
+          context.commit("tablesLoading", `Loading ${tables.length} tables`)
+
+          // TODO: I think a the table parameter is missing
+          // @ts-ignore
+          const tableColumns = await context.state.connection.listTableColumns()
```
rathboma, line 247: I've coded this to work with or without a table. If no table - it fetches all columns for all tables. So this TODO is fine

src/store/index.ts
```diff
@@ -0,0 +1,346 @@
...
+          for (let index = 0; index < materialized.length; index++) {
+            const view = materialized[index]
+            const columns = await context.state.connection.listMaterializedViewColumns(view.name, view.schema)
+            viewColumns = viewColumns.concat(columns)
+          }
+
+          const allColumns = tableColumns.concat(viewColumns)
+
+          tables.forEach((table) => {
+            // TODO: Check the following 2 lines case is not been used
```
rathboma, line 258: All, good, thanks.

tsconfig.json
```diff
@@ -0,0 +1,44 @@
+{
+  "compilerOptions": {
+    "target": "es2018",
```
rathboma, line 3: I don't really know anything about this file, but some naive questions:

es2018 - this works in browsers?
Any thing I specifically need to know about here?

rathboma (changes requested):
Hey, answered all your TODO questions. I think there are a few minor tweaks to make -- you ok making them?

I'm still a little nervous about all the extra stuff I'll need to learn, but excited to add some more stability to the library code.

src/common/appdb/Connection.ts
```diff
@@ -34,11 +33,11 @@ export default class Connection {
       subscriptions: [
```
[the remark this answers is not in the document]
geovannimp: I checked and passwords has been saved without encryption. I will change to subscribers and check if it resolves the issue, but maybe we will need some migration to encrypt already saved connections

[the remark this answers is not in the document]
geovannimp: I had to do the migration to fix this, now passwords are being saved encrypted at the database. I will open another PR because its a "critical" bug and can be merged already without typescript.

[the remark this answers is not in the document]
geovannimp: https://github.com/beekeeper-studio/beekeeper-studio/pull/322

src/common/appdb/models/saved_connection.ts
```diff
@@ -56,22 +57,22 @@ export class DbConnectionBase extends ApplicationEntity {
   }
 
   @Column({type:"varchar", nullable: true})
-  host = 'localhost'
+  host: string = 'localhost'
 
   @Column({type: "int", nullable: true})
-  port
+  port: Nullable<number> = null
```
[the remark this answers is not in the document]
geovannimp: `port?: number` allows only `number` and `undefined`, but in many places of the code is used `someVaraible = null` and I would need to change every place to `someVaraible = undefined` so I created this `Nullable<>` type to keep code standard.

In my option, refactor all the code to remove those `Nullable<>` and use `someVaraible?` should be the first thing to do after this being merged at master.

src/common/appdb/models/saved_connection.ts
```diff
@@ -80,21 +81,25 @@ export class DbConnectionBase extends ApplicationEntity {
...
-  set sshMode(value) {
+  set sshMode(value: string) {
     this._sshMode = value
-    if (!this._sshMode != 'userpass') {
-      this.sshPassword = null
-    } else if (this._sshMode != 'keyfile') {
+    // TODO: This code dont make sense, have to be refactored
+    // @ts-ignore
+    if (!this._sshMode !== 'userpass') {
+      // TODO: sshPassword exist only at SavedConnection, it should moved here or this method moved to SavedConnection
```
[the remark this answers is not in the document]
geovannimp: I moved, but still a TODO.

`if (!this._sshMode !== 'userpass') {`

`!this._sshMode` return a `boolean` and `'userpass'` is a `string`, so this code always return true.

src/shims-tsx.d.ts
geovannimp: This is create automatically when I convert the project to typescript. It can be removed I think (or moved to typings) but dont cause any harm.

src/store/index.ts
geovannimp: But can I remove those lines?

Because `query` is never used.

tsconfig.json
geovannimp: I think we could use some lower target, but this is the default when converting the project to typescript and the code seems to be working. With a higher target we will have a transpiled JS code closest to typescript without crazy workarounds.

src/lib/menu/ClientMenuActionHandler.ts
geovannimp: Do you mean something like this?

```
type IMenuActionHandler = IClientMenuActionHandler | INativeMenuActionHandler
```

But what is the diference between `IClientMenuActionHandler` and `INativeMenuActionHandler`?

src/common/appdb/Connection.ts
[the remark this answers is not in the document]
rathboma: you rock!

geovannimp:
Fixed almost all TODOs.

I commented every change request/doubt and I'm waiting answers or to be marked as resolved.

And I think you missed some TODOs, so I placed my nickname in all remaining TODOs to be easier to find.

src/common/appdb/models/saved_connection.ts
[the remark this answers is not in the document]
rathboma: oh I see. Hmm I'll look into that.

src/store/index.ts
rathboma: Yep!

src/common/appdb/models/saved_connection.ts
```diff
@@ -217,13 +196,38 @@ export class SavedConnection extends DbConnectionBase {
...
+  sshPassword: Nullable<string> = null
+
+
+
+  _sshMode: string = "agent"
+
+  @Column({name: "sshMode", type: "varchar", length: "8", nullable: false, default: "agent"})
+  set sshMode(value: string) {
+    this._sshMode = value
+    // TODO: (geovannimp) This code dont make sense, have to be refactored
```
rathboma, line 208: Yes, I think this should just be:

`if(this._sshMode !== 'userpass')`

IE - clear the SSH password if not authenticating with password

src/lib/menu/ClientMenuActionHandler.ts
rathboma: Ah, yea I wasn't clear here. There's no difference! They should use the same interface

Both NativeMenuActionHandler and ClientMenuActionHandler have the same interface. So there should be 3 files:

```
common/.../IMenuActionHandler
app/.../ClientMenuActionHandler (implements IMenuActionHandler)
background/.../NativeMenuActionHandler (implements IMenuActionHandler)
```

src/store/index.ts
```diff
@@ -0,0 +1,344 @@
...
+    async fetchUsername(context) {
+      const name = await username()
+      context.commit('setUsername', name)
+    },
+
+    async connect(context, config: SavedConnection) {
+      if (context.state.username) {
+        const server = ConnectionProvider.for(config, context.state.username)
+        // TODO: (geovannimp) Check case connection is been created with undefined as key
+        const connection = server.createConnection(config.defaultDatabase || undefined)
```
rathboma, line 188: I think this is ok?

src/store/index.ts
```diff
@@ -0,0 +1,344 @@
...
+            const view = materialized[index]
+            const columns = await context.state.connection.listMaterializedViewColumns(view.name, view.schema)
+            viewColumns = viewColumns.concat(columns)
+          }
+
+          const allColumns = tableColumns.concat(viewColumns)
+
+          tables.forEach((table) => {
+            // TODO: (geovannimp) Check the following 2 lines case is not been used
+            // const query = { tableName: table.name }
```
rathboma, line 257: Yes these lines can be removed.

rathboma:
ok, added more comments to the TODOS

src/shims-tsx.d.ts
rathboma: kk, we can just leave it

src/common/appdb/models/saved_connection.ts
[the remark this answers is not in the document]
rathboma: Yeah that shoudn't have the first `!`. Typo.

src/common/appdb/models/saved_connection.ts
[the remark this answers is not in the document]
rathboma: I think typeorm is fine with it being either undefined or null and setting to null? Is the problem that we're using both undefined and null in different places?

tsconfig.json
MasterOdin: You could go as high as `es2019` given that beekeeper uses chromium 80 (through electron 8.2.5) and a minimum node version of 12. Unless you wanted to support a lower version of node or electron, there is no reason to not make this as high as possible, using a combination of https://kangax.github.io/compat-table/es2016plus/ and https://node.green/ to see what features the environment matches.

src/common/appdb/models/saved_connection.ts
geovannimp: Ok, but with this the `else if` don't work.

If `_sshMode` is not different from `"userpass"`, that means that `_sshMode` is `"userpass"` and cant be `"keyfile"`.

Typescript give me the following erro: 

```
This condition will always return 'true' since the types '"userpass"' and '"keyfile"' have no overlap.ts(2367)
```

Btw, I changed the if/else to only else. Only confirm to me that I catch this right.

src/common/appdb/models/saved_connection.ts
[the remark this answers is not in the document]
geovannimp: Is not a problem, but with undefined with we do things like someObject?.propertie and with null we can't do this.

And to make type correctly based in what typeorm accepts, the type should be `port?: Nullable<number>` and we would need to check every place by null and undefined. So is better to use only null or only undefined, and undefined is more JS friendly.

tsconfig.json
geovannimp: I would stay this value for now, based that this is the default. But after this being merged we can play with this value for tests and check if we have any gain with higher/lower values.

src/store/index.ts
geovannimp: The problem is this:
```
if (server.db[dbName]) {
```

As dbName can be `undefined`, this can happen: `server.db[undefined]`.

It's actually working I think, but typescript complains about using `undefined` as key.

```
Type 'undefined' cannot be used as an index type.ts(2538)
```

We can keep this way for now, but would be good to refactor any time.

geovannimp:
Cool, I fixed every change request I think.

Still 3 TODOs at `lib/db/client.ts`

src/lib/db/client.ts
```diff
@@ -0,0 +1,419 @@
...
+function listDatabases(server: IDbConnectionServer, database: IDbConnectionDatabase, filter: IDbFilter) {
+  checkIsConnected(server , database);
+  return database.connection?.listDatabases(filter);
+}
+
+
+async function getQuerySelectTop(server: IDbConnectionServer, database: IDbConnectionDatabase, table: string, schema: string, limit: number) {
+  checkIsConnected(server , database);
+  let limitValue = limit;
+  // TODO: (geovannimp) _limit is not declared
```
rathboma, line 322: I don't use this function, it's a leftover, so this is fine

rathboma: IE - don't worry about this function at all. You could just remove it if you like.

src/lib/db/client.ts
```diff
@@ -0,0 +1,419 @@
...
+    if (database.connection) {
+      database.connection.disconnect();
+    }
+
+    // reuse existing tunnel
+    if (server.config.ssh && !server.sshTunnel) {
+      logger().debug('creating ssh tunnel');
+      server.sshTunnel = await connectTunnel(server.config);
+
+      // TODO: (geovannimp) address and port dont exist at sshTunnel
```
rathboma, line 187: good catch. This is a leftover from a refactor. Totally fine to just omit this.

rathboma (commented):
I think these are all fine to ignore honestly.

rathboma:
ok, github hid that diff from me, so I didn't see them.

Added comments -- basically we can ignore those if we want.

rathboma:
@geovannimp One merge conflick, then I think we're good to go \0/

geovannimp:
All done! :D

PR #308, opened 2020-08-29
Added the option to remove (multiple) saved queries and queries from the history

### LINKED ISSUE #305: FEAT: Button to clear history of queries or saved queries
An option that allows you to delete queries you no longer need, either individually or as a group

UtechtDustin:
Added the option to remove (multiple) saved queries and queries from the history

Fix #305

Peek 2020-08-30 01-52

### DIFF
--- a/src/components/sidebar/core/FavoriteList.vue
+++ b/src/components/sidebar/core/FavoriteList.vue
@@ -19,10 +19,10 @@
           </x-button>
         </a>
       </div>
-      <div class="toolbar" v-show="this.checkedFavorites.length > 0">
+      <div class="toolbar" v-show="checkedFavorites.length > 0">
         <div class="flex flex-right">
           <a class="btn btn-link" @click="discardCheckedFavorites">Discard</a>
-          <a :title="this.removeTitle" class="btn btn-primary" @click="removeCheckFavorites">Remove</a>
+          <a :title="removeTitle" class="btn btn-primary" @click="removeCheckedFavorites">Remove</a>
         </div>
       </div>
     </nav>
@@ -53,16 +53,15 @@
       async remove(favorite) {
         await this.$store.dispatch('removeFavorite', favorite)
       },
-      async removeCheckFavorites() {
-        this.checkedFavorites.forEach(favorite => {
-          this.remove(favorite)
-        })
+      async removeCheckedFavorites() {
+        for(let i = 0; i < this.checkedFavorites.length; i++) {
+          await this.remove(this.checkedFavorites[i])
+        }
         this.checkedFavorites = [];
       },
       discardCheckedFavorites() {
         this.checkedFavorites = [];
       }
     }
-
   }
 </script>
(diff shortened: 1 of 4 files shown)


UtechtDustin:
@rathboma 
It's working fine for the saved queries, so i tried to implement the same for the quries in the history but i get the error
```
 Property or method "discardCheckedHistoryQueries" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property.
```
I dont understand that error, because there is a method called `discardCheckedHistoryQueries`, also its working for the saved queries.
Could you help with that, did i miss something ?!

rathboma:
Sure I'll take a look. @ghstng can you do a ui review?

ghstng:
I can pull into a branch and cleanup, if you both are good with everything.

UtechtDustin:
Its fine for me, im not a UI expert :-)

rathboma:
Sure then I can fix the vue problem

UtechtDustin:
> Sure then I can fix the vue problem

Thanks, please then tell me what was the issue was :-)

rathboma:
@UtechtDustin Your bug is that `discardCheckedHistoryQueries` is outside of the `methods {}` block, simply move it inside `methods`, and it will work (also do that for the other function you added)

UtechtDustin:
> @UtechtDustin Your bug is that `discardCheckedHistoryQueries` is outside of the `methods {}` block, simply move it inside `methods`, and it will work (also do that for the other function you added)

Oh... How i can't saw that.. Sometimes the small issues are the hardest one..

rathboma:
If you fix that little bug we'll merge it and do a style tidy up 👍

src/components/sidebar/core/FavoriteList.vue
```diff
@@ -28,15 +35,32 @@
...
     methods: {
       click(item) {
         this.$root.$emit('favoriteClick', item)
       },
       async remove(favorite) {
         await this.$store.dispatch('removeFavorite', favorite)
+      },
+      async removeCheckedFavorites() {
+        this.checkedFavorites.forEach(favorite => {
+          this.remove(favorite)
```
rathboma, line 58: Should you have an `await` here? You will need to do a for() loop to do that

src/components/sidebar/core/HistoryList.vue
```diff
@@ -35,7 +57,18 @@
           return text
         }
       }
+    },
+    async remove(historyQuery) {
+      await this.$store.dispatch('removeHistoryQuery', historyQuery)
+    },
+    async removeCheckedHistoryQueries() {
+      this.checkedHistoryQueries.forEach(historyQuery => {
+        this.remove(historyQuery)
```
rathboma, line 66: same deal with `await` here

rathboma (changes requested):
minor change - do an `await` wherever you can

UtechtDustin:
Lets wait for @ghstng for the design then i will look at this :-)

rathboma:
We chatted -- the ux is mostly good, we're just going to tweak a few things after it's merged -- it's hard to do that with someone else's branch

UtechtDustin:
Oh okay, i pushed a new version of that.
Currently i have one last problem (except from the design), if i remove saved query somehow ALL queries will be removed from the list but no one will be removed from the store.
I try to figure out what the problem is, but if you have time it would great if you also could look.

UtechtDustin:
@rathboma its ready for merge

PR #309, opened 2020-08-30
Other DB for dev

UtechtDustin:
This PR will change the db which will be used for the development.

My problem right now is that i try to help as good i can, but i also want to use beekeeper for my project's.
This would mean i always have a "dirty" connection list.


UtechtDustin:
@rathboma I pushed a new way how to figure out the env.
I use the prop [isPackaged](https://www.electronjs.org/docs/api/app#appispackaged-readonly) from electron.
```
A Boolean property that returns true if the app is packaged, false otherwise.    
For many apps, this property can be used to distinguish development and production environments.
```

rathboma:
Cool. Never seen that before

UtechtDustin:
Is it ready to merge, or is there something to do ?

src/common/platform_info.ts
```diff
@@ -24,14 +25,15 @@ const platformInfo = {
...
+  isDevEnv: isDevEnv,
   sshAuthSock: p.env.SSH_AUTH_SOCK,
   environment: process.env.NODE_ENV,
   debugEnabled: !!process.env.DEBUG,
   platform: easyPlatform,
   darkMode: testMode? true : e.nativeTheme.shouldUseDarkColors || windowPrefersDarkMode,
   userDirectory,
   testMode,
-  appDbPath: path.join(userDirectory, 'app.db'),
+  appDbPath: path.join(userDirectory, isDevEnv ? 'app.db' : 'app-dev.db'),
```
rathboma, line 36: isn't this backwards? isDevEnv === isPackaged, but shouldn't it be reversed?

UtechtDustin: your right, fixed!

src/common/platform_info.ts
```diff
@@ -24,14 +25,15 @@ const platformInfo = {
   isSnap: p.env.ELECTRON_SNAP,
   isDevelopment: process.env.NODE_ENV !== 'production',
   isAppImage: p.env.DESKTOPINTEGRATION === 'AppImageLauncher',
+  isDevEnv: isDevEnv,
```
rathboma, line 28: You should update `isDevelopment` here instead of making a new key. Checking `isPackaged` is better than checking the NODE_ENV

src/main.ts
```diff
@@ -40,7 +41,7 @@ import VueClipboard from 'vue-clipboard2'
     TimeAgo.addLocale(en)
     // @ts-ignore
     Tabulator.prototype.defaultOptions.layout = "fitDataFill";
-    const appDb = path.join(config.userDirectory, 'app.db')
+    const appDb = path.join(config.userDirectory, platformInfo.isDevEnv ? 'app-dev.db' : 'app.db')
```
rathboma, line 44: please just make this look at `platformInfo.appDbPath`

src/migration/dev-1.js
```diff
@@ -48,7 +48,7 @@ export default {
     {
       name: "Beekeeper's Database",
       connectionType: "sqlite",
-      defaultDatabase: path.join(platformInfo.userDirectory, 'app.db')
+      defaultDatabase: path.join(platformInfo.userDirectory, platformInfo.isDevEnv ? 'app-dev.db' : 'app.db')
```
rathboma, line 51: Please just have this look at platformInfo.appDbPath.

We should def have a single place we define this.

rathboma (changes requested):

rathboma:
@UtechtDustin let me know if you have time to change these few final things!

UtechtDustin:
@rathboma I will do it today :-)

UtechtDustin:
@rathboma reminder :-)

rathboma:
Hehe. I'm super tight on time right now but I'll try to look tonight

PR #312, opened 2020-08-31
Added the option to use a socket

### LINKED ISSUE #300: feature: Support database connections using a socket path rather than a host and port
It looks like there are hints at connecting directly with a socket path https://github.com/beekeeper-studio/beekeeper-studio/blob/9ff3131030057364f31f3034677fc0f6a9bd48b5/src/lib/connection-provider.js#L23 

https://github.com/beekeeper-studio/beekeeper-studio/blob/2521b8c66721c7e634a006a992bad4e4f1f6153e/src/lib/db/server.js#L22

But there does not appear to be anything in the UI to indicate how you'd go about doing this. Nor where one might be able to create their own config file to manually define this.

UtechtDustin:
This PR will add an new option which allows connecting to a database using a socket.

Fix #300

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -35,6 +35,10 @@ export default {
       { name: 'Amazon Redshift', value: 'redshift' },
       { name: 'CockroachDB', value: 'cockroachdb' }
     ],
+    connectionMethods: [
+      { name: 'Default', value: 'default' },
+      { name: 'Socket', value: 'socket' }
+    ]
   },
   maxResults: 50000
 }
--- a/src/lib/db/server.js
+++ b/src/lib/db/server.js
@@ -19,7 +19,7 @@ export function createServer(serverConfig) {
 
     config: {
       ...serverConfig,
-      host: serverConfig.host || serverConfig.socketPath,
+      host: serverConfig._connectionMethod === 'default' ? serverConfig.host : serverConfig.socketPath,
     },
   };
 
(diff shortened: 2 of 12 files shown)


src/config.js
```diff
@@ -35,6 +35,10 @@ export default {
       { name: 'Amazon Redshift', value: 'redshift' },
       { name: 'CockroachDB', value: 'cockroachdb' }
     ],
+    connectionMethods: [
```
rathboma, line 38: we could move this to the `saved_connection` and have a `supportsSocket` getter that checks if it is mysql, mariadb, or psql?

src/lib/db/server.js
```diff
@@ -19,7 +19,7 @@ export function createServer(serverConfig) {
 
     config: {
       ...serverConfig,
-      host: serverConfig.host || serverConfig.socketPath,
+      host: serverConfig._connectionMethod === 'default' ? serverConfig.host : serverConfig.socketPath,
```
rathboma, line 22: I don't think _connectionMethod is available here, I think this is the config provided by `connection_provider.js`

rathboma:
Let me know if you have time to do the slight refactor here! Otherwise probably won't merge until I have time to work on it (I'm partially homeschooling my kid for September as well as full time working :-S)

UtechtDustin:
I guess this PR is easier to reimplement (after the core moved to typescript) instead of rebasing it.

Issue #197, opened 2020-06-11
Settings options for settings screen [please help contribute]


rathboma:
Right now we're not building a settings screen, because we have no need for it.

However we will build one when we have a lot of things to put in it.

Current ideas:

- Connection timeout
- SSH Timeout
- Firewall (for app updates)
- Theme
- Font size
- Interface zoom level

GammaGames:
Would editing shortcuts be on this screen too? It would be nice to be able for users to change the run or new query shortcuts.

And the ability to change the default editor contents would be nice. I don't want 10 blank lines, but defaulting to a blank `Select ... From` would be nice

seisenreich:
A zoom functionality would be great. Even when no settings screen is available it would be fantastic to be able to zoom with `Ctrl/Cmd` + `+`  or zoom out with `Ctrl/Cmd` + `-`, similar to how web browsers do it.
Also, you could show the current installed version of the app in this screen.

rathboma:
Zoom is coming soon.

I like the default query idea, and about screen, but they won't make it into v1.

chupetoide:
Hi guys. Is there any way to change the sql editor font family ?

fellipegpbotelho:
Any update regarding on change font family?

bobhenkel:
This tool is great keep up the awesome work!

Please add changing the font and I'll upgrade to the payed version the second that feature lands.

Thanks!

martinschilliger:
And about settings screen: Ghostty just opens the settings file if I press `CMD + ,` (default Keybinding for Settings on macOS), maybe this would be an easy solution too? Or your website about configurations: https://docs.beekeeperstudio.io/user_guide/configuration/

It took me quite some time to find out about it, as I searched all over the UI for the settings.

PR #319, opened 2020-09-05
Change openTable to only focus table when already opened

geovannimp:
Peek 04-09-2020 22-33

### DIFF
--- a/src/components/CoreTabs.vue
+++ b/src/components/CoreTabs.vue
@@ -144,15 +144,22 @@
         if (!table && tableName) {
           resolvedTable = this.$store.state.tables.find(t => t.name === tableName)
         }
-        const t = {
-          id: uuidv4(),
-          type: 'table',
-          table: resolvedTable || table,
-          connection: this.connection,
-          initialFilter: filter,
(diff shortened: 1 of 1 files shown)


UtechtDustin:
Does this make sense, what if you want multiple tabs with the same table for filtering multiple stuff ?
I understand that this makes sense for some of us, but i guess this should be a option not the default (or the other way around).

#197

geovannimp:
If some filter has been made, it opens a new tab.
Peek 05-09-2020 04-21

UtechtDustin:
Okay then it makes absolute sense!
I like it!

geovannimp:
I made another PR to open the table with double click.
Maybe double click could focus if already opened and clicking at this button always open a new one.

rathboma:
I've played with this before and decided not to add this behavior. I think in some ways it makes sense, but having multiple behaviors for the same button (focus vs open) based on an unknown context can be confusing.

I think I'd prefer not to do this and keep tab-opening actions fairly simple.

In the future I want to have a global-search which would allow you to jump to an open tab or open a new tab

PR #320, opened 2020-09-05
Add double click action to list-item-btn at TableListItem

geovannimp:
As we open a database with double click, I think open a table should be the same.

### DIFF
--- a/src/components/sidebar/core/table_list/TableListItem.vue
+++ b/src/components/sidebar/core/table_list/TableListItem.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="list-item">
-    <a class="list-item-btn" role="button" v-bind:class="{'active': selected,'open': showColumns }" @mousedown.prevent="toggleColumns">
+    <a class="list-item-btn" role="button" @dblclick="openTable()" v-bind:class="{'active': selected,'open': showColumns }" @mousedown.prevent="toggleColumns">
       <span class="btn-fab open-close" >
         <i class="dropdown-icon material-icons">keyboard_arrow_right</i>
       </span>


rathboma:
The problem is that is also double-triggers the mousedown event, so it would need to be handled at that level I think (I've tried to add it before, but didn't due to that)

geovannimp:
I understand, but personally I prefer to have the accordion to open/close than have to click in a little button. (That's my friend opinion too btw)

I think the primary use of the application is to search at tables, so maybe the first action should be the search and the button could expand the accordion.

I saw at issues that this is a old discussion and when configurations are done we could have this configurable. But in the meantime, do we have any platform feedback to community? Something like a twitter we could ask for community what do they expect from this or another features.

UtechtDustin:
I agree @geovannimp, almost every database tools handle it that way, because it makes sense.

In my opinion the table structure can be removed from the table list view. It would make more sense to create a own view for that, where you also can edit the table structure as well.

rathboma:
@UtechtDustin the table structure is there for those writing SQL (in my case this is my primary use-case for Beekeeper Studio and most of the UX is built around this assumption).

I think honestly we haven't really re-evaluated the sidebar behavior with table-tabs in mind, but I agree double-click should open the table view. It may be we want an explicit arrow for expanding, but we haven't decided yet.

We should also have it in a right-click menu, which I will add as soon as we have a 'table details' view (with columns, indexes, ddls, etc)

If you can get it so that double click doesn't re-close the structure dropdown, then I'll merge it. It may be as easy as adding a `stop` to the event, or you might need to handle the double click inside the `mousedown` event handler.

rathboma:
@geovannimp you up for seeing if you can get the double click to not close the table structure dropdown? If you can get it doing that I will merge it!

geovannimp:
I will try, but my Vue knowledge is really superficial.

geovannimp:
Btw, @rathboma You need some help with 'table details' implementation?

UtechtDustin:
```
<a class="list-item-btn" role="button" v-bind:class="{'active': selected,'open': showColumns }" @mousedown.prevent="clickTableName">
```
and
```
clickTableName() {
        const that = this;
        if (this.tableNameClickTimeout !== undefined) {
          clearTimeout(this.tableNameClickTimeout);
          this.tableNameClickTimeout = undefined;
          this.openTable();
          return;
        }

        this.tableNameClickTimeout = setTimeout(() => {
          that.toggleColumns()
          that.tableNameClickTimeout = undefined;
        },200)
      },
```
works fine for me.
Peek 2020-09-15 21-55

rathboma:
@geovannimp Yes I would! Thanks for offering! I'll make an issue and outline what I think we need on there.

rathboma:
closing in favor of #336

PR #322, opened 2020-09-07
Bug/connection passwords being saved without encryption

geovannimp:


src/migration/20200907-add_encrypted-columns.js
```diff
@@ -0,0 +1,48 @@
...
+          isNullable: true
+        }),
+        new TableColumn({
+          name: 'sshPassword_decrypted',
+          type: 'varchar',
+          isNullable: true
+        }),
+      ]
+
+      await runner.startTransaction()
```
rathboma, line 26: this might not work (rollback correctly) because sqlite cannot remove columns? Not sure.

rathboma: I think it's best to break this out into two migrations:
1. Create new columns
2. run the update queries (with transaction)

rathboma (changes requested):
I'm still actually not sure if it's worth encrypting this stuff or just treating the user environment as trusted, but let's go with my original intent!

Good find here with your typescript changes!

I've added a few comments based on the unfortunate limitations of sqlite.

src/migration/20200907-add_encrypted-columns.js
```diff
@@ -0,0 +1,48 @@
...
+
+      const entities = await SavedConnection.find();
+      entities.forEach(e => {
+        e.password = connections.find(c => c.id == e.id).password_decrypted
+        e.sshKeyfilePassword = connections.find(c => c.id == e.id).sshKeyfilePassword_decrypted
+        e.sshPassword = connections.find(c => c.id == e.id).sshPassword_decrypted
+      });
+      SavedConnection.save(entities);
+
+      await runner.dropColumns('saved_connection', temp_columns)
```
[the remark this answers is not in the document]
geovannimp: I can test it again, but I'm almost sure that the columns were dropped.

I tested here again and the columns were dropped. Typeorm must have some tricks to remove the columns.

geovannimp:
@rathboma I looked at de migration again and we don't really need a temp collumn. Take a look at the migration again, it's a lot simpler now

rathboma:
Nice

src/migration/20200907-add_encrypted-columns.js
```diff
@@ -0,0 +1,27 @@
...
+      await runner.query('UPDATE saved_connection SET password = NULL, sshKeyfilePassword = NULL, sshPassword = NULL;')
+
+      const entities = await SavedConnection.find();
+      entities.forEach(e => {
+        e.password = original_connections.find(c => c.id == e.id).password
+        e.sshKeyfilePassword = original_connections.find(c => c.id == e.id).sshKeyfilePassword
+        e.sshPassword = original_connections.find(c => c.id == e.id).sshPassword
+      });
+
+      SavedConnection.save(entities);
```
rathboma, line 19: I think you need an `await` here!

geovannimp: Yeap, fixed! :D

rathboma:
tested. Works great. Legend

Issue #327, opened 2020-09-12
BUG: Say something here


thiagola92:
**Describe the bug**
Quoting SQL keywords and wrong quoting columns and table names.

```sql
CREATE TABLE "ASDF.A.B" ("a.b" text)
CREATE TABLE asdf ("a.b" text)
CREATE TABLE "asdf.a" (asdf text)
CREATE TABLE "texto com espaco" (abc text)

-- Try autocomplete SELECT
-- expectation: SELECT
-- outcome: "SELECT"
SEL

-- Try autocomplete table name with dot or space or capital letter
-- expectation: "ASDF.A.B"
-- outcome: "ASDF"."A"
SELECT * FROM A

-- expectation: "texto com espaco"
-- outcome: texto com espaco
SELECT * FROM tex

-- Try autocomplete column name with dot
-- expectation: asdf."a.b"
-- outcome: asdf.a
SELECT asdf.

-- Try autocomplete column name after quoted table name
-- expectation: "asdf.a".asdf
-- outcome: no result
SELECT "asdf.a". FROM "asdf.a"
```

**To Reproduce**
1. Connect to **postgres** database
2. Run the first 4 lines from the code above
3. Try autocomplete the code

**Expected behavior**
Expected not to quote keywords  
Expected quote only table/column name that have dot, capital letter or space
Expected to find the column name after quoted table name

**Screenshots / Gifs**
https://www.youtube.com/watch?v=6_1GUrd7zCA

**Version Information (please complete the following information):**
 - OS: Ubuntu 20.04
 - App Version 1.7.5

thiagola92:
On postgres when you want your table/column name to have dot/capital letter/space you have to quote it.  
`CREATE TABLE "EXAMPLE"(name text)`  
`CREATE TABLE "example.two"(name text)`  
`CREATE TABLE "example three"(name text)`  

Right now beekeeper try to decide if need to quote or not the text received from CodeMirror.  
https://github.com/beekeeper-studio/beekeeper-studio/blob/c13ee7274d61741a451a127559f3e8c75e37c7db/src/components/TabQueryEditor.vue#L526-L535  

The problem start when CodeMirror shows table/columns names that should have quote, without quote on the autocomplete  
https://github.com/beekeeper-studio/beekeeper-studio/blob/c13ee7274d61741a451a127559f3e8c75e37c7db/src/components/TabQueryEditor.vue#L222-L234  

[...]

PR #328, opened 2020-09-12
Fix autocomplete

thiagola92:
## Propose to fix autocomplete problems #327  

Still doesn't complete when the name is on quotes  
```sql
SELECT "asdf.a". -- no suggestion appears
```

### DIFF
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -227,7 +227,7 @@
           })
 
           // add quoted option for everyone that needs to be quoted
-          if (this.connectionType === 'postgresql' && /[^a-z0-9_]/.test(table.name))
+          if (this.connectionType === 'postgresql' && (/[^a-z0-9_]/.test(table.name) || /^\d/.test(table.name)))
             result[`"${table.name}"`] = cleanColumns
           
           // don't add table names that can get in conflict with database schema 
@@ -531,9 +531,9 @@
             if (origin === 'complete' && keywords[text[0].toLowerCase()] != true) {
               const names = text[0]
                 .match(/("[^"]*"|[^.]+)/g)
+                .map(n => /^\d/.test(n) ? `"${n}"` : n)
                 .map(n => /[^a-z0-9_]/.test(n) && !/"/.test(n) ? `"${n}"` : n)
                 .join('.')
-              console.log(names)
   
               co.update(from, to, [names], origin)
             }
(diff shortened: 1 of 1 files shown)


thiagola92:
There is too much ambiguity problem.

rathboma:
Thanks for the PR! I'll take a look soon

thiagola92:
I know that i commit to much so i will try to explain.  

Check if is a keyword before deciding if should quote or not.  
Try to split the schema to get the table/column names.  
If the table/column name need to be quote, quote it  
```javascript
        if (this.connectionType === 'postgresql')  {
          this.editor.on("beforeChange", (cm, co) => {
            const { to, from, origin, text } = co;

            const keywords = CodeMirror.resolveMode(this.editor.options.mode).keywords

            // quote names when needed
            if (origin === 'complete' && keywords[text[0].toLowerCase()] != true) {
              const names = text[0]
                .match(/("[^"]*"|[^.]+)/g)
                .map(n => /^\d/.test(n) ? `"${n}"` : n)
                .map(n => /[^a-z0-9_]/.test(n) && !/"/.test(n) ? `"${n}"` : n)
                .join('.')
  
              co.update(from, to, [names], origin)
            }
          })
        }
```

Removed wrapIdentifier

Quote columns names with dot, others columns that need quote will be quoted inside "beforeChange"  
Add quoted option for tables names  
If we accept table name with dot and without quote, we will never be able to discover the table/column name later
```javascript
hintOptions() {
        const result = {}
        this.tables.forEach(table => {
          const cleanColumns = table.columns.map(col => {
            return /\./.test(col.columnName) ? `"${col.columnName}"` : col.columnName
          })

          // add quoted option for everyone that needs to be quoted
          if (this.connectionType === 'postgresql' && (/[^a-z0-9_]/.test(table.name) || /^\d/.test(table.name)))
            result[`"${table.name}"`] = cleanColumns
          
          // don't add table names that can get in conflict with database schema 
          if (!/\./.test(table.name))
            result[table.name] = cleanColumns
        })
        return { tables: result }
      },
```

For curiosity, i notice that we check a lot if the conection type is postgresql, there is some reason for this? outside postgresql this operations are not allowed?

rathboma:
Feels like our auto complete is getting more sophisticated, we should probably figure out a way to set up a test harness for this.

rathboma:
I finally managed to review all this code. Looks awesome. Merging. Thanks for all the complex regex work. Feels like we're going to need a real SQL AST parser soon.

Issue #333, opened 2020-09-15
BUG: Invalid IV length


UtechtDustin:
**Describe the bug**
I will get the following error for the newest commit/master branch.
I'm not able to figure out the issue, but it could be a issue with #322 

@rathboma 
Does it makes sense to create a community chat e.g. slack/discord etc. for smaller questions ?
```
vue.runtime.esm.js?2b0e:619 [Vue warn]: Error in mounted hook (Promise/async): "Error: Invalid IV length"

found in

---> <ConnectionInterface> at src/components/ConnectionInterface.vue
       <App> at src/App.vue
         <Root>
```

UtechtDustin:
After deletion the beekeeper database it works, so it seems it's an issue with the encryption.
I'm not sure if this is a issue because of the branch hopping or not, this should be checked twice (otherwise the connections could be corrupted)

rathboma:
Damn. Any errors in the terminal?

rathboma:
Yeah don't want to corrupt connections at all

UtechtDustin:
Errors from the dev tools:
```
vue.runtime.esm.js?2b0e:619 [Vue warn]: Error in mounted hook (Promise/async): "Error: Invalid IV length"

found in

---> <ConnectionInterface> at src/components/ConnectionInterface.vue
       <App> at src/App.vue
         <Root>
warn @ vue.runtime.esm.js?2b0e:619
logError @ vue.runtime.esm.js?2b0e:1884
globalHandleError @ vue.runtime.esm.js?2b0e:1879
handleError @ vue.runtime.esm.js?2b0e:1839
eval @ vue.runtime.esm.js?2b0e:1856
Promise.catch (async)
invokeWithErrorHandling @ vue.runtime.esm.js?2b0e:1856
callHook @ vue.runtime.esm.js?2b0e:4219
insert @ vue.runtime.esm.js?2b0e:3139
invokeInsertHook @ vue.runtime.esm.js?2b0e:6346
patch @ vue.runtime.esm.js?2b0e:6565
Vue._update @ vue.runtime.esm.js?2b0e:3945
updateComponent @ vue.runtime.esm.js?2b0e:4066
get @ vue.runtime.esm.js?2b0e:4479
Watcher @ vue.runtime.esm.js?2b0e:4468
mountComponent @ vue.runtime.esm.js?2b0e:4073
Vue.$mount @ vue.runtime.esm.js?2b0e:8415
[...]

PR #335, opened 2020-09-17
Password encryption is back

rathboma:
I already use simple-encryption and it is more careful about having a fixed-length iv (which was the earlier problem).

@UtechtDustin if you have a copy of your DB pre-decryption problem can you test this PR? It works fine for me.

Fix: #333


UtechtDustin:
Works ! :-)

PR #356, opened 2020-09-28
Remove duplicate database.connecting check

MasterOdin:
Upstream PR: https://github.com/sqlectron/sqlectron-core/pull/115

This removes a duplicate `database.connecting` check. It was probably supposed to be `server.connecting`, except that is not used anywhere else in the codebase. I suspect it might have been planned to be added in the server tunnel connection code, but given that it's not exactly a race condition anyone is struggling with, can probably get away with just getting rid of the check altogether.


rathboma:
thanks!

PR #357, opened 2020-09-29
Added ability to select all (Cmd/Ctl+A) in the background process

christianpatrick:
Thanks so much for making such a beautiful app! I've been using it for the past week and absolutely love it!

One point of friction for me is the inability to select all text from inputs when creating/updating connections. For a bit of context, one of the apps I work in has a local database port that changes from time to time. I'd love to simply paste in the new one without having to backspace every character beforehand.

I noticed that Electron already had `Select All` partially setup. So I simply added in the final steps and updated the native menu to include it under the "Edit" section. Here's a preview of the new functionality:
select-all-command

If there was a reason for not including it, or if I am not including it incorrectly, please feel free to let me know. I'll be super happy to adjust as needed. Thanks for your time!

### DIFF
--- a/src/background/NativeMenuActionHandlers.ts
+++ b/src/background/NativeMenuActionHandlers.ts
@@ -37,6 +37,9 @@ export default class NativeMenuActionHandlers implements IMenuActionHandler {
   paste(_1: Electron.MenuItem, win: Electron.BrowserWindow) {
     win.webContents.paste()
   }
+  selectAll(_1: Electron.MenuItem, win: Electron.BrowserWindow) {
+    win.webContents.selectAll()
+  }
 
   setZoom = async (level: number) => {
     getActiveWindows().forEach(window => {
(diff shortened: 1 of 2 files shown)


rathboma:
Ah, good find! Yes this is an oversight. It works on Windows/Linux, but MacOS is really funny about global shortcuts (they have to be in the menu to work).

My only ask is that you remove the `separator` -- I re-use this code for client-side menus (windows/linux) and separators totally break my code, lol. :-S

rathboma (changes requested):
Just need to remove the separator because `NewAppMenu.vue` can't handle it yet

christianpatrick:
Hey @rathboma, thanks for the perspective here! I'll make a fix for that separator now 😄

christianpatrick:
@rathboma, the separator has been addressed, and I've requested re-review. Thanks for your time!

PR #368, opened 2020-10-04
feat(readme): translate readme to Brazilian Portuguese

brenocastelo:
This Pull Request is a translation from README to Brazilian Portuguese


rathboma:
Thanks so much!

PR #376, opened 2020-10-08
Add ability to add SSL certficate files

trevorrjohn:
This is the same as https://github.com/beekeeper-studio/beekeeper-studio/pull/109, I have just rebased the branch and made the necessary changes. 

Solves: https://github.com/beekeeper-studio/beekeeper-studio/issues/89

### DIFF
--- a/src/common/appdb/models/saved_connection.ts
+++ b/src/common/appdb/models/saved_connection.ts
@@ -101,6 +101,15 @@ export class DbConnectionBase extends ApplicationEntity {
   @Column({type: 'boolean', nullable: false, default: false})
   ssl: boolean = false
 
+  @Column({type: 'varchar', nullable: true})
+  sslCaFile: Nullable<string> = null
+
+  @Column({type: 'varchar', nullable: true})
+  sslCertFile: Nullable<string> = null
+
+  @Column({type: 'varchar', nullable: true})
+  sslKeyFile: Nullable<string> = null
+
   // GETTERS
   get hash() {
     const str = [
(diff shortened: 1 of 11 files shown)


trevorrjohn:
I have tested this with both a server that requires certificates and one that only requires SSL, both connections worked fine. However, I only tested with PostgreSQL (CloudSQL).

rathboma:
Wow, thanks for adding this! I'll review over the weekend at make sure we're at least passing the correct files through to MySQL and SQL Server

rathboma (changes requested):
First - these changes look fantastic. Thanks so much.

So the one remaining item is to add support for SqlServer. Could you do the same thing for the SQL Server driver?

The mssql node package uses 'tedius' as the Sql Server driver. Like PSQL and MySQL you can set the Ca file and such.

StackOverflow example: https://stackoverflow.com/questions/37987786/verify-certificate-from-sql-server-in-node

Docs: http://tediousjs.github.io/tedious/api-connection.html
Look for ` options.cryptoCredentialsDetails ` and ` options.trustServerCertificate `

trevorrjohn:
@rathboma I have updated the PR. I was only able to go off the documentation you linked, so if you could test it with sql server that would be awesome.

rathboma:
Looks great. One final thing -- I think you need to add the new cert fields to used_connection.ts and also copy them as part of the `copy` mechanism?

trevorrjohn:
Good catch, there is one warning error I want to look at as well. Might be nothing.

trevorrjohn:
@rathboma this should be good to go now.

mholt:
Awesome! Does this work for CockroachDB as well?

Thanks so much for working on this.

trevorrjohn:
@mholt my guess is it should, it uses the postgresql client. I cannot test it though. 

@rathboma I also added support for cassandradb.

trevorrjohn:
@rathboma any chance you could put out a new release with  this? I have some non-technical co-workers that would love to start using this.

rathboma:
@trevorrjohn Yes! I'll be putting together a release this week. I have quite a few things to test (thanks to lots of community contributions), including a major move to Typescript (!) for many components.

Stay tuned! It will be out soon :-D

PR #392, opened 2020-10-15
Filter tables using pipe (|) as an OR clause.

carlson001:
Similar to a feature in HeidiSQL. Adds the ability to use the table filter with an OR type clause using a pipe character as the separator. 

Split terms based on presence of pipe.
Sort results alphabetically, and return unique array.


rathboma:
WHAAAT!! This is awesome. Going to review over the next couple of days, but thanks for opening a PR. I'll get back to you soon :-).

PR #393, opened 2020-10-15
convert sql_tools to typescript

MasterOdin:
A quick update to typescript for this file as I tackle the larger clients.


rathboma:
Good to see you back in the repo!

PR #395, opened 2020-10-16
Delete rows via row context menu on table view

fabscav:
I'd like to contribute to this project. You guys have done a fantastic job so far imho. I'd also like use beekeeper-studio on a daily basis. More options to manipulate data on the table view is something I'm missing the most, so I thought I'd implement it.

What I've implemented so far is the ability to delete rows on the table view via a row context menu (tabulator functionality).

Mind you, the implementation of this feature is rather rudimentary and done in parallel to the existing update functionality. Going forward I'd probably like to consolidate all data changing operations in the table component and in the client classes, but I thought I might submit this functionality as a first contribution before digging deeper.

delete_row


UtechtDustin:
Looks awesome for the first step, later it should be possible to select the columns through headers on the left side (which are also required for e.g. exports).
What happens if you edit a cell, mark the row for deletion and then commit the change ? (Is that possibel ? Could this cause problems ?)

fabscav:
> What happens if you edit a cell, mark the row for deletion and then commit the change ? (Is that possibel ? Could this cause problems ?)

It's possible and would result in the update query being executed before the delete query. So an unnecessary operation, but other than that it shouldn't result in any issues. Definitely something I'd optimize when consolidating table view change operations though.

UtechtDustin:
Okay that is fine for the first step i guess.
I thought it maybe deletes the row first then tries to update it, which maybe could result in issues.
But if that is not the case, good job :-)

rafmst:
This would definitely be a nice addition to the software. My dream context menu would be something like Sequel Pro has. 

image

I could help and make this if anyone wants to join forces.

fabscav:
> This would definitely be a nice addition to the software. My dream context menu would be something like Sequel Pro has. 

I've already started working on adding create and clone row as context options in a different branch and it's basically done. I'm just waiting for this PR to go through, since it's based on it.

The copy options would be a nice addition as well tho.

rathboma:
Wow, this is awesome!

I'm putting together a 1.8 release, but will review this with the aim to merge once I've finished that.

:heart: all the amazing contributions from you all! Thanks for helping to make the project amazing.

src/components/tableview/TableTable.vue
```diff
@@ -427,25 +442,62 @@ export default {
...
     async saveChanges() {
       try {
-        // throw new Error("This is an error")
-        const newData = await this.connection.updateValues(this.pendingEditList)
-        const updateIncludedPK = this.pendingEditList.find(e => e.column === e.pkColumn)
-        if (updateIncludedPK) {
+        let replaceData = false
+
+        // Commit Deletes
+        if (this.pendingDeleteList.length > 0) {
```
rathboma, line 469: I think probably this should check if this row has been edited and if so, remove it from the pendingEdits list.

Currently if you edit a row, then mark for deletion it will delete it then try to edit it, which could have side effects.

src/components/tableview/TableTable.vue
```diff
@@ -152,6 +153,12 @@ export default {
     pendingEditList() {
       return Object.values(this.pendingEdits)
     },
+    pendingDeleteList() {
```
rathboma, line 156: Any reason to not combined edits and deletes into a `pendingOperations`?

src/components/tableview/TableTable.vue
```diff
@@ -427,25 +442,62 @@ export default {
...
-        if (updateIncludedPK) {
+        let replaceData = false
+
+        // Commit Deletes
+        if (this.pendingDeleteList.length > 0) {
+          const deleteSuccess = await this.connection.deleteRows(this.pendingDeleteList)
+          if (deleteSuccess) {
+            replaceData = true
+          } else {
+            this.$noty.error("Error deleting rows")
```
rathboma, line 474: It should probably `return` here so as not to commit the edits?

src/lib/db/clients/postgresql.js
```diff
@@ -486,6 +487,34 @@ export async function updateValues(conn, updates) {
   return results
 }
 
+export async function deleteRows(conn, updates) {
+  const deleteQueries = updates.map(update => {
+    let where = {}
+    where[update.pkColumn] = update.primaryKey
+    return knex(update.table)
```
rathboma, line 494: can we use Knex for mysql also?

I think it's much nicer using knex vs manually building queries (which always makes me nervous no matter how often I do it)

src/lib/db/clients/sqlite.js
```diff
@@ -152,6 +153,33 @@ export async function updateValues(conn, updates) {
   return results
 }
 
+export async function deleteRows(conn, updates) {
+  const deleteCommands = updates.map(update => {
+    return {
+      query: `DELETE FROM ${update.table} WHERE ${update.pkColumn} = ?`,
```
rathboma, line 159: Lets use knex here too, so these get wrapped properly

src/lib/db/clients/mysql.js
```diff
@@ -321,6 +322,34 @@ export async function updateValues(conn, updates) {
   return results
 }
 
+export async function deleteRows(conn, updates) {
+  const deleteCommands = updates.map(update => {
+    return {
+      query: `DELETE FROM ${wrapIdentifier(update.table)} WHERE ${wrapIdentifier(update.pkColumn)} = ?`,
```
rathboma, line 328: Can we use knex here like you did for postgres?

rathboma (changes requested):
This is AMAZING.

I have a couple of requests, just to neaten things up (knex) and address a couple of bugs (inconsistent delete/edit behavior).

src/components/tableview/TableTable.vue
fabscav: I did actually refactor it that way in preparation for implementing the add row functionality. I'll add the refactor part to this PR.

src/components/tableview/TableTable.vue
fabscav: Yes absolutely. It's also part of the refactoring mentioned above. I'll add it to this PR shortly.

src/components/tableview/TableTable.vue
fabscav: Sure, that makes sense. Otherwise we'd need the possibility to display multiple errors I guess.

src/lib/db/clients/mysql.js
fabscav: Absolutely. I just saw that it wasn't included in this connector, so I thought there might be a reason for that. I'll update it soon.

fabscav:
@rathboma I've made some updates that should take care of the things you mentioned in your feedback.

rathboma:
Awesome! I'll take a look later today

src/components/tableview/TableTable.vue
```diff
@@ -484,51 +519,135 @@ export default {
...
+        pkColumn: this.primaryKey,
+        primaryKey: pkCell.getValue()
+      }
+
+      this.addPendingChange(CHANGE_TYPE_DELETE, payload)
+    },
+    addPendingChange(changeType, payload) {
+      if (changeType === CHANGE_TYPE_UPDATE) {
+        // remove existing pending updates with identical pKey-column combo
+        let pendingUpdates = _.reject(this.pendingChanges.updates, { 'key': payload.key })
```
rathboma, line 547: const

src/components/tableview/TableTable.vue
```diff
@@ -484,51 +519,135 @@ export default {
...
+        pendingUpdates.push(payload)
+
+        this.$set(this.pendingChanges, 'updates', pendingUpdates)
+      }
+
+      if (changeType === CHANGE_TYPE_DELETE) {
+        // remove pending updates for the row marked for deletion
+        let filter = { 'primaryKey': payload.primaryKey }
```
rathboma, line 555: also should be consts

src/components/tableview/TableTable.vue
```diff
@@ -484,51 +519,135 @@ export default {
...
+            this.$noty.error("Error updating rows")
+
+            return
+          }
+        }
+
+        // handle deletes
+        if (this.hasPendingDeletes) {
+          try {
+            await this.connection.deleteRows(this.pendingChanges.deletes)
```
rathboma, line 615: I think both the updates, and the deletes should happen inside a single transaction. We don't really want changes to be partially applied.

src/lib/db/clients/mysql.js
```diff
@@ -325,6 +329,27 @@ export async function updateValues(conn, updates) {
   return results
 }
 
+export async function deleteRows(conn, deletes) {
+  const deleteCommands = buildDeleteQueries(knex, deletes)
+
+  const commands = ['START TRANSACTION', ...deleteCommands];
```
rathboma, line 335: Inside of here we're using a transaction, but the edits and deletes happen inside separate transactions.

I think it's reasonable for the user to assume they both operate inside a single transaction. Having updates apply, but not row deletes would be pretty confusing I think.

rathboma (changes requested):
I think the UI components look great. Nice work.

On the backend, we probably need both updates and deletes to apply inside the same transaction

rathboma:
Quick ping here to see if you have time to update the transaction behavior, would love to merge this.

fabscav:
I might be able to implement the changes over the holidays, but I can't promise anything unfortunately.

rathboma:
I'm going to do a release today, but if you get around to this I'll merge and release it in a point release!

Matthew

src/components/tableview/TableTable.vue
fabscav: In this case a variable should be appropriate, since in the next line additional data is being pushed to the array `pendingUpdates`.

rathboma:
looks like you've been doing some work! Let me know when you're ready for another review. :+1:

fabscav:
Yeah I did refactor it to use a single transaction and it works fine except for SQL Server. It won't let me span the transactions over multiple queries, unless executed as a joint statement. Am I missing something or is that just not possible?

src/lib/db/clients/sqlserver.js
```diff
@@ -446,24 +446,54 @@ export async function getPrimaryKey(conn, database, table, schema) {
...
+export async function applyChanges(conn, changes) {
+  let results = []
 
-  const { updateQueries, selectQueries } = buildUpdateAndSelectQueries(knex, updates)
-  const sql = ['set xact_abort on', 'BEGIN TRANSACTION', ...updateQueries, 'COMMIT'].join(";")
-  const results = []
   await runWithConnection(conn, async (connection) => {
     const cli = { connection }
-    await driverExecuteQuery(cli, { query: sql })
+    await driverExecuteQuery(cli, { query: 'set xact_abort on; BEGIN TRANSACTION' })
```
rathboma, line 454: so yes there is this limitation with SQL Server. You'll have to compile them into a single statement, similar to how the old `updateValues` worked:

const sql = ['BEGIN TRANSACTION', ...updateQueries, ...deleteQueries, 'END TRANSACTION']

fabscav: Ah, that's what I was afraid of. I made the necessary changes.

bediu:
Is this feature coming close to being shipped? I might be the biggest fan of this software and I'm grateful. 
There's a lot of need to manipulate db via ui because it's very useful for backend development.

rathboma:
Yes! Should be doing a release this week.

rathboma:
@fabscav looks like you actually broke the `mysql.spec.js` and `postgres.spec.js` tests with this change (and I broke the PR automated test workflow)

By renaming `updateValues` to `applyChanges` a couple of tests need fixing. Any chance you could do that?

PR #397, opened 2020-10-18
Update saved_connection.ts

vitaly-t:
Better use of the `ConnectionString`.

Defaults are properly passed within the constructor, as they should be.


rathboma:
looks like this breaks some of the null checks?

vitaly-t:
Yeah, those are obsolete <Nullable> things today :)

Issue #388, opened 2020-10-15
BUG (minor): Error on connection is carried across connections views


danilopolani:
**Describe the bug**
Whenever a connection fails and an error appears below the connection box, that error is carried across ALL the connections when you click them on the sidebar (even before trying to connect to them).

**To Reproduce**
1. Create a connection (even a fake one is fine, it just needs to be on the sidebar)
2. Create another failing connection or just use the quick connect. This must fail.
3. Connect to the failing connection.
4. When the error appears below the box, navigate across other connection (e.g. the created at step 1)

**Expected behavior**
The message should disappear or, better, be linked to that connection-view only.

**Screenshots / Gifs**
https://streamable.com/hlfwsr

**Version Information (please complete the following information):**
 - OS: Ubuntu 20.04
 - App Version: 1.7.4

**Additional context**
-

rathboma:
Excellent UI bug report, thank you!

rathboma:
this is in master.

PR #399, opened 2020-10-19
Assign connection errors to specific connection

fabscav:
This PR stores connection errors on a per-connection basis, thus fixing the bug mentioned in #388.


rathboma:
Thanks for the PR!!

I actually just solved this by clearing the error when the config is modified. See: https://github.com/beekeeper-studio/beekeeper-studio/commit/58370fef635ab13021ffe26fd6086d67200a9bb1

PR #409, opened 2020-10-23
upgrade mysql2 to ^2.2.0

### LINKED ISSUE #406: FEAT: Parse errors for 'unknown plugin sha256_password' and add a helpful message
Mysql sometimes give you an error message (see #405 ) about sha256_password, a deprecated password storage plugin.

Many times this simply means you are using the wrong username. Other times it means you need to change how your passwords are stored.

Goal: if we see this error, add two helpful things:
1. Suggest that the username might be wrong
2. Provide a link to our docs on how to solve this.

MasterOdin:
Closes #406 

This upgrades the `mysql2` dependency to its latest version (2.2.5) with a minimum of `^2.2.0`. In this minor release, support for the deprecated `sha256_password` authentication scheme was added (amongst a handful of other things), and that it does not appear to have affected the general usage of it within beekeeper beyond the added authentication mode support.

Testing can be done with: `docker run --rm -e MYSQL_ROOT_PASSWORD=password -e MYSQL_DATABASE=beekeeper -e MYSQL_USER=test -e MYSQL_PASSWORD=test -p 3306:3306 --name mysql mysql:8 --default-authentication-plugin=sha256_password`

On current master, get the error as reported in #406. With this patch, it connects fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "material-icons": "^0.3.1",
     "mkdirp": "^1.0.4",
     "mssql": "^6.2.0",
-    "mysql2": "^2.1.0",
+    "mysql2": "^2.2.0",
     "node-sass": "^4.12.0",
     "node-ssh-forward": "^0.7.2",
     "pg": "^7.12.1",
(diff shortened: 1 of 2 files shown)


dolmen:
I still have issue connecting to a `sha256_password` account. See #1488.

`mysql2` 3.0.0-rc1 has one more `sha256_password` fix.

PR #412, opened 2020-10-25
switch back to mainline pgpass

MasterOdin:
[pgpass@1.0.4](https://npmjs.com/package/pgpass) was released which contains the patch from the beekeeper-studio fork which fixed throwing on empty home setting.


rathboma:
just jumped on to do this, thanks :-D

Issue #352, opened 2020-09-26
FEAT: Tab Reordering


GeorgeWells:
It would be nice if we could have the ability to be able to reorder the tabs by clicking and dragging.

Also, it would be awesome if the open SQL tabs were remembered per connection when closing the connection/window and restored when that connection is opened again. Maybe this is something that is regularly saved behind-the-scenes in case of a power cut or crash to minimise losing any queries that have been typed up.

rathboma:
Yes, tab reordering should be a thing. As should saving tab contents. Lets split these into two tickets.

PR #425, opened 2020-10-30
Tab reordering

fabscav:
This PR adds the ability to reorder tabs by dragging and dropping them, as mentioned in #352, by using the package vue-draggable.

Testing this on Windows 10 using WSL2 and VcXsrv, I noticed a white border around the dragged element, which I found to be impossible to remove. This border exists on all dragged elements, even if using native drag and drop (which vue-draggable relies on anyways), so I'm unsure what the cause or solution to this would be.

Edit: It turned out to be an issue with VcXsrv. There's no border when running it on Windows or Linux natively.
tab_sort

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "vue-clipboard2": "^0.3.1",
     "vue-js-modal": "^1.3.33",
     "vue-property-decorator": "^8.4.2",
+    "vuedraggable": "^2.24.2",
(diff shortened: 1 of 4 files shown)


rathboma:
This also looks awesome! Need to test it then I'll get back to you :-)

rathboma:
Damn. This 'just works'. Amazing.

PR #427, opened 2020-10-31
Raw query support for table view

### LINKED ISSUE #429: BUG: Can't filter for null values in table view
Can't search for NULL values when in data view

Add WHERE <fieldName> IS NULL and WHERE <fieldName> IS NOT NULL to dropdown

sql is possible of course

**Additional context**
This gives an error when executing
image

### LINKED ISSUE #430: FEAT: Allow search IN () and NOT IN ()
A lot of times, I want to search for a few records with specific values (mostly primary keys). Currently, I need to write a query or search one at the time using 'equals'.

To be able to search on all values at once, the pulldown should allow for `IN` and `NOT IN`.

**Additional context**
image

NB This could even be expanded with JSON, REGEXP etc.

fabscav:
This PR implements an alternative way to filter table view results: by raw `where` condition.
raw_query

Fixes #430
Fixes #429


rathboma:
Nice! We'd thought about this. I'll review this week! Thanks

rathboma:
Couple of requests

- When someone goes from simple toggle to complex toggle, is there a way to pre-populate it with the value of the simple toggle? It should be pretty easy to piece it together.
- Also if possible (and you can figure it out), what do you think about adding a unit test to db.ts in tests/* for checking both types of filters work?

Question:

- What happens if the query is invalid? Is there an error that pops up somewhere?

fabscav:
@rathboma Thanks for your kind words and feedback 👍 

> When someone goes from simple toggle to complex toggle, is there a way to pre-populate it with the value of the simple toggle? It should be pretty easy to piece it together.

Definitely possible without much effort. However, I'm unsure what the best behavior would be. Would it only pre-populate the first time you toggle to raw query? Or would it always overwrite it? Or only stop once the raw query is manually changed? Since it's not possible to always keep both methods in-sync, I'm not sure there's a good way to do it.

> What happens if the query is invalid? Is there an error that pops up somewhere?

Good call. The `dataFetch()` method simply threw the error, so I've updated the PR to include better error handling. Now, any errors related to the queries executed by the `dataFetch()` method are displayed on the bottom left of the table view.

> Also if possible (and you can figure it out), what do you think about adding a unit test to db.ts in tests/* for checking both types of filters work?

Sure thing, the more tests the better. However, since my changes on the backend side are contained within the `clients/utils.js`, would it perhaps be better to create a `utils.spec.js` instead, with a test for the method I changed?

rathboma:
> @rathboma Thanks for your kind words and feedback 👍
> 
> > When someone goes from simple toggle to complex toggle, is there a way to pre-populate it with the value of the simple toggle? It should be pretty easy to piece it together.
> 
> Definitely possible without much effort. However, I'm unsure what the best behavior would be. Would it only pre-populate the first time you toggle to raw query? Or would it always overwrite it? Or only stop once the raw query is manually changed? Since it's not possible to always keep both methods in-sync, I'm not sure there's a good way to do it.

Yes, I've been thinking about this. I think a sensible solution would be to populate the rawquery if and only if:
- Raw query entered by user is currently empty when switching
- When a user enters a raw query and switches back to structured query, should it clear the structured query? My guess is no?

> 
> > What happens if the query is invalid? Is there an error that pops up somewhere?
> 
> Good call. The `dataFetch()` method simply threw the error, so I've updated the PR to include better error handling. Now, any errors related to the queries executed by the `dataFetch()` method are displayed on the bottom left of the table view.
Great!

> > Also if possible (and you can figure it out), what do you think about adding a unit test to db.ts in tests/* for checking both types of filters work?
> 
> Sure thing, the more tests the better. However, since my changes on the backend side are contained within the `clients/utils.js`, would it perhaps be better to create a `utils.spec.js` instead, with a test for the method I changed?
Yes, that'd be great. Wherever is sensible to add a test is fine by me, that's already better than 90% of the codebase :-)

fabscav:
@rathboma I've added the pre-populate function as described and also included tests for the modified `buildSelectTopQuery` method in `utils.js`.

rathboma:
This looks super good. Going to merge and we can tidy up styles in master if we need to.

PR #525, opened 2021-01-23
Add and clone rows via row context menu on table view

fabscav:
This PR expands on the changes in #395 by adding add row and clone row options to the table view row context menu.

There are still a couple of things to consider:

- ~~SQL Server implementation isn't working yet (will be solved when solved in #395)~~
- Add row only accessible via row context menu, thus unreachable for inserting the first row.
- No validation of column content or primary key duplicate.
- ~~The green background is possibly too bright~~.
- ~~Context menu styling doesn't match other context menus (anymore?)~~.

insert_row


UtechtDustin:
As soon this PR is merged and beekeeper has a function to export/import tables/database, i could start to use it in production 👍🏻

src/components/tableview/TableTable.vue
```diff
@@ -392,7 +414,47 @@ export default {
...
+            this.tabulator.addRow({}, false).then(row => { 
+              this.addRowToPendingInserts(row)
+              this.tabulator.scrollToRow(row, 'bottom', false)
+            })
+          }
+        },
+        {
+          label: '<i class="item-icon material-icons">content_copy</i> Clone row',
+          action: (e, row) => {
+            this.tabulator.addRow(row.getData(), false).then(row => {
```
rathboma, line 442: Two things here:

- I think one nice thing to do here would be to clear the `primaryKey` column if there is one, as you cannot insert with a duplicate primary key. This should be as easy as `data[this.primaryKey] = undefined`
- Probably `row.getData()` returns the actual object used by the row, so might need to make a copy of it? `const data = { ...row.getData() }`

src/components/tableview/TableTable.vue
```diff
@@ -486,51 +566,142 @@ return dt.split("(")[0]
...
+      this.addPendingChange(CHANGE_TYPE_DELETE, payload)
+    },
+    addPendingChange(changeType, payload) {
+      if (changeType === CHANGE_TYPE_INSERT) {
+        this.pendingChanges.inserts.push(payload)
+      }
+
+      if (changeType === CHANGE_TYPE_UPDATE) {
+        // remove existing pending updates with identical pKey-column combo
+        let pendingUpdates = _.reject(this.pendingChanges.updates, { 'key': payload.key })
```
rathboma, line 617: const

rathboma (changes requested):
Minor change request. This looks amazing!

src/components/tableview/TableTable.vue
fabscav: `pendingUpdates` is being modified on the next line.

rathboma:
@fabscav Merged the first PR! Great work!! 

Can you merge origin/master into this branch? That way we can see the only new changes :-)

fabscav:
> 
> 
> @fabscav Merged the first PR! Great work!!
> 
> Can you merge origin/master into this branch? That way we can see the only new changes :-)

Done 😃

ReynoldsCahoon:
Looking forward to when this gets added!

src/lib/db/clients/utils.js
```diff
@@ -123,6 +123,23 @@ export async function genericSelectTop(conn, table, offset, limit, orderBy, filt
...
 
+export function buildInsertQueries(knex, inserts) {
+
+  return inserts.map(insert => {
+    // remove empty pkColumn data if present
+    insert.data = _.omitBy(insert.row.getData(), (value, key) => {
+      return (key === insert.pkColumn && !value)
+    })
+    
+    console.log("inserts", inserts)
```
rathboma, line 134: we need to remove the console logs, and replace with `electron-log`.

src/lib/db/clients/sqlite.js
```diff
@@ -124,6 +130,10 @@ export async function applyChanges(conn, changes) {
     await driverExecuteQuery(cli, { query: 'BEGIN'})
 
     try {
+      if (changes.inserts) {
```
rathboma, line 133: One thing that I think would be very helpful (this has the ability to wreck someone's database if there's a bug) would be to have these functions covered by the integration test suite.

Covering:
- inserting good data
- Trying to insert a bad row (transaction shouldn't commit)
- delete, update, and inserts mixed together, should do the right thing
- Bad insert or update mixed with a good version of another (transaction should revert it)

rathboma (changes requested):
This change looks AMAZING.

If able, would be be willing to add an integration test? It might be possible to cover all databases with the same code.

fabscav:
> If able, would be be willing to add an integration test?

Sure, I can do that.

rathboma:
Thanks! That'd be awesome

rathboma (changes requested):
These tests look great!

Looks like you've added them to all the database client integration tests?
There's an easier way to include them in the test for all clients -- in `tests/lib/db.ts` I run common tests for all databases.

See `testdb()` -> This calls specific test functions.

You could move these new test functions, into here and call them from within `testdb()`, that way we don't need to duplicate the test code for all the specs.

fabscav:
> There's an easier way to include them in the test for all clients -- in tests/lib/db.ts I run common tests for all databases.

Ah, didn't see that one, was only looking at the integration tests folder.

> You could move these new test functions, into here and call them from within `testdb()`, that way we don't need to duplicate the test code for all the specs.

Fair enough, but wouldn't all the assertions then fail under the same name, "Should pass standard tests"? That seems quite inconvenient when debugging failed tests.

rathboma:
That is also kind of annoying yes. I think this is good as-is and I can tidy up that stuff as needed. Great work! Merging

Issue #216, opened 2020-06-25
Table and schema names do not autocomplete properly


rathboma:
Currently the editor auto completes column names, but does not auto complete schmas or tables in certain circumstances.

- Schemas should auto-complete along with table names after typing `from`, `join`, etc.
- Tables should auto-complete after typing `schema.`, same as columns auto-complete after typing `table.`

infraweavers:
Just to chime in on this; it's pretty critical for PostgreSQL uses. PG databases tend to use schemas a lot compared to MySQL/SQLServer etc so it basically makes the autocompletion useless as it stands for most databases.

PR #533, opened 2021-01-29
Some quality of life improvements for autocomplete

mintsoft:
In my experience, the autocomplete of queries is the single most important feature of any DB querying tool; yet they're all terrible.

I've changed this as 2 small tweaks to always prefer user created tables first (users _very rarely_ are querying any of the builtins) and to fully qualify tables with their schema. 

image

Sort of heading towards fixing/improving https://github.com/beekeeper-studio/beekeeper-studio/issues/216

Happy to change tac or whatever as required; just let me know


rathboma:
Nice! Let me take a look on Monday. Thanks for the PR :-)

rathboma:
So I like the idea of auto-completing `schema.table`, a hack would be to add both of them to the table_hint results, what do you think?

I think in many cases folks don't bother writing `public`, so I don't want to force that.

mintsoft:
Sure, I can try that @rathboma !

mintsoft:
> So I like the idea of auto-completing schema.table, a hack would be to add both of them to the table_hint results, what do you think?

@rathboma actually I'm confused now I've had another look at the code, I think the above is what I've done?


> I think in many cases folks don't bother writing public, so I don't want to force that.
Good point! I've updated it so that for `public` it will now add it both with and without the schema so users can autocomplete both `public.blah` and `blah`.

rathboma:
Nice. One last question -- for other DBs schema is usually `null`, so that probably needs a null check

PR #535, opened 2021-01-31
Connection write mode option

fabscav:
This PR adds a new option to connections: **write mode**, It can have one of three values:

- Enabled (default): Current behavior, full write access.
- Confirm: All queries that change data or structure have to be confirmed.
- Read-only: Queries that change data or structure are not allowed.

connection_writemode

### DIFF
--- a/src/components/TabQueryEditor.vue
+++ b/src/components/TabQueryEditor.vue
@@ -76,7 +76,7 @@
 
     <!-- Parameter modal -->
     <modal class="vue-dialog beekeeper-modal" name="parameters-modal" @opened="selectFirstParameter" @closed="selectEditor" height="auto" :scrollable="true">
-      <form @submit.prevent="submitQuery(queryForExecution, true)">
+      <form @submit.prevent="submitQuery(queryForExecution, true, true)">
         <div class="dialog-content">
           <div class="dialog-c-title">Provide parameter values</div>
           <div class="dialog-c-subtitle">Don't forget to use single quotes around string values</div>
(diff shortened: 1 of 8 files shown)


src/components/TabQueryEditor.vue
```diff
@@ -121,6 +145,7 @@
   import QueryEditorStatusBar from './editor/QueryEditorStatusBar.vue'
   import rawlog from 'electron-log'
   const log = rawlog.scope('query-editor')
+  const writeVerbs = ["alter", "create", "delete", "drop", "insert", "rename", "truncate", "update", "upsert"]
```
rathboma, line 148: My concern is that there could be select statements containing some of these words legitimately.  We already have a dependency on sql-query-identifier, we should probably rely on that.

I think I'd also be uncomfortable (right now) saying that beekeeper can prevent write operations, I think warning is fine, but sql-query-identifier would probably need some new features.

read-only mode should probably be implemented database-by-database, for example in postgres - https://github.com/brianc/node-postgres/issues/569

To do that we'd need to run a statement before each query is executed.

fabscav: I updated the component to utilize the `sql-query-identifier` instead of the verb array. However I did notice it doesn't mark `alter table` queries as modifications.

Implementing write-protection on a connection-level should be possible. SQLite and SQL Server support this as a connection param afaik. Others such as MySQL would probably require the query to be wrapped in a read only transaction. 

I could look into implementing it, but if you're reluctant with implementing a read-only mode in general I'd be fine with dropping this PR as well. Tbh I only need this functionality to be present in my fork in order to start using beekeeper in production environments.

rathboma: I'm 100% open to this feature. I think it would be a great feature.

Yes, if we can do read-only mode at the connection-level that would be fantastic, although I know that's a lot of work.

We could use the following:
- user selects read-only
- if connection support it, works as expected
- If not, we provide a warning that it might not be 100%, and implement at the app level

For 'warn me' mode:
- provide a warning on the connection screen that the feature is in beta and might not be 100%.


If we do those things, I'm good with it.

With alter table -- ideally we can add basic detection to sql-query-identifier

rathboma:
Hey! I think this is a good draft, but it's been a while since you've been working on it. I think it best I close it for now, eventually we'll come back to this as part of a real permissions system.

PR #540, opened 2021-02-04
feat: shows and inputs postgres bytea values as base64 string

moisespsena:
This PR add feature to shows and inputs Postgres `BYTEA` values as base64 string.
- shows in data table
- edit value in data table


rathboma:
Awesome PR, thanks!

PR #567, opened 2021-02-19
Export tables and table tabs to file

fabscav:
This PR adds export functionality for tables and table view tabs. The data can be exported as `csv`, `json` or `sql`, each with their own set of output options, such as including a header row (csv) or table create query (sql), from either the table list item context menu or from an export button at the top-right of a table tab (better placement suggestions very welcome). When exporting from a table view tab, the active query filters are used for the export.

There are still quite a few things to do before this PR is ready, including adding error handling, tests, better TypeScript structure, general code refactorings and so on, but I thought I'd add it as a draft in order to get some initial feedback.

Edit: I've made some updates, including removing the blocking progress modal and instead adding a new sidebar tab to keep track of exports. I've updated the .gif to reflect these changes.

export_3


rathboma:
Holy moly this is awesome! Looking forward to reviewing it when you're ready

src/components/tableview/TableTable.vue
```diff
@@ -27,6 +27,7 @@
           </div>
           <div class="btn-wrap">
             <button class="btn btn-primary" type="submit">Search</button>
+            <button class="btn btn-flat" type="button" @click.prevent="exportTable()">Export</button>
```
fabscav, line 30: Not happy with adding another button to this bar. It's getting more and more crowded. However, I don't really know where else it could go.

src/filters/pretty-bytes-filter.js
```diff
@@ -0,0 +1,28 @@
+// usage: {{ file.size | prettyBytes }}
```
fabscav, line 1: Had to add this as a custom filter, as none of the packages worked.

src/lib/db/client.ts
```diff
@@ -64,8 +64,8 @@ export interface IDbInsert {
 
 export interface TableResult {
   result: any[],
-  fields: string[]
-  totalRecords: Number
```
fabscav, line 68: `totalRecords` not being a primitive caused trouble down the line and I saw no usage of the non-primitive `Number` type, so I changed it.

src/lib/export/formats/csv.ts
```diff
@@ -0,0 +1,39 @@
...
+
+interface OutputOptionsCsv {
+  header: boolean,
+  delimiter: string
+}
+
+export default class CsvExporter extends Export {
+  readonly format: string = 'csv'
+
+  constructor(
```
fabscav, line 12: I don't have much TypeScript experience and I didn't find a way to avoid having to declare a constructor override only for the format-specific `OutputOption` interface. I tried declaring a generic `OutputOptions` interface in the `export.ts` and extending that, but it doesn't work that way apparently.

tsconfig.json
```diff
@@ -21,7 +21,7 @@
     ],
     "paths": {
       "@/*": [
-        "src/*"
```
fabscav, line 24: Since `baseUrl` already points to `src`, this alias was pointing to `src/src`.

src/assets/styles/app/vendor/xel.scss
```diff
@@ -1632,8 +1632,8 @@ x-box {
  */
 
 x-progressbar {
-  background: rgba(59, 153, 251, 0.27);
+  background: rgba(250, 216, 59, 0.27);
   height: 5px;
-  --bar-background:#3B99FB;
-  --bar-box-shadow: 0px 0px 0px 1px #4392F9;
+  --bar-background: #fad83b;
```
fabscav, line 1637: I was unable to use variables here for some reason.

src/components/export/ExportNotification.vue
```diff
@@ -0,0 +1,73 @@
+<template>
+  <span></span>
```
fabscav, line 2: Yeah, this component doesn't have a template. It's a kind of workaround for Noty not supporting reactive content.

src/lib/export/export.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    md5sum.update(Date.now().toString(), 'utf8')
+    md5sum.update(this.table.name)
+    md5sum.update(this.filePath)
+
+    return md5sum.digest('hex')
+  }
+
+  async getChunk(offset: number, limit: number): Promise<TableResult | undefined> {
+    const result = await this.connection.selectTop(
```
fabscav, line 62: It would possibly be better to implement a separate `select` method that doesn't return the row count. However, when testing it the performance gain from not querying the row count was negligible.

src/lib/export/export.ts
```diff
@@ -0,0 +1,209 @@
...
+      this.status = Export.Status.Error
+      this.error = error
+
+      if (this.options.deleteOnAbort) {
+        await this.deleteFile()
+      }
+    }
+  }
+
+  calculateTimeLeft(): void {
```
fabscav, line 160: Ideally this would do the calculations based on the durations of the last n chunks instead.

src/lib/export/formats/sql.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (!this.connection.connectionType || !this.knexTypes[this.connection.connectionType]) {
+      throw new Error("SQL export not supported on connectiont type " + this.connection.connectionType)
+    }
+
+    this.knex = knexlib({ client: this.knexTypes[this.connection.connectionType] || undefined })
+  }
+
+  async getHeader(firstRow: any) {
+    if (this.outputOptions.createTable) {
+      const schema = this.table.schema && this.outputOptions.schema ? this.table.schema : ''
```
fabscav, line 40: `this.table.schema` is possibly not available in all client implementations.

rathboma:
I'll take a look at this PR this week! Don't want you to think I've forgotten about it :-)

fabscav:
Hey no rush, take your time 🙂

src/components/export/ExportNotification.vue
rathboma: Yeah, we'll probably move off-of noty at some point

src/components/tableview/TableTable.vue
rathboma: Don't worry about that! We'll deal with it :-).

We're likely going to coalesce on some sort of `actions` button in the statusbar for export / adding rows / viewing table schema information, etc, so it will go there, but we need to experiment with it some more.

src/filters/pretty-bytes-filter.js
rathboma: nice.

src/lib/export/export.ts
rathboma: So three things to address here:

- I just added an updated rowCount implementation that relies on an approximate count rather than the real count, as on very large tables this is prohibitively expensive. So we do probably want to avoid that.

- Because the sequence of selectTop operations don't take place inside a single transaction there could be duplicates as it pages through the rows when new rows are inserted.
- Most database clients provide a way to maintain a pointer and page through the results as they stream from the database, we're best to use this in order to get a consistent snapshot.

The amount of work you have done already exceeds my expectations. Really really good stuff. So I'm happy taking on this backend code to make it 'do the right thing'.

src/lib/export/export.ts
rathboma: Totally fine I think except for MongoDB which famously slows down the deeper into an export you get :laughing:

rathboma: (we don't even support mongo yet)

src/lib/export/formats/csv.ts
rathboma: I'll fix this, don't worry :-)

src/lib/export/formats/csv.ts
```diff
@@ -0,0 +1,40 @@
+import { Export, ExportOptions } from "@/lib/export";
```
rathboma, line 1: Love the way you've structured the export files. I'm very glad we moved to typescript honestly

src/lib/export/formats/json.ts
```diff
@@ -0,0 +1,42 @@
...
+  async getFooter() {
+    return ']'
+  }
+
+  formatChunk(data: any): string[] {
+    const formattedChunk = []
+
+    for (const row of data) {
+      const spacing = this.outputOptions.prettyprint ? 2 : undefined
+      const content = indentString(JSON.stringify(row, null, spacing), 2) + ','
```
rathboma, line 35: How does this (and the other exporters) handle things like byte arrays? I'll probably need to combine some of our existing tabulator logic for the display of non-readable table data

rathboma (approved):
This is an amazing PR, thank you so much for contributing it.

I'm going to change the merge target and merge it into a local Beekeeper branch. That way Gregory and I can take what you've done and do some of the final tidy up before merging it.

I've done the same with your other PR (clone / add rows), so we can integrate more user-feedback when saving.

You rock!

PR #569, opened 2021-02-19
sidebar-toggling, hotkey

ghstng:


src/components/CoreInterface.vue
```diff
@@ -36,7 +37,12 @@
           this.$refs.content
         ]
       },
-
+      keymap() {
```
rathboma, line 40: This should be in `data`

src/components/sidebar/CoreSidebar.vue
```diff
@@ -2,7 +2,8 @@
 
   <div class="sidebar-wrap row">
     <global-sidebar
-      @click="click"
+      @selected="click"
+      v-on="$listeners"
```
rathboma, line 6: Nice, I've never used this, cool.

rathboma (changes requested):
One minor thing - moving the keymap to `data`.

rathboma:
One thing we've forgotten:

Adding a menubar entry for this under `view`. I should be able to do that before we merge.

rathboma:
@ghstng it is now not possible to resize the sidebar at all.

rathboma:
image

If you make the window smaller

rathboma (changes requested):
Problems:

1. Can't resize the sidebar at all
2. It seems to start really wide
3. Make a small window and it squishes into the side

rathboma:
I'm unfortunately going to undo your keymap too -- ctrl+b breaks standard mac and ubuntu keyboard shortcuts when editing (that's back a character in emacs-land, which both OSes bind to)

ghstng:
@rathboma rgr will look.

PR #582, opened 2021-03-04
 Some quality of life improvements for autocomplete 

mintsoft:
Replaces https://github.com/beekeeper-studio/beekeeper-studio/pull/533 as I got completely screwed over by git.


mintsoft:
@rathboma I've added the null check around the schema now so that should be good to go I think.

mintsoft:
@rathboma any particular feedback?

PR #600, opened 2021-03-19
Re enable addclone

rathboma:
Still outstanding things to discuss:

- [x] Some sort of user feedback when an insert succeeds
- [x] Some sort of user feedback when a delete succeeds
- [x] Save Errors don't wrap properly (look at the width of the inner span)
- [x] Add/Clone row in light mode has no coloring?
- [x] Right click menu icons for add/clone/delete actions?
- [x] Do we need an 'actions' button for `add record`? Seems weird in right-click menu


rathboma:
@ghstng lots to discuss with you here

ghstng:
rgr

PR #601, opened 2021-03-19
Table exports

rathboma:
Continuation of https://github.com/beekeeper-studio/beekeeper-studio/pull/567/

All credit goes to @fabscav .

Still TODO:

- [x] Exports should use a streaming select with database-pointers rather than paging through selectTop
- [x] CSV exports should use papaparse to avoid edge cases
- [x] Complex objects should be serialized the same as when displayed in the app - byte arrays, json, hex, etc.
- [x] Figure out where to put the 'export' button
- [x] See Original PR for more comments and minor cleanup issues
- [x] formatChunk should simply return a `string`, not a `string[]`
- [x] SQL export is missing semi-colons between statements
- [ ] Exports should run in a WEB WORKER, because currently they lock up the UI thread

UX Questions
- [x] Should exports show in a tab rather than a modal?
- [x] Thoughts on the export notifications?


UtechtDustin:
After this feature is implemented, i finally can use beekeeper in production 🎉 
Does this PR will also contains a import function?

rathboma:
> After this feature is implemented, i finally can use beekeeper in production tada
> Does this PR will also contains a import function?

It does not contain an import function, but we'll be building that soon :-)

PR #602, opened 2021-03-19
(fix) auto-updater typo: availble -> available

gabesullice:
Screen Shot 2021-03-19 at 16 01 41

P.S. I _love_ this software. Thank you! I've been looking for something like this for years :D


rathboma:
OMG embarrassing typo. Thanks for the PR :-).

PR #605, opened 2021-03-20
FIX: Copied cell values have a new line separator

SarTais:
When copying a value of the column from the results table it adds a new line separator to the data and it's very annoying when you need to copy the column value and paste it to the string constant for example.

Here is a live example:
before

After reviewing I found that it passes table cell which contains ```pre``` tag with basically column data and also ```div``` elements related to column resizer. And in a result Selection class selects all those unneeded elements.

Here is how it works now for result-table cells:
after

### DIFF
--- a/src/components/editor/ResultTable.vue
+++ b/src/components/editor/ResultTable.vue
@@ -101,7 +101,7 @@
     },
     methods: {
       cellClick(e, cell) {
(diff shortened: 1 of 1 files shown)


rathboma:
Great fix thanks! I'd never have found that. :-)

Issue #606, opened 2021-03-21
Can't execute the query :(


MohamedJakkariya:
**Describe the bug**
I just open my studio and try to execute the following query it's thrown some error like shown below.

```
TypeError: Cannot read property 'map' of undefined
```

I think this happened after update the latest release. Before it was working pretty fine.

**To Reproduce**
1. Just opened studio.
2. Click on connect to the remote database server.
3. Open the query tab and execute the following query.

```
SET FOREIGN_KEY_CHECKS = 0
```
Error : 
![error_1](https://user-images.githubusercontent.com/66210702/111896506-22b8ab80-8a40-11eb-86bb-a86885ca09f0.PNG)
![image](https://user-images.githubusercontent.com/66210702/111896516-306e3100-8a40-11eb-9cb5-c3973b0685aa.png)

**Expected behavior**
Just need to disable my foreign key check but can't do it.

**Screenshots / Gifs**
I'll add screenshot along with it.

**Version Information (please complete the following information):**
 - OS: [Windows 64 bit]
 - App Version [1.10.1]

rathboma:
This was fixed in #607! I'll do a point release today

MohamedJakkariya:
Thank you so much because of this issue I downgraded my version. 

Hopefully, It won't happen again I think 👍

PR #607, opened 2021-03-21
[Issue #606] undefined check before calling map

MarcelloCalligaris:
It's my first time using this application, and I ran into the same issue as [described in issue 606](https://github.com/beekeeper-studio/beekeeper-studio/issues/606) while performing simple statements like `use personal_finance;`

I made this quick fix to get it working on my end and thought I'd share; however I only did minimal testing with some `use,` `select`, `create`, and  `insert` statements and did not notice any side effects.


rathboma:
Thanks!

PR #658, opened 2021-05-19
feat: directly open url or file

### LINKED ISSUE #629: FEAT: Allow double-click on a .db or .sqlite file to open the file in Beekeeper Studio
I want to be able to work with sqlite as easily as with a csv. With a csv, I double-click it, and I can see it.

I'd like to be able to use macos's native "open with" for db files. For example:

1. Right-click on a file, press Open with..., have "Beekeeper Studio" appear as an option. When I click it, it opens and connects to the db.
2. Open the inspector for a file, select Beekeeper, and choose "Always open with"
3. In a terminal, `open -a Beekeeper\ Studio somefile.db` and have it work

**Additional context**
I tested with [chinook.db](https://cdn.sqlitetutorial.net/wp-content/uploads/2018/03/chinook.zip). 
Changing the filename to `chinook.sqlite` did not help. 
Selecting a connection type and then selecting the file did work.

branchvincent:
This registers (*macOS only*) event handlers for:
* [Opening a url](https://www.electronjs.org/docs/api/app#event-open-url-macos)
* [Opening a file](https://www.electronjs.org/docs/api/app#event-open-file-macos)

Enabling direct connections to a database via:

```shell
$ open postgres://host/database # or any other supported url scheme
$ open path/to/sqlite.db # or double-clicking the file in Finder
```

Closes #629

### Caveats

1. I'm not sure if this is the desired approach, notably because it only benefits macOS users. However, it was the most straightforward and what I was familiar with as a [TablePlus](https://tableplus.com/) user.
2. I'm not familiar with Vue so I know I missed some obvious things 😄 please point me in the right direction


src/background.ts
```diff
@@ -107,6 +107,20 @@ app.on('ready', async () => {
   createFirstWindow()
 })
 
+// Open a connection from a file (e.g. ./sqlite.db)
+app.on('open-file', (event, file) => {
+  event.preventDefault();
+  const win = getActiveWindows()[0];
```
branchvincent, line 113: This should probably be creating a new window?. I'm also not handing this:

> open-file is also emitted when a file is dropped onto the dock and the application is not yet running. Make sure to listen for the open-file event very early in your application startup to handle this case

rathboma: Hey! Yes I think it should definitely open a new window. I can do that.

rathboma:
Amazing PR! I'm going to tidy it up a little

vue.config.js
```diff
@@ -56,6 +56,50 @@ module.exports = {
...
           }
         ],
+        fileAssociations: [
+          {
+            name: "SQLite File",
+            ext: ["db", "sqlite", "sqlite3"],
+            role: "Editor"
+          }
+        ],
+        protocols: [
```
rathboma, line 66: Looking at the docs, it seems like `protocols` is meant to be a part of a fileAssociation?

I might be wrong (the docs are confusing). Have you tested this as working?

rathboma (commented):
Looks great. I tidied some stuff up with file opening and creating new windows.

One outstanding question though -- (does it work?)

vue.config.js
rathboma: I think you have it right, looking at this commit - https://github.com/electron-userland/electron-builder/blob/28cb86bdcb6dd0b10e75a69ccd34ece6cca1d204/packages/app-builder-lib/src/options/PlatformSpecificBuildOptions.ts

branchvincent: yes this was working for me! I built it and tested locally, as well as manually inspected the generated `Info.plist`.

Actually, I was very confused at first as well, I think the docs are misformatted. I had to read the [source](https://github.com/electron-userland/electron-builder/blob/28cb86bdcb6dd0b10e75a69ccd34ece6cca1d204/packages/app-builder-lib/src/options/PlatformSpecificBuildOptions.ts#L74-L81) to confirm.

branchvincent: we are not the only ones confused 😂: https://github.com/electron-userland/electron-builder/issues/5286

rathboma: lol yes.

rathboma:
I'm not sure my background.ts changes actually work, so going to test those, then if it's good I'll merge!

rathboma:
@branchvincent can you test this on your mac to see if it still works? It should even work when dragging items to the dock item.

branchvincent:
sure thing!

rathboma:
Thanks! I'm testing linux/windows -- it has to parse the `process.argv` to open files here. No idea how protocols work on linux

branchvincent:
just tried by double click a file, and wasn't able to get it working. logging `query` [here](https://github.com/branchvincent/beekeeper-studio/blob/df7522ce8b16f0f759d4999d7f892ee2be2ab902/src/App.vue#L49) gives me
```javascript
query: {"?file":"/Users/Branch/Code/branchvincent/beekeeper-studio/dev/sakila.db"}
```

so, it's including the `?`

rathboma:
lol, yes this is a typo, used the wrong package.

I'll be fixing that tonight with working linux and windows. I'll ping you when I push :-)

rathboma:
should work now!

branchvincent:
works great for me! only thing I did notice is that `new ConnectionString(url)` [here](https://github.com/branchvincent/beekeeper-studio/blob/cabc7bd48456114b799b48617406f35fe7fe5a73/src/common/appdb/models/saved_connection.ts#L220) will fail when the url is a file with a space in the name. not sure how you want to handle that

src/background.ts
```diff
@@ -104,9 +108,40 @@ app.on('ready', async () => {
...
+  }
 })
 
+// Open a connection from a file (e.g. ./sqlite.db)
+app.on('open-file', async (event, file) => {
+  event.preventDefault();
+  console.log('opening file')
+  const settings = await initBasics()
+  
+  await buildWindow(settings, { url: file })
```
branchvincent, line 134: maybe we can do something like this?
```suggestion
  await buildWindow(settings, { url: `file://${encodeURI(file)}` })
```

rathboma:
@branchvincent That would only work for mac -- for linux and windows I have to just parse `process.argv`, so I get no indication if it is a url or a path.

I think the solution I have now is opinionated `enough` to work. If a file doesn't end in `.db` or such, it assumes it is not a sqlite file, unless it has ://.

It's a little hacky, but I think it's probably 99% good enough for all platforms.

The platform differences suck.

src/background.ts
rathboma: that only works for mac unfortunately. I think we have enough of a solution

rathboma:
@branchvincent if this works well for you, let me know and I'll merge it! You can also remove `draft`

src/App.vue
```diff
@@ -42,13 +45,46 @@ export default {
...
+
+
+    if (this.file) {
+      try {
+          await this.$store.dispatch('openFile', this.file)
+      } catch (ex) {
+        console.error(ex)
+        this.$noty.error(`Error opening ${this.file}: ${ex.message}`)
+      }
+    }
```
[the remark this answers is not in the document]
branchvincent: (btw, happy to help by removing that if it's indeed not needed!)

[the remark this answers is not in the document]
branchvincent: went ahead and removed what i thought was unneeded: 722d565
lmk if anything looks off 😅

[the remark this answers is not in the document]
rathboma: Looks great, thanks!

PR #684, opened 2021-06-08
Connection ssh label

ghstng:
Sidebar connection labeling when ssh display
`bastion > ssh > host:dbname`


src/common/appdb/models/saved_connection.ts
```diff
@@ -131,19 +131,26 @@ export class DbConnectionBase extends ApplicationEntity {
...
     } else {
-      let connectionString = `${this.host}:${this.port}`;
       if (this.defaultDatabase) {
         connectionString += `/${this.defaultDatabase}`
       }
       return connectionString
     }
   }
 
+  get bastionHostString() {
```
rathboma, line 146: why do you need these? Can't you access `sshBastionHost` directly?

src/components/sidebar/connection/ConnectionListItem.vue
```diff
@@ -9,8 +9,18 @@
     >
       <span :class="`connection-label connection-label-color-${labelColor}`"></span>
       <div class="connection-title flex-col expand">
-        <span class="title">{{label}}</span>
-        <span class="subtitle">{{subtitle}}</span>
+        <div class="title">{{label}}</div>
+        <div class="subtitle"> 
+          <span class="bastion" v-if="this.config.sshBastionHost">
+            <span class="truncate">{{ subtitleBastion }}</span>&nbsp;>&nbsp;
```
rathboma, line 15: no need for the computed properties, just use `this.config.sshBastionHost` here too

rathboma (changes requested):
I don't think you need the getters or the computed properties, lets fix that, then it can be merged.

PR #692, opened 2021-06-14
Folder-System

No1Lik3U:
I created a rename button for the favorite queries which are saved.
They can easily be renamed by right-clicking them and then choose rename with the new name.

if there is any inefficiency or error please let me know.

And I apologize to write the function on the master branch it was not intended.


rathboma:
This looks awesome! I'll review tomorrow AM :-)

No1Lik3U:
> This looks awesome! I'll review tomorrow AM :-)

Thanks, mate. I have a idea that I am currently trying to implement which is that we have a explorer structure in the editor. This will have two advantages first of all we will have a folder structured query system and second if you think fast forward some months/years then we can also create a database-design tool that is connected to the query system make it possible to create also, beside the .query "file types" also .design for example. 

Just one question. Can you maybe send me the ORM we are using on this project because I could not find it. 
Thank you.

rathboma:
Oh, nice, so folders for queries? Yeah we've been wanting to work on that, let us know how it goes!

The ORM is TypeORM - https://typeorm.io/

rathboma:
So I did a review - I think generally the flow looks good, but I think you'd simplify things a lot if you moved the rename functionality to it's own component, then you don't need to track refs with ids, or anything like that, the whole thing is self-contained.

You can see we do this a lot with components like `table-list-item`.

So I'd create something like `QueryNameEditor` or similar.

No1Lik3U:
> So I did a review - I think generally the flow looks good, but I think you'd simplify things a lot if you moved the rename functionality to it's own component, then you don't need to track refs with ids, or anything like that, the whole thing is self-contained.
> 
> You can see we do this a lot with components like `table-list-item`.
> 
> So I'd create something like `QueryNameEditor` or similar.

Thanks for the review I will implement it with the folder system.

No1Lik3U:
> So I did a review - I think generally the flow looks good, but I think you'd simplify things a lot if you moved the rename functionality to it's own component, then you don't need to track refs with ids, or anything like that, the whole thing is self-contained.
> 
> You can see we do this a lot with components like `table-list-item`.
> 
> So I'd create something like `QueryNameEditor` or similar.

Hi mate its me again. Just a quick question. Where do we save data that is needed even though the application was closed (persistence)?

rathboma:
Hey! The data is saved in a SQLite database, we use TypeORM as the object mapping layer.

rathboma:
Let me know when you're done with this and I'll take another look!

No1Lik3U:
> Let me know when you're done with this and I'll take another look!

Thanks I am still on it got a lot done. I am now trying to make rename functionality work as you suggested it and then I will try to work ob the saving part of the queries. Thanks for the help :)

No1Lik3U:
I am finished with the first version of the foldertree structure but there is still a lot to do. I just want to show how far I am and don't build upon some bad fundamental. Thanks for the review.

rathboma:
Ok, I'll take a look tomorrow!

rathboma:
Hey! I took a look at the code. I see what you're trying to do - manage files on the filesystem rather than the database?

I definitely think we need a way to manage files on the filesystem, but I don't think we want to replace all saved queries with an explorer type thing.

As we move to add plugins, query syncing, and such, it'll be a benefit to have the queries in the database. Also having them on the filesystem means we need to implement a file watcher to check routinely for changes which is also a lot of work!

I'm super happy to merge a PR with your initial changes - renaming a query, but I think we'd need to unwrap this explorer work before that.

If you are interested in continuing to add an explorer pane, lets do it as a separate global tab to begin with, so that we don't break current saved query behavior!

No1Lik3U:
Thanks for the review! I can rewrite that way but i have got an idea. We can still save the files in the database but have a intern path system with the explorer. This will have the benefit of nice looking and organized queries but also clear out the points you have mentioned as concerns. Let me think what you think about that system.

No1Lik3U:
Hi mate its me again. Sorry to get on your nerves. Could you mabye tell me what the meaning of the column "valueType" is in the table user_setting?

Thanks.

PR #701, opened 2021-06-19
Clarified When Results are Truncated

### LINKED ISSUE #693: BUG: no indicator that results are truncated
The UI should tell you that results are truncated to 50,000 when that happens

JaceBayless:
Close #693 

When results are truncated, the result count now shows as "${truncated row count} / ${total row count}" and the tooltip message shows that the results were truncated.

When not truncated:
Screen Shot 2021-06-18 at 10 39 47 PM

When truncated:
Screen Shot 2021-06-18 at 10 31 20 PM


rathboma:
Fantastic, thanks!

PR #706, opened 2021-06-23
Added the ability to refresh again

### LINKED ISSUE #700: FEAT: Bring back the quick refresh functionality
In older versions, you could click on the clock icon as seen in the picture below, and it would refresh the current table.
refresh

Now this feature has been moved to a sub-menu and there is one extra step to refresh a table: 
refresh-new

There is absolutely no reason why would you make such a useful feature harder to access.

JaceBayless:
Close #700 

Clicking the time since last refreshed now refreshes the table again.


rathboma:
:tada:

PR #736, opened 2021-07-23
Update saved_connection.ts

vitaly-t:
In this day and age, conditional chains should be used properly :wink:


rathboma:
holy crap I didn't know about this syntax. TIL.

PR #742, opened 2021-07-27
tools: fix wrong mssql conversion of bit datatype

emersonsm:
SQL Server bit is an integer data type unlike others databases.
I have added an exception for mssql returning original value.

Before:
before

After:
after


rathboma:
Thank you!

Issue #714, opened 2021-07-02
Azure SQL Server - Tables will not load


cj-alert:
I create a connection.
I test the connection (success).
I connect (success).
I see a list of tables, etc.
I try to open a table, and receive the error: "error loading data. see error below"
I see below: "Cannot read property '1' of null."
[1625266135980.log](https://github.com/beekeeper-studio/beekeeper-studio/files/6757338/1625266135980.log)

rathboma:
Can you let me know exactly what type of database you started on Azure? A screenshot of the specific page and the configuration you are using would be very helpful.

cj-alert:
Hi. It's was a s0 sql server database on azure. Standard configuration.
What would you like a screenshot of? Beekeeper or azure? Thanks so much!

rathboma:
This is fixed for the next release!

PR #757, opened 2021-08-08
Fix tables not loading on Azure SQL Server

Kalissaac:
I also encountered the issue in #714, so I took a look and it seems to occur when Beekeeper runs the `SELECT @@VERSION as version` query. Azure runs a specialized version of SQL Server so the output of that command is along the lines of:
```
Microsoft SQL Azure (RTM) - 12.0.2000.8 
	Jul 23 2021 13:14:19 
	Copyright (C) 2019 Microsoft Corporation
```
Which is different from what the version parsing RegEx is expecting (`/SQL Server (\d+)/g`), resulting in a null value since no matches are found.

This PR fixes it by ignoring the version number parse result if it's null and instead falling back to the hardcoded `2017` value in the code. I'm not sure what effect (if any) this will have on features that rely on a specific SQL Server version.


rathboma (approved):
Thank you!

PR #762, opened 2021-08-11
Copy Row As Insert Statement

sridharswain:
Added a feature to copy a row as insert statement.

UI Changes:
1. Added a button on right click menu on table page and with label `Copy as insert`.

Creates insert statement and registers it to clipboard of the system.

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -752,9 +752,7 @@ export default Vue.extend({
     },
     cellCopyRowAsInsert(e, cell) {
       const row = cell.getRow()
-      console.log(row)
       const data = { ...row.getData() }
-      console.log(data)
       const payload = {
         table: this.table.name,
         schema: this.table.schema,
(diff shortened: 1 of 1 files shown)


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -744,6 +750,33 @@ export default Vue.extend({
...
+      const data = { ...row.getData() }
+      const payload = {
+        table: this.table.name,
+        schema: this.table.schema,
+        data
+      }
+      const insertQuery = this.createInsertStatement(payload);
+      this.$native.clipboard.writeText(insertQuery);
+    },
+    createInsertStatement(payload) {
```
rathboma, line 764: Nice! Love this functionality. However I'm trying to avoid building SQL in the front-end as it doesn't always work the same across all databases.

Would you be open to adding a new method to the `Connection` interfaces to support this? Something like `buildInsertStatements(table, values, schema?)`, it would need implementing on postres, sql-server, mysql, and sqlite, although probably a generic version can be created and shared between them.

rathboma (changes requested):
Let me know if you're able to do this extra stuff?

azmy60:
I would like to help!

rathboma:
@azmy60 I'd love your help! With the new right click menu we could easily add a `copy as insert`, so long as the insert code is generated by the client code. Is this something you're able to look at doing?

azmy60:
I would love to :grin:! I will look into it this afternoon. This feature is really helpful when you need a quick look at a query instead of exporting a table.

PR #772, opened 2021-08-20
depending on libayatana-appindicator3-1 

### LINKED ISSUE #562: BUG: missing dependency on Debian 11 - bullseye
**Describe the bug**
Missing dependency on libappindicator3-1 in Debian 11

**To Reproduce**
1. Attempt to install as per the readme on a Debian 11 machine

**Expected behavior**
It should install.

 - OS: Debian 11 bullseye

**Additional context**
It appear that the package was dropped and that you are supposed to use libayatana-appindicator instead.
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=895037

janusheide:
resolves #562 

using libayatana-appindicator3-1 on Ubuntu and Debian instead of obsolete libappindicator3-1

### DIFF
--- a/apps/studio/vue.config.js
+++ b/apps/studio/vue.config.js
@@ -23,7 +23,7 @@ module.exports = {
 
           .options({
             presets: [
-              ['@babel/preset-env', { 
+              ['@babel/preset-env', {
                 modules: false,
                 targets: {
                     esmodules: true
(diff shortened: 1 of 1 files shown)


rathboma:
looks good. Seems to be on Ubuntu as far back as 18.04. Is this in Debian 10?

rathboma:
ok, looks like it's on debian 10, but not 9. I think that's a fine compromise.

PR #775, opened 2021-08-23
Integrated-Explorer

No1Lik3U:
Hi its me again. 

To get started here I am pretty new to this whole open-source-collabaration-system but I am trying my best to be as efficient as possible. I started the beekeeper project with the simple idea of renaming the favorite_query that are being saved after they have been saved. But then I started to think further and created a whole explorer to have a directory and query structure based on the windows explorer which was rejected (absolutely justified) because it was not really well suited for future plugins. After I got this message I started the whole projected again but this time with the intention to integrate the whole system and not building my own thing (abvioulsy its the better solution). So now I can gladly present you the first prototype of the explorer system fully functioning with the already built in favorite_query system.

I also added workspaces into the whole mix to have even more organization power. But remember thats only the beging. For the future (after this system has been merged with the real software) I was planing on to do:

1. Drag-system for the directories
2. User-Prefences
3. automatic save intervalls 
4. Opening all the files and directories as they have been left in the last work session
5. and more ...

I hope I dont get on your nerves with this and I could build something that is also going with the future ideas.


rathboma:
Hey! I know you've done a lot of work on the filesystem explorer in this PR and that there is a pretty big merge conflict now due to how files have been rearranged for the new monorepo.

I appreciate all the work and vision you've put into this feature, but I honestly don't think we want to have a filesystem browser built into Beekeeper Studio. We're moving towards offering remote syncing of queries, and this just isn't as easy if we don't own the data storage.

If I'm misunderstanding what this PR does, please let me know. It's possible we could re-use some of this work as a 'sql importer' or something similar, but I think it's unlikely we'll ever have a full filesystem browser like VSCode.

Matthew

No1Lik3U:
Hi. Thanks for the answers. Can you maybe explain what you mean by "[...] and this just isn't as easy if we don't own the data storage".

rathboma:
So if users can edit and move around files on the filesystem it becomes a lot harder to manage those in the app, as we have to account for many more change patterns

No1Lik3U:
> So if users can edit and move around files on the filesystem it becomes a lot harder to manage those in the app, as we have to account for many more change patterns

I understand. I would say that this problem could be prevented if we dont allow the moving aroung the files. because we have two seperate tables (directories and files) we know all about the files and always know about the parent directory. But I can understand what you mean by this if its not worth the challenge then you shouldnt take it into the project. But I would say that this folder-system would be a great way to organize the whole structure from the user perspective. You could also but this as a fourth tab. That would habe the advantage that this tab would be in the beta phase and we could see how the whole thing is developing.

No1Lik3U:
.

Issue #372, opened 2020-10-07
FEAT: automatically sort saved connections alphabetically


evs-xsarus:
Currently, the saved connections are sorted on creation order. Could this be changed to have them sorted alphabetically?

rathboma:
Two options:
- created date
- name

This choice should persist.

phillpafford:
I would group by color as well, or even a drag and drop list order would be good. as I tend to group database connection for environments by color, it's an easy visual to know I'm on a production connection if I see Red

AlexandreHiroyuki:
Hi. I would like to work on it.

rathboma:
Awesome!

It should be a fairly self-contained feature, and we can tidy up the UI easily afterwards, so no need to make it 'perfect'.

nmelnick:
Hi, all! I've done a little work on this to solve for it locally, since it was bothering me. :) @rathboma With you working on #440, would it still be worth it do try a PR for this? If so, if I have some questions about the UI, would you prefer to have it within that PR? Thanks!

PR #782, opened 2021-08-28
Persisted sorting for the connection list

nmelnick:
This PR is to address issue #372 .

I'm new to Vue and this project, so this PR is open to feedback, and if there is ongoing work that replaces this, I'm happy to let it go, too.

The default sort order for the connection list is by ID, which kinda correlates to when it was created. This becomes unwieldy over time. This PR adds a new button to change sort order, and persists the sort order in settings on each change. I added to the initial list of sort options in the issue, providing:

* Color (labelColor)
* Created (id)
* Name (name)
* Connection type/protocol (type)


nmelnick:
I see you were busy! Congratulations on 2.0, @rathboma! :)

I have merged in the latest changes from master and tested.

apps/studio/src/migration/20210811_add_sort_to_settings.js
```diff
@@ -0,0 +1,16 @@
+export default {
+  name: "20210811_add_sort_to_settings",
+  async run(runner) {
+    const queries = [
+      `
+        INSERT INTO user_setting (key, defaultValue, valueType)
+        VALUES
+        ('sortOrder', 'id', '0')
```
rathboma, line 8: could we call this `connectionSortOrder` rather than `sortOrder`? That way we can have a different sort order for other stuff.

apps/studio/src/store/modules/settings/SettingStoreModule.ts
```diff
@@ -54,6 +54,10 @@ const SettingStoreModule: Module<State, any> = {
     menuStyle(state) {
       if (!state.settings.menuStyle) return 'native'
       return state.settings.menuStyle.value
+    },
+    sortOrder(state) {
```
rathboma, line 58: I think this adds much needed functionality. But it does mean that connections cannot be dragged / dropped for sorting. Do you think that's a good trade off?

rathboma (changes requested):
one minor change, one higher order question

rathboma:
Know what? I think this is good enough to merge.

nmelnick:
@rathboma Hah -- sorry, didn't get to those notifications in time! I'll do a follow up PR taking your comments into account, I think they're all good. I think allowing a 'manual' sort order makes a ton of sense, I was just blinded to it because I didn't use it.

rathboma:
It's all good! I think this is very useful as-is. Don't worry about the manual sorting, we'll be adding folders soon so that will require some changes anyway.

PR #834, opened 2021-10-07
should support clear text plugin out of the box

rathboma:
Maybe this should be opt-in?

@MasterOdin what do you think? Should this just be supported out of the box if a Server requests authenticating in this way?

For context:
https://github.com/sidorares/node-mysql2/issues/438

https://github.com/sqlectron/sqlectron-db-core/issues/21

### DIFF
--- a/apps/studio/src/lib/db/clients/mysql.js
+++ b/apps/studio/src/lib/db/clients/mysql.js
@@ -786,6 +786,13 @@ function configDatabase(server, database) {
     supportBigNumbers: true,
     bigNumberStrings: true,
     connectTimeout  : 60 * 60 * 1000,
+    authPlugins: {
+      mysql_clear_password: pluginOptions => ({ connection, command }) => {
+        const password =
+          command.password || pluginOptions.password || connection.config.password;
+        return Buffer.from(`${password}\0`)
+      }
+    }
   };
 
   if (server.sshTunnel) {


MasterOdin:
I am of the belief that the user should have to opt into sending clear text passwords for it to work. Having it on by default makes Beekeeper at default less secure option for connecting to MySQL databases (especially in a possible MiTM situation) if it's not actually necessary.

Could be made easier than using an environment variable (e.g. checkbox in the UI) though to do so.

MasterOdin:
Datagrip on the other hand requires you to explicitly opt-in.

MasterOdin:
A follow-up here would be asking "how many users need this setting enabled?" vs total number of mysql connections. It's not exactly an oft requested thing in my experience.

ethan-dowler:
@rathboma YES! I am able to log in to my AWS RDS Postgres DB instance (that requires the cleartext plugin!) with [the build from this PR](https://github.com/beekeeper-studio/beekeeper-studio/actions/runs/1318025218). This would fix #834 . I still think cleartext should be an option instead of default, but this would huge for my team.

silverbullettruck2001:
@rathboma is this PR going to be merged anytime soon?

arthurju:
I'm also really interest by this merge !
Any news ?

luisfix:
Have there been any updates on this?

silverbullettruck2001:
@luisfix @rathboma there is a PR out ([PR](https://github.com/sidorares/node-mysql2/pull/1552)).  I believe @sidorares is working on tests?

rathboma:
Yes I saw that! I'd much rather rely on an upstream implementation vs my own. Wish I had enough time to push all these things forward myself!

@sidorares :pray: thanks for your work on the plugin

karloluis:
Did this get a resolution?

engineerpatrickedward:
Hi Team, Is there any update regarding this?

karloluis:
Did we miss some update? Seeing this be closed without it is confusing.

Issue #926, opened 2021-12-22
QUESTION: How to turn off spell check?


EdenHell:
**Details:**
When I was editing in the form, a red wavy line appeared. I think this is a spell check of chrome. I did not find a switch to turn off the spell check in the program. If not, I hope to add this option.

**Version Information**
Operating System: Windows 11
App Version: 2.1.5

**Screenshot**:
![image](https://user-images.githubusercontent.com/5779655/147056040-1660bd80-979f-4f29-a9c4-be32d7577b55.png)

rathboma:
This is done!

PR #929, opened 2021-12-23
Turn off built-in support for speckcheck #926

Likenttt:
Hi, I think the spellcheck is really annoying.  
Also, I saw someone reported it in the issues. https://github.com/beekeeper-studio/beekeeper-studio/issues/926
Following the docs from electron, I disabled the spellcheck when the window is created. 
https://www.electronjs.org/docs/latest/tutorial/spellchecker


rathboma:
Thank you!

PR #1020, opened 2022-02-07
Can't edit text inputs after window.confirm - dialog called from the main process

przemyslawkleszcz:
Found a problem with window.confirm:

beekeeper

Since this is the function that blocks the execution thread of the script - I modified the solution to use the electron dialog.

### DIFF
--- a/apps/studio/src/background.ts
+++ b/apps/studio/src/background.ts
@@ -67,11 +67,25 @@ async function initBasics() {
   log.info("Building the window")
   log.info("managing updates")
   manageUpdates()
+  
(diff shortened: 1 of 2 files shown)


rathboma:
Thanks for spotting this! This seems to be a weird windows problem, so I'm going to implement it with a modal in app, rather than dealing with native dialogs. I'll tag you on the commit!

rathboma:
Thanks for finding this! https://github.com/beekeeper-studio/beekeeper-studio/commit/991afe01b6e6bc91b607b4f0c70d388ce3b250aa

PR #1044, opened 2022-02-18
Improve PostgreSQL create table script column options

lafriks:
* Automatically detect SMALLSERIAL/SERIAL/BIGSERIAL column options
* Add default values
* Add basic support for identity columns (without sequence options)


rathboma:
Thanks!

PR #1045, opened 2022-02-18
Bump node-sass and sass-loader to support compilation on node v16

lafriks:


lafriks:
There is just no such file, I don't know why it did not catch them earlier in other theme this part was already commented out. Is it really needed?

rathboma:
Oh! I guess the previous node-sass either didn't care about that or used a different resolution method.

Does the file not exist, or is the path wrong? If the path is wrong I guess it just needs correcting, otherwise, yeah just kill that import.

lafriks:
yes, there is no such file, see: https://github.com/beekeeper-studio/beekeeper-studio/blob/master/apps/sqltools/src/assets/styles/themes/light/theme.scss#L2
and https://github.com/beekeeper-studio/beekeeper-studio/tree/master/apps/sqltools/src/assets/styles/app/vendor

Issue #1034, opened 2022-02-14
BUG: Unable to boot with Fedora libva error


matdave:
**Describe the bug**
The application is unable to boot in Fedora 35 with Nvidia GPU

**Logging**
./Beekeeper-Studio-3.1.0.AppImage 
11:04:48.024                     › initializing background
running migrations
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
[9842:0214/110449.975335:FATAL:gpu_data_manager_impl_private.cc(415)] GPU process isn't usable. Goodbye.
Trace/breakpoint trap (core dumped)

**Temporary Solution**
I am able to get it to run if i add the argument ` --in-process-gpu` 

**Version Information (please complete the following information):**
 - OS: Fedora release 35 (Thirty Five) x8
 - App Version 3.1.0
 - Nvidia Driver 510.47.03

rathboma:
I think this is related to Wayland, judging by the internet breadcrumbs I see. Something about Chromium not supporting something but the driver not knowing what to do.

rathboma:
eg - https://bugzilla.redhat.com/show_bug.cgi?id=1491186

There should be a way to run Beekeeper Studio natively in Wayland (not XWayland) soon as that is merged into upstream chromium now I think

rathboma:
Don't think there's really much I can do until then.

rathboma:
Oh yeah Nvidia is notorious for terrible wayland support. I don't think they even supported wayland in their drivers until 2022.

I'd suggest running your desktop with X for a while if you're using Nvidia drivers.

Going to close this issue as I don't think I can really debug this.

rathboma:
@matdave there will be a linux build created as part of #1094 - can you test that and let me know if it fixes the issue?

Apakottur:
Just adding that I've just encountered the error on a fresh install of Ubuntu 22.04 with NVIDIA driver 510 on app version 3.2.1.

Vam-Jam:
You can get it to run on wayland by adding in some arguments. 

```
--no-sandbox --ignore-gpu-blocklist --disable-features=UseOzonePlatform --enable-features=VaapiVideoDecoder --use-gl=desktop --enable-gpu-rasterization --enable-zero-copy
```

PR #1094, opened 2022-03-05
Update Electron to latest 13.x release

cgwyllie:
Fixes https://github.com/beekeeper-studio/beekeeper-studio/issues/1034

Electron 13 breaking changes are minimal, from their release notes:

- Fixed so window.open() parameter frameName is no longer set as window title. #27481
- Changed session.setPermissionCheckHandler(handler) to allow for handler's first parameter, webContents to be null. #19903

See: https://www.electronjs.org/releases/stable?version=13&page=6#13.0.0

No usages of those APIs appear to be in this project.

EDIT: I see there were more extensive breaking changes planned: https://www.electronjs.org/docs/latest/breaking-changes#planned-breaking-api-changes-130

---

Hi! I ran into the same error referenced in #1034 and did a little research for similar issues in other Electron apps.

There was a thread in Ubunbtu's tracker about a similar issue, that appears to be an upstream Chromium problem that was patched.

Seems like the Electron team backported the fix for the issue, but not into the 12.x tree.

See https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1944468.

(EDIT: Not actually sure _which_ change it was that was needed, the errors in that thread are a bit different)

Because I'm not very familiar with the project, I'm not sure if there's anything else that needs updated along side the Electron library.

The only change from the app I can see is that the dev tools are now open by default when launching the `electron:serve` command.

I've not given the app a very thorough test as yet, so let me know what I can do to help get this released :)

P.S. on Fedora I'm using an X11 session, not Wayland, so not 100% sure it's identical to the other issue, however I do have the same Nvidia driver installed, and the `--in-process-gpu` did resolve for me before updating Electron.

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -122,7 +122,7 @@
     "@vue/test-utils": "^1.0.3",
     "aws-sdk": "^2.645.0",
     "babel-eslint": "^10.0.1",
-    "electron": "12.1.0",
+    "electron": "13.6.9",
     "electron-builder-notarize": "beekeeper-studio/electron-builder-notarize#master",
     "electron-devtools-installer": "^3.1.1",
     "electron-updater": "^4.2.5",
--- a/yarn.lock
+++ b/yarn.lock
@@ -5508,10 +5508,10 @@ electron-util@^0.14.1:
     electron-is-dev "^1.1.0"
     new-github-issue-url "^0.2.1"
 
-electron@12.1.0:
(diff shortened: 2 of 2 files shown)


rathboma:
Thanks! Let me take a look at this PR.

rathboma:
I just did a once over, it seems...fine? Thanks for the upgrade, these are usually a lot harder :laughing:

PR #1098, opened 2022-03-08
add 'in' condition for table filter

yogski:
### Changes
- Added `IN` option for filtering table
- Changed placeholder a little bit for easier UX

### Notes
I took a bit from raw filtering option and put it as new option.
`IN` should be able to handle multiple values. It is different than other options (equal, less than, like, etc.), so I returned it as string like in raw filter instead of arrays. Added parentheses for wrapping.

### Demo
new-in-filter

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -68,7 +68,7 @@
                   class="form-control"
                   type="text"
                   v-model="filter.value"
-                  placeholder="Enter Value"
+                  :placeholder=builderPlaceholder
                   ref="valueInput"
                 />
                 <button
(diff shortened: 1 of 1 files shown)


rathboma:
Thanks! This looks awesome.

I think thee are a couple of minor things that need changing:

Looks like for `in` filters you are squishing it into a string. I try to avoid doing that in the UI as it can lead to accidental sql injection (eg if you put a ; in your value).

I think if `filter.value` is made an array it should work. It will need a few more changes:

- change TableFilter in src/lib/models, so that the type of `value` is `string | string[]`
- add a test case in `tests/lib/db.ts#filterTests` to cover the `in` condition. This will make sure it works across all databases

Thanks for contributing!

rathboma (changes requested):
see comment above

rathboma:
@yogski checking to see if you have time to make these changes, or do you need me to take the PR over?

ianberryman:
@rathboma I see your comment above, I'll update this PR unless you think a new one would be better lmk

ianberryman:
Duh nm, I've got my own fork

PR #1146, opened 2022-04-05
Set the default port after the database type has been changed

### LINKED ISSUE #1086: BUG: Unable to fully delete port when creating a new connectoin
When creating a new connection, you can not fully clear the 'port' setting because it auto-populates back to 3306 once it is cleared. 

**To Reproduce**
1. Create a new connection
2. Click Port Number input
3. Try and delete the entire number

**Expected behavior**
The port number input should clear and remain clear.

UtechtDustin:
This PR fixes #1086.
The current version set the default port every time the input is empty.
With this PR the default port will be set once while changing the database type.


rathboma:
Thanks!

PR #1153, opened 2022-04-08
Add an option to use a socket

azmy60:
Option to use a socket path instead of host/port for database connections. #300

Peek 2022-04-08 21-19


rathboma:
Wow, this is awesome! Thanks!

Looks like the one thing it is missing is the implementation of the socket connection when a connection is made. Is that something you can add also?

Files you'll need to look at:

- ConnectionProvider.ts - pass through the socket path in the database config (this file translates from IConnection to IDBServer)
- lib/db/clients/{postgres,sqlserver,mysql} - connect via socket rather than host/port, it's a little different for each database type


Some things that probably need changing:
- I actually don't think SQL Server allows socket connections

azmy60:
Thanks for the reply! I will try to fix them.

rathboma:
Great! You can run integration tests by running: `yarn workspace beekeeper-studio test:integration`

azmy60:
Hey @rathboma. I'm running into some issues now. I get [four failing tests](https://gist.github.com/azmy60/35658006e7f349eecfa86449ae333a54) instead of one as in the CI checks. I've tried many different ways and thought it was an environment issue. So I reinstalled my OS :sweat_smile:, but the result is still the same (I'm using Debian 11). Can I get some ideas to solve this? I'll keep working on local commits for now.

rathboma:
Thanks for all the work! I'll review this tomorrow and hopefully get it merged :tada:

rathboma:
I've also enabled CI for your code, so the CI jobs should tell us if there are any test issues

jc00ke:
Awesome, thanks for adding this!

PR #1170, opened 2022-04-16
Added support for selecting multiple rows to delete and clone

Andrey-Ba:
Added support for selecting multiple rows, deleting and cloning them.
select-delete_clone

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -198,7 +198,6 @@ import {AppEvent} from '../../common/AppEvent';
 import { vueEditor } from '@shared/lib/tabulator/helpers';
 import NullableInputEditorVue from '@shared/components/tabulator/NullableInputEditor.vue';
 import { mapState } from 'vuex';
-import tabulator from '../../../../../shared/src/lib/tabulator';
 
 const log = rawLog.scope('TableTable')
 const FILTER_MODE_BUILDER = 'builder'
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -198,6 +198,7 @@ import {AppEvent} from '../../common/AppEvent';
 import { vueEditor } from '@shared/lib/tabulator/helpers';
 import NullableInputEditorVue from '@shared/components/tabulator/NullableInputEditor.vue';
 import { mapState } from 'vuex';
+import tabulator from '../../../../../shared/src/lib/tabulator';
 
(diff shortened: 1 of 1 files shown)


rathboma:
Awesome! Thanks for the PR, I'll review on Monday

ad-on-is:
I just became aware of that one. Maybe you can merge my implementation of the selection, where I added an empty column at the beginning, which also reduces the amount of click-events on the cells.

rathboma:
Yeah let me figure out how to merge these prs together.

I want to make all the tables 'feel like excel', this is a big start to that.

ad-on-is:
I think it'd make more sense for @Andrey-Ba to merge my PR into his, since the multi-row actions have a "more meaningful purpose".

Andrey-Ba:
@ad-on-is I have merged your PR into mine

rathboma:
Going to look at this today. So sorry for the delay (dad visited, then I got COVID :-S)

rathboma (changes requested):
So I've been poking around with this. I think it's a good start but needs some work, in particular I don't think click-to-multi-select is the right model.

I don't know if you have time to do more work on this PR, but here are the things I think need changing. If you don't end up doing these, I'll probably end up doing them as part of a future body of work, albeit much later than now.

1. Clicking on a row header only selects a single row
2. CTRL+Click does the multi select (the current behavior
4. Click row 1, then shift-click row 9 selects rows 1->9
5. Freeze the row header on the left so when you scroll it's still there
6. Add row numbers to the row headers
7. Right clicking a row header should show the normal right-click menu
8. Clicking a cell should cancel the selection
9. Click and drag should work to select multiple rows when CTRL is held down

Outside the scope of this PR:
- Column headers should work the same way as row headers. moving sorting to a separate 'button' within the header.
- Eventually - arbitrary range selection (click and drag a cell / shift&ctrl click), just like Google Sheets

rathboma:
Please let me know if either of you are able to do the additional work. I'll leave this PR open for a while and if neither of you picks it up in the next 2-3 weeks I'll see if I can fit it in.

Andrey-Ba:
Thanks for your feedback! Hope you had a smooth recovery from COVID. Unfortunately, I will not be able to do it in the following weeks as I have few exams left at university.

ad-on-is:
I too am busy right now with other work. Will see when the time allows it.

If this is good enough to have the basics covered, can't it be merged now, and create a new PR for the change-requests?

rathboma:
Closing in favor of #1326, although that is also blocked on a tabulator feature

PR #1171, opened 2022-04-17
Copy Row as Insert

azmy60:
Allow user to copy a row insert query to clipboard.

Peek 2022-04-16 15-38

Tasks:
- [x] Add buildInsertQuery method to connection interface
- [x] Update db clients (mysql, postgres, etc) to use [this helper function](https://github.com/beekeeper-studio/beekeeper-studio/blob/50e8ac31e66ff268acd2a701d5750da515e28531/apps/studio/src/lib/db/clients/utils.js#L146)
- [x] Update dropdown item
- [x] Add tests

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -311,6 +311,18 @@ export default Vue.extend({
           },
         },
         {
+          label: '<x-menuitem><x-label>Copy Row (Query)</x-label></x-menuitem>',
+          action: async (_e, cell) => {
+            const tableInsert = {
+              table: this.table.name,
+              schema: this.table.schema,
+              data: [cell.getRow().getData()],
+            }
+            const query = await this.connection.getInsertQuery(tableInsert)
+            this.$native.clipboard.writeText(query)
+          }
+        },
+        {
           label: '<x-menuitem><x-label>Copy Row (JSON)</x-label></x-menuitem>',
           action: (_e, cell) => {
             const data = cell.getRow().getData()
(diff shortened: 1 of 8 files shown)


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -328,6 +329,17 @@ export default Vue.extend({
...
           action: (_e, cell) => this.$native.clipboard.writeText(Papa.unparse([cell.getRow().getData()], { header: false, delimiter: "\t", quotes: true, escapeFormulae: true }))
         },
+        {
+          label: '<x-menuitem><x-label>Copy as Insert</x-label></x-menuitem>',
+          action: async (_e, cell) => {
+            const exporter = new SqlExporter("./tmp/sql.export", this.connection, this.table, [], { chunkSize: 100 }, {})
+            await exporter.exportToFile();
+
+            const cells = cell.getRow().getCells().map(cell => cell.getValue())
+            const formatted = exporter.formatRow(cells)
```
azmy60, line 339: This might be the simplest way I found to generate an insert query, though it still exports a new file as a side effect.

I'm thinking of extending the Exporter so that probably we can do `new SqlExporter(this.connection, this.table, ...).getExportString()`? What do you think about this problem @rathboma ?

azmy60:
It looks like we already have a helper function for [building an insert query](https://github.com/beekeeper-studio/beekeeper-studio/blob/50e8ac31e66ff268acd2a701d5750da515e28531/apps/studio/src/lib/db/clients/utils.js#L146). Now I'm looking for a way to connect this with the context menu.

rathboma:
This looks awesome, thanks! I'm going to review on Monday :tada:

rathboma:
This looks awesome, thanks! I'm going to review on Monday :tada:

rathboma:
@azmy60 Thank you for making such a thorough pull request! Even the tests look :rocket:

Issue #1174, opened 2022-04-18
BUG: Right click doesn't copy field values


NathanFirmo:
**Sometimes a value is not copied from clipboard**
When you try copy an int id value, an error happens. So the value isn't copied.

**To Reproduce**
1. Open a table and try to copy an int id value using right click
2. Paste on other tab using Ctrl+V
3. Verify if the value was paste correctly

**Expected behavior**
To constitute a bug, the pasted value must to be diferent the one you try to copy.

**Screenshots / Gifs**
As you can see here, when you try copy an int id value, an error happens. So the value isn't copied.

### Example in a Table
![PR_Bekeeper_Error](https://user-images.githubusercontent.com/79997705/163884948-0f6f9067-0066-49e3-81eb-74b2b5dd8deb.gif)

### Example in a query
![PR_Bekeeper_Error_2](https://user-images.githubusercontent.com/79997705/163884953-aa4d0841-7954-4dc1-b4c3-9f366d06f068.gif)

**Version Information (please complete the following information):**
 - OS: Pop_OS! 21.10
 - App Version 3.3.8
 - Database type and version mysql

rathboma:
Thank you for the bug report! I'll look into fixing this right now.

PR #1175, opened 2022-04-18
fix error copying data

NathanFirmo:
# 🐛 Fix error copying data
Related issue #1174

## Error description
As you can see here, when you try copy an int id value, an error happens. So the value isn't copied.

### Example in a Table
PR_Bekeeper_Error

### Example in a query
PR_Bekeeper_Error_2

## Solution preview
On the gif's bellow, you can see that the values are copied correctly, both in the table and in the query.

### Example in a table
PR_Bekeeper_Ok

### Example in a query
PR_Bekeeper_Ok_2


rathboma:
Hey! Thanks for the fix, I used it, but moved the logic to the native module instead

PR #1206, opened 2022-05-03
Allow unpin table from "yellow pin" icon

Guilledll:
Hi, first of all awesome App!

I Just installed is a few minutes ago and I have a little suggestion. I was using the "pin" in some tables and I was expecting to **unpin a table clicking the "yellow pin"**, but nothing happens.
I think it will be nice to have this feature to allow users to **"pin / unpin" faster**, rather than moving to the upper tab to unpin them from the "X" button.


rathboma:
Hey! Hmm that's weird, it should work based on line 12 in your file. Let me double check I didn't break something, but if not I'll merge, thanks for the PR!

UtechtDustin:
@rathboma 
This PR should be merged, can confirm that this is not working (yet).
I'm also not sure what do you mean with ` it should work based on line 12 in your file.`
Line 12 is the item when it is NOT pinned.

Guilledll:
Yeah @rathboma i just entre git, searched where the logic was and copy/paste on the "yellow pin" the method that unpins where clicking the "x".

rathboma:
I've implemented this separately while also adding a few minor style tweaks, but used this code as a base. Thanks for this!

PR #1231, opened 2022-05-19
feat: add toggle for expand/collapse table list

ad-on-is:
beekeeper-expandtoggle

I switched the two icons for a toggle

### DIFF
--- a/apps/studio/src/components/sidebar/core/TableList.vue
+++ b/apps/studio/src/components/sidebar/core/TableList.vue
@@ -245,17 +245,6 @@
         }
         }
       },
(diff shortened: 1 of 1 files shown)


rathboma:
Looks great! @ad-on-is are the `expandAll` and `collapseAll` methods still needed for anything? If not - can you remove them?

rathboma (changes requested):
Remove the collapseAll and expandAll methods if no longer used

ad-on-is:
> Remove the collapseAll and expandAll methods if no longer used

Done!

PR #1250, opened 2022-05-30
Fix can't delete rows with binary IDs

### LINKED ISSUE #839: BUG: can't delete row with `binary` IDs in MySQL
**Describe the bug**
I tried deleting a row of a table. I right-clicked it, clicked on delete row and then on the apply button. Nothing happend

**To Reproduce**
1. install [Shopware](https://www.shopware.com/) locally
1. Open a random table
2. Try to delete a row

**Expected behavior**
I expect that the row is deleted after clicking on apply.

**Screenshots / Gifs**

 - OS: Windows 10
 - App Version: 2.1.4

**Additional context**
Shopware is using `binary` for their IDs and I assume that this is the problem.
I tried it with a similar software, which showed the SQL statement it will run and there I saw something like  `... WHERE id = NULL;`.
Therefore the binary is probably the problem.

Adminer and MySQL Workbench are able to delete rows with binary IDs.

azmy60:
Fixes #839

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -906,12 +906,15 @@ export default Vue.extend({
 
       row.getElement().classList.add('deleted')
 
+      const column = this.table.columns.find(c => c.columnName === pkCell.getField())
+      const isPrimaryKeyBinary = column.dataType.toUpperCase().includes('BINARY')
+
       const payload = {
         table: this.table.name,
         row: row,
         schema: this.table.schema,
         pkColumn: this.primaryKey,
-        primaryKey: pkCell.getValue()
+        primaryKey: isPrimaryKeyBinary ? Buffer.from(pkCell.getValue(), 'hex') : pkCell.getValue()
       }
 
         // remove pending updates for the row marked for deletion
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -505,6 +505,7 @@ export default Vue.extend({
         this.filterMode === FILTER_MODE_BUILDER &&
         this.filter.type && this.filter.field && this.filter.value
       ) {
+        
         return [this.filter]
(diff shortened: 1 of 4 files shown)


apps/studio/src/lib/data/tools.ts
```diff
@@ -42,7 +42,7 @@ export const Mutators = {
    */
   genericMutator(value: any, preserveComplex: boolean = false): JsonFriendly {
     const mutate = Mutators.genericMutator
-    if (_.isBuffer(value)) return value.toString()
```
azmy60, line 45: The return value showed incorrect representation of the real buffer. We need to use a specific encoding like hexadecimals for example.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -906,12 +906,15 @@ export default Vue.extend({
...
+      const column = this.table.columns.find(c => c.columnName === pkCell.getField())
+      const isPrimaryKeyBinary = column.dataType.toUpperCase().includes('BINARY')
+
       const payload = {
         table: this.table.name,
         row: row,
         schema: this.table.schema,
         pkColumn: this.primaryKey,
-        primaryKey: pkCell.getValue()
+        primaryKey: isPrimaryKeyBinary ? Buffer.from(pkCell.getValue(), 'hex') : pkCell.getValue()
```
azmy60, line 917: knex supports handling Buffer object

apps/studio/src/lib/data/tools.ts
Zerotask: Probably toString(16)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toString

rathboma: Yeah hex is the right way to do it. Although probably we need to prefix hex strings with `s/` when displayed (although not edited), this is a convention used in many UIs for hex'd binary values

PR #1272, opened 2022-06-21
Updated deb-postinstall

### LINKED ISSUE #1218: BUG: Installing from .deb uses deprecated apt-key
**Describe the bug**
When installing Beekeeper Studio on a recent version of Ubuntu using the .deb file, it sets up the key for the apt repo using `apt-key`, which has been [deprecated](https://manpages.ubuntu.com/manpages/jammy/man8/apt-key.8.html). Output contains the following lines:

```
Warning: apt-key should not be used in scripts (called from postinst maintainerscript of the package beekeeper-studio)
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
```

**To Reproduce**
Download the .deb and install it.

**Expected behavior**
Installation completes without warnings. The repo key is added to `/etc/apt/trusted.gpg.d`. Or better yet, to `/usr/share/keyrings` and the .list file refers to it using the "signed-by" option:

```
deb [signed-by=/usr/share/keyrings/beekeeper-studio.gpg] https://deb.beekeeperstudio.io stable main
```

 - OS: Ubuntu 22.04
 - App Version: 3.4.2

IndrekHaav:
Hopefully fixes #1218 

Edit: I've little to no experience packaging software for Debian/Ubuntu, I just tested deb-postinstall as a standalone script.

### DIFF
--- a/apps/studio/build/deb-postinstall
+++ b/apps/studio/build/deb-postinstall
@@ -2,15 +2,11 @@
 
 OLD_FILE=/etc/apt/sources.list.d/beekeeper-studio.list
 NEW_FILE=/etc/apt/sources.list.d/beekeeper-studio-app.list
+KEY_FILE=/usr/share/keyrings/beekeeper-studio.gpg
 
-available()
-{
-        command -v "$1" >/dev/null 2>&1
+available() {
+  command -v "$1" &>/dev/null
 }
-
-if [ -f $NEW_FILE ]; then
-  exit 0
-fi
 
 if [ -f $OLD_FILE ]; then
   echo "--> Beekeeper Studio has migrated to a new apt repo, adding it, and removing the old one"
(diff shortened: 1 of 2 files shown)


IndrekHaav:
Note: I used `/usr/share/keyrings/` because that's what the [Debian wiki recommends](https://wiki.debian.org/DebianRepository/UseThirdParty#OpenPGP_Key_distribution):

> If future updates to the key will be managed by an apt/dpkg package as recommended below, then it SHOULD be downloaded into `/usr/share/keyrings` using the same filename that will be provided by the package.

rathboma:
Can we also check the existence of the keyringsdirectory? I know the Debian Wiki only talks about checking /etc/apt/keyrings, but it seems like a sensible check?

rathboma:
I want to merge this, but do you know if this will break the deb install for existing users?

IndrekHaav:
> I want to merge this, but do you know if this will break the deb install for existing users?

It shouldn't. I did the same changes manually on my machine and have been able to update from the apt repo just fine (except of course that my changes get overwritten with each update). But like I said, I haven't tested the entire package with the changes, just the deb-postinstall script.

lode:
Would be great to get this through! I keep getting the warning mentioned.

I there something I can do to help? @rathboma, can you describe the steps I could take to test this?

Btw, I just make one small change which did fix the deb installation & warning. Not sure about the updates later on.

```patch
-wget --quiet -O - https://deb.beekeeperstudio.io/beekeeper.key | sudo apt-key add -
+wget --quiet -O - https://deb.beekeeperstudio.io/beekeeper.key | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/beekeeper.key.gpg > /dev/null
```

rathboma:
Hey all, I know folks get warnings when installing the DEB because of this, but I just don't have the bandwidth to fully test this right now. I'm going to close for the timebeing, but will probably re-open sometime in the future.

Issue #1246, opened 2022-05-26
FEAT: View record in {table} doesn't account for multiple foreign keys.


theOofDoctor:
**Is your feature request related to a problem? Please describe.**
If you have multiple foreign key relations on the same column, the view record tooltip action seems to just pick the last relation as the table to navigate to, even if that key doesn't go to that specific column

**Describe the solution you'd like**
I would like that tooltip to navigate to the proper table, depending on which relation is actually 'active' in that case.

**Describe alternatives you've considered**
You could possibly just show all of the tables that are linked in that tooltip

**Additional context**
In this particular case, we have 3 relations on the 'TargetId' column, with only one of them being linked to the 'User' table:
![image](https://user-images.githubusercontent.com/58712803/170579319-206646fb-9470-451f-87da-64fe667026ba.png)

Whenever one hovers over that column in the data view it only allows you to navigate to the 'User' table:
![signal-2022-05-26-145301](https://user-images.githubusercontent.com/58712803/170578993-c53090da-f2d7-4e8d-8e78-01fc3077fab9.jpeg)

When you get to that table, there is no data to be found:
![image](https://user-images.githubusercontent.com/58712803/170579150-29316422-ebe2-49f9-a61f-a35ea53543af.png)

This does work as expected if the link is to the 'User' table though (I didn't put this as a bug because I'm assuming this just wasn't thought of so it is as designed?)

rathboma:
@theOofDoctor I know this is merged, but would you be able to give me some SQL to create tables like this so I can demo it?

What database supports multiple relations on the same column?

rathboma:
Oh, looks like SQL Server. Either way - some SQL to help me demo it would be amazing!

theOofDoctor:
@rathboma yeah for sure! I'm at work rn but I should be able to get to this tonight

theOofDoctor:
@rathboma Here ya go!

```sql
CREATE TABLE Table1 (Id INT NOT NULL PRIMARY KEY, MultipleFKs INT NOT NULL);
CREATE TABLE Table2 (FK1 INT NOT NULL PRIMARY KEY);
CREATE TABLE Table3 (FK2 INT NOT NULL PRIMARY KEY);
CREATE TABLE Table4 (FK3 INT NOT NULL PRIMARY KEY);

ALTER TABLE Table1 WITH NOCHECK ADD CONSTRAINT FK_TABLE2 FOREIGN KEY (MultipleFKs)
[...]

PR #1274, opened 2022-06-21
Add support for multiple foreign keys on a column in table view.

theOofDoctor:
I made these changes with respect to #1246.

In my day to day use of beekeeper, I have run into this a couple times: I have multiple foreign keys on a column, but beekeeper can only navigate to one of them. I elected to just add a menu to navigate to any of the foreign keys on that column, as evaluating the keys on a row by row basis would probably be super expensive (the example table I have below has two columns with 4 FKs each, and we have thousands of records in production right now).

These are the tooltips that have changed:
tooltips

And here is the clickMenu for columns with multiple FKs.
navigation

The default behaviour for this has not changed. If there is 1 FK, simply clicking the FK button will navigate. But if there are multiple FKs, then you get this click menu and you have to choose which table you'd like to navigate to.

Hopefully this PR is clear enough, but if you have any questions/comments, let me know. I also completely understand if this isn't an ideal solution for you, but I'd be happy to work on it further if required. Thanks for reading!

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -401,7 +401,8 @@ export default Vue.extend({
     tableKeys() {
       const result = {}
       this.rawTableKeys.forEach((item) => {
-        result[item.fromColumn] = item
+        if (!result[item.fromColumn]) result[item.fromColumn] = [];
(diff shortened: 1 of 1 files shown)


rathboma:
Damn, that's awesome. Going to look at it now.

PR #1286, opened 2022-06-28
Show database version in footer

### LINKED ISSUE #1156: FEAT: show database version in the footer
It's useful to get some additional information about the connected database.

Show additional information like the database version at the footer bar
Example:
image

MySQL but if possible it should work for all supported databases.

niklas-stegmann:
I added a way to display the database version.
There is not much space in the footer, so I added a tooltip for the connection type where the version is displayed.

Maybe someone has a better idea?

### Demo
postgres

mariadb

sqlserver

sqlite

Fixes #1156

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -155,6 +155,7 @@ export default async function (server: any, database: any): Promise<DatabaseClie
   return {
     /* eslint max-len:0 */
     supportedFeatures: () => features,
+    versionString: () => getVersionString(version),
     wrapIdentifier,
     disconnect: () => disconnect(conn),
     listTables: (_db: string, filter: FilterOptions | undefined) => listTables(conn, filter),
@@ -1500,6 +1501,10 @@ async function runWithConnection<T>(x: Conn, run: (p: PoolClient) => Promise<T>)
   } finally {
     connection.release();
   }
+}
+
+function getVersionString(version: VersionInfo): string {
+  return version.version.split(" on ")[0];
 }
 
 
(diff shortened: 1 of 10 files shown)


Zerotask:
Would be great to see a summary directly in the footer, e. g. PostgreSQL 14.8

rathboma:
I think this looks great. I'd actually like to utilize the window titlebar more for this type of information, although that doesn't really work on Mac, one step at a time. :-)

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1502,6 +1503,10 @@ async function runWithConnection<T>(x: Conn, run: (p: PoolClient) => Promise<T>)
   }
 }
 
+function getVersionString(version: VersionInfo): string {
+  return version.version.split(" on ")[0];
```
rathboma, line 1507: Does this work properly for Cockroach DB and Redshift do you think? Maybe it should return the full version for those?

apps/studio/src/lib/db/clients/sqlserver.js
```diff
@@ -37,8 +37,11 @@ export default async function (server, database) {
   // light solution to test connection with with the server
   await driverExecuteQuery(conn, { query: 'SELECT 1' });
 
+  const version = await getVersion(conn);
```
rathboma, line 40: You can probably delete the previous line (SELECT 1) as getting the version will test the connection also

rathboma:
@niklas-stegmann Question - can you double check cockroachdb & redshift versions will split properly given your `getVersion` function? I think the version strings are listed online in a bunch of places. Maybe add a comment with examples from each.

niklas-stegmann:
@rathboma here are the results:

**CockroachDB:**

full version string:
image

displayed version:
image


**AWS Redshift**

full version string:
image

displayed version:
image

Issue #884, opened 2021-11-18
FEAT: IN operator


gmcrocetti:
Hi all. Congrats for the amazing software :)

### Is your feature request related to a problem? Please describe.

It would be amazing to have the `IN` operator available in this drop-down list:
![Screenshot_20211118_152729](https://user-images.githubusercontent.com/24530683/142475332-dc362d7e-9e85-431a-ae0f-060f3b632d2b.png)

---

### Describe the solution you'd like

Have the `IN` operator available as a option.

---
### Describe alternatives you've considered

Write a raw query by myself.

---

rathboma:
Right now the best way to do this is to click the `<>` button and write a clause in SQL. I realize that's not perfect, but it is the easiest.

I think this would be a good beginner ticket for someone actually.

udhaykumarbala:
Hey, I am interested in working on this issue. can you please help me get started on the final requirement

I think "tableview/TableTable.vue" is the file, that i have to change for this issue

devcydo:
Hi! I saw some comments on the PR, can I help on this?

gmcrocetti:
Recently tested and it's working. Thank you for your time ! :)

PR #1296, opened 2022-07-09
Add IN operator to table filter #884

ianberryman:


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -553,8 +557,15 @@ export default Vue.extend({
         this.filterMode === FILTER_MODE_BUILDER &&
         this.filter.type && this.filter.field && this.filter.value
       ) {
-        
-        return [this.filter]
+        if (this.filter.type === 'in') {
+          const vals = this.filter.value.split(/,/)
```
rathboma, line 561: I think this should probably include whitespace either side of the comma - `/\s*,\s*/`

rathboma (changes requested):
looks awesome! I think we just need a single change to the regex and we're good.

rathboma:
@ianberryman looks like Cockroach (and probably redshift) doesn't support the `?` syntax for parameters. I think that's why it was using `$1`, `$2`. See the failed unit tests

apps/studio/src/lib/db/clients/utils.ts
```diff
@@ -81,6 +81,8 @@ export function buildFilterString(filters, columns = []) {
       const column = columns.find((c) => c.columnName === item.field)
       if (column && column.dataType.toUpperCase().includes('BINARY')) {
         return `HEX(${wrapIdentifier(item.field)}) ${item.type} ?`
```
rathboma, line 83: What if this is both Binary and 'IN'?

rathboma (changes requested):
You are amazing, I have two more super minor changes, then we're good to merge!

apps/studio/package.json
```diff
@@ -19,7 +19,7 @@
     "postinstall": "electron-builder install-app-deps",
     "postuninstall": "electron-builder install-app-deps",
     "release:prs": "./build/get-prs.sh",
-    "test:integration": "TEST_MODE=1 ELECTRON_RUN_AS_NODE=1 yarn electron ../../node_modules/jest/bin/jest.js --config ./jest.integration.config.js"
```
rathboma, line 22: Not sure why this was deleted? This is needed to have the tests run in the right way

rathboma: Oh, I guess this doesn't work on Windows right?

ianberryman: I'll fix that, I removed that to try to get tests working locally on Windows

Issue #1018, opened 2022-02-07
FEAT: Filter and reorder columns


bediu:
**Is your feature request related to a problem? Please describe.**
I have two problems: 
* Some columns tend to have big (jsond) data which slow beekeeperstudio down. These big data columns are absolutely unnecessary in the visual table almost all the time.  
* A lot of columns are purely functional/meta, I have basically never interacted with 90% of my columns through beekeeper studio

**Describe the solution you'd like**
The ability to ignore columns. But not just ignore them visually. It would be nice if beekeeper studio ignored it when communicating with the database as well -- this would make the app so much smoother when working with big data. 

The ability to re-arrange columns. I have asked about this feature before. This is very necessary. It is very poor UX to always scroll horizontally past 20 columns you rarely touch, just to access a target column! 

**Describe alternatives you've considered**
I haven't considered alternatives. The other tools try too hard, the beauty of beekeeper studio is simplicity compared to the rest of the monoliths out there and that is why I believe these feature requests will only add to the feeling of why people choose beekeeper studio over, say, dbweaver.

**Additional context**
Beekeeper studio has always been open in my taskbar for over half a year now. Please consider adding these power user features. 

**Database type, if appropriate**
Postgresql

azmy60:
@rathboma I'll try to make PR for the filter. It will look similar to this.
![Screenshot_2022-07-18_15-13-52](https://user-images.githubusercontent.com/38707148/179471763-6b80cc07-e20f-4abd-aa48-b89c9edc04bc.png)

bediu:
@azmy60 I'll buy you a beer if this PR makes it in next version

PR #1221, opened 2022-05-10
feat: make rows selectable

ad-on-is:
beekeeper-selectable

I implemented the selectable option for tabulator and added styling to the selected rows.

PR #1310, opened 2022-07-20
Add Column Filter

azmy60:
Mentioned in #1018, this pr only has column filter (does not include column reorder).

Peek 2022-07-21 23-50

Tasks:
- [x] Implement Column Filter
- [x] Implement tests for custom select query


bediu:
@azmy60 I think drag and drop functionality inside the very same list as it is in the GIF would be more than enough for now. 
 
The other option is to re-order columns as they are in the main table with drag and drop, but this one might have a lot of implications & side effects. 

Just my opinion.

azmy60:
@bediu You're right. The first option is simpler. It also makes more sense to put two similar concepts together in one group, in my opinion.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1089,6 +1104,28 @@ export default Vue.extend({
       pendingUpdate.cell.getElement().classList.remove('edited')
       pendingUpdate.cell.getElement().classList.remove('edit-error')
     },
+    showColumnFilterModal() {
+      this.$modal.show('column-filter-modal', {
+        allColumns: [...this.table.columns],
+        selectedColumnNames: [...this.selectedColumnNames],
+        onApply: (selectedColumnNames) => {
```
rathboma, line 1111: So instead of providing `allColumns`, `selectedColumns`, and `onApply` to the modal function (which I recognize as the React-way to do this), in Vue you're best to provide properties and callbacks to the component. This will also make it more consistent with the rest of the App. The component itself can filter 


```vue
<ColumnFilterModal :columns="columnsWithFilterAndOrder" @changed="applyColumnChanges" />

```

apps/studio/src/lib/db/clients/sqlserver.js
```diff
@@ -118,7 +118,8 @@ function buildFilterString(filters) {
   return filterString
 }
 
-function genSelectOld(table, offset, limit, orderBy, filters, schema) {
+function genSelectOld(table, offset, limit, orderBy, filters, schema, selects) {
+  const selectString = selects.join(', ')
```
rathboma, line 122: The column names need escaping here - `selects.map((s) => wrapIdentifier(s)).join(", ")`

apps/studio/src/lib/db/clients/sqlserver.js
```diff
@@ -174,12 +175,13 @@ function genCountQuery(table, filters, schema) {
...
 }
 
-function genSelectNew(table, offset, limit, orderBy, filters, schema) {
+function genSelectNew(table, offset, limit, orderBy, filters, schema, selects) {
   const filterString = _.isString(filters) ? `WHERE ${filters}` : buildFilterString(filters)
 
   const orderByString = genOrderByString(orderBy)
   const schemaString = schema ? `${wrapIdentifier(schema)}.` : ''
 
+  const selectSQL = `SELECT ${selects.join(', ')}`
```
rathboma, line 184: same here

apps/studio/src/lib/db/clients/utils.ts
```diff
@@ -117,6 +117,7 @@ export function buildSelectTopQuery(table, offset, limit, orderBy, filters, coun
     filterParams = filterBlob.filterParams
   }
 
+  const selectSQL = `SELECT ${selects.join(', ')}`
```
rathboma, line 120: same here!

apps/studio/tests/lib/db.ts
```diff
@@ -305,6 +305,27 @@ export class DBTestUtil {
     expect(result).toMatchObject(['pears'])
   }
 
+  async columnFilterTests() {
```
rathboma, line 308: You++ Amazing adding tests!!

rathboma (changes requested):
This is amazing! I only have a couple of minor changes for readability + SQL safety.

Once you're done, let me know, and I can add code to persist the changes to the database so it persists across sessions.

azmy60:
@rathboma Thanks for the feedbacks! I'm making the fixes today.

azmy60:
@rathboma Some tests fail when we use `wrapIdentifier()`

rathboma:
@azmy60 Can you fix the conflict, then I'll merge it!

azmy60:
@rathboma I'm on it

rathboma:
I'm combining this with #1221 and adding a settings column header to control filtered columns and column ordering. check it out:
row-selection

*coming soon

azmy60:
@rathboma I've been thinking about this that moving the column header of the table to re-order is probably the simplest ux, in my opinion. I've seen DataGrip has this feature too. Also, the implementation is made very simple! Thanks to Tabulator for providing the [moveableColumn](http://tabulator.info/docs/5.3/move#columns) feature.

bediu:
@azmy60  @rathboma 

What happened to this feature? Still haven't seen it live?

azmy60:
There was an [issue](https://github.com/olifolkerd/tabulator/issues/3886) where the columns are misaligned when the filter is performed. Once the tabulator 5.4 is released, I think it'd be safe to launch this feature :)

PR #1315, opened 2022-07-26
Feature/issue 1236 primary key null on clone

### LINKED ISSUE #1236: FEAT: Set primary key cell to NULL while cloning a row
Currently while cloning a row, one of the things that needs to be done for primary keys (especially when they are auto incremented) is to delete the current value and keep it as NULL before applying your changes. 

It would be great if when cloning a row the primary key cell would be set to NULL auomatically on the new cloned row, at least when the primary key has auto increment setting.

Now that multiple rows selection is around the corner, it feels it would have even more need.

Thank you

wmontgomery:
Fix #1236

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -928,7 +928,15 @@ export default Vue.extend({
       const row = cell.getRow()
       const data = { ...row.getData() }
 
-      this.tabulator.addRow(data, true).then(row => {
+      const dataParsed = Object.keys(data).reduce((acc, d) => {
+        if (!this.primaryKeys.find(element => element === d)) {
+          acc[d] = data[d]
+        }
+        return acc
+      }, {})
+
+
+      this.tabulator.addRow(dataParsed, true).then(row => {
         this.addRowToPendingInserts(row)
         this.tabulator.scrollToRow(row, 'center', true)
       })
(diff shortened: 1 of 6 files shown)


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -927,8 +927,14 @@ export default Vue.extend({
     cellCloneRow(_e, cell) {
       const row = cell.getRow()
       const data = { ...row.getData() }
+      const dataParsed = Object.keys(data).reduce((acc, d) => {
+        if (!this.primaryKeys.find(element => element === d)) {
```
rathboma, line 931: This should just use `includes` -> `if(!this.primaryKeys?.includes(d))`

rathboma (changes requested):
Find => includes, then we're good!

apps/studio/src/components/tableview/TableTable.vue
wmontgomery: Includes it is!

apps/studio/src/store/index.ts
```diff
@@ -424,6 +423,9 @@ const store = new Vuex.Store<State>({
             })
           })
 
+          // eslint-disable-next-line no-debugger
+          // debugger
```
rathboma, line 427: Three leave, one enters :laughing:

PR #1322, opened 2022-07-31
Feature/issue 1270 drop database table

### LINKED ISSUE #1270: FEAT: Add "Drop database" and "Drop table" on the context menu (right click) (and truncate, rename?)
I'm using an ORM during development that is migrating the database.
I need to DROP the table(s) or the database(s) some time to mage the migration pass.
A "Rename" and "Truncate" option would be great as well.

You right click on the table/database name, a context menu appears with few options like "Drop" / "Rename" / "Refresh" /  "Truncate"

I'm using SQL query for now.

**Additional context**
image

All (I'm using it only for Postgresql till now).

wmontgomery:
Fix #1270 part 1/2


apps/studio/src/lib/db/client.ts
```diff
@@ -165,6 +175,10 @@ export class DBConnection {
   getRoutineCreateScript = getRoutineCreateScript.bind(null, this.server, this.database)
   truncateAllTables = truncateAllTables.bind(null, this.server, this.database)
   setTableDescription = setTableDescription.bind(null, this.server, this.database)
+
+  // delete stuff
+  dropElement = dropElement.bind(null, this.server, this.database)
```
rathboma, line 180: Rather than writing a whole new method, you should use two helper methods I recently wrote - `bind` and bindAsync`, these are the first step to removing some of the duplication here:

image

apps/studio/src/lib/db/client.ts
```diff
@@ -467,6 +481,11 @@ async function getTableColumnNames(server: IDbConnectionServer, database: IDbCon
   }
 }
 
+function dropElement(server: IDbConnectionServer, database: IDbConnectionDatabase, elementName: string, typeOfElement: DatabaseElement, schema:string) {
```
rathboma, line 484: see above comment, this can be removed

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1347,6 +1348,15 @@ export async function truncateAllTables(conn: Conn, schema: string) {
   });
 }
 
+export async function dropElement (conn: Conn, elementName: string, typeOfElement: DatabaseElement, schema: string): Promise<void> {
```
rathboma, line 1351: good stuff, but see above regarding wrapLiteral

apps/studio/src/lib/db/clients/mysql.js
```diff
@@ -722,6 +724,15 @@ export async function truncateAllTables(conn) {
   });
 }
 
+export async function dropElement (conn, elementName, typeOfElement) {
+  await runWithConnection(conn, async (connection) => {
+    const connClient = { connection }
+    const sql = `DROP ${typeOfElement} ${wrapIdentifier(elementName)}`
```
rathboma, line 730: `typeOfElement` needs wrapping as a literal here, and in all the other dialects, `${MysqlData.wrapLiteral(typeOfElement}`

apps/studio/src/lib/db/clients/postgresql.ts
rathboma: What if there's no schema specified?

apps/studio/src/lib/db/clients/sqlite.js
```diff
@@ -475,6 +478,15 @@ export async function truncateAllTables(conn) {
   });
 }
 
+export async function dropElement (conn, elementName, typeOfElement) {
+  await runWithConnection(conn, async (connection) => {
+    const connClient = { connection };
+    const sql = `DROP ${typeOfElement} ${wrapIdentifier(elementName)}`
```
rathboma, line 484: same regarding wrapLiteral

apps/studio/src/lib/db/clients/sqlserver.js
```diff
@@ -833,6 +834,15 @@ export async function truncateAllTables(conn) {
   });
 }
 
+export async function dropElement (conn, elementName, typeOfElement, schema) {
+  await runWithConnection(conn, async (connection) => {
+    const connClient = { connection };
+    const sql = `DROP ${typeOfElement} ${wrapIdentifier(schema)}.${wrapIdentifier(elementName)}`
```
rathboma, line 840: What if schema isn't provided?

apps/studio/src/mixins/TableListContextMenus.ts
```diff
@@ -62,8 +62,16 @@ export default {
...
-
-
+        {
+          type: 'divider'
+        },
+        {
+          name: "Delete",
+          slug: 'sql-delete',
+          handler: ({ item }) => {
+            this.$root.$emit('deleteDatabaseElement', item, 'Delete') // please format last argument in a nice way because it's shown in a modal (can move to a compute in the modal, but that's not today)
```
rathboma, line 72: For emitting events, can we have the second argument be an object with all the values in it? This makes it consistent with all the other app events.

Also please add `deleteDatabaseElement` to `AppEvent.ts`

So something like:
```js
this.$root.$emit(AppEvent.deleteDatabase, { item, action: 'delete' })
```

Also it's easy to capitalize this in the modal with LoDash, we should do that:
```
 title() {
    return _.capitalize(this.action)
 }
```

rathboma (changes requested):
A few minor org changes and a SQL Injection fix, then we're good to go!

apps/studio/src/lib/db/clients/mysql.js
wmontgomery: the `typeOfElement` was getting wrapped when passed into the function, but I get what you're saying. Handle it in each function so it does what it has to do closer to where it's being executed.

apps/studio/src/lib/db/clients/postgresql.ts
wmontgomery: I was under the impression schema had to be passed in with postgres (at least that's how I've always done it but I don't leave much to chance)? If not, I can add an if in there to get different sql generated.

apps/studio/src/lib/db/clients/sqlserver.js
wmontgomery: Same as above with postgres, will take the same approach.

apps/studio/src/lib/db/clients/postgresql.ts
rathboma: So probably the app always passes schema, but maybe not. It's possible I'm not consistent about this either, but I try to be. Everything will work without schema, it will just default to `public`

wmontgomery:
@rathboma changes have been made and ready for review. I added the issue number in the description for the PR so should be ok with the automation as well.

PR #1326, opened 2022-08-01
Row selection in all tables

### LINKED ISSUE #928: QUESTION: Deleting multiplew rows in a table.
**Details:**
Is there a way to delete multiple rows from a table? I have looked around but couldn't find any.  
Thanks

### LINKED ISSUE #1394: FEAT: The ability to delete multiple rows at one time
During development I sometimes want to delete one, multiple or all rows of a table.

1. Option to delete all:
Right click on the table name -> delete all rows
Option to press ctrl + a and then press del
Possibly a button somewhere (at the bottom?) to enter a delete mode (where I can check specific rows to delete them)

2. Option to delete a single row
A the moment you have to right click on the row -> Delete Row. It would be great to simply select a row and the press del

3. Option to delete multiple rows
Select with the mouse multiple rows, then a) press del or right click -> Delete Selected Rows
Probably a delete mode (see above) where you can check specific rows (or all) and then delete them; kinda lika a bulk mode

- Right click and delete one by one
- Run a SQL statement
- Use a different program

In my case MySQL/MariaDB, but would be a use case for other DBMS' too.

rathboma:
Fix #928 
Fix #1394 

This is the first set of changes to modernize the data table implementation. This PR will provide:

- Row selection (click, ctrl+click, shift+click)
- right click copy works with the selection
- Click a cell to dismiss row selection
- Move the column configuration to the top left of the table

https://github.com/beekeeper-studio/beekeeper-studio/assets/279769/da29461a-c3cc-41a5-9d3f-2e9c0659d52a


kgolenko:
how are you doing with this task

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1325,7 +1453,7 @@ export default Vue.extend({
       pendingUpdate.cell.getElement().classList.remove('edit-error')
     },
     showColumnFilterModal() {
-      this.$modal.show(this.columnFilterModalName)
+      this.$modal.show(this.columnFilterModalName, )
```
wmontgomery, line 1456: Since there's no second argument being passed in, take the comma out of there, please and thanks

wmontgomery (approved):

PR #1327, opened 2022-08-01
feat: Database dropdown search

bmartinsdev:
I added support for a customizable dropdown with search.
Following the TODO on Databasedropdown.vue, I opted by using this package:
https://vue-select.org/

Implementation uses theme variables and can be used anywhere with the class "dropdown-search".
demo


rathboma:
Awesome, thank you!

Issue #1345, opened 2022-08-15
FEAT: Support IAM Authentication for Amazon Redshift


bcelenza:
**Is your feature request related to a problem? Please describe.**

AWS allows an end user to use temporary credentials federated by an IAM user or role. You can read more about it [here](https://docs.aws.amazon.com/redshift/latest/mgmt/generating-user-credentials.html).

**Describe the solution you'd like**

I'd like for the ability to _optionally_ use IAM to authenticate to an Amazon Redshift cluster.

**Describe alternatives you've considered**

The current alternative or workaround is to manually [call the `GetClusterCredentials` API using the AWS CLI](https://docs.aws.amazon.com/cli/latest/reference/redshift/get-cluster-credentials.html).

**Additional context**
Add any other context or screenshots about the feature request here. (we love screenshots with red circles if the request is for something visual)

**Database type, if appropriate**

Amazon Redshift only (for now!).

rathboma:
I''m

Issue #832, opened 2021-10-05
FEAT: Clear Text Authentication for MySQL


BrysonRafael:
**Is your feature request related to a problem? Please describe.**
The database I am trying to connect to requires MySQL's [clear text pluggable authentication](https://dev.mysql.com/doc/refman/8.0/en/cleartext-pluggable-authentication.html) and I cannot connect to it using Beekeeper Studio.

**Describe the solution you'd like**
Beekeeper Studio allows me to authenticate with a MySQL database that requires clear text authentication. I believe I would need to install the clear text plugin somehow in Beekeeper. Perhaps an extension option that is similar to VSCode's that allows me to easily install plugins?
![image](https://user-images.githubusercontent.com/19216177/136063325-2a3e243a-bc68-4ba3-aba0-26e2254c991c.png)

**Describe alternatives you've considered**
I'm currently using DBeaver to connect to the database that I need to, but I don't like the interface and it feels too heavy for what I use it for.

**Additional context**
Here is an example of the SSL config that I have in DBeaver for my issue:
![image](https://user-images.githubusercontent.com/19216177/136063026-2a685b20-be62-4f4c-89a0-b82bc6f08538.png)
I'm not doing anything fancy in the connection config except for using SSL.

Thanks for building this project and keeping it open source! I'm loving the look and feel of it, I just wish I could use it for all of my database connections 🥲

ethan-dowler:
@rathboma Well, I guess I got a bit mixed up. I did mean Postgres, but I guess the cleartext plugin is not what fixed my issue haha. @BrysonRafael will be back in the office tomorrow. He can follow up on this.

rathboma:
Ok, I clearly need more testing of the auth plugin. I'll try my best to fix this on Monday.

Do you know the SQL I need to set up a user to require this auth plugin?

BrysonRafael:
@MasterOdin that's exactly how I'm trying to connect. Sorry @rathboma, I probably should have mentioned this to begin with 🤦🏻‍♂️

PR #1346, opened 2022-08-15
Amazon Redshift: Add support for IAM authentication (#1345)

bcelenza:
This pull request adds support for using [IAM authentication with Amazon Redshift](https://docs.aws.amazon.com/redshift/latest/mgmt/generating-user-credentials.html). Addresses #1345. 

How it works, roughly:

1. An API call is made to Amazon Redshift's control plane asking for temporary cluster credentials. 
2. If successful, the response returns a modified username and temporary password (valid for 15 minutes by default).
3. The temporary credentials are immediately used to create a connection to the cluster. The connection is active as long as the temporary credentials are valid, after which an error is returned.

I wasn't quite sure where to put the credential exchange code. At the moment it looks like a few database-specific features are handled in `postgresql.ts`, so I've started with the functionality in there. Happy to move it as desired. I've also added validation for required fields so friendly error messages are returned.

For the UI, I re-used the layout for SSH settings since the functionality is quite similar (an optional toggle with required internal fields). See screenshots (and pardon my screen darkener):

The initial screen will show with the IAM authentication section toggled off.
Screen Shot 2022-08-15 at 10 34 21 AM

When toggled on, the fields appear.
Screen Shot 2022-08-15 at 11 17 29 AM

## Testing

I've verified this works with two different redshift clusters, and verified that all error messages I could trigger came back and were understandable.

## Future Direction

If we like this sort of change, there are a handful of things we can do to make it better in the future:
1. Add support for credentials that were already configured using the AWS CLI on the user's machine. This would make most of these fields optional.
2. Add support for AWS Single Sign-On, largely supported with (1) but may require a little extra credential handling.
3. Attempt to automatically refresh the credentials behind the scenes to keep the active connection window truly active.

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -1377,18 +1377,19 @@ async function configDatabase(server: { sshTunnel: boolean, config: IDbConnectio
   let passwordResolver: () => Promise<string>;
 
   // For Redshift Only -- IAM authentication and credential exchange
-  if (server.config.client === 'redshift' && server.config.options?.useIAM) {
+  const redshiftOptions = server.config.redshiftOptions;
+  if (server.config.client === 'redshift' && redshiftOptions?.iamAuthenticationEnabled) {
     const awsCreds: AWSCredentials = {
-      accessKeyId: server.config.options.awsAccessKeyId,
-      secretAccessKey: server.config.options.awsSecretAccessKey
+      accessKeyId: redshiftOptions.accessKeyId,
+      secretAccessKey: redshiftOptions.secretAccessKey
     };
 
     const clusterConfig: ClusterCredentialConfiguration = {
-      awsRegion: server.config.options.awsRegion,
-      clusterIdentifier: server.config.options.clusterIdentifier,
+      awsRegion: redshiftOptions.awsRegion,
+      clusterIdentifier: redshiftOptions.clusterIdentifier,
       dbName: database.database,
       dbUser: server.config.user,
-      dbGroup: server.config.options.databaseGroup,
+      dbGroup: redshiftOptions.databaseGroup,
       durationSeconds: server.config.options.tokenDurationSeconds
     };
 
@@ -1396,7 +1397,7 @@ async function configDatabase(server: { sshTunnel: boolean, config: IDbConnectio
 
     // We need resolve credentials once to get the temporary database user, which does not change
     // on each call to get credentials.
-    // This is usually something like "IAM:<user>:<group>" or "IAMA:<user>:<group>".
+    // This is usually something like "IAMA:<user>" or "IAMA:<user>:<group>".
     tempUser = (await credentialResolver.getClusterCredentials(awsCreds, clusterConfig)).dbUser;
 
     // Set the password resolver to resolve the Redshift credentials and return the password.
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -9,6 +9,7 @@ import knexlib from 'knex'
 import logRaw from 'electron-log'
 
 import { DatabaseClient, IDbConnectionServerConfig, DatabaseElement } from '../client'
+import { AWSCredentials, ClusterCredentialConfiguration, RedshiftCredentialResolver } from '../authentication/amazon-redshift';
 import { FilterOptions, OrderBy, TableFilter, TableUpdateResult, TableResult, Routine, TableChanges, TableInsert, TableUpdate, TableDelete, DatabaseFilterOptions, SchemaFilterOptions, NgQueryResult, StreamResults, ExtendedTableColumn, PrimaryKeyColumn, TableIndex, IndexedColumn, } from "../models";
 import { buildDatabseFilter, buildDeleteQueries, buildInsertQuery, buildInsertQueries, buildSchemaFilter, buildSelectQueriesFromUpdates, buildUpdateQueries, escapeString, joinQueries } from './utils';
 import { createCancelablePromise } from '../../../common/utils';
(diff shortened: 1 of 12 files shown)


rathboma:
Wow, this is amazing! Thanks for all the work. I'll review it presently

We can probably do the same thing for RDS databases too right? (Although that requires #832 I think)

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1358,10 +1359,46 @@ function configDatabase(server: { sshTunnel: boolean, config: IDbConnectionServe
     }
   }
 
+  // For Redshift Only -- IAM authentication and credential exchange
```
rathboma, line 1362: Could we move this logic into a file like `lib/db/authentication/amazon.ts`? That way we can re-use it for other Amazon IAM-authenticated databases.

rathboma: Then we can do something like:

```
const { tempUser, tempPassword } = iamAuthenticate(server.config.options)
```

rathboma (changes requested):
Only really one requested change regarding code-org. Good stuff!

rathboma:
@bcelenza looks like you need to remove `package-lock.json` and do a `yarn install`!

Question:
- What happens if someone is using the IAM connection and the credentials expire? Should there be a way to quickly reconnect?

MasterOdin:
The postgres adapter allows you to pass a function to `password` for that use case (which is what we do in @popsql).

rathboma:
@bcelenza looks like we should do it the way @MasterOdin suggests! That way reconnect will work.

We should still break out the iam code into a separate file, but structured the way suggested then auto reconnects will work

MasterOdin:
As another point, when we added IAM support to @popsql in similar fashion (requiring to type in access key id + secret access key), the users we had most interested in the feature immediately started asking for the ability to put in a profile and have the app automatically pull the info from their `~/.aws/credentials file, similar to how they were using other DB applications.

apps/studio/src/lib/db/clients/postgresql.ts
bcelenza: Yeah that makes sense. One caveat to using `amazon.ts` is that this particular credential exchange is specific to Redshift. RDS has a slightly different approach from what I recall. I'll move this out into its own file and make sure it's obvious this is a Redshift-specific operation.

bcelenza:
Using the password function approach makes sense, will check that out.

I had listed supporting the AWS CLI config (~/.aws/credentials) as a potential future enhancement, but I can take a look at feasibility in this go-around. It can complicate what's considered required/optional in the configuration screen, but nothing a little hinting / field disabling can't resolve.

Thanks for the feedback!

apps/studio/src/lib/db/clients/postgresql.ts
bcelenza: After digging into this a little more, the credential processes are slightly different, though all could conform to a single interface. I'm thinking that's a future enhancement since RDS isn't a formal database option right now.

I've created a `RedshiftCredentialResolver` which will handle exchanging the credentials. While implementing the password callback for `pg` I noticed it was being invoked far more frequently than needed, which could result in throttling from AWS.

At the risk of over-blowing it a little, I created the resolver using a singleton pattern to prevent accidentally having two of these in runtime retrieving independent credentials, and a light-weight caching mechanism that will return good credentials if invoked frequently.

bcelenza:
> Using the password function approach makes sense, will check that out.
> 
> I had listed supporting the AWS CLI config (~/.aws/credentials) as a potential future enhancement, but I can take a look at feasibility in this go-around. It can complicate what's considered required/optional in the configuration screen, but nothing a little hinting / field disabling can't resolve.
> 
> Thanks for the feedback!

If it's alright with you all, I would like to take sourcing from AWS CLI configration separately. It will effectively be a relaxing of the requirements of the current form, but I want to make sure we account for all the flavors of credentials there (static, profile-based, and AWS SSO).

Other feedback should have solutions in the code now. Thanks!

bcelenza:
I'm noticing some issues with the toggle not activating the form, which leads me to believe I've missed something with storage. Will take a look later today.

bcelenza:
@rathboma Before I go further with this, I could use your help in confirming my current understanding.

The toggle not working was indeed related to the storage for a saved connection. Because I was using the existing `options` structure, I found a scenario where the client already has that general structure but not the redshift options within it -- so the embedded `redshift` structure was undefined, leading to an error trying to see the value of `useIAM` from that structure in the view.

Instead of embedding it within an existing `options` structure, would it make sense to carve out a new column in the saved connection, something like `redshiftOptions`? Then we could persist all the redshift-specific things there and provide a default structure (with `useIAM` set to false by default), which should work for this scenario.

Thanks!

rathboma:
@bcelenza This is AMAZING. One last thing - can you merge in `master` and fix the merge conflicts? Once that's done I'll merge this and include it in a new release next week

bcelenza:
@rathboma No problem, done. Thanks!

bcelenza:
No worries! Thanks for all the feedback, glad we could make it work.

rathboma:
@bcelenza Few questions for you for user sanity:

1. Token Duration - should this have a default? How should a user figure out what to set this to?
2. Does this work with SSH tunnels, or should I hide SSH if this is active?
3. If IAM authentication is active, presumably they no longer need username and password?
4. Do they still need the `default database`?

Thanks for your help!

bcelenza:
> Token Duration - should this have a default? How should a user figure out what to set this to?

If unspecified it defaults to 15 minutes, which would be my personal recommendation. The constraints on the timeout are documented here: (see `--duration-seconds`): https://docs.aws.amazon.com/cli/latest/reference/redshift/get-cluster-credentials.html

> Does this work with SSH tunnels, or should I hide SSH if this is active?

Yup! Should work just fine. 

> If IAM authentication is active, presumably they no longer need username and password?

They'll still need to provide a username, but not a password. The password, if entered, would be ignored.

> Do they still need the default database?

Yup! It will be used in the process of getting temporary credentials and to connect to the actual DB.

PR #1355, opened 2022-08-25
Exclude entities

### LINKED ISSUE #1320: FEAT: hide system tables
There may be users who like to see system tables but I can't remember a time when I've used them. There are collections of system tables in some databases that clutter the view and just add a bit more of unnecessary data on screen that means your eyes have additional filtering to do.

Preferably there could be some way of hiding the system tables from view. I personally wouldn't mind this in settings as I use system tables so infrequently. Others may want that per database, maybe as part of the connection info, or just a live UI toggle somewhere, perhaps even in the filters list "System" could be distinguished from "Tables".

azmy60:
Hiding/excluding entities works exactly like `PinnedEntity`, except it's hidden from the sidebar. I choose the word "exclude" in case that it'll be easier to add an "include" feature in the future.

Peek 2022-08-26 06-11

Tasks:
- [ ] Implement modal to un-exclude entities
- [ ] Inherit PinModule from useBaseEntityModule

Fixes #1320

### DIFF
--- a/apps/studio/src/common/AppEvent.ts
+++ b/apps/studio/src/common/AppEvent.ts
@@ -20,7 +20,8 @@ export enum AppEvent {
   quickSearch = 'quickSearch',
   promptLogin = 'cloud_signin',
   promptQueryImport = 'cloud_q_import',
-  promptConnectionImport = 'cloud_c_import'
+  promptConnectionImport = 'cloud_c_import',
+  hideEntity = 'hideEntity',
 }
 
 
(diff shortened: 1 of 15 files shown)


rathboma:
Nice! Looking forward to seeing how this turns out. Again we'll need a way to visually indicate that there are hidden entities. Also it'd be great to have a 'hide' for schemas too (for psql/sql-server)

azmy60:
Yes. I will add this to the schemas as well.

azmy60:
Closed in favor of #1376

PR #1356, opened 2022-08-26
Bugfix/issue 1264 bit conversion error

### LINKED ISSUE #1264: BUG: Bits are not converted back to proper form for insert scripts.
**Describe the bug**
Currently in results view/data view, all bits are converted to true/false values (which I do find helpful). One frustrating thing is when you export something as a script (copy row (insert) for instance) those values are not converted back to bits, so I have to manually go through and fix those values before running the script. A small thing, but can be annoying.

**To Reproduce**
1. Open data view of some table that has bit columns
2. Right click on a row
3. Click "Copy Row (insert)"
4. Paste this in a new query window

**Expected behavior**
Bits should be converted back to 1/0 form instead of true/false for insert.

**Screenshots / Gifs**
image
image
The second screenshot is of the resulting script, which fails as 'true' is not a valid value for a bit column in SQL server

 - OS: Windows 11
 - App Version 3.5.8
 - Database type and version SQL Server 15.0.2080.9

wmontgomery:
Fix #1264


wmontgomery:
First part is making sure copy insert script works. Moving to the second issue which was the bits getting converted to b-somethingsomething

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -373,7 +373,7 @@ export default Vue.extend({
         {
           label: '<x-menuitem><x-label>Copy Row (Insert)</x-label></x-menuitem>',
           action: async (_e, cell) => {
-
+            console.log(this.tableColumns)
```
rathboma, line 376: @wmontgomery can we remove this.

wmontgomery: Oops, you bet.

wmontgomery:
@rathboma fix has been made

apps/studio/src/lib/data/tools.ts
```diff
@@ -62,13 +62,16 @@ export const Mutators = {
...
    * Stringify bit data for use in json / UIs
+   * @param {DBConnection|null} connection: The connection object used within Vue
    * @param  {any} value
    * @returns JsonFriendly
    */
-  bitMutator(value: any): JsonFriendly {
-    // SQL Server bit data type is an integer data type
-    if (this.connection?.connectionType === "sqlserver") return value;
-
+  bitMutator(connection: DBConnection|null, value: any): JsonFriendly {
```
rathboma, line 69: Ok, couple more changes here. Sorry for late comments.

1. This doesn't need the full connection, it should just take the `dialect` which is typed and can be matched. It should default to `undefined` using the question mark. See dialects/index.ts :
```ts
bit1Mutator(value: any, dialect?: Dialect): JsonFriendly {
 if (dialect && dialect === 'sqlserver') return value

|

```
2. Also because this is optional, it should be the last argument.

apps/studio/src/lib/data/tools.ts
```diff
@@ -9,26 +9,26 @@ function dec28bits(num: any): string {
...
     if (dataType && dataType.startsWith('bit')) {
-      return this.bitMutator.bind(this)
+      return this.bitMutator.bind(this, conn)
     }
     return this.genericMutator.bind(this)
   },
 
 
-  mutateRow(row: any[], dataTypes: string[] = [], preserveComplex: boolean = false): JsonFriendly[] {
+  mutateRow(row: any[], dataTypes: string[] = [], preserveComplex: boolean = false, connection: DBConnection = null): JsonFriendly[] {
```
rathboma, line 28: I think you just need `dialect` here. See below. Also use `dialect?: Dialect` for optional

rathboma (changes requested):

apps/studio/src/lib/data/tools.ts
wmontgomery: dialect will need to be the first argument in there because it is passed in as the argument when setting up the `bind` function in `resolveDataMutator`

wmontgomery:
@rathboma , changes have been made. Passing the dialect in vs the whole connection is the biggest change.

PR #1365, opened 2022-09-05
Feature/840 create db and schema

### LINKED ISSUE #840: FEAT: Be able to create databases
It can be sometimes annoying to go into command line to create new database

Be able not to just connect to database but also create it.

Well using the command line lol

wmontgomery:
fix #840


rathboma:
@wmontgomery looks like you need to merge master in to this branch first

rathboma:
Also for these types of features, can you add a gif of it working?

apps/studio/src/components/connection/AddDatabaseForm.vue
```diff
@@ -0,0 +1,77 @@
...
+      {{error}}
+    </div>
+
+    <div class="form-group">
+      <input class="form-control" v-model="databaseName" @keydown.enter.prevent.stop="save" type="text" placeholder="Database Name">
+    </div>
+
+    <div class="form-group" v-if="charSets.length > 0">
+      <select v-model="selectedCharset">
+        <option v-for="charset in charSets" v-bind:key="charset" :selected="charset === defaultCharSet" :value="charset">{{charset}}</option>
```
rathboma, line 15: Just as an FYI `v-bind:key=` is the same as `:key=`

rathboma: Should this attribute be `defaultCharSet` or `selectedCharSet`? I think it should be `selected`

apps/studio/src/components/connection/AddDatabaseForm.vue
```diff
@@ -0,0 +1,77 @@
...
+</template>
+
+<script>
+  export default {
+    props: ['connection'],
+    data() {
+      return {
+        charSets: [],
+        collations: [],
+        defaultCharSet: null,
```
rathboma, line 39: CharSet vs Charset, don't care which, but should pick one and stick with it

apps/studio/src/components/connection/AddDatabaseForm.vue
```diff
@@ -0,0 +1,77 @@
...
+    methods: {
+      async updateCollations() {
+        this.collations = await this.connection.listCollations(this.selectedCharset)
+        this.selectedCollation = this.collations[0]
+      },
+      async save() {
+        const dbNameRegex = /^[a-zA-Z0-9_-]*$/
+        try {
+          if (!this.databaseName.match(dbNameRegex)) {
+            throw new Error('Database name invalid, must be alphanumeric / have only _ or - special characters')
```
rathboma, line 61: I'm not sure this is always true? Can special characters in database names be escaped like table names in some databases?

Is there a reason we can't just rely on the database creation function to provide us with an error here?

apps/studio/src/components/sidebar/core/DatabaseDropdown.vue
```diff
@@ -5,12 +5,21 @@
       <a class="refresh" @click.prevent="refreshDatabases" :title="'Refresh Databases'">
         <i class="material-icons">refresh</i>
       </a>
+      <a class="refresh" @click.prevent="$modal.show('config-add-database')" :title="'Add Database'">
+        <i class="material-icons">add</i>
+      </a>
     </div>
+    <modal class="vue-dialog beekeeper-modal save-add-database" name="config-add-database" height="auto" :scrollable="true">
```
rathboma, line 12: We need to start using `<portal to="modals">` to wrap all modals so they render properly above other elements

apps/studio/src/components/sidebar/core/DatabaseDropdown.vue
```diff
@@ -5,12 +5,21 @@
...
+      <div class="dialog-content">
+        <add-database-form :connection="connection" @databaseCreated="databaseCreated" @cancel="$modal.hide('config-add-database')"></add-database-form>
```
rathboma, line 14: So if you click this, fill it out, then cancel, then go to create a new DB again it will be partially filled out. Is that right? If not we should add a v-if for whether the modal is open. It will also load all charsets on first render

apps/studio/src/lib/db/client.ts
```diff
@@ -31,6 +31,12 @@ export interface DatabaseClient {
   getTableKeys: (db: string, table: string, schema?: string) => void,
   query: (queryText: string) => CancelableQuery,
   executeQuery: (queryText: string) => void,
+  // create database
+  listCharsets: () => Promise<string[]>,
+  getDefaultCharSet: () => Promise<string>,
```
rathboma, line 36: same comment regarding `Charset` vs `charSet`

apps/studio/src/lib/db/clients/sqlite.js
```diff
@@ -656,6 +662,19 @@ function getVersionString(version) {
...
+  if (!checkValidityOfDatabaseName(databaseName)) {
+    throw new Error('Database name invalid, must be alphanumeric / have only _ or - special characters')
+  }
+  // because this is a convenience for an otherwise ez-pz action, the location of the db file will be in the same location as the other .db files.
+  // If the desire for a "but I want this in another directory" is ever wanted, it can be included but for now this feels like it suits the current needs. 
+  const fileLocation = conn.dbConfig.database.split('/')
+  fileLocation.pop()
+
+  const db = new Database(`${fileLocation.join('/')}/${databaseName}.db`)
+  db.close()
```
rathboma, line 675: Should this switch you to using the new database file if you do this in SQLite? Maybe it should open a new window with this database open?

apps/studio/src/lib/db/clients/sqlserver.js
```diff
@@ -73,6 +84,21 @@ export default async function (server, database) {
     alterTableSql: (change) => alterTableSql(conn, change),
     alterTable: (change) => alterTable(conn, change),
 
+    // db creation
+    /*
+      SQL Server doesn't use character sets as these are part of the collation used (set at server or as you please)
+      https://stackoverflow.com/questions/7781103/sql-server-set-character-set-not-collation
```
rathboma, line 90: Love all these comments on collations and charsets, ++

apps/studio/src/lib/db/clients/utils.ts
```diff
@@ -233,6 +233,11 @@ export async function withClosable<T>(item, func): Promise<T> {
 
 }
 
+export function checkValidityOfDatabaseName(databaseName: string): boolean {
+  const dbNameRegex = /^[a-zA-Z0-9_-]*$/
```
rathboma, line 237: is this always true?

Also - don't need this in the front end as you do it here

shared/src/lib/dialects/postgresql.ts
```diff
@@ -16,6 +16,11 @@ const defaultLength =(t: string) => {
   return 8
 }
 
+// copied from https://www.postgresql.org/docs/12/multibyte.html as there is no way to get this via sql command unlike mysql
+const charsets = [
+  'BIG5', 'EUC_CN', 'EUC_JP', 'EUC_JIS_2004', 'EUC_KR', 'EUC_TW', 'GB18030', 'GBK', 'ISO_8859_5', 'ISO_8859_6', 'ISO_8859_7', 'ISO_8859_8', 'JOHAB', 'KOI8R', 'KOI8U', 'LATIN1', 'LATIN2', 'LATIN3', 'LATIN4', 'LATIN5', 'LATIN6', 'LATIN7', 'LATIN8', 'LATIN9', 'LATIN10', 'MULE_INTERNAL', 'SJIS', 'SHIFT_JIS_2004', 'SQL_ASCII', 'UHC', 'UTF8', 'WIN866', 'WIN874', 'WIN1250', 'WIN1251', 'WIN1252', 'WIN1253', 'WIN1254', 'WIN1255', 'WIN1256', 'WIN1257', 'WIN1258'
```
rathboma, line 21: omg, yuck, lol

apps/studio/src/components/connection/AddDatabaseForm.vue
wmontgomery: I was basing it largely on the postgres rules since that's where I was having the most issue. Totally fine with knocking it back to the database creation function to be the final word, was thinking from a user perspective give them something prior, but I get the issue. I can take this out no problem.

apps/studio/src/components/connection/AddDatabaseForm.vue
```diff
@@ -0,0 +1,77 @@
...
+        databaseName: null,
+        selectedCharset: null,
+        selectedCollation: null,
+        error: null
+      }
+    },
+    async mounted(){
+      this.charSets = await this.connection.listCharsets()
+      this.defaultCharSet = await this.connection.getDefaultCharSet()
+      this.selectedCharset = this.defaultCharSet
```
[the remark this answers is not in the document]
wmontgomery: This was left over from the previous PR so I kept it in. There was a difference on the mysql side, but I honestly don't know off hand. Can do some more research into it.

apps/studio/src/components/connection/AddDatabaseForm.vue
wmontgomery: relic of the previous PR for the first one, can def change it. Good eye as I missed it.

apps/studio/src/lib/db/clients/sqlite.js
wmontgomery: Was taking the same approach we had with the others where it created it but didn't make any of the switches over. Can totally do as suggested otherwise.

shared/src/lib/dialects/postgresql.ts
wmontgomery: I hated it too

apps/studio/src/lib/db/clients/utils.ts
wmontgomery: Was taking the approach of validate on front end and "back end" This is going to be going away anyhow. Like it never happened

apps/studio/src/components/connection/AddDatabaseForm.vue
```diff
@@ -0,0 +1,77 @@
...
+
+    <div class="form-group" v-if="charSets.length > 0">
+      <select v-model="selectedCharset">
+        <option v-for="charset in charSets" v-bind:key="charset" :selected="charset === defaultCharSet" :value="charset">{{charset}}</option>
+      </select>
+    </div>
+
+    <div class="form-group" v-if="collations.length > 0">
+      <select v-model="selectedCollation">
+        <option v-for="collation in collations" v-bind:key="collation" :value="collation">{{collation}}</option>
```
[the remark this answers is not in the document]
wmontgomery: All I can figure is the charset select was given a default to match the charset used in the server itself to make it easier to know which one to select instead of giving too many options? As far as a "default collation" though, not sure if there is one for the charset chosen

apps/studio/src/components/connection/AddDatabaseForm.vue
wmontgomery: Also switched it over to `selectedCharset` because that gets set on mounted anyway to the default one. Not sure why `defaultCharset` was even there.

apps/studio/src/lib/db/clients/sqlite.js
rathboma: Here's what this should do for SQLite:
Back in the UI, check for SQLite, then after Create Database, submit the `AppEvent.menuClick` to the backend for 'newWindow'. You'll need to add an argument to this for the URL. Places that will be useful to look:

ClientMenuActionHandler (Example of dispatching the event)
NativeMenuActionHandler (where this action is actually handled)
WindowBuilder (see we add query params to the window URL here, this is to open the database on-open)

tl;dr - improve the handling of the `newWindow` MenuClick event to handle `OptionOptions`

For the other engines:
- it should switch to the new database

apps/studio/src/components/connection/AddDatabaseForm.vue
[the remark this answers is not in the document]
wmontgomery: portals have been added. I went through and put a portal around all the other modals except for the Export modal because that one was inconsistent in when it popped up. There's more to that sucker that needs some looking into, but most of the other ones are handled.

rathboma:
Going to do a release then merge this one after.

PR #1390, opened 2022-10-01
J52

Crosier52:
as long as we cant see Telegram on either side of dev, i believe we in the good old USA


rathboma:
I don't understand this PR -- it's against an old branch and it is the same as master.

I'm going to close it as I have no way to review it, but if you're trying to submit a real PR, please reopen one with the right files, Happy to help if something isn't working right.

Issue #1389, opened 2022-09-28
FEAT: Columns should be loaded dynamically, not all loaded up front


rathboma:
Currently the behavior is:

    On establishing a connection a call is made to refresh the schema. This does two things:
        Loads all table names in the current database, for all schemas
        Loads ALL COlUMNS for all tables
        Maps columns to tables
        Stores the whole thing in the VueX store.

Why do we do this?
We use this local state in two ways:

    We use it to display columns in the sidebar
    We use it to provide autocomplete

Both of these are important features

New Behavior (this is what we want)

    On establishing a connection a call is made to refresh the schema. This should:
        Load all table names for the current database.
        That's it.
    When a table is expanded in the sidebar we should dynamically load the columns. It's fine if we keep them in the store at this point
    When someone types `table.` in the editor it should dynamically load the columns. It's fine if we keep them in the store at this point.

Other Stuff That Should Work
When clicking the refresh button in the sidebar it should wipe out the column cache, reload tables, and reload columns for any expanded table in the sidebar.

PR #1404, opened 2022-10-21
Feature/1389 dynamic load columns

wmontgomery:
#1389 
query_tab

column_load

### DIFF
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -340,12 +340,10 @@
         
         try {
           if (table.columns.length === 0) {
-            // Also, add a loader here like the tables loading one
             await this.$store.dispatch('updateTableColumns', table)
           }
           const existing = this.tabItems.find((tab) => tab.matches(t))
           if (existing) return this.$store.dispatch('tabs/setActive', existing)
-  
           this.addTab(t)
         } catch(err) {
           this.$noty.error('There was an error loading the table columns, please try again.')
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -339,7 +339,10 @@
         t.title = table.name
         
         try {
-          await this.$store.dispatch('updateTableColumns', table)
+          if (table.columns.length === 0) {
+            // Also, add a loader here like the tables loading one
+            await this.$store.dispatch('updateTableColumns', table)
+          }
           const existing = this.tabItems.find((tab) => tab.matches(t))
           if (existing) return this.$store.dispatch('tabs/setActive', existing)
   
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -332,16 +332,21 @@
         tab.unsavedChanges = true
         this.addTab(tab)
       },
-      openTableProperties({ table }) {
+      async openTableProperties({ table }) {
         const t = new OpenTab('table-properties')
         t.tableName = table.name
         t.schemaName = table.schema
         t.title = table.name
-
-        const existing = this.tabItems.find((tab) => tab.matches(t))
-        if (existing) return this.$store.dispatch('tabs/setActive', existing)
-
-        this.addTab(t)
+        
+        try {
+          await this.$store.dispatch('updateTableColumns', table)
+          const existing = this.tabItems.find((tab) => tab.matches(t))
+          if (existing) return this.$store.dispatch('tabs/setActive', existing)
+  
+          this.addTab(t)
+        } catch(err) {
+          this.$noty.error('There was an error loading the table columns, please try again.')
+        }
       },
       openTable({ table, filter}) {
         const tab = new OpenTab('table')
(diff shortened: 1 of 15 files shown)


apps/studio/src/components/CoreTabs.vue
```diff
@@ -332,16 +332,22 @@
...
         t.title = table.name
-
-        const existing = this.tabItems.find((tab) => tab.matches(t))
-        if (existing) return this.$store.dispatch('tabs/setActive', existing)
-
-        this.addTab(t)
+        
+        try {
+          if (table.columns.length === 0) {
+            await this.$store.dispatch('updateTableColumns', table)
```
rathboma, line 343: I'm not sure we need to do this - doesnt table properties update the table columns on load? If not, that's the place to do this because this way won't work after closing the app and restoring tabs.

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -778,6 +779,30 @@ import { FavoriteQuery } from '@/common/appdb/models/favorite_query'
       },
       fakeRemoteChange() {
         this.query.text = "select * from foo"
+      },
+      getColumnsForAutocomplete() {
+        const cm = this.editor.getValue() 
+        if (cm.toLowerCase().search(/[from(\s)?|join(\s)?]/g) === -1) return
```
rathboma, line 785: Is this how sqlhint identifies tables? Can we steal that logic instead?

This won't work with aliases, updates, deletes, etc.

apps/studio/src/components/tableinfo/TableSchema.vue
```diff
@@ -394,8 +412,13 @@ export default Vue.extend({
...
     }
   },
-  mounted() {
+  async mounted() {
     this.tabState.dirty = false
+    if (this.table.columns.length === 0) {
+      // If this tab was already opened when someone connected to the DB, there wouldn't be columns and you'd be sitting on an empty tab with no structure.
+      // well no more.
+      await this.$store.dispatch('updateTableColumns', this.table)
+    }
```
rathboma, line 421: We can just rely on this initializer and not don this work again in core tabs

apps/studio/src/vendor/sql-hint/index.js
```diff
@@ -0,0 +1,313 @@
...
+    if (!getTable(table)) {
+      var oldTable = table;
+      table = findTableByAlias(table, editor);
+      if (table !== oldTable) alias = true;
+    }
+
+    if (globalEditorOptions?.getColumns) {
+      await globalEditorOptions.getColumns()
+    }
+    var columns = getTable(table);
```
rathboma, line 173: This is where we should integrate our async column fetch, by providing a function to override `getTable`.

Their code does a much better job of table name resolution that I do not want to sidestep.

rathboma: You can just take a name in `getTableColumns` and just drop it in here.

rathboma (changes requested):
Looking pretty good.

Big points:
1. We should not be parsing tables ourselves, we should rely on SQL hint.
2. Component initializer should load columns, not tab launcher

apps/studio/src/vendor/sql-hint/index.js
rathboma: Eg:

Var columns = getTableColumns ? getTableColumns(table) : getTable(table)

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -778,6 +779,30 @@ import { FavoriteQuery } from '@/common/appdb/models/favorite_query'
...
+            if (triggerWords.includes(word.toLowerCase())) {
+              // will need to clean up the word if it's wrapped up in stuff
+              if (allTables.tables[arr[index + 1]]?.length === 0){
+                acc.push(arr[index + 1])
+              }
+            }
+            return acc
+          }, [])
+          .forEach(async(table) => {
+            const tableToFind = this.tables.find(t => t.name === table)
```
rathboma, line 802: Need to filter by schema here too

wmontgomery: I don't believe schema is passed through sql hint at this time, so would need to add that. If you'd like me to for this release I certainly can.

apps/studio/src/vendor/sql-hint/index.js
wmontgomery: Yup, overthought this one

apps/studio/src/components/TabQueryEditor.vue
rathboma: @wmontgomery Well you still need to filter to make sure the table is either in the default schema, or has no schema no?

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -412,9 +412,9 @@ import { FavoriteQuery } from '@/common/appdb/models/favorite_query'
         const [markStart, markEnd] = this.locationFromPosition(editorText, from, to)
         this.marker = this.editor.getDoc().markText(markStart, markEnd, {className: 'highlight'})
       },
-      hintOptions() {
+      tables() {
```
rathboma, line 415: don't think we need this anymore do we?

PR #1455, opened 2022-12-16
Multiple filters in builder mode

mahmudmihon:
This is an image
This is an image

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -6,8 +6,8 @@
       </div>
     </template>
     <template v-else >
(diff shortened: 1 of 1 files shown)


rathboma:
Hey! Thanks for your PR submission. I'll take a look at it next week

rathboma:
Hey! This looks awesome. I'm going to play around with it a little more this week. I'm worried about how much vertical screen space it will take up. I'm thinking maybe it should collapse when there are more than one filter. Let me figure out how that should look.

mahmudmihon:
Hi, the vertical space is fixed. So, for mutiple filters it will be scrollable.

PR #1463, opened 2022-12-26
PR to Fix Issue 1399 (sqlite int64s)

davidkaufman:
- Add new sqlite regression test that proves the error
- Fix bug by calling db.defaultSafeIntegers(true) on new connections
- Configue eslint-loader to target es2020 and recognize BigInt's as a type to fix startup error
- Update columnFilterTests in integration/lib/db/clients/all.js to handle sqlite BigInts
- Added useNullAsDefault option to knex connection to silence (some of) the "sqlite does not support inserting default values" warnings
- Fix (unrelated) error in QueryEditorStatusBar.vue's shortNum() function (when creating tables)

### DIFF
--- a/apps/studio/src/lib/db/clients/sqlite.js
+++ b/apps/studio/src/lib/db/clients/sqlite.js
@@ -25,10 +25,6 @@ const sqliteErrors = {
 
 const PD = SqliteData
 
-// Fix (part 1 of 3) Issue #1399 - int64s not displaying properly
-// Fixes "TypeError: Do not know how to serialize a BigInt"
-BigInt.prototype.toJSON = function() { return this.toString() }
-
 export default async function (server, database) {
   const dbConfig = configDatabase(server, database);
   logger().debug('create driver client for sqlite3 with config %j', dbConfig);
--- a/apps/studio/src/lib/db/clients/sqlite.js
+++ b/apps/studio/src/lib/db/clients/sqlite.js
@@ -643,10 +643,10 @@ async function runWithConnection(conn, run) {
   try {
     db = new Database(conn.dbConfig.database)
 
-    // Fix (part 2 of 3) Issue #1399 - int64s not displaying properly
+    // Fix (part 1 of 2) Issue #1399 - int64s not displaying properly
     // Binds ALL better-sqlite3 integer columns as BigInts by default
     // https://github.com/WiseLibs/better-sqlite3/blob/master/docs/integer.md#getting-bigints-from-the-database
-    // (Part 3 of 3 is in shared/src/lib/dialects/index.ts)
+    // (Part 2 of 2 is in apps/studio/src/common/initializers/big_int_initializer.ts)
     db.defaultSafeIntegers(true);
 
     const results = await run(db)
--- a/apps/studio/src/lib/db/clients/sqlite.js
+++ b/apps/studio/src/lib/db/clients/sqlite.js
@@ -14,7 +14,10 @@ import { ClientError } from '../client';
 const log = rawLog.scope('sqlite')
 const logger = () => log
 
-const knex = knexlib({ client: 'better-sqlite3'})
+const knex = knexlib({ client: 'better-sqlite3',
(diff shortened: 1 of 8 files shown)


apps/studio/src/lib/db/clients/sqlite.js
```diff
@@ -14,14 +14,20 @@ import { ClientError } from '../client';
...
 
 const sqliteErrors = {
   CANCELED: 'SQLITE_INTERRUPT',
 };
 
 const PD = SqliteData
 
+// Fix (part 1 of 3) Issue #1399 - int64s not displaying properly
+// Fixes "TypeError: Do not know how to serialize a BigInt"
+BigInt.prototype.toJSON = function() { return this.toString() }
```
rathboma, line 30: This modifies BigInts for the whole app if I'm reading this right?

If so we probably need to put this somewhere more obvious. I think somewhere like `src/common/initializers/big_int_initializer.ts`. You'll then also need to include and run this in `main.ts` and also in the test runners.

rathboma (changes requested):
Good job tracking this down! One change - can we move `BigInt` modification to somewhere more central so it is easy to find in the future.

apps/studio/src/lib/db/clients/sqlite.js
davidkaufman: It didn't work in the `.ts` file, which turned out to be A Good Thing because it reminded me that comments on the StackOverflow question that I got it from had this exchange:

> This works with pure JavaScript, but not TypeScript. Compiling with tsc gives error `TS2339: Property 'toJSON' does not exist on type 'BigInt'.  
> [Bruce](https://stackoverflow.com/users/4361237/bruce)

> There's a TypeScript solution in that link if you keep reading:
> - https://github.com/GoogleChromeLabs/jsbi/issues/30#issuecomment-1006088574  
> 
> [Adam Spiers](https://stackoverflow.com/users/179332/adam-spiers)

So I added the more complete JS-and-TS solution to the new initializer.  Fixing the test runner now

davidkaufman: Hmm, not sure why that build failed.  Looks like an Apple-y thing: 

>  skipped macOS application code signing  reason=cannot find valid "Developer ID Application" identity or custom non-Apple code signing certificate, it could cause some undefined behaviour, e.g. macOS localized description not visible, see https://electron.build/code-signing allIdentities=     0 identities found
>                                                 Valid identities only
> Skipping notarization: Authentication environment variables for notarization are missing. Either APPLE_ID and APPLE_ID_PASSWORD, or API_KEY_ID and API_KEY_ISSUER_ID must be defined.
> Error: Exit code: ENOENT. spawn /usr/bin/python ENOENT

But, actually, it seems like this should be moved into that new `/initializers/` file, so i'll psh that and see if it builds then

shared/src/lib/dialects/index.ts
```diff
@@ -5,6 +5,14 @@ import { PostgresData } from "./postgresql";
...
 import { SqlServerData } from "./sqlserver";
 
+// Fix (part 3 of 3) Issue #1399 - int64s not displaying properly
+// Fixes error 'BigInt' is not defined in lib/db/clients/sqlite.js
+declare global {
+  interface BigIntConstructor {
+    toJSON:()=>BigInt;
+  }
+}
+
```
davidkaufman, line 15: actually, it seems like this `declare global` TS code should be moved into that new `/initializers/` file, too

davidkaufman: (reviewing myself... 🙄)

apps/studio/src/lib/db/clients/sqlite.js
davidkaufman: Ok despite the (apparently unrelated) build failures, I think this is good to go.  Sorry for all the back and forth (e.g., dialects/index.ts is no longer even modified).  I'm happy to create a new branch and PR that is cleaner to commit to master, if you'd like! let me know

rathboma (approved):

Issue #747, opened 2021-07-29
FEAT: Find/replace in the query editor


skoshy:
**Is your feature request related to a problem? Please describe.**
I want to be able to find/replace in the query editor

**Describe the solution you'd like**
Pressing Ctrl/Cmd+F should show a find dialog with optional replacing. Right now it makes the app fullscreen. It'd be nice if it could even support regex like VSCode.

rathboma:
Yes, for sure.

wmontgomery:
PR available here https://github.com/beekeeper-studio/beekeeper-studio/pull/1464

PR #1464, opened 2022-12-28
Feature/747 find replace in editor

wmontgomery:
#747  
The styling presented is the default as given by codemirror with  an alteration to the length of the input field (inline style set it at 10em, I jumped it to 20em). I had considered having the label and input inline, but replace is written in a much different way that styling was becoming difficult to handle both sets, so decided to keep it the standard load out.

find-replace

find-light-theme

Update to button styles
image


rathboma:
@wmontgomery can you add styling for the buttons they add in so they match our other buttons?

rathboma (changes requested):
Lets try and apply our button styles here, make sure it works in both dark and light mode.

wmontgomery:
@rathboma , Fix has been made. Issue getting the PR build for macos since it seems to want some kind of apple id/key to build stuff.

rathboma:
Quick a before merging - does it look good in light mode also?

wmontgomery:
Ah nuts, the dialog.scss file isn't bringing in the light variables as I expected it would, so the buttons are white when they should be black. I'll look into it.

wmontgomery:
OK figured it out. Light theme update added

PR #1479, opened 2023-01-15
Bugfix/1477 create table

### LINKED ISSUE #1477: BUG: new update broke the 'create table' function
**Describe the bug**
new update broke the 'create table' function

**To Reproduce**
1. Go to a DB
2. Click on the + icon to create a new table
3. See error --> page is blank

**Expected behavior**
should have a table creation gui

**Screenshots / Gifs**
Screen_Shot_2023-01-13_at_16_35_10

 - OS: [e.g. Windows/Mac/Ubuntu]
 - App Version [e.g. help -> About Beekeeper Studio]
 - Database type and version [eg Posgresql 9.3]

**Additional context**

wmontgomery:
fix #1477 

Issue was mysql and sqlite were missing the defaultSchema functions, so those were added to each client/*dialect*.js file as an empty string.

Also fixed a style issue when closing a table-builder tab.

### DIFF
--- a/apps/studio/src/lib/db/clients/sqlite.js
+++ b/apps/studio/src/lib/db/clients/sqlite.js
@@ -38,6 +38,7 @@ export default async function (server, database) {
     supportedFeatures: () => ({ customRoutines: false, comments: false, properties: true }),
     versionString: () => getVersionString(version),
     wrapIdentifier,
+    defaultSchema: () => '',
     disconnect: () => disconnect(conn),
     listTables: () => listTables(conn),
     listViews: () => listViews(conn),
--- a/apps/studio/src/assets/styles/app/_layout.scss
+++ b/apps/studio/src/assets/styles/app/_layout.scss
@@ -609,6 +609,9 @@ input[type=file] {
   &.table-builder {
     color: $brand-info;
   }
+  &.table-builder-icon {
+    color: $brand-info;
+  }
 }
 
 // Split.js Resizer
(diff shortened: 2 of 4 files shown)


rathboma (approved):

rathboma:
@wmontgomery does that need adding to Oracle and Cassandra in the ultimate edition?

wmontgomery:
@rathboma There is a defaultSchema function in the oracle client, and I'll be sure to make note of it for the Cassandra side when I'm working through Write. Chances are it will need to be since most of the client pieces were taken from MySQL's.

apps/studio/src/lib/db/clients/sqlite.js
```diff
@@ -38,6 +38,7 @@ export default async function (server, database) {
     supportedFeatures: () => ({ customRoutines: false, comments: false, properties: true }),
     versionString: () => getVersionString(version),
     wrapIdentifier,
+    defaultSchema: () => '',
```
rathboma, line 41: @wmontgomery i think if there is no schema this should return null, no?  Aren’t we going to want to do checks like if(schema).

rathboma: This is how we handle no schema elsewhere in the code. Can we make this change and check table creation still works?

rathboma: I'm changing it to return null

PR #1482, opened 2023-01-20
fix(tabletable): move the gathering of the filters until after the columns are returned

wmontgomery:
select-dropdown


wmontgomery:
@rathboma , fix for the email issue that you sent over.

rathboma:
Dang. Easy fix. The table table thanks you

rathboma (approved):

Issue #1410, opened 2022-10-26
FEAT: Show comments on the table structure


Karlies:
Hey, I'm using beekeeper to connect my postgresql, I comment a lot for my co-workers ,but I can't see it when I click [View Structure] . Now there are only [Name, Type, Nullable, Default Value and Primary]
If you can add a column of  [Comment], it would be very useful for development.

vioulo:
> 

hi， how did you do it？
Also, how to display comments for tables？

GKyomen:
Hi, I would really like to see comments too. Is there a way or a prevision to when this feature will be done?

PR #1487, opened 2023-01-31
feat: adds comment column on table info view

leduard:
Hello! Thats my first open-source PR, so please go easy on me 😄 

Resolves: #1410 and #823 (partially i think?)

Changelog:
- a new field on table viewer that show the comments on columns
- retrieves comments from postgreSQL tables
  - something like that was already implemented for the mysql client, but it was never displayed on the schema viewer

Recording 2023-01-31 at 00 30 48

### DIFF
--- a/apps/studio/src/components/tableinfo/TableSchema.vue
+++ b/apps/studio/src/components/tableinfo/TableSchema.vue
@@ -231,14 +231,14 @@ export default Vue.extend({
           width: 70,
           cssClass: "read-only never-editable",
         },
-        {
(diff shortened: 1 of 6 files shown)


rathboma:
Thanks for adding this! Yes I originally had it but worried it made the table too wide. I guess it's still better to have it than to not.

rathboma (approved):

nicoxiang:
Why this commit is reverted? this is a important feature.

not-night-but:
@nicoxiang This merge was reverted as it caused other issues (see #1568).

Issue #1461, opened 2022-12-23
BUG: SQL: Create doesn't work for materialized views


krystxf:
**Describe the bug**
In the `Entities` section, if I open an options menu on the entity, that is a materialized view, in which there is an option `SQL: Create`. It works just fine for views, but for the materialized view I get this error: `Can't find script for example_materialized_view (materialize-view)`

**To Reproduce**
1. Go to the ENTITIES section
2. Right-click on an entity that is a materialized view (open the dropdown menu)
3. Click on `SQL: Create`
4. See the error

**Expected behavior**
Open a new tab with SQL query which creates the same materialized view (same as with view).

**Screenshots / Gifs**
![image](https://user-images.githubusercontent.com/48121710/209299585-4d140ec1-7761-4882-b06a-1f76f687b31e.png)

**Version Information (please complete the following information):**
 - OS: Arch Linux x64, kernel 5.15
 - App Version 3.7.9
 - PosgreSQL 13.9

**Additional context**
`SELECT pg_get_viewdef('example_materialized_view', true);` returns the SQL query

rathboma:
Thanks for bringing this up!

krystxf:
there are also other issues with view and materialized views
1. there's truncate option even tho it returns warning message that you can't truncate views 
2.  same problem as with create but with drop materialized view

should I create another issue, or should we connect it to one?

rathboma:
No, we can leave it here and I'll rename the issue. Thanks!

krystxf:
This is my favorite software for working with databases so I'm glad I can help you make it even better

krystxf:
Here is a solution prototype #1502

PR #1502, opened 2023-02-17
FIX: Postgres create and delete materialized view

krystxf:
this fixes some issues from #1461 

- fix `SQL: create` in context menu functionality for materialized views
- fix `Drop` in context menu functionality for materialized views

### DIFF
--- a/apps/studio/src/lib/db/client.ts
+++ b/apps/studio/src/lib/db/client.ts
@@ -12,6 +12,7 @@ const logger = createLogger('db');
 export enum DatabaseElement {
   TABLE = 'TABLE',
   VIEW = 'VIEW',
+  "MATERIALIZED-VIEW" = 'MATERIALIZED VIEW',
   DATABASE = 'DATABASE'
 }
 
--- a/apps/studio/src/lib/db/client.ts
+++ b/apps/studio/src/lib/db/client.ts
@@ -65,7 +65,7 @@ export interface DatabaseClient {
   getTableProperties: (table: string, schema?: string) => Promise<TableProperties | null>,
   getTableCreateScript: (table: string, schema?: string) => Promise<string>,
   getViewCreateScript: (view: string) => void,
-  getMaterializedViewCreateScript: (view: string) => void,
+  getMaterializedViewCreateScript?: (view: string) => Promise<string[]>,
   getRoutineCreateScript: (routine: string, type: string, schema?: string) => void,
   truncateAllTables: (db: string, schema?: string) => void,
   listMaterializedViews: (filter?: FilterOptions) => Promise<TableOrView[]>,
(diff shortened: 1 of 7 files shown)


apps/studio/src/lib/db/client.ts
```diff
@@ -484,6 +487,11 @@ function getViewCreateScript(server: IDbConnectionServer, database: IDbConnectio
   return database.connection?.getViewCreateScript(view);
 }
 
+function getMaterializedViewCreateScript(server: IDbConnectionServer, database: IDbConnectionDatabase, view: string /* , schema */) {
```
rathboma, line 490: Given that this is only implemented for postgresql, this function needs to check if this function exists before calling it, and otherwise return a default value, probably `null`.

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1372,7 +1385,7 @@ export async function truncateAllTables(conn: Conn, schema: string) {
 export async function dropElement (conn: Conn, elementName: string, typeOfElement: DatabaseElement, schema: string = 'public'): Promise<void> {
   await runWithConnection(conn, async (connection) => {
     const connClient = { connection };
-    const sql = `DROP ${PD.wrapLiteral(typeOfElement)} ${wrapIdentifier(schema)}.${wrapIdentifier(elementName)}`
+    const sql = `DROP ${PD.wrapLiteral(DatabaseElement[typeOfElement])} ${wrapIdentifier(schema)}.${wrapIdentifier(elementName)}`
```
rathboma, line 1388: Forgive my ignorance here - how is this change different in function to the prior version?

apps/studio/src/lib/db/client.ts
```diff
@@ -64,6 +65,7 @@ export interface DatabaseClient {
   getTableProperties: (table: string, schema?: string) => Promise<TableProperties | null>,
   getTableCreateScript: (table: string, schema?: string) => Promise<string>,
   getViewCreateScript: (view: string) => void,
+  getMaterializedViewCreateScript: (view: string) => void,
```
rathboma, line 68: This should return the type `Promise<String>`, and it should be optional also:

`getMaterializedViewCreateScript?:`, similar to `defaultSchema`

rathboma (changes requested):
Thank you so much for your PR :pray: .

I have a couple of minor changes I'd like for you to make before I merge it, but this is an awesome contribution. Thanks so much.

apps/studio/src/lib/db/clients/postgresql.ts
krystxf: it works like a key-value this way, it didn't make any difference since `typeOfElement` were same as values (one word), but with materialized view `typeOfElement` is `materialized-view`, but value that has to be filled in here is `MATERIALIZED VIEW`, I thought this was the easiest way how fix it.

apps/studio/src/lib/db/client.ts
krystxf: The function is pretty similar to `getViewCreateScript` function so I used the same types, isn't  in that case mistake also in `getViewCreateScript` function?

apps/studio/src/lib/db/clients/postgresql.ts
rathboma: Ok, sounds good. Looking forward to the other minor changes then will merge :+1:

krystxf:
Here [e5e55ca](https://github.com/beekeeper-studio/beekeeper-studio/pull/1502/commits/e5e55cad649ce2e545a7b937e44769a734d845e5) I added check if function exists

I also added a function to the other databases files (`mysql.ts`, etc.), following the pattern of how functions that the database does not have were handled

apps/studio/src/lib/db/client.ts
rathboma: Yeah the interface is wrong for many functions! A contributor converted us to TS and some stuff is half converted :-)

krystxf: so is this conversation resolved or should I change something?

rathboma (approved):
You rock! Thanks so much for this

PR #1509, opened 2023-02-17
#1484 - Editing Disabled message.

### LINKED ISSUE #1484: BUG: We call a table 'read only' but what we really mean is 'no editing or deleting'. Fix the messages to make this clear.
**Describe the bug**
Tables without primary key column have a little note at the bottom right saying 'Read only: Only tables with a single primary key column are editable'.
However, it is very much possible to add empty rows to such a table using either `Ctrl-N` or the little plus button on the  bottom right and to commit those changes, making the table very much not read-only.

**To Reproduce**
- Create a table without primary key (and without any NOT NULL) constraints, eg `CREATE TABLE test(some_text text);`
- Open the table 
- Press `Ctrl+N` or the little plus button on the bottom right
- Press `Apply`

**Expected behavior**
Well, taking the 'Read Only' seriously, I would expect the plus button the not be present/grayed out, and `Ctrl+N` to have no effect.
However, it's not clear to me why **adding** rows to a table without primary key is forbidden in the first place -- so a better solution would be to allow adding rows /with content/ (not just NULLs as is the case now...).
And while I'm here, even more awesome would be to allow editing rows if their initial values allow unique identification in the table -- but I can see how that's full of ugly, as well as some potentially quite expensive queries, so that's clearly beyond this bug report....

 - OS: Ubuntu
 - App Version 3.7.10
 - Database Type: PostgreSQL 14.5 (Ubuntu 14.5-0ubuntu0.22.04.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.2.0-19ubuntu1) 11.2.0, 64-bit

not-night-but:
Fix #1484, Message now reads "Editing Disabled" and the tooltip has been changed to reflect this:

image

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -504,7 +504,7 @@ export default Vue.extend({
     },
     readOnlyNotice() {
       return this.dialectData.notices?.tableTable ||
-        "Tables without a primary key column only support insert and delete. Editing of existing records is disabled."
+        "Tables without a primary key column only support inserts. Editing of existing records is disabled."
     },
     // it's a table, but there's no primary key
     missingPrimaryKey() {
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -155,7 +155,7 @@
           </x-button>
         </template>
         <template v-if="!editable">
-          <span class="statusbar-item" :title="readOnlyNotice"><i class="material-icons-outlined">info</i> Read Only</span>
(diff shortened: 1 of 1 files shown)


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -504,7 +504,7 @@ export default Vue.extend({
     },
     readOnlyNotice() {
       return this.dialectData.notices?.tableTable ||
-        "Only tables with a single primary key column are editable."
+        "Tables without a primary key column only support insert and delete. Editing of existing records is disabled."
```
rathboma, line 507: Actually they don't support delete :-)

rathboma (changes requested):
delete isn't supported actually!

not-night-but:
@rathboma agh my bad 🤦

rathboma (approved):

PR #1512, opened 2023-02-18
1472 selectable fields

### LINKED ISSUE #1472: FEAT: selectable fields in the table structure view
When coding, I often need to go find the exact name of a column in a table. Using Beekeeper, I open the "table structure" view. Ideally I want to be able to highlight a column name and copy it to clipboard. But I can't even highlight text fields in this view.

Would it be possible to make it happen?

not-night-but:
Fix #1472. Now this tooltip will show when a column name cannot be edited:
image

And the user can click the column name to copy it to clipboard


rathboma:
Can we also add the `selected` style that we use for non-editable cells in the TableTable view?

Peek 2023-02-20 21-38

rathboma (changes requested):
Looks good, but for consistency, lets add the cell-selected style we use in TableTable too.

apps/studio/src/components/tableinfo/TableSchema.vue
```diff
@@ -417,6 +420,18 @@ export default Vue.extend({
         data: this.tableData,
         placeholder: "No Columns",
       })
+    },
+    columnNameCellClick(_e: any, cell: CellComponent) {
```
rathboma, line 424: If you click away from the table does it unselect the cell? Take a look at table table for how to do that.

rathboma (changes requested):
This needs to handle the case when someone clicks away from the cell, in that case it should de-select it. See how we do this in tabletable (we add a global click watcher).

not-night-but:
@rathboma handling global clicks now

rathboma (approved):
Actual change looks good! Need to fix the weird other changes before merging though.

not-night-but:
> The diff on the pr looks like it has a bunch of other changes in it? Can you merge in master and push up?
> 
> this might be a GitHub oddity

Actually looking at the commit history I'm super confused as to what's happened. I try to keep my branches up to date with main, something must've happened during one of the merges?

Issue #1524, opened 2023-02-24
BUG: Materialized View not showed on Windows PG


MaximeRaynal:
**Describe the bug**
Materialized view are not displayed in the entity explorer when the database is PostgreSQL installed on Windows (tested on 14.2).

**To Reproduce**
1. Install PostgreSQL on Windows (https://www.postgresql.org/download/windows/)
2. Create a Materialized View
3. You can't see it in the entity list

**Version Information (please complete the following information):**
 - OS: Windows
 - App Version: 3.8.10
 - Database type and version: PostgreSQL 14.2

**Why**

It's cause of version interpretation.

Version "number" is interpreted as follow (version is response of `SELECT version()`):

```js
parseInt(
      version.split(" ")[isPostgres ? 1 : 2].replace(/^v/i, '').split(".").map((s: string) => s.padStart(2, "0")).join("").padEnd(6, "0"),
      10
    )
```

On a Windows PostgreSQL installation there is a comma just after the version number: `PostgreSQL 14.2, compiled by Visual C++ build 1914, 64-bit`.

So `version.split(" ")[isPostgres ? 1 : 2].replace(/^v/i, '').split(".").map((s: string) => s.padStart(2, "0"))` 
is `['14', '2,']`  and
`version.split(" ")[isPostgres ? 1 : 2].replace(/^v/i, '').split(".").map((s: string) => s.padStart(2, "0")).join("").padEnd(6, "0")` => `142,00`.

So version number became `142` after a `parseInt` instead of `142000` excepted.

MaximeRaynal:
[PR](https://github.com/beekeeper-studio/beekeeper-studio/pull/1525)

MaximeRaynal:
PR merged I can close

PR #1525, opened 2023-02-24
FIX bad version interpretation on Windows

MaximeRaynal:
Due to a comma in version output causing this issue https://github.com/beekeeper-studio/beekeeper-studio/issues/1524

I just update the replace regex to remove comma if present.

It would be better to refacto the whole function based on regex but need to etablish a test set first.


MaximeRaynal:
Tests fails due to:
`error: the database system is starting up` => `There is no connection available.`  
For all queries.

Is it possible to re try ? Seem like a bug in test env.

rathboma:
Does this break version strings if psql is running on Linux?

MaximeRaynal:
No I just modify the regex to remove the comma if present (as it's done for the `v`).

PR #1528, opened 2023-02-26
feat: migrate redshiftOptions to iamAuthOptions

ndom91:
- Updated and fixed typescript + eslint + vue setup
- It seems eslint/TS was not working correclty for who knows how long
- Added `@aws-sdk/*` to vue's `transpileDependencies` because newer features like nulllish coalescing inside those packages was breaking the build, I guess this electron version is on a slightly older version of node(?)


ndom91:
Error

PR #1529, opened 2023-02-26
fix: eslint/TS setup and all initial errors

ndom91:
- Updated and fixed typescript + eslint + vue setup
- It seems eslint/TS was not working correclty for who knows how long
- Added `@aws-sdk/*` to vue's `transpileDependencies` because newer features like nulllish coalescing inside those packages was breaking the build, I guess this electron version is on a slightly older version of node(?)
- Some of the files got automatically formatted based on the newly applied (again) eslint settings. Meaning there are some style changes in here already. More will come as people save additional files with the settings applied again.
    - If you don't like any of them, we'll have to change them in the config files (`.eslintrc.json`), right.


rathboma:
I don’t think this pr includes the eslintrc file? Although I am on an iPad so I might just not be able to find it.

ndom91:
Yeah its definitely not ready yet, I just pulled out some of the initial commits from that other PR to here. I'll update it to "ready for review" once I've double checked everything :wink:

rathboma:
I’ll try and take a look on Friday. Thanks for this. It’s a big pr for sure.

ndom91:
Just wanted to pop in with a friendly reminder - the longer it takes, the more out of sync this gets with `main` and the harder (maybe impossible :sweat_smile:) it gets to merge this haha 

CC: @rathboma

rathboma:
@not-night-but is this possible to merge? Wanna take a shot?

not-night-but:
The build is passing!!! Spent all day dealing with merge conflicts and trying to figure out how far we really wanted to take the linting. In the end I ended up adding the following to the disabled rules:

1. no-explicit-any (this would mean that we would have to remove every `any` in the codebase)
2. explicit-module-boundary-types (this would mean that we couldn't use `any` on any exported code)
3. require-prop-types (this would require us to redo all our props in every vue component)
4. require-default-prop (this would require us to redo all our props in every vue component)
5. no-non-null-assertion (for the amazon_redshift authenticator)

If anyone disagrees with any of the above disabled rules, we can eventually re-enable them in the future, I just think this PR was already getting way too big, and adding on that many more changes just seems like asking for trouble.

I also added a new script: `all:lint`. This runs eslint for sqltools, beekeeper-studio, and the shared folder. Just makes it a little easier to check that everything is in order before pushing.

ndom91:
Awesome, @not-night-but! Thanks for coming back to this. 

Happy we can correctly enable the linter and TS again here 🥳 

Regarding the rule changes, I think that's totally fine for getting this out initially. We could follow-up with tightening those up / cleaning up afterward

not-night-but (approved):
Just gonna approve this and merge as most of the changes are formatting anyways.

PR #1530, opened 2023-02-26
Pinned connections

### LINKED ISSUE #1496: FEAT: The ability to pin frequently used connections (same as we pin tables)
In the connections tab we should be able to `pin` connections to the top, just like we do with tables in the core interface

not-night-but:
Closes #1496

This PR adds the ability for users to pin frequently used connections in the connections sidebar.
image
image

Based on how we handle pinning of tables in the table list, pinned connections are stored in a table in the beekeeper database, and are available in the app through a vue store.

I left all of the actions that are enabled for saved connections on the pinned connections, let me know if this should change.


rathboma:
Looks good, quick question, can we hide the pinned area when there are no pinned connections? I think we do this with tables also

not-night-but:
@rathboma yeah I can definitely do that! Might just have to mess around with split.js for a bit

rathboma:
My guess is we reinitialize split when someone pins something

not-night-but:
@rathboma Changes made! The pinned section doesn't show up unless there are pins to show. We reinitialise the split object if there are less than 2 pins after a change.
image
image

rathboma (changes requested):
Did a little testing on this. Overall - nice work!

Needs fixing:
- Pin some connections, then ctrl+shift+r to reload the app, no pinned section appears, but they're still marked as pinned in the saved area. This can be solved by removing all but one pinned connection.
- If I deleted a saved connection which is pinned it should also delete the pin.

Questions:
- Have you tested this with workspaces in the ultimate edition?

not-night-but:
@rathboma Both of those issues fixed now. I have not tested this in ultimate, I will now tho. I am storing the workspaceId and filtering pins based on that, so I expect it should be good?

apps/studio/src/components/sidebar/ConnectionSidebar.vue
```diff
@@ -217,29 +268,43 @@ const log = rawLog.scope('connection-sidebar');
...
-        elementStyle: (dim, size) => ({
-          'flex-basis': `calc(${size}%)`
-        }),
-        direction: 'vertical',
-        sizes: this.sizes
-      })
+      this.split = this.getSplit()
     },
     methods: {
+      getSplit() {
```
rathboma, line 289: So sometimes you destroy the existing split before creating a new one and sometimes you don't. I think the reason for that is that you use `getSplit` to simply return a split, but really I think it would be more idiomatic to have it be called `buildSplit`.

I think you can merge it, but here's a minor change that you should make:


```js
methods {
  buildSplit() {
    if (this.split) this.split.destroy()
    this.split = new Split(...)
  },
  ...
}
```
Then you just call it, rather than using it to assign

apps/studio/src/components/sidebar/ConnectionSidebar.vue
```diff
@@ -217,29 +268,43 @@ const log = rawLog.scope('connection-sidebar');
...
       importFromLocal() {
         console.log("triggering import")
         this.$root.$emit(AppEvent.promptConnectionImport)
       },
       refresh() {
         this.$store.dispatch('data/connectionFolders/load')
         this.$store.dispatch('data/connections/load')
+        this.$store.dispatch('pinnedConnections/loadPins').then(() => this.$store.dispatch('pinnedConnections/reorder', this.pinnedConnections))
+        this.split.destroy()
+        this.split = this.getSplit()
```
rathboma, line 307: You don't need to call it explicitly, it will be called with the watcher.

In Vue, it's best to rely on watchers to dispatch events whenever possible, that way it's harder to forget something if you make changes in several different places.

apps/studio/src/store/modules/PinConnectionModule.ts
```diff
@@ -0,0 +1,77 @@
...
+      if (existing) {
+        return;
+      }
+
+      const newPin = new PinnedConnection(item);
+      newPin.position = (context.getters.orderedPins.reverse()[0]?.position || 0) + 1;
+      await newPin.save();
+      context.commit('add', newPin);
+    },
+    async reorder(_context, pins: PinnedConnection[]) {
```
rathboma, line 65: I see we do something similar in the PinModule. One difference is that you're not calling `set` here. You should be!

apps/studio/src/store/modules/PinConnectionModule.ts
```diff
@@ -0,0 +1,77 @@
...
+      pins.forEach((p, idx) => p.position = idx);
+      await PinnedConnection.save(pins);
+    },
+    async remove(context, item: SavedConnection) {
+      const existing = context.state.pins.find((p) => p.connectionId === item.id);
```
rathboma, line 70: You also need to filter for workspaceId

apps/studio/src/store/modules/PinConnectionModule.ts
```diff
@@ -0,0 +1,77 @@
...
+          workspaceId: context.rootState.workspaceId
+        }
+      });
+      context.commit('set', pins || []);
+    },
+    async unloadPins(context) {
+      context.commit('set', []);
+    },
+    async add(context, item: SavedConnection) {
+      const existing = context.state.pins.find((p) => p.connectionId === item.id)
```
rathboma, line 55: you also need to check workspaceId

apps/studio/src/store/modules/PinConnectionModule.ts
```diff
@@ -0,0 +1,77 @@
...
+    pins: []
+  }),
+  getters: {
+    pinned(state: State): PinnedConnection[] {
+      return state.pins;
+    },
+    orderedPins(_state, getters, rootState): PinnedConnection[] {
+      const connections = rootState['data/connections'].items;
+      return getters.pinned.sort((a, b) => a.position - b.position).map((pin: PinnedConnection) => {
+        const c = connections.find((c) => c.id === pin.connectionId);
```
rathboma, line 22: You also need to check workspaceId

apps/studio/src/components/sidebar/ConnectionSidebar.vue
```diff
@@ -149,6 +184,18 @@ const log = rawLog.scope('connection-sidebar');
...
     watch: {
+      // If we load with some pins, this will reinitialize split to reflect that
+      noPins(value) {
+        if (!value)
+          this.split = this.getSplit()
+      },
+      // Check if we need to reinitialize split based on pinnedConnections length
+      pinnedConnections(value) {
+        if (value.length < 2) {
+          this.split?.destroy()
```
rathboma, line 195: if you always need to call destroy before setting a new split, change the method to `buildSplit` and do it in there. See below comments.

rathboma (changes requested):
Couple of minor changes still to make here, then it's good to merge.

apps/studio/src/store/modules/PinConnectionModule.ts
not-night-but: Do we? The only pins in the state are the ones for the current workspace.

rathboma: In theory, no. In reality it’s probably possible to get in a weird state where there is some data in the store for a different workspace.

I had many such bugs when I launched workspaces!

not-night-but: @rathboma ah okay, will do

not-night-but:
@rathboma changes made, tested in ultimate with workspaces, everything seems to work fine.

rathboma (approved):

PR #1533, opened 2023-03-01
Added Vim Keybindings to the Editor

### LINKED ISSUE #317: FEAT: vim key bindings for query editing
I like to do as much of my coding in vim as possible. Vim does not have a great story when it comes to SQL editing and Beekeeper Studio looks like an awesome app otherwise.

I would love to be able to continue using vim key bindings in the Beekeeper Studio Query Editor

There is a way how you can replicate this app in vim, but I found it confusing, I prefer to have a native app for this task.
I'm using TablePlus now, but it does not have vim support either.

austinwilcox:
Close #317 

This commit adds the option for keymaps to the connection interface and allows a user to select default, or vim. With vim I added some additional code in the TabQueryEditor so that Esc is not bound to cancel a query. That way You do not need to use CTRL+[ to exit insert mode. We may need to bind cancel query to a different keypress, thoughts?


rathboma:
Hmm, yes good point, I don't use VIM so I don't know what keyboard shortcuts are common, but what about ctrl+esc or something like that?

austinwilcox:
Ctrl-Esc would be a good one for vim users to cancel a query. Let me go in and add that keybinding.

austinwilcox:
Added that additional line to cancel a query with the ctrl+esc combo. I ran my tests on it and everything worked as expected.

rathboma:
Thanks for contributing this! I think probably the best home for this setting is on the sql editor screen on the same line as the run button.

Also looks like this value doesn’t persist between reloads — that’s probably something we’ll want to do also using the Setting model and the settings vuex store.

If you’re able to make these changes that’d be great, otherwise I’ll probably need a couple of weeks to get to it.

rathboma (changes requested):
See above comment about location and persistence

austinwilcox:
I worked on changing where the keymap is selected, I was not able to get the styles the exact same, and I'm not confident in the way that I am handling the event to reload the editor, could I get a bit of guidance on that? My first thoughts were to just reload the instance of the editor, and I can trigger that by setting  a key on the parent div and then re-initialize the component when I change the keymap selection. Once that's in place, I will then add the Keymap to the settings model in the vuex store.

rathboma:
Let me take a look, thanks!

rathboma:
So I took a look at the code. I think we probably need to integrate this with the UserSetting model and either the VueX store, or the typrorm event subscription system so that query editor tabs can effectively listen for setting changes.

Let me think about this and figure out a way to implement. To do it right now would require me to write some new framework code that we don't have right now

not-night-but:
@rathboma @austinwilcox 

Hey! Made a couple changes to this to better fit with the style of the app/how other settings are set in the app.

There's now a settings cog in the bottom left of the query editor for switching keymaps. The keymap is saved to the UserSettings and will stay as the last selected keymap until it is change again :)
vim

Thanks!
Day

rathboma (approved):
SHIP IT :tada:

rathboma:
@not-night-but fix conflicts and merge away!

rathboma:
For a future PR, here's a MacOS-like Emacs mode I use on duckdocs.com:

https://gist.github.com/rathboma/cc2276a4730262ff6acb45f8396bc318

not-night-but:
Coolio! I was also planning on eventually adding a helix keymap because ... well I love helix lol. Shouldn't be too hard to integrate those with the official CM keymaps

austinwilcox:
@not-night-but Thank you for pushing this to the finish line. This is going to be incredible to have. Also love the idea of the helix keybindings, something I think I'll look at adding in the future is the ability to add your own custom keybindings based on a .vimrc. Thanks to Helix, I have to have gl and gh to navigate to the beginning and end of a line, and I add that in every vimrc I use now.

not-night-but:
@austinwilcox Yeah for sure! Thanks for doing the initial work! I love the idea of using the .vimrc, or at the very least custom keybindings.

rathboma:
Yeah I think Beekeeper Studio needs a whole suite of configuration and customization options in the future :-)

PR #1537, opened 2023-03-04
Interval Sorting

### LINKED ISSUE #1466: BUG: Interval column sorting is incorrect
**Describe the bug**
Having a (postgres) database column of type 'interval' the front-end sorting in beekeeper studio is not correct --  see screenshot (the fourth entry of 27mins is in the wrong place).

image

with the changes proposed in #1465 the behaviour is still incorrect:

image

(see #1442 for some background to this report).

**To Reproduce**
- create a postgres table with a column of type `Interval`
- insert some values into the table (see screenshots above for values)
- display table in beekeeper studio, and try to sort by the value of the interval column

**Expected behavior**
Sorting is in increasing/decreasing length of the interval.

- OS: Ubuntu
- App Version 3.7.2
- PostgreSQL 14.5 (Ubuntu 14.5-0ubuntu0.22.04.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.2.0-19ubuntu1) 11.2.0, 64-bit

not-night-but:
Fixes #1466

This solution does fix interval sorting, but I feel like it's kinda hacky?

adds two new small packages: postgres-interval and tinyduration

postgres-interval is used to parse postgres intervals into ISO 8601 Duration strings.
tinyduration is used to parse those ISO strings to an object.

image
image


apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -304,6 +311,19 @@
...
+      // HACK (day): this is probably not the best way of doing things, but postgres intervals are dumb
+      intervalSorter(a, b, aRow, bRow, column, dir, sorterParams) {
+        try {
+          const durationA = td.parse(intervalParse(a).toISOString());
+          const durationB = td.parse(intervalParse(b).toISOString());
+          const dateA = new Date(durationA.years, durationA.months, durationA.days, durationA.hours, durationA.minutes, durationA.seconds);
+          const dateB = new Date(durationB.years, durationB.months, durationB.days, durationB.hours, durationB.minutes, durationB.seconds);
+          return dateA - dateB;
+        } catch {
+          console.error('ERROR SORTING INTERVALS');
```
rathboma, line 324: So if this errors it could create O(n*n) console errors right? Maybe just have it return 0 without doing a console error. Not ideal but console spam can really slow the app.

rathboma (approved):
This looks great. Merge it!

I'd remove the console.error if you can though

PR #1541, opened 2023-03-09
Copy to sql

### LINKED ISSUE #1534: Add 'Copy to SQL' to the table editing interface
We allow a user to copy the SQL of an action rather than applying it automatically, we should do the same thing for table data changes.

See as reference: alterTable and alterTableSql

not-night-but:
Close #1534 

Copy to SQL for data view.
I mainly went off of the implementation for this feature in the tableSchema view.

Adds a dropdown to the apply button in data view:
image

When clicked, changes that are pending are converted to sql and opened in a new tab:
image
image

Note: I ran into an issue with updating data in sqlite tables. It was an issue with the knex client, which I did fix, but the solution seems kinda hacky to me.

This has been tested on Sqlite3, Sql Server, Postgres, and MySql

### DIFF
--- a/apps/studio/src/lib/db/client.ts
+++ b/apps/studio/src/lib/db/client.ts
@@ -59,6 +59,7 @@ export interface DatabaseClient {
   alterRelationSql: (changes: RelationAlterations) => string | null
   alterRelation: (changes: RelationAlterations) => Promise<void>
 
+  getChangesSql: (changes: TableChanges) => Promise<string>,
   getInsertQuery: (tableInsert: TableInsert) => Promise<string>,
   getQuerySelectTop: (table: string, limit: number, schema?: string) => void,
   getTableProperties: (table: string, schema?: string) => Promise<TableProperties | null>,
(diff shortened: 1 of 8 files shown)


apps/studio/src/lib/db/client.ts
```diff
@@ -59,6 +59,7 @@ export interface DatabaseClient {
   alterRelationSql: (changes: RelationAlterations) => string | null
   alterRelation: (changes: RelationAlterations) => Promise<void>
 
+  getChangesSql: (changes: TableChanges) => Promise<string>,
```
rathboma, line 62: These changes look great.

one nitpick: to be consistent with the other methods where we have a x, and a xSql can we name it applyChangesSql?

rathboma (changes requested):
Great work!

super minor change requested.

This also needs a basic integration test to at least call the function and check some basic text is present. Don’t try and match the whole string, just look for “insert”, etc.

also, when you merge this you’ll need to add support to Cassandra and oracle in the ultimate edition.

not-night-but:
@rathboma Changes made.  I also made applyChangesSql synchronous, as it doesn't actually need to be async at all.

rathboma (approved):
🎉

PR #1543, opened 2023-03-11
Feat: duplicate table

lanzoninicola:
The new feature adds an item to the contextual menu that appears when the user right-clicks on a table. This item allows the user to duplicate the table, copying both the structure and the data (except indexes, triggers, constraints) (it works doing a "select * into <new_table> from <table_name>).

The video that has been provided demonstrates how the feature works in action, allowing users to see exactly how the new functionality will benefit them. Overall, this feature should make it much easier and more efficient for users to duplicate tables, saving them time and effort in their database editing work.

Link to the video: https://youtu.be/cn0dG4YSIZk

Furthermore, it is worth mentioning that in the future, there is also a possibility to add a flag that allows the user to choose how much data to copy when duplicating a table. Specifically, the flag could enable the user to copy all data, only a certain number of records (X records), or no data at all. This added functionality would provide even more flexibility and control to users when using the duplicate table feature.

Tested with MYSQL, POSTGRES, MSSQL.

Thanks

### DIFF
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -89,7 +89,7 @@
           <span class="expand"></span>
           <button ref="no" @click.prevent="$modal.hide(duplicateTableModal)" class="btn btn-sm btn-flat">Cancel</button>
           <!-- <button @focusout="sureOpen && $refs.no && $refs.no.focus()" @click.prevent="completeDuplicateTableAction" class="btn btn-sm btn-primary">{{this.titleCaseAction}} {{this.dbElement}}</button> -->
-          <pending-changes :submit-apply="duplicateTable" :submit-sql="duplicateTableSql" />
+          <pending-changes-button :submit-apply="duplicateTable" :submit-sql="duplicateTableSql" />
         </div>
       </modal>
     </portal>
@@ -116,7 +116,7 @@
   import TabWithTable from './common/TabWithTable.vue';
   import TabIcon from './tab/TabIcon.vue'
   import { DatabaseEntity } from "@/lib/db/models"
-  import PendingChanges from '../components/common/PendingChanges.vue'
+  import PendingChangesButton from './common/PendingChangesButton.vue'
 
   export default Vue.extend({
     props: [ 'connection' ],
@@ -131,7 +131,7 @@
       TableBuilder,
       TabWithTable,
       TabIcon,
-      PendingChanges
+      PendingChangesButton
     },
     data() {
       return {
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -78,11 +78,12 @@
 
       <modal :name="duplicateTableModal" class="beekeeper-modal vue-dialog sure header-sure" @opened="sureOpened" @closed="sureClosed" @before-open="beforeOpened">
         <div class="dialog-content">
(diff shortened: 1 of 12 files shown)


apps/studio/src/components/CoreTabs.vue
```diff
@@ -238,6 +263,48 @@
...
+
+        this.$modal.hide(this.duplicateTableModal)
+
+        this.$nextTick(async() => {
+          try {
+            if (this.dbAction.toLowerCase() !== 'duplicate') {
+              return
+            }
+
+            await this.connection.duplicateTable(tableName, this.duplicatedTableName, schema)
```
rathboma, line 292: Do you need to clear duplicatesTableName and duplicatedTableParams here in a finally block?

apps/studio/src/lib/db/client.ts
```diff
@@ -82,6 +82,9 @@ export interface DatabaseClient {
   // delete stuff
   dropElement: (elementName: string, typeOfElement: DatabaseElement, schema?: string) => Promise<void>
   truncateElement: (elementName: string, typeOfElement: DatabaseElement, schema?: string) => Promise<void>
+
+  // duplicate table
```
rathboma, line 86: Could you possibly also add a ‘duplicateTableSql’ method here (similar to alterTableSql vs alterTable) so users can edit the sql themselves if they like?

apps/studio/src/lib/db/clients/mysql.js
```diff
@@ -755,6 +759,14 @@ export async function truncateElement (conn, elementName, typeOfElement) {
   });
 }
 
+export async function duplicateTable(conn, tableName, newTable) {
+  let sql = `CREATE TABLE ${wrapIdentifier(newTable)} LIKE ${wrapIdentifier(tableName)};`;
```
rathboma, line 763: I didn’t even know about this syntax! Thanks for this

rathboma (changes requested):
Fantastic PR thanks so much!

I’ve added a couple of requests inline with the code.

one last ask:
would you be able to add an integration test in the db.ts and all.is file? That way we can make sure it works for all databases we add going forward :-)

apps/studio/src/components/CoreTabs.vue
```diff
@@ -73,6 +73,24 @@
...
+          <div class="form-group">
+            <label for="duplicatedTableName">New table name</label>
+            <input type="text" name="duplicatedTableName" class="form-control" required v-model="duplicatedTableName" autofocus>
+          </div>
+          <small>* Indexes, relations or triggers will not be duplicated</small>
+        </div>
+        <div class="vue-dialog-buttons">
+          <span class="expand"></span>
+          <button ref="no" @click.prevent="$modal.hide(duplicateTableModal)" class="btn btn-sm btn-flat">Cancel</button>
+          <button @focusout="sureOpen && $refs.no && $refs.no.focus()" @click.prevent="completeDuplicateTableAction" class="btn btn-sm btn-primary">{{this.titleCaseAction}} {{this.dbElement}}</button>
```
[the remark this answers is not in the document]
lanzoninicola: Thank you for all your comments, I will work on it in the next few days....

lanzoninicola:
> 

I'm glad you like it. If all goes well, I should be able to let you know next week when the updated version is ready for your review.

lanzoninicola:
Hello @rathboma,
I have finished all the tasks that were assigned to me and have also included the tests as per your suggestion. If you are ready to move forward with the review, I would greatly appreciate it.
Thank you.

rathboma (approved):
Amazing! Sorry it took so long, it's been a rough Q1 :-)

Issue #1498, opened 2023-02-06
FEAT: Pinned Tables Sorting


mattnSwimd:
The tool is great, but one problem I have is using pinned tables. When I open a database with pinned tables on a database of (Example) 600 tables, I need to pin them to find them or I have to scroll for 30 seconds. I use the pin feature to see the tables I am actually in on a daily basis to avoid all of the needless scrolling, but the problem is that I still end up with (Example) 40 tables pinned. That's not the problem. The problem is that the tables are pinned in the order I added them. It can be very difficult to find them when I am jumping between (Example) 10 databases to remember where they are. It normally ends up being quicker to scroll through the entire list alphabetically sorted.

My proposed feature would be just adding a sort button to the pinned list. It does not have to sort alphabetically all the time, Even if it only sorts when the button is clicked it would be a massive help.

I have considered sorting them all manually but it ends up growing into a very time consuming and tedious task very very quickly. As once you have even 2 database connections and 10 tables in each, the task becomes a bit of a bear. Let alone adding to the list.

The databases I am currently using to see this:
    MariaDB
    MySQL

I am unsure if this is the same in the others as I only use these two but It would be a massive improvement for me.

rathboma:
Sounds like a good intro ticket, thanks! Can't say this is a priority, but it would be a good ticket for a new contributor to pick up.

tomaskudlicka:
I added a button but I would add an asc,desc option and then an option to sort by assignment date or entity name. 
<img width="268" alt="Screenshot 2023-03-16 at 20 32 31" src="https://user-images.githubusercontent.com/103044576/225733028-c72a2a8b-7515-4b6f-915d-323929c799b5.png">
So how should the shifting behave correctly?

mattnSwimd:
For my purposes, all I need is an alphabetical sort ascending or descending, doesn't matter, whichever you get working first. If you want to get fancy, toggle it and every time you click it, it switches between ascending and descending. For the dragging I'd say let people drag it and come out of the ordering, and if they click the sort button again it will sort again.

tomaskudlicka:
I made a working version, if it can be like this. https://github.com/beekeeper-studio/beekeeper-studio/pull/1551

mattnSwimd:
That looks great to me! Thank you.

FaroJoaoFaro:
I believe this issue was solved with pull request #1551.
Should be closed

mariusa:
@not-night-but FYI to close

PR #1551, opened 2023-03-17
Pinned Tables Sorting #1498

tomaskudlicka:
Add sorting by position or entity name.
Screenshot 2023-03-17 at 11 53 52
Screenshot 2023-03-17 at 11 54 04


rathboma (approved):
Great! Thanks so much!

PR #1558, opened 2023-03-30
Partitions fix 96

not-night-but:
Removed the ability to edit partitions on v < 10
Peek 2023-03-30 01-39
(Apologies for my background bleeding through on click, not sure why that's happening, screencaps hate me)

I'm not really sure if having it as a supported feature is really what we want to do? I think it's between that and having it as a disabled feature on the dialect?

Also, should I just hide the partition expression column for now?

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -159,7 +159,9 @@ export default async function (server: any, database: any): Promise<DatabaseClie
 
   const version = await getVersion(conn)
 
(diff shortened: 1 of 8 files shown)


rathboma:
@not-night-but I think it's fine. Lets ship it

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -288,10 +290,25 @@ export async function listTablePartitions(conn: HasPool, tableName: string) {
     WHERE base_tb.oid = ?::regclass
   `, [tableName]).toQuery();
```
rathboma, line 291: @not-night-but this is where we need to also filter by schema. This only filters by tableName

rathboma (changes requested):
Still need to filter by schema name!

not-night-but:
@rathboma Changes made!

PR #1565, opened 2023-04-07
#1564 - Fix table list dups

### LINKED ISSUE #1564: BUG: Tables appear twice in UI
**Describe the bug**

For tables that have the same name _in different schemas_, the table name will appear twice in beekeeper's UI. The duplicates appear for all schemas that share table names. Using a Postgres 15 backend.

**To Reproduce**

1. Create two tables with the same name, in different schemas. A simple example might look like this:

```sql
CREATE SCHEMA schema1;

CREATE TABLE schema1.my_table (
  "id" INTEGER PRIMARY KEY
);

CREATE SCHEMA schema2;

CREATE TABLE schema2.my_table (
  "id" INTEGER PRIMARY KEY
);
```

2. Refresh and check the Beekeeper UI: the tables show up two times.

**Expected behavior**

Tables should show only once, per schema.

**Screenshots / Gifs**

image

 - OS: Ubuntu 22.04.2 LTS
 - App Version: 3.9.4
 - Database type and version: Postgres 15.2

**Additional context**

I often update Beekeeper using apt, so I'm not sure when this started happening. It feels like this is a very new issue that appeared in a version within the last 14 days or so.

not-night-but:
Fix #1564 
image

This issue was caused by the new partitions script, which was just going off of table name in one of the joins :facepalm:.

Now joining on tablename and schema name, as well as oid in a different join.

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -249,10 +249,10 @@ export async function listTables(conn: HasPool, filter: FilterOptions = { schema
     sql += `
         pc.relkind as tabletype
       FROM information_schema.tables AS t
+      JOIN pg_class AS pc
+      ON t.table_name = pc.relname AND t.table_schema = pc.relnamespace::regnamespace::text
       LEFT OUTER JOIN pg_inherits AS i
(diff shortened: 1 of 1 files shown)


rathboma:
@not-night-but test?

PR #1584, opened 2023-04-25
Fix save and parameter modals in query editor.

not-night-but:
Fix some issues with modals when multiple query editors are open


rathboma (approved):
Feel free to merge. One question first, does modal.show accept a $ref? If so that would be much safer in general

not-night-but:
@rathboma good point! I'll take a looksie

not-night-but:
@rathboma unfortunately it has to be a string (the name of the modal), or a component with component props etc, which is a much bigger undertaking. Merging as is.

PR #1585, opened 2023-04-26
#1583 - Fix partitions so they don't hide inherited tables

### LINKED ISSUE #1583: BUG: Only parent table shows in UI when using INHERITS
**Describe the bug**
When a child table `INHERITS` from another parent table, only the parent table shows up in Beekeeper's UI. The child table is totally hidden, and unfamiliar users connecting to the DB might never be aware that it exists.

___

**To Reproduce**

Step 1: In a fresh database, run for instance the following SQL:

```sql
/* Create schema just for kicks */
CREATE SCHEMA my_schema;

/* Create parent table */
CREATE TABLE my_schema.parent (
  id INTEGER PRIMARY KEY
);

/* Create child table, this does *NOT* show in Beekeeper UI. */
CREATE TABLE my_schema.child (
  name VARCHAR(100)
) INHERITS (my_schema.parent);

/* This is not needed to show the bug, but perhaps makes it more clear...*/
INSERT INTO my_schema.child(id, name) VALUES (1, 'Alice');
```

Step 2: check the Beekeeper UI - the child does not show.

___

**Expected behavior**

At least by default, child tables should show up in the UI instead of being invisible. Preferrably they should also somehow indicate which table is their parent.

___

**Screenshots / Gifs**

**Image 1 below:** Screenshot after running the above example sql; the child table is not shown, while the parent is.

image

**Image 2 below:** The inserted element (Alice) correctly shows up for the parent, when clicked. However, the child table can't be selected in the menu to the left since it is not shown at all.

image

___

 - OS: Ubuntu 22.04 LTS
 - App Version: Beekeeper 3.9.9
 - Database type and version: Postgres 15.2

not-night-but:
Fix #1583 

Partitions filtering was also filtering out inherited tables that aren't partitions.

This PR changes how we filter partitions, as we were filtering them in the query itself before. Now we are just grabbing the parent table's relkind and using that to determine if it's a partition or not client side.

TODO:
- [x] Integration test
- [x] Disable pg < 10

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -291,7 +291,7 @@ export async function listTablePartitions(conn: HasPool, tableName: string, sche
       JOIN pg_class pt                ON pt.oid = i.inhrelid
       JOIN pg_stat_all_tables ps      ON ps.relid = i.inhrelid
       JOIN pg_namespace nmsp_parent   ON nmsp_parent.oid = base_tb.relnamespace 
-    WHERE nmsp_parent.nspname = ? AND base_tb.relname = ?;
+    WHERE nmsp_parent.nspname = ? AND base_tb.relname = ? AND base_tb.relkind = 'p';
   `, [schemaName, tableName]).toQuery();
 
   const data = await driverExecuteSingle(conn, { query: sql });
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -105,7 +105,7 @@ async function getVersion(conn: HasPool): Promise<VersionInfo> {
     isCockroach,
     isRedshift,
     number,
-    hasPartitions: (isPostgres && number >= 90000) //for future cochroach support?: || (isCockroach && number >= 200070)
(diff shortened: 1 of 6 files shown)


apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -250,18 +250,21 @@ export async function listTables(conn: HasPool, filter: FilterOptions = { schema
...
-        pc.relkind as tabletype
+        pc.relkind as tabletype,
+        parent_pc.relkind as parenttype
       FROM information_schema.tables AS t
       JOIN pg_class AS pc
         ON t.table_name = pc.relname AND ${schemaStatement}
       LEFT OUTER JOIN pg_inherits AS i
         ON pc.oid = i.inhrelid
+      LEFT OUTER JOIN pg_class AS parent_pc
+        ON parent_pc.oid = i.inhparent
```
rathboma, line 261: integration test for this change?

rathboma (changes requested):
looks good, but lets add an integration test to make sure we don't have a regression on this in the future.

Also test that `getPartitions` doesn't return inherited tables that are not partitions

rathboma (approved):

PR #1598, opened 2023-05-16
1545 editor context menu

### LINKED ISSUE #1545: FEAT: right-clicking in the editor should show common text-editing options
When right-clicking in the SQL editor Beekeeper should show the common commands a user expects. Some of these actions may have to call the codemirror APIs to be implemented. That is something the developer of this feature will need to figure out

- Undo
- Redo
- Cut
- Copy
- Paste
- Delete
- Select All
- Horzontal line
- Format SQL

Things to keep in mind:
- Grey-out actions if they cannot be performed. For example - if no text is selected, you cannot cut, copy, or delete, so these items should be disabled with an appropriate style.

- If easy, also display the appropriate keyboard shortcut next t

not-night-but:
Close #1545 

This PR adds all of the actions to the right click menu specified in #1545.

image

It also adds the ability to add a disabled class to elements in the context menu, as well as add keyboard shortcuts for those actions.

### DIFF
--- a/apps/studio/src/assets/styles/components/context-menu.scss
+++ b/apps/studio/src/assets/styles/components/context-menu.scss
@@ -47,6 +47,9 @@
     opacity: 0.6;
     margin-right: $gutter-h!important;
   }
+  .shortcut {
+    color: darken($text-dark, 30%);
+  }
 }
 .vue-simple-context-menu__divider, hr {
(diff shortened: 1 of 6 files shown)


not-night-but:
Wow my lang server really did not like some of these files

PR #1599, opened 2023-05-16
Create Routine sql syntax errror fix

### LINKED ISSUE #1522: BUG: opening the SQL:Create view of a stored procedure or function has a syntax error
**Describe the bug**
The code block is escaped by a named dollar quote (`$function$` code block `$function$`). Regardless of the name given to the code block in the function definition, the dollar quotes are automatically renamed to `$function$`, however a space is introduced between the word function and the second dollar sign in the opening tag: `$function $`, making it a non-valid opening quote. This syntax error leads to the whole code block not being escaped at all and having to remove the space before being able to run the generated sql query.

**To Reproduce**
1. Run this query:

```sql
drop function if exists public.just_fail();

create function public.just_fail() returns void 
as $foo$
  begin
    raise exception 'I refuse!'
    using detail = 'Pretty simple',
    hint = $hint$ There's nothing you can do. $hint$ ;
  end 
$foo$ language plpgsql;
```

2. Right-click on the newly created function in the left side-bar and choose `SQL: Create`
3. See the wrongly inserted space in the opening dollar quotes `$function $` (instead of $foo$) and `$hint $` instead of `$hint$`
5. Also note the incorrect syntax highlighting by the misinterpretation of the single quote within the dollar-quoted string. One important reason (probably the most important reason) for dollar-quoting to exist at all is to allow for the usage of single quotes within the quoted text.
6. Last and least important the semi-colon at the end of the function definition is missing (after the last `$function$`. 

grafik

**Expected behavior**
I would expect the dollar-quotes to not be renamed and to have no spaces introduced in their name.

 - OS: Windows Portable
 - App Version 3.8.9
 - Database type and version Posgresql 14.7

**Additional context**
https://www.postgresqltutorial.com/postgresql-plpgsql/dollar-quoted-string-constants/

not-night-but:
Fix #1522 

This PR updates sql-formatter, and removes the call to formatting for create routine.

The reason this was done is sql-formatter does not actually support formatting routines, so it was screwing up the scripts. This means we will just display the routine as it was formatted by the user.

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -66,7 +66,7 @@
     "sass-loader": "^10.1.1",
     "simple-encryptor": "^3.0.0",
     "split.js": "^1.5.11",
-    "sql-formatter": "^6.1.1",
+    "sql-formatter": "^10.0.0",
     "sql-query-identifier": "^2.2.0",
     "ssh2": "^1.11.0",
     "tabulator-tables": "~5.4.1",
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -445,8 +445,8 @@ export default Vue.extend({
     },
     async loadRoutineCreate(routine) {
       const result = await this.connection.getRoutineCreateScript(routine.name, routine.type, routine.schema)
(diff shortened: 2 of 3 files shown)


rathboma:
SQL formatter tested as working elsewhere? If so, merge away

rathboma (approved):

PR #1600, opened 2023-05-16
Warn user when filtering that it will discard changes.

### LINKED ISSUE #1019: FEAT: Warn users that sorting or filtering will discard pending changes
**Describe the bug**
When adding data to a database and then click any column all pending changes are lost

**To Reproduce**
1. Create new Table
2. Select Table
3. Click 'Add row' a couple of times
4. Add data to rows
5. Click on any column

**Expected behavior**
Either only existing data gets sorted by column and changes stay on top (maybe in themselves sorted by column).
Or pending changes are mixed with existing data sorted by column
Or a warning appears that this will discard your pending changes with an option to cancel

not-night-but:
Close #1019 

This PR adds a modal to TableTable that warns the user if they attempt to filter while they still have pending changes:

image

From this point you can either cancel the filter, or you can discard the changes and proceed with the filter.

Still to do:
- [ ] Check if header sorting needs to be disabled/prompt before applying the sort somehow.
- [x] Check to see if filtering columns also causes this issue


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -201,6 +201,21 @@
       </div>
 
     </statusbar>
+
+    <portal to="modals">
+      <modal class="vue-dialog beekeeper-modal" :name="`discard-changes-modal-${tab.id}`">
```
rathboma, line 206: Unrelated: how do we generalize this stupid behavior of the modal library? It's super easy to make a mistake. Does it accept a `ref` instead of a `name`?

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -201,6 +201,21 @@
...
+        <div class="dialog-content">
+          <div class="dialog-c-title">Confirmation</div>
+          <div class="modal-form">
+            Sorting or Filtering now will discard {{pendingChangesCount}} pending change(s). Are you sure?
+          </div>
+        </div>
+        <div class="vue-dialog-buttons">
+          <button class="btn btn-flat" type="button" @click.prevent="$modal.hide(`discard-changes-modal-${tab.id}`)">Cancel</button>
```
rathboma, line 214: Also unrelated: Using the name is so ugly :vomiting_face:

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -201,6 +201,21 @@
...
+          <button class="btn btn-primary" type="button" @click.prevent="forceFilter">I'm Sure</button>
```
rathboma, line 215: can this have focus on load, so users just need to hit `enter` to confirm? (same as vscode)

rathboma (approved):
:shipit:, one minor question, but good to merge either way.

What is our policy on focus for popup modals like this? My thought is we should focus the primary action by default.

apps/studio/src/components/tableview/TableTable.vue
not-night-but: Agreed, I hate it 😂.

apps/studio/src/components/tableview/TableTable.vue
not-night-but: I believe we've had this conversation before? I'll look into it, but I don't think it supports refs

not-night-but: Maybe I'll look into filing an issue/PR on their repo cause we've run into this issue like 4 or 5 times now with the save modal, params modal, filter modal, and now this one lol

not-night-but:
Before merge: focus primary modal action

PR #1601, opened 2023-05-19
#423 - Remove quotes on paste into query editor.

### LINKED ISSUE #423: FEAT: Auto-remove quotes around pasted query strings
When I work with the debugger in VS Code I often need to test the queries that are generated by the code. Since the quires are stored in variable, I can copy the value of this variable to copy the query. However, they are stored as a string, and is thus copied with surrounding quotation marks. When pasting them into Beekeeper, I have to then remove the quotation marks, before I can run the query, which is inconvenient.

In dbeaver, which I come from, the quotation marks are automatically removed. Would be nice if Beekeeper could do the same. 

Thanks for a great project :)

not-night-but:
Close #423 

This PR adds a listener to code mirror that intercepts paste events, checks if the text is quoted and if it is a query, and if both of those things are true it removes the quotes and then pastes it into the editor.


not-night-but:
@rathboma Just some notes for you:

- I didn't add unit tests, as it was just a couple lines in the UI, and I'm not sure where you would want the code to go. Should be an easy move if we want to move it elsewhere, it's just a function that'll have to be called by the listener.
- This also adds the ability to format on paste, do we want to add that?

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -544,6 +544,22 @@
             }
           })
 
+          this.editor.on("paste", (cm, e) => {
+            e.preventDefault(); 
+            const quotes = ["'", '"', '`'];
+            let clipboard = this.$native.clipboard.readText();
+            const isQuoted = quotes.includes(clipboard[0]) && quotes.includes(clipboard[clipboard.length - 1]) && clipboard[0] === clipboard[clipboard.length - 1];
```
rathboma, line 551: So I think you should get the pasted text from CM, not the native clipboard, otherwise they could be different.
Questions:

- If someone is legit copying a quoted string from one editor to another, what will happen?
- As a new dev I have noooo idea what this is doing, can you add some comments to walk through why this is here and how it works.
- Can you abstract this to another file and add a unit test?

rathboma (changes requested):
- testing
- docs
- edge cases

apps/studio/src/components/TabQueryEditor.vue
not-night-but: Okay so after coming back from this, I'm not sure we should get the pasted text from CM, or even if we can. This event is fired before the paste is handled by code mirror, so the content isn't actually there for us to manipulate. 

Also, since this was implemented, we've added vim mode which is gonna need its own solution, as this works with copy/paste but not with yank/paste, as yank is handled by the internal vim api, and yanked text is not yanked to the system clipboard (even when it should be ie. when using `"+y`	 ). We can listen for vim commands and do much the same thing on paste, even possibly fix the lack of global yank in the process. 

Thoughts?

not-night-but: I forgot to add answers to the questions lol:
1. copying from one codemirror to another works fine (there was a whitespace issue, but I've fixed that), as codemirror just copies/pastes to/from the system clipboard, unless in vim mode and you're using yank/put, which doesn't work yet
2. I'll document it for sure!
3. Same with tests :smile:

PR #1610, opened 2023-06-07
feat: support schema autocomplete

### LINKED ISSUE #216: Table and schema names do not autocomplete properly

### LINKED ISSUE #976: BUG: Code Completion doesn't autocomplete schemaname or schemaname.tablename
I am connected to my database which has roughly 1200 objects and about 20 schemas.  The objects load and then are displayed on the left navigation window listing tables, functions etc.

I type out the following in the sql editor:

select * from schema.tablename;

While I am typing, my expectation is that as soon as I start typing "schema", I would be provided recommendations of schema.  It does provide recommendations for table, but I do not like setting my search_path so I like code sql with the full declaration of schema.tablename.  And after I type "schema." there are no tablename recommendations.

I like your app, but this is my main issue with postgres sql tools.  Looks good and I will continue to follow your app.

Also, when I right click on a function in the navigation window and select SQL:Create, nothing happens where I would expect it to open the DDL in a tab.

SYSTEM : Ubuntu 20.04
BEEKEEPER : 3.0.7

### LINKED ISSUE #1184: FEAT: Postgres autocomplete all schemas
Only the public schema tables get autocomplete right now, and there's no autocomplete if you type `schema.`, but it should autocomplete tables here too.

Ripped from #173

azmy60:
fix #976 | fix #216 | fix #1184 

This is an attempt to support schema autocomplete with table alias and columns. I replaced some logics in the vendor file as it can't be modified anywhere outside nor tested. It's far easier to solve this when I can unit test the codes :smile:. Let me know if there are things that need to be worked around. I've also tested this in postgres and sqlite (making sure it doesn't break in database without schemas).

bks-support-schema

### DIFF
--- a/apps/studio/src/lib/editor.ts
+++ b/apps/studio/src/lib/editor.ts
@@ -89,12 +89,7 @@ export function makeDBHint(
   };
 }
 
-const SCHEMA_TABLE = new RegExp(
-  "(?:(?<schema>[quote]?.+?[quote]?)\\.(?<table1>[quote]?.+[quote]?)|(?<table2>[quote]?.+[quote]?))".replaceAll(
-    "quote",
-    "'\"`"
-  )
-);
+const SCHEMA_TABLE = /(?:(?<schema>(['"`]?).+?\2?)\.(?<table1>(['"`]?).+\4?)|(?<table2>(['"`]?).+\6?))/
 
 export function queryTable(dbHint: DBHint, query: string) {
   const schemaTable = SCHEMA_TABLE.exec(query);
@@ -162,5 +157,5 @@ export function isQuote(str: string) {
 const WORD_SPLITTER = /`[^`]*`[^\s]*|'[^']*'[^\s]*|"[^"]*"[^\s]*|[^\s]+/g;
 
 export function splitWords(text: string) {
-  return text.match(WORD_SPLITTER);
+  return text.match(WORD_SPLITTER) || [];
 }
--- a/apps/studio/src/lib/editor.ts
+++ b/apps/studio/src/lib/editor.ts
@@ -2,14 +2,26 @@ import type { TableOrView } from "@/lib/db/models";
 import { DialectData } from "@shared/lib/dialects/models";
 
 interface Word {
+  /**
+   * The completion text for codemirror. This is wrapped with quotations if
+   * needed, e.g. '"special+table"'.
+   **/
   text: string;
+  /**
+   * The name of the corresponding type. This is not wrapped with quotations.
+   **/
   name: string;
   type: "schema" | "table" | "column";
+  /**
+   * The name of the corresponding schema of the table without quotations (if
+   * the database supports schemas).
+   **/
   schema?: string;
 }
 
 /**
- * A list of all words in key-value form, where the key is the word that's case sensitive.
+ * A list of all words in key-value form, where the key is the word that's
+ * case sensitive.
  */
 type WordList = Record<string, Word>;
 
@@ -61,7 +73,7 @@ export function makeDBHint(
       ...acc,
       [name]: {
         name,
-        text: name,
+        text: dialect.friendlyNormalizedIdentifier(name),
         type: "schema",
       },
     }),
@@ -77,12 +89,10 @@ export function makeDBHint(
   };
 }
 
-const quot = "'\"`";
-
 const SCHEMA_TABLE = new RegExp(
-  "(?:(?<schema>[quot]?.+?[quot]?)\\.(?<table1>[quot]?.+[quot]?)|(?<table2>[quot]?.+[quot]?))".replaceAll(
-    "quot",
-    quot
+  "(?:(?<schema>[quote]?.+?[quote]?)\\.(?<table1>[quote]?.+[quote]?)|(?<table2>[quote]?.+[quote]?))".replaceAll(
+    "quote",
+    "'\"`"
   )
 );
 
@@ -96,7 +106,9 @@ export function queryTable(dbHint: DBHint, query: string) {
   if (!table) return undefined;
 
   const unwrappedTable = dbHint.dialect.unwrapIdentifier(table);
-  const unwrappedSchema = schema ? dbHint.dialect.unwrapIdentifier(schema) : undefined;
+  const unwrappedSchema = schema
+    ? dbHint.dialect.unwrapIdentifier(schema)
+    : undefined;
 
   if (
     dbHint.defaultTableWordList[unwrappedTable] &&
@@ -106,7 +118,7 @@ export function queryTable(dbHint: DBHint, query: string) {
   }
 
   return dbHint.tableWords.find(
-    (table) => table.schema === schema && table.name === unwrappedTable
+    (table) => table.schema === unwrappedSchema && table.name === unwrappedTable
   );
 }
 
@@ -121,6 +133,23 @@ export function findTableOrViewByWord(tableOrViews: TableOrView[], word: Word) {
   );
 }
 
+export interface Editor {
+  display: { view: { [line: number]: { line: { text: string } } } };
+}
+
+export interface Cursor {
+  line: number;
+  ch: number;
+}
+
+export function getTextNearCursor(editor: Editor, cursor: Cursor) {
+  const fullText = editor.display.view[cursor.line].line.text;
+  const textBeforeCursor = fullText.substring(0, cursor.ch);
+  const text = splitWords(textBeforeCursor).pop();
+  const startAt = textBeforeCursor.length - text.length;
+  return { text, startAt };
+}
+
 export function splitSchemaTable(str: string) {
   const [schema, ...table] = str.split(".");
   return [schema, table.join(".")];
@@ -128,4 +157,10 @@ export function splitSchemaTable(str: string) {
 
 export function isQuote(str: string) {
   return str === "'" || str === '"' || str === "`";
+}
+
+const WORD_SPLITTER = /`[^`]*`[^\s]*|'[^']*'[^\s]*|"[^"]*"[^\s]*|[^\s]+/g;
+
+export function splitWords(text: string) {
+  return text.match(WORD_SPLITTER);
 }
--- a/apps/studio/src/lib/editor.ts
+++ b/apps/studio/src/lib/editor.ts
@@ -1,4 +1,5 @@
 import type { TableOrView } from "@/lib/db/models";
+import { DialectData } from "@shared/lib/dialects/models";
 
 interface DBHintTable {
   name: string;
@@ -22,16 +23,13 @@ interface DBHint {
 
 export function makeDBHint(
   tables: TableOrView[],
-  connectionType: string
+  dialectData: DialectData,
 ): DBHint {
   const schemaSet: Set<string> = new Set();
   const hintTables: DBHintTable[] = [];
   tables.forEach((table) => {
     hintTables.push({
-      name:
-        connectionType === "postgresql"
-          ? sanitizeTableName(table.name)
-          : table.name,
+      name: dialectData.maybeWrapIdentifier(table.name),
       schema: table.schema,
     });
     if (table.schema) schemaSet.add(table.schema);
@@ -95,10 +93,4 @@ export function pushTablesToResult(
       };
     })
   );
-}
-
-const tableNameValidationRegex = /(?:[^a-z0-9_]|^\d)/;
-
-export function sanitizeTableName(name: string): string {
-  return tableNameValidationRegex.test(name) ? `"${name}"` : name;
 }
--- a/apps/studio/src/lib/editor.ts
+++ b/apps/studio/src/lib/editor.ts
@@ -14,38 +14,36 @@ interface Word {
 type WordList = Record<string, Word>;
 
 export interface DBHint {
+  defaultSchema?: string;
   /**
-   * A list of all tables in key-value form, where the key is the table name.
+   * A list of tables in key-value form for fast queries, where the key
+   * is the table name. If the database supports schemas, all tables are from
+   * the default schema.
    **/
-  tableWordList: WordList;
-  /**
-   * A list of all tables in array form.
-   */
+  defaultTableWordList: WordList;
+  /** list of all tables in array form. */
   tableWords: Word[];
   schemaWordList: WordList;
-}
-
-export function findWord(wordList: WordList, word: string) {
-  return wordList[word];
+  dialect: DialectData;
 }
 
 export function makeDBHint(
   tables: TableOrView[],
-  dialectData: DialectData,
+  dialect: DialectData,
   defaultSchema?: string | null
 ): DBHint {
   const schemaSet = new Set<string>();
   if (defaultSchema) schemaSet.add(defaultSchema);
-  if(defaultSchema === null) defaultSchema = undefined;
+  if (defaultSchema === null) defaultSchema = undefined;
 
   const tableWords: Word[] = [];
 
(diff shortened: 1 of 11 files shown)


rathboma:
Wow, this is awesome! Going to take a look today. Thanks for the killer PR :-)

apps/studio/src/lib/editor.ts
```diff
@@ -0,0 +1,104 @@
...
+        text: table.schema ? `${table.schema}.${tableName}` : tableName,
+        displayText: table.name,
+      };
+    })
+  );
+}
+
+const tableNameValidationRegex = /(?:[^a-z0-9_]|^\d)/;
+
+export function sanitizeTableName(name: string): string {
```
rathboma, line 102: Can this be something added to the `DialectData` for all connection types? Then we can just pass in the dialect data for each connection.

We already have a `wrapIdentifier`, you could add this as `maybeWrapIdentifier` or `friendlyNormalizedIdentifier`? [See this file as an example](https://github.com/beekeeper-studio/beekeeper-studio/blob/e405db30297950a4295c9d1a15e78aa2713e9a45/shared/src/lib/dialects/postgresql.ts#L4)

rathboma: The dialect data is available in the store as a getter, [see the getter here](https://github.com/beekeeper-studio/beekeeper-studio/blob/e405db30297950a4295c9d1a15e78aa2713e9a45/apps/studio/src/store/index.ts#L128).

So you could do something like:

```js
computed: {
 ...mapGetters('dialectData')
}

```

Then: `makeDbHint(tables, this.dialectData)`

apps/studio/tests/unit/lib/editor.spec.js
```diff
@@ -0,0 +1,130 @@
...
+
+  beforeAll(() => {
+    dbHint = makeDBHint(tableOrViews, "postgresql");
+  });
+
+  describe("makeDBHint", () => {
+    it("should make arrays of tables and schemas", () => {
+      expect(makeDBHint(tableOrViews, "postgresql")).toEqual({
+        tables: [
+          { name: "my_table", schema: "public" },
```
rathboma, line 23: In postgres if you create a normal table using quotes, like `'"normal_table"'`, it always has to be referenced with the quotes I think? Does it work in that case?

(didn't test this, just pulling it from my memory)

apps/studio/tests/unit/lib/editor.spec.js
```diff
@@ -0,0 +1,130 @@
...
+  parseDBHintTable,
+  parseDBHintTables,
+  queryTable,
+  findTablesBySchema,
+  pushTablesToResult,
+  sanitizeTableName,
+} from "../../../src/lib/editor";
+import { tableOrViews } from "../../fixtures/tables";
+
+describe("lib/editor", () => {
```
rathboma, line 12: Please correct me if I'm wrong here, but it looks like this won't work if the DB has schemas, and no schema is provided in the query?

Postgres by default uses the `public` schema, and SQL Server uses `dbo`, and not specifying this is totally legit, and really common honestly.

A good unit test would be to check that it can resolve:

- `schema.table`
- `schema.table.column`
- `table`
- `table.column`

The complex bit is that `a.b` could be either `schema.table` or `table.column`.

It gets a little more complex in that there could be both `schema: a`, and `table: public.a`, so the hint system would need to provide some ability to handle this situation. So if someone types `a.`, it should probably autocomplete both table names from `schema: a` and column names from `table: public.a`.

Default schema is [also provided in the store as a getter](https://github.com/beekeeper-studio/beekeeper-studio/blob/e405db30297950a4295c9d1a15e78aa2713e9a45/apps/studio/src/store/index.ts#L104), so should be easy to integrate if not already?

rathboma (changes requested):
This is 🚀 🚀 🔥 🔥 . Really amazing work to figure this out.

I have two real pieces of feedback:

1. Can we generalize the table name sanitization?
2. Can we handle default schemas and ambiguous table/schema matches?

apps/studio/tests/unit/lib/editor.spec.js
azmy60: Oh, you've got a point. I forgot to add the default schema.

I also think this part is a little bit confusing because there are [two ways](https://github.com/beekeeper-studio/beekeeper-studio/pull/1610/files#diff-8ada96660a46ee9c6784d974ac1043ced5db7c7f1be4eab25c9e25ecb43fbd3bR66-R70) to find tables right now: one by looking up the key in `tables` object, the other one is by using the `hintTables` array. I add the second one because the `tables` can overwrite the same table names that have different schemas, while the `hintTables` can not.

I'll try to add the test to clarify this one and fix the other feedbacks soon.

apps/studio/tests/unit/lib/editor.spec.js
azmy60: Yup. Table names that are case sensitive will need to be referenced with quotes. I'll add this to the test.

azmy60:
@rathboma I've added the default schema and `maybeWrapIdentifier` to the `dialectData`. The trickiest part here is to resolve `schema.table.column` where the table uses quotes like `public."table"`. Is this good enough to cover the edge cases? I'm not entirely aware of all databases syntax.

apps/studio/tests/unit/lib/editor.spec.js
azmy60: Your test suggestions should now be covered by the [queryTable test](https://github.com/beekeeper-studio/beekeeper-studio/pull/1610/files#diff-4e044bf2110531d177d19691735abb6cea1e27c92e792eed2734814d8b274ab9R144). As long as it resolves the `schema.table`, it should be able to resolve `schema.table.column` too which depends on the column lookup [here](https://github.com/beekeeper-studio/beekeeper-studio/pull/1610/files#diff-8bd68c134db51e45491610ee0f614c52db0be8412bf23eded25bf6aeffa086a9R918). It should also work when schema is not provided or `undefined` [(test)](https://github.com/beekeeper-studio/beekeeper-studio/pull/1610/files#diff-4e044bf2110531d177d19691735abb6cea1e27c92e792eed2734814d8b274ab9R126-R142).

azmy60:
I just notice that if you write <code>schema.\`table\`</code> with quotation instead of `schema.table`, it won't show the autocomplete. Would you like this to be fixed as well?

not-night-but:
@azmy60 If we support other quotations (`'` and `"`) then I think it's best we support backtick as well

azmy60:
Ok. I'll start working on that tomorrow.

not-night-but:
@azmy60 I'm just messing around with this a bit, and I have a thought:

I think the default behaviour (without specifying a schema) should remain the same. As it stands right now, if I don't specify a schema in postgres I get all the tables from information_schema and pg_catalog before I get any tables from public or any schemas.

Old behaviour: 
oldAutoComplete

New behaviour:
newAutoComplete

Note that in the new behaviour, choosing one of the autocomplete options from other schemas actually creates an incorrect autocompletion, as you'll get a `relation doesn't exist` error because the correct schema isn't specified. It should probably show 
1. Tables from the default schema
2. schemas
3. keywords

not-night-but (changes requested):
@azmy60 I'm just messing around with this a bit, and I have a thought:

I think the default behaviour (without specifying a schema) should remain the same. As it stands right now, if I don't specify a schema in postgres I get all the tables from information_schema and pg_catalog before I get any tables from public or any schemas.

Old behaviour: 
oldAutoComplete

New behaviour:
newAutoComplete

Note that in the new behaviour, choosing one of the autocomplete options from other schemas actually creates an incorrect autocompletion, as you'll get a `relation doesn't exist` error because the correct schema isn't specified. It should probably show 
1. Tables from the default schema
2. schemas
3. keywords

apps/studio/src/lib/editor.ts
```diff
@@ -0,0 +1,131 @@
...
+ */
+type WordList = Record<string, Word>;
+
+export interface DBHint {
+  defaultSchema?: string;
+  /**
+   * A list of tables in key-value form for fast queries, where the key
+   * is the table name. If the database supports schemas, all tables are from
+   * the default schema.
+   **/
```
not-night-but, line 22: Love that you're documenting your code, thanks so much!

apps/studio/src/lib/editor.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+const quot = "'\"`";
+
+const SCHEMA_TABLE = new RegExp(
+  "(?:(?<schema>[quot]?.+?[quot]?)\\.(?<table1>[quot]?.+[quot]?)|(?<table2>[quot]?.+[quot]?))".replaceAll(
+    "quot",
+    quot
+  )
+);
```
not-night-but, line 87: Unless you have a reason I'm not seeing here, could you replace `quot` with `quote`?

apps/studio/src/vendor/sql-hint/index.js
```diff
@@ -231,22 +257,42 @@
         var lineText = query[i];
         eachWord(lineText, function(word) {
           var wordUpperCase = word.toUpperCase();
-          if (wordUpperCase === aliasUpperCase && getTable(previousWord))
+          if (wordUpperCase === aliasUpperCase && (tableWord = getTable(previousWord)))
```
not-night-but, line 260: If possible I would prefer if we didn't assign within conditionals. I'm also not sure that `table` is entirely necessary anymore, but I haven't dug around in this area of the code.

apps/studio/src/vendor/sql-hint/index.js
```diff
@@ -231,22 +257,42 @@
...
+    const lineText = editor.display.view[cur.line].line.text;
+    const lineTextBeforeDot = lineText.substring(0, cur.ch - 1)
+    const wordStart = lineTextBeforeDot.lastIndexOf(" ") + 1
+    const word = lineTextBeforeDot.split(" ").pop()
+    return [word, wordStart]
+  }
+
+  function addTablesBySchema(result, search) {
+    const [schema] = splitSchemaTable(search);
+      const unwrappedSchema = dbHint.dialect.unwrapIdentifier(schema);
```
not-night-but, line 282: Indentation :stuck_out_tongue_winking_eye:

not-night-but (changes requested):
@azmy60 Great job! Unfortunately I think I found another bug :face_exhaling: 

It seems if I have a quoted schema, it won't autocomplete the tables:
quotedSchemaIssue

Hopefully it's an easy fix, and then we can see what @rathboma thinks :grinning:

apps/studio/src/vendor/sql-hint/index.js
azmy60: I assigned within conditionals because I thought it could minimize the lines of changes. But I'll fix that if it's preferred no to.

apps/studio/src/vendor/sql-hint/index.js
azmy60: lol 😆

apps/studio/src/vendor/sql-hint/index.js
not-night-but: I personally just think it makes the code a bit harder to read.

azmy60:
@not-night-but I've added the fixes to your feedbacks. The problem was if the schema/table contains spaces like `SELECT * FROM "my schema".table`, it would parse it as `["SELECT", "*", "FROM", '"my', 'schema".table']`. It should be fixed based on [this test](https://github.com/beekeeper-studio/beekeeper-studio/pull/1610/files#diff-4e044bf2110531d177d19691735abb6cea1e27c92e792eed2734814d8b274ab9R272-R277) and works for any quotation characters.

However, the autocompletion wouldn't work if the table name contains a quote like `my'table` (it's a valid name in mysql and postgres, though it's probably not a good practice 😅). This could be fixed by updating [this regex](https://github.com/beekeeper-studio/beekeeper-studio/pull/1610/files#diff-ff50d40fa14d8bc2b3c86de9f768caed39751c7b3fed18958816733cd1ed409bR162) or the [splitWords()](https://github.com/beekeeper-studio/beekeeper-studio/pull/1610/files#diff-ff50d40fa14d8bc2b3c86de9f768caed39751c7b3fed18958816733cd1ed409bR164) function. But I haven't found any solution yet.

not-night-but:
@azmy60 I'm so sorry I missed this!!! Thanks so much. I'll take a look in a bit and see if I can figure out a way to support that weird syntax. In my experience there is a lot of stuff that is supported in postgres that really shouldn't be 😂. When I found out you could put spaces in schema/table names I almost had an aneurysm.

apps/studio/src/lib/editor.ts
```diff
@@ -77,12 +89,10 @@ export function makeDBHint(
...
-
 const SCHEMA_TABLE = new RegExp(
-  "(?:(?<schema>[quot]?.+?[quot]?)\\.(?<table1>[quot]?.+[quot]?)|(?<table2>[quot]?.+[quot]?))".replaceAll(
-    "quot",
-    quot
+  "(?:(?<schema>[quote]?.+?[quote]?)\\.(?<table1>[quote]?.+[quote]?)|(?<table2>[quote]?.+[quote]?))".replaceAll(
+    "quote",
+    "'\"`"
   )
 );
```
not-night-but: @azmy60 As I'm looking at this again, I'm wondering what the `table2` capture group is for here. Maybe I'm just bad at regex, but it seems to me like `table1` will always be the group that matches, yes?

not-night-but: Nevermind I'm dumb lol

not-night-but:
@azmy60 as for fixing the regex to allow for `my'table`, I believe this should fix it:

```regex
(?:(?<schema>(['"`]?).+?\2?)\.(?<table1>(['"`]?).+\4?)|(?<table2>(['"`]?).+\6?))
```

azmy60:
@not-night-but Thanks for the help! I just took a quick look, it turned out that the issue was not from the regex🤦. The issue was from the CodeMirror (probably the sql mode module?) that didn't [get the token](https://github.com/beekeeper-studio/beekeeper-studio/blob/94b1cee9c15397c6c0f60ca6d530728221a1bc0c/apps/studio/src/vendor/sql-hint/index.js#L312) correctly.

So If I write `"my'table".`, it gives me `'table".` as the token. The correct token should just be the `.` character. You can see that it also breaks the highlight and the autocomplete. I recorded this using the current version of Beekeeper Studio.

[screen-recorder-thu-jul-13-2023-09-54-46.webm](https://github.com/beekeeper-studio/beekeeper-studio/assets/38707148/8f2b2953-bad9-4824-9d50-f4aa87a85253)

not-night-but:
@azmy60 hmmmm. Yeah there are a couple quirks with our highlighting system.

I would still advise that we switch the regex to something like what I sent above, as it ensures that the closing quote is the same as the opening quote

azmy60:
@not-night-but done :)

not-night-but:
@azmy60 Coolio, thanks! So beyond the weird highlighting issues with that one example, were there any other issues you saw?

azmy60:
I didn't notice any other issues so far.

not-night-but:
@azmy60 alrighty cool. I'll play around with it a bit more in a bit, and if I don't find anything I'll ask Matthew if we're good to merge 😃

not-night-but (approved):

Issue #83, opened 2020-05-07
Big Query support


danipolo:
Nice project guys, why not BigQuery support?

rathboma:
If you can persuade google to give us a bunch of test credits I'll happily build a connection provider for it :-)

danipolo:
how much do you need? 👯

therustybear:
you just need to auth a credit card onto a google account and the first x (lots) of credits are free. Plenty to test GBQ

rathboma:
I have a GCP trial account. I don't have time to dedicate to this right now, but when I do, my primary concern is on-going support for the platform. As we add new features to the app we'll want to test them against BigQuery. I'll also want to integrate BigQuery into my (upcoming) integration test suite.

So I'd need a more permanent way to test it :-)

go-dustin:
Did you ever get the work env worked out? If not DM me on linkedin and I'll see what I can do to help out. 

https://www.linkedin.com/in/dustinwilliamsatx/

jbarneson:
hey Matt -- I'm an avid BQ consumer as well. I'd be happy to help get some testing on here if Dustin connected you with some GCP resources

StijnHoeke-MH:
Would love BQ to be a part of this, +1.

nicodemus-opon:
+1 would help

adampresley:
+1 here too. Then I could stop using DBeaver :)

MaxHalford:
+1

mickaelandrieu:
I'd become an Ultimate Edition Customer once it's done, this is the only missing feature I needed to uninstall DataGrip and only work with VSCode & Beekeeper Studio.

chekolyn:
Hola @rathboma,

I was wondering if you are taking any volunteers for this feature;  I was just doing some research on what's out there and we can probably use Google's Nodejs client: https://github.com/googleapis/nodejs-bigquery. For integration testing  this emulator looks promising https://github.com/goccy/bigquery-emulator ( I have not tried this yet). 

I was doing a small poc  with my gcp account and I was able to do a basic bigquery query with the above client using an IAM service account. 
![Screenshot from 2023-05-02 21-02-18](https://user-images.githubusercontent.com/1329157/235830827-e6a1bd85-4df7-437a-8a5e-d332d74052c2.png)

rathboma:
That's awesome! Would love to have you commit a barebones implementation -- we'd be happy to then take it over and tidy it up :-).

[...]

PR #1635, opened 2023-07-12
Initial Bigquery Client Support 

chekolyn:
Basic implementation for BigQuery client support: https://github.com/beekeeper-studio/beekeeper-studio/issues/83 

Only Service Account authentication is working; since Google Authentication can be implemented in numerous ways.


rathboma:
Whaaaaat!!? This is awesome. We'll get this reviewed ASAP. Thanks for the amazing PR!

apps/studio/src/lib/db/clients/bigquery.js
```diff
@@ -0,0 +1,276 @@
...
+}
+
+
+function prettyJSON(obj) {
+  logger().debug(JSON.stringify(obj, null, 2))
+}
+
+
+async function listTablesOrViews(client, db, type) {
+  // Lists all tables or views in the dataset
```
not-night-but, line 191: This throws an error if there is no dataset currently selected (connecting without a default dataset for instance.

not-night-but (changes requested):
@chekolyn Hey! Sorry it took a bit to get back to you on this, we've had a lot going on.

I haven't gone super in depth into the code yet, but some initial issues I'm finding when messing around with this locally:

- [x] It seems the implementation isn't checking that the connection was successful before proceeding to the core interface (ie query editor, tables list). So I can actually 'connect' to a bigquery database even if I don't have one running.
- [x] I'm also having trouble with just running queries in general, but I suspect this has to do with the above issue, so I'm not even sure if my connection is valid. I will continue messing around with it
- [x] Add database doesn't currently work `The database could not be created: Cannot read properties of undefined (reading 'apply')`
- [ ] Once we get the basic functionality working we should probably write some integration tests similar to how the other databases are implemented if possible. Would help with making sure everything is functioning correctly. 

This is a great start! Looks like all the foundations are there, we just need to figure out what is going on with the connection interface issue and then I'm sure the rest of the issues will become much easier to debug.

I did pull down the PR and fix the build/lint issues that are preventing the workflows from running properly.

I will go through your implementation soon and give you feedback on that, but I think these issues are more pressing than code style

chekolyn:
@not-night-but thanks for the feedback will get working on the above.

apps/studio/src/lib/db/clients/bigquery.js
chekolyn: I have found the bug for this; this line is not needed in the implementation. 
```server.config.client = client // save the client for later use```

**Comment by not-night-but** [comment]
@chekolyn yay! I'll take a look today!

> I was hopping you can provide more insight regarding the built/lint issues; not sure were to look regarding that.

Sorry what build/lint issues are you experiencing? The workflows seem to have passed? When developing if you want to check linting you should be able to run `yarn run all:lint` and that will lint/autofix any issues

File: apps/studio/src/migration/dev-1.js [hunk 2]
```diff
@@ -44,6 +44,14 @@ export default {
...
       defaultDatabase: 'sakila'
     },
+    {
+      name: "[DEV] Docker Bigquery",
+      connectionType: 'bigquery',
+      port: 9050,
+      host: 'localhost',
+      projectId: 'bks',
+      defaultDatabase: 'world',
+    },
```
not-night-but, line 54: This should be in it's own migration, look at dev-3.js on how to do that. The reason is, for existing development builds this won't run.

not-night-but (changes requested):
@chekolyn Great work so far! I still have to go through a bunch of the code, but right now there is one big issue: we need a dialect data file for this. You can look in `shared/src/lib/dialects/` for examples of this, and that is where it should go. 
This file provides a lot of functionality for the app (you may notice the query editor doesn't work as well as the other dialects right now, this is the reason). It seems that it is just defaulting to the SQLite dialect data, which is not what we want:
image

I'll continue to go through the code and add feedback. If you could resolve conversations that you have addressed that would be great.

I merged master into your branch so there are no conflicts at the moment. You're welcome to ping me if you need any clarification on things.

Thanks again!
Day

chekolyn:
Hola @not-night-but ,

I've been looking at the dialect code and it's going to take me a little while to figure it out. This part of the code also uses the knex module, the good news is that the sql-formatter does have a dialect for **bigquery** out of the box. 
 
> @chekolyn Great work so far! I still have to go through a bunch of the code, but right now there is one big issue: we need a dialect data file for this. You can look in `shared/src/lib/dialects/` for examples of this, and that is where it should go. This file provides a lot of functionality for the app (you may notice the query editor doesn't work as well as the other dialects right now, this is the reason). It seems that it is just defaulting to the SQLite dialect data, which is not what we want: image

not-night-but:
@chekolyn alrighty. I'll look into that knex extension you linked, but it looks like it's not maintained so we might have to fork it 🤷

not-night-but:
@chekolyn Hey! So just to keep you updated, we are gonna get this merged pretty soon. This will be an "alpha" release of the feature though, as there are probably still some things to clean up. Great work though! 

Thank you so much for contributing,
Day

chekolyn:
Hola @not-night-but,

That's great news! Please let me know if any additional help is needed.

rathboma:
@not-night-but thought we were merging this then fixing? Lets get it merged.

not-night-but:
Yeah sorry, I'm working on another branch, just haven't clicked the button yet.

not-night-but (approved):

Issue #1517, opened 2023-02-18
BUG: MS SQL - precision and scale missing for decimal values


rathboma:
![image](https://user-images.githubusercontent.com/279769/219902756-d7b4c844-a489-4b53-8d7e-674b6b73f3f3.png)

davidkaufman:
+1 for this!

PR #1664, opened 2023-08-11
Fix Issue #1517: Add Precision and Scale for Decimal Values in MS SQL

tkudlicka:
This pull request addresses Issue #1517, which was about missing precision and scale values for decimal columns in MS SQL.

Changes Made:

Modified the SQL query in apps/studio/src/lib/db/clients/sqlserver.js   to include precision and scale values for decimal columns.
Tested the changes in a local environment with MS SQL.

Screenshot: 
Screenshot 2023-08-11 at 11 44 02


not-night-but:
Hi! Thanks so much for filing this PR! I'll try to fool around with it today and see if we can get it merged!

not-night-but (approved):
Looks good to me! Works well too, thanks so much!

PR #1670, opened 2023-08-16
add: Working on adding vimrc support for beekeeper

austinwilcox:
I would like some help setting up the ipcrenderer procces to actually read in the vimrc, which on the flip side of that I still need to set up the reading of the file to pull from the users config folder, on Linux or Mac it should be the same but that place would be different on a windows machine. @not-night-but Any chance I can get your help on the ipcrenderer piece of this? I spent some time reading through the code and wasn't sure where the best place to place this would be. Most of the code needed should be in here, once the ipc section is added I can do some testing on things and see how well it works.


rathboma:
@austinwilcox hey! Quick question -- you asked about a `ipcRenderer`, what do you need that for? You should be able to read the file directly from the vue app (we have nodeIntegration enabled)

austinwilcox:
@rathboma That might be the most exciting piece of news I have heard all day! I had no idea that you could do that. I will try and have the PR updated today with a working version.

austinwilcox:
Alright, I have a working version of the code now running. This support basic commands like the following:
```
nmap gl $
nmap gh ^
nmap n nzz
```
The parser I wrote for the config is very bare bones, just to get something off the ground, I would like to improve it so that we can read in cmd mode commands as well so that we can have more sophisticated remaps. I have not written the code yet for unmapping, I think I will plan to do that when I make the parser more sophisticated. If you want that done before this code gets merged in let me know and I can work on that.

not-night-but (approved):
Just tested this myself and I'm SO HAPPY. I always miss my helix keybindings whenever I'm in beekeeper, so I'm excited to at least have some of the simpler ones around now!

If you have any other additions you want to make you are more than welcome to open another PR, but I'm gonna merge this pretty soon here.

Thanks so much!!!
Day

PR #1673, opened 2023-08-18
Add unsigned column to table builder & fix overflowing notice

### LINKED ISSUE #1660: FEAT: unsigned column
While creating/altering a table, I need to specify if the data will be signed or unsigned. I should be able to set that property visually.

Captura de pantalla de 2023-08-08 14-49-47

azmy60:
Resolves #1660

- [x] Add unsigned column to table builder
image

- [ ] Add unsigned column to table edit

- [x] Fix a breaking notice in status bar when the screen is smaller

image

Updated:
image

### DIFF
--- a/apps/studio/src/assets/styles/app/_base.scss
+++ b/apps/studio/src/assets/styles/app/_base.scss
@@ -23,3 +23,6 @@ body {
 .material-icons {
   transform: rotate(0.03deg);
 }
+.invisible {
+  visibility: hidden;
+}
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -263,10 +263,10 @@
           </x-buttons>
         </template>
         <template v-if="!editable">
-          <span
-            class="statusbar-item"
-            :title="readOnlyNotice"
-          ><i class="material-icons-outlined">info</i> Editing Disabled</span>
+          <span class="statusbar-item item-notice" :title="readOnlyNotice"
+            ><i class="material-icons-outlined">info</i>
+            <span> Editing Disabled</span>
+          </span>
         </template>
 
         <!-- Actions -->
@@ -315,6 +315,11 @@
   position: absolute;
   right: 50%;
   top: 200px;
+}
+.item-notice > span {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 }
 </style>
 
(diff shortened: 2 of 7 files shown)


rathboma (commented):
Question - instead of adding a new column, what if we just add `INT UNSIGNED` to the list of column types a user can select from?

azmy60:
Oh, that's actually better cause it's much simpler!

rathboma:
@azmy60 let me know when this is done and ready for review again!

azmy60:
I think I have the changes in my local. I'll update you soon today.

azmy60:
@rathboma I think unsigned columns are only supported in MySQL/MariaDB, right?

rathboma:
> @rathboma I think unsigned columns are only supported in MySQL/MariaDB, right?

Yes? I usually have to google this type of thing, but I think so

shared/src/lib/dialects/mysql.ts
```diff
@@ -12,7 +12,9 @@ import {
 
 const types = [
   ...SpecialTypes,
-  'bit', 'int', 'integer', 'tinyint', 'smallint', 'mediumint', 'bigint', 'float', 'double', 'double precision', 'dec', 'decimal', 'numeric', 'fixed', 'bool', 'boolean', 'date', 'datetime', 'timestamp', 'time', 'year', 'char', 'nchar', 'national char', 'varchar', 'nvarchar', 'national varchar', 'text', 'tinytext', 'mediumtext', 'blob', 'longtext', 'tinyblob', 'mediumblob', 'longblob', 'enum', 'set', 'json', 'binary', 'varbinary', 'geometry', 'point', 'linestring', 'polygon', 'multipoint', 'multilinestring', 'multipolygon', 'geometrycollection'
+  'bit', 'int', 'int unsigned', 'integer', 'integer unsigned', 'tinyint', 'tinyint unsigned',
```
rathboma, line 15: When we query a table for types, do we appropriately bring back the `unsigned` component? Like if you load a table in the TableStructure tab, does it show unsigned integers?

Can you add a test to the mysql integration tests to make sure that you can both create, and retrieve unsigned columns.

rathboma (changes requested):
Let's add a unit test, then we're all good.

shared/src/lib/dialects/mysql.ts
azmy60: Fortunately, it just shows the `unsigned`. Pretty handy! Also in the main table view. And I'm adding the tests now.

image

rathboma (approved):

PR #1687, opened 2023-09-04
Open filtered table in a new query tab

### LINKED ISSUE #1677: FEAT: Open SQL query from table view
I would like to see a feature where a button/link/option is added to the table view that would open a new query tab that has the same SQL used by the table view.

SELECT --* or selected fields-- FROM --open table-- WHERE --basic or advanced search-- ORDER BY --Ordered column--

This would save me a lot of time if I do not need to recreate the SQL.

azmy60:
Closes #1677 

Animation4

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -304,7 +304,7 @@
               <x-label>Show or hide columns</x-label>
             </x-menuitem>
             <x-menuitem @click="openQueryTab">
-              <x-label>Open in query tab</x-label>
+              <x-label>Copy view to SQL</x-label>
             </x-menuitem>
           </x-menu>
         </x-button>
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -303,6 +303,9 @@
             <x-menuitem @click="showColumnFilterModal">
               <x-label>Show or hide columns</x-label>
             </x-menuitem>
+            <x-menuitem @click="openQueryTab">
+              <x-label>Open in query tab</x-label>
+            </x-menuitem>
           </x-menu>
         </x-button>
       </div>
@@ -375,6 +378,7 @@ import { TableUpdate } from '@/lib/db/models';
 import { markdownTable } from 'markdown-table'
 import { dialectFor, FormatterDialect } from '@shared/lib/dialects/models'
 import { format } from 'sql-formatter';
+import { safeSqlFormat } from '@/common/utils'
 const log = rawLog.scope('TableTable')
 const FILTER_MODE_BUILDER = 'builder'
 const FILTER_MODE_RAW = 'raw'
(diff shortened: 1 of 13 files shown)


wmontgomery (approved):
Looking good!

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -303,6 +303,9 @@
             <x-menuitem @click="showColumnFilterModal">
               <x-label>Show or hide columns</x-label>
             </x-menuitem>
+            <x-menuitem @click="openQueryTab">
+              <x-label>Open in query tab</x-label>
```
rathboma, line 307: Not sure what copy we should use here.

We can use one of two conventions:
- How we have 'apply' and 'copy to sql' when performing an action (like editing)
- How we have `SQL: xyz` when right clicking a table.

I'm included to use the second convention here because this isn't the secondary action on an `apply` button.

Options:
- :cow:  `Open in query tab`
- :cat:  `Open table filter as SQL`
- :dog: `Copy to SQL`
- :deer: `Copy view to SQL`
- :snail:  `SQL: Table query`
- :skunk:  `SQL: Table view`

@not-night-but @wmontgomery @azmy60 vote with the emoji as a response to this comment!

rathboma: My vote: :deer:

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -525,6 +526,30 @@ async function selectTopStream(
...
+    table,
+    offset,
+    limit,
+    orderBy,
+    filters,
+    selects,
+    schema,
+    version,
+  });
+  return knex.raw(query.replaceAll(/\$\d+/g, '?'), params).toQuery();
```
rathboma, line 550: Hmm, isn't there a way with knex to have it substitute params?

apps/studio/src/components/tableview/TableTable.vue
not-night-but: My vote: 🦌

azmy60: I vote 🦌, it looks right by convention.

apps/studio/src/lib/db/clients/postgresql.ts
azmy60: Yeah, I just found that by digging into the repo a little bit. Let me see if I can fix it.

apps/studio/src/components/tableview/TableTable.vue
wmontgomery: 🦌 is my vote. Would it make sense to change "view" to whatever it is whether it's "table" or "view"?

rathboma (changes requested):
Once we've removed the regex param substitution, we're good to go.

@azmy60 this will also need a PR on the ultimate edition to implement for Oracle and Cassandra

apps/studio/src/lib/db/clients/postgresql.ts
azmy60: I think I misread something. I couldn't find a way to do it with knex. It only let me to use Positional bindings (`?` or `??`) and Named bindings (`:`). I instead copy paste the query builder and hardcode the params.

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -444,6 +444,53 @@ function buildSelectTopQueries(options: STQOptions): STQResults {
   }
 }
 
+function buildParameterizedSelectTopQuery(
```
rathboma, line 447: Hey, it's not a good idea to have duplicate code like this, if we need to fix a bug, we'd have to fix it twice. Feels like you're only going with this solution because it was quick.

It would be better to make `buildSelectTopQueries` take an additional parameter like `inlineParams = false`.

Seems like the only thing that needs to be changed is the calculation of `value` during the filter creation.

Then you can add some tests for `buildSelectTopQueries` with both sets of flags.

rathboma (changes requested):
I think we solved one problem but introduced another :-). Let's not have duplicate code, but instead add an optional parameter for the original function that changes it's behavior

apps/studio/src/lib/db/clients/postgresql.ts
azmy60: Sorry, that was a careless decision of me, I don't know what I was doing 😞 .

rathboma: Not a big deal :-)

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -390,15 +391,19 @@ function buildSelectTopQueries(options: STQOptions): STQResults {
     let paramIdx = 1
     filterString = "WHERE " + filters.map((item) => {
       if (item.type === 'in' && _.isArray(item.value)) {
-        const values = item.value.map((_v, idx) => {
-          return `$${paramIdx + idx}`
+        const values = item.value.map((v, idx) => {
+          return options.inlineParams
+            ? knex.raw('?', [v]).toQuery()
```
rathboma, line 396: smart way to quote the values!

rathboma (approved):

rathboma:
@azmy60 this will need a follow-up PR on ultimate for the ultimate-only database drivers

azmy60:
Ok! Do I need to wait for this to be merged there? I can make the PR this week, if you'd like to.

PR #1689, opened 2023-09-04
Fix invalid array syntax in postgres

### LINKED ISSUE #1596: BUG: Invalid SQL for "Copy insert" action of dropdown in table for Postgres
**Describe the bug**
Invalid SQL for "Copy Insert" action of dropdown in table data with array column for Postgres.

**To Reproduce**
1. Create table with array column and insert row.
2. Open table.
3. Click RMB on table row then click "Copy Row (insert)"
4. Paste and run copied query.

**Expected behavior**
The insertion was successful.

 - OS: macOS Ventura 13.3.1
 - App Version: 3.9.9
 - Database: Postgres 10.4

**Additional context**
Expected SQL, ex.:
```sql
insert into "public"."test" ("some_array") values ('{1,9,2,10,11}');
```
Actual SQL with error, ex.:
```sql
insert into "public"."test" ("some_array") values ('[1,9,2,10,11]');
```

azmy60:
Fix invalid array syntax by making it written and shown with `{}` instead of `[]`.

image

Closes #1596


apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1189,7 +1189,9 @@ async function updateValues(cli: any, rawUpdates: TableUpdate[]): Promise<TableU
   const updates = rawUpdates.map((update) => {
     const result = { ...update}
     if (update.columnType?.startsWith('_')) {
-      result.value = JSON.parse(update.value)
+      const extracted: string | undefined = update.value.match(/\{(.*)\}/)?.[1]
```
rathboma, line 1192: Hmm, this seems pretty brittle. This would match strings with `{` legitimately in them.

How about we use postgres-array? It's a package on NPM - https://github.com/bendrucker/postgres-array

That way we don't need to own the parsing. Generally if I see a regex with database values that makes my spidey-sense go off a little bit because I did this a lot in the beginning and caused a lot of bugs :-D

apps/studio/src/mixins/data_converter.js
```diff
@@ -13,9 +13,14 @@ export default {
...
             */
           const columnNamesOnly = columns.map((c) => c.field)
-          return data.rows.map((row) => {
-          return _.pick(row, columnNamesOnly)
-          })
+          return data.rows.map((row) => _
+            .chain(row)
+            .pick(columnNamesOnly)
+            .transform((result, val, key) => {
+              result[key] = Array.isArray(val) ? `{${val.join(',')}}` : val
```
rathboma, line 20: Do you need to escape any  characters in the string? Maybe no, but worth a check.

rathboma (changes requested):
lets check for escaping characters and avoid parsing results using a regex, and instead use a package.

apps/studio/src/lib/db/clients/postgresql.ts
azmy60: lol, now I see that you mentioned it.

azmy60: Since we are changing the way we write array value from `[1,2,3]` to `{1,2,3}`, should we supply them as they are, or do we still need to parse them into arrays?

not-night-but: Sorry, do you mean should we change how they are displayed to the user in results/table?

azmy60: I feel like it's more natural to write the original syntax `{1,2,3}` in the table view rather than `[1,2,3]`. We can also avoid issues with parsing since we probably don't need it, but it's up to you.

azmy60: I had a little issue with the library you mentioned. It gave me some bad result when I typed: 
- parse(`{{1,2}, {3,4}}`) => `[ [ '1', '2' ], [ '3', '4' ], ' ' ]` ❌ 
- parse(`{{1,2},{3,4}}` )=> `[ [ '1', '2' ], [ '3', '4' ] ]` ✅

azmy60: @not-night-but yes, and also to send them to the client. Right now we send array to the client.

Sorry, I didn't see your reply @not-night-but, github didn't show it after I refresh the page 😅

not-night-but: Oh no problem! With the amount of hours between us we're bound to have very scattered conversations 😂. I agree with you, we should show arrays in the postgres syntax. It means those values can be directly pasted into a query, and may save us some conversions.

With that other thing, that is incredibly confusing, but it isn't necessarily wrong? I'm not super familiar with postgres arrays, but we could probably just filter the result for any empty values, no?

azmy60: That could do 🤔 . But it gets even messier if I type `{ "a" , "b" }`, and it gives me this result `[ ' a', ' ', ' b', ' ' ]`.

not-night-but: Hmmm yeah that's annoying. Seems to be an issue with the parser itself. Probably gets even worse with multidimensional arrays as well 😡. 

From the examples you've given it seems that the parser is expecting no spaces at all... What happens if you input something like `{"hello there","hi"}`?

azmy60: Yeah I think it expects no spaces. That's just `[ 'hello there', 'hi' ]`.

not-night-but: Are you inputting these values manually? Or are they coming from a Postgres database. I wonder if postgres removes the extraneous spaces when it stores and that's why the lib doesn't parse the spaces properly. If that's not the case then we're probably blocked on whether we want to just file a PR on the original repo, fork it and make the necessary changes ourselves, or just implement our own parser 🤷

azmy60: Yes, it's manual. I mean, when you try to add rows of records, or applying some changes in the table view, some users probably would use extra spaces which is reasonable. And that's ok in postgres. I'll add some tests related to this to consolidate the case.

not-night-but: I just mean if you create a table with an array column, insert something like `{ "1" , "2" , "3" }`, what gets returned when selected? Does postgres do any sanitization before storing the record?

azmy60: I think postgres remove the spaces internally, if I run `insert into ... ('{ "1" , "2" }')` and select it , it will return `{"1","2"}`.

not-night-but: So at least that explains why this was overlooked in the library. I'm sure @rathboma will have some ideas when he wakes up lol

rathboma: Hi Azmy,
I think for now, if it's easier to keep it as `{}`, then lets do that.

apps/studio/src/mixins/data_converter.js
```diff
@@ -1,4 +1,7 @@
 import _ from 'lodash'
+import knexlib from 'knex'
+
+const knex = knexlib({ client: 'pg' })
```
rathboma, line 4: Hey! This looks to be postgres-specific, but this mixin is used for all the different databases, so this isn't going to work

rathboma (changes requested):
Everything looks great aside from the use of `pg` specific logic in a Vue mixin.

azmy60:
I dig a little more to find that the exporter needs a little bit of adjustment too.

apps/studio/src/lib/db/clients/index.ts
```diff
@@ -1,12 +1,21 @@
...
+import mysql, { knex as mysqlKnex } from './mysql';
+import postgresql, { knex as postgresKnex } from './postgresql';
+import sqlserver, { knex as sqlServerKnex } from './sqlserver';
+import sqlite, { knex as sqliteKnex } from './sqlite';
 import cassandra from './cassandra';
 import bigquery from './bigquery.js';
+import { Dialect } from '@shared/lib/dialects/models';
+import { Knex } from 'knex';
 
+export function knexOf(dialect: Dialect): Knex | undefined {
```
rathboma, line 12: I don't like these types of centralized switches on database type as they're easy to miss when adding a new database, but I think this will go away if we keep arrays as JSON arrays when in the UI.

apps/studio/src/lib/data/tools.ts
```diff
@@ -40,11 +41,17 @@ export const Mutators = {
...
    */
-  genericMutator(value: any, preserveComplex = false): JsonFriendly {
-    const mutate = Mutators.genericMutator
+  genericMutator(dataType: string, dialect: Dialect, value: any, preserveComplex = false): JsonFriendly {
+    const mutate = Mutators.genericMutator.bind(this, dataType, dialect)
     if (_.isBuffer(value)) return value.toString('hex')
     if (_.isDate(value)) return value.toISOString()
-    if (_.isArray(value)) return preserveComplex? value.map((v) => mutate(v, preserveComplex)) : JSON.stringify(value)
+    if (_.isArray(value)) {
+      if (preserveComplex) return value.map((v) => mutate(v, preserveComplex))
```
rathboma, line 49: I think this file has changed my mind. I think we should show arrays in the front end as looking like JSON arrays `[a, b, c]`, and convert them in the drivers when writing to the database.

Having them appear as db-specific strings just moves too much database-knowledge into the front-end.

This will also help with data importing where we need to convert arrays for insert.

rathboma (changes requested):
Sorry, I've changed my mind and think we should just show arrays the same way across all databases because I don't like so much database-specific logic living in the UI.

I think you did this PR well, but we're working around the problem that the front-end really shouldn't have so much database specific serialization.

apps/studio/src/lib/data/tools.ts
azmy60: That makes sense.

azmy60:
Closing in favor of #1708.

This PR goes into a different direction and has too many changes, so it's much easier to start over.

PR #1690, opened 2023-09-04
fix: add a default column name title of Result

### LINKED ISSUE #1662: FEAT: Sensible default for blank or whitespace only result column titles
When running a query such as
```sql
select count(*) from sys.procedures
```

The results column header shows up as `&nbsp;` unless I also alias the result like in the following

```sql
select count(*) as stored_procedures from sys.procedures
```

I'd like unaliased results for unnamed results to be given a default name such as "result count", "results" or similar instead of the escaped string value.

While always aliasing results is probably best practice, it would be nice to have a sensible default.

**Additional context**
Screenshot 2023-08-10 at 9 22 15 AM

This affects all databases.

wmontgomery:
fix #1662

record-count

### DIFF
--- a/apps/studio/src/components/editor/ResultTable.vue
+++ b/apps/studio/src/components/editor/ResultTable.vue
@@ -177,11 +177,9 @@
       },
       tableColumns() {
         const columnWidth = this.result.fields.length > 30 ? globals.bigTableColumnWidth : undefined
-        return this.result.fields.map((column) => {
-          console.log('COLUMN: ', column);
-          console.log(column.name)
+        return this.result.fields.map((column, index) => {
           const result = {
-            title: column.name || 'Result',
+            title: column.name || `Result ${index}`,
             titleFormatter: 'plaintext',
             field: column.id,
             titleDownload: escapeHtml(column.name),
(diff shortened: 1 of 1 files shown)


apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -179,8 +179,9 @@
         const columnWidth = this.result.fields.length > 30 ? globals.bigTableColumnWidth : undefined
         return this.result.fields.map((column) => {
           console.log('COLUMN: ', column);
+          console.log(column.name)
```
not-night-but, line 182: I'd just get rid of the logs here

not-night-but (approved):

rathboma (changes requested):
just kill the logs and we're good to go!

apps/studio/src/components/editor/ResultTable.vue
wmontgomery: Oops. Consider them gone!

wmontgomery:
> what if there are multiple unnamed columns? Are they all `Result`?

Yeah there would be. It's not a frequent case (at least I hope not as you should be aliasing your columns). I can add some kind of counter to it if wanted. That's actually fairly easy since it's a map.

Something like this just to get the index in there
```js
return this.result.fields.map((column, index) => {
  const result = {
    title: column.name || `Result ${index}`,
```

rathboma (approved):

PR #1691, opened 2023-09-04
fix: allowing adding records to tables with no primary key

### LINKED ISSUE #1688: BUG: Unable to add record from table that has no primary key
**To Reproduce**
1. Create a table without primary key (e.g. `CREATE TABLE sal_emp (first_name varchar, last_name varchar);`)
2. Open the table view
3. Add new record / ctrl + n
4. Cells are disabled and can't fill new data

I think it should be possible to visually add a record in tables without primary key. We probably need some kind of rewrite to disable the cells when editing existing records, but not when adding new records.

iifawzi:
Hey, This PR should fix #1688

Sep-04-2023 22-28-02

Closes: #1688

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -561,7 +561,7 @@ export default Vue.extend({
               cell.setValue(null);
             }
           },
-          disabled: !this.editable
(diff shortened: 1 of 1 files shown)


rathboma (approved):
awesome, thank you so much for the great contribution!

PR #1692, opened 2023-09-05
Fix invalid bool and tabulator update data

### LINKED ISSUE #1408: BUG: sqlite3 bool field problem
**Describe the bug**
When I want to change the value of a boolean field in the sqlite3 database, I encounter the following error:
`SQLite3 can only bind numbers, strings, bigints, buffers, and null`
beekeeper only allows entering `true` and `false` values.

**Screenshots / Gifs**

ssbeekeeper

**Version Information:**
 - OS: Arch Linux x86_64 6.0.2
 - App Version 3.6.2
 - Database type and version sqlite3

azmy60:
This introduces three fixes 🎉:
-  The boolean data hasn't been converted to `1` or `0` when applying changes. We have no issue when inserting new data because it's converted by `knex`.
- The `tabulator.updateData` is ignored due to the data supplied to it didn't contain the [index column.](https://tabulator.info/docs/5.5/update#alter-update:~:text=This%20function%20takes%20an%20array%20of%20row%20objects%20and%20will%20update%20each%20row%20based%20on%20its%20index%20value.%20Options%20without%20an%20index%20will%20be%20ignored%2C%20as%20will%20items%20with%20an%20index%20that%20is%20not%20already%20in%20the%20table%20data.)
- Add false/true options when adding `boolean` value

Closes #1408

### DIFF
--- a/apps/studio/src/lib/db/clients/sqlite.js
+++ b/apps/studio/src/lib/db/clients/sqlite.js
@@ -252,7 +252,7 @@ export async function applyChanges(conn, changes) {
 
 export async function updateValues(cli, updates) {
   const commands = updates.map(update => {
-    const params = [update.value];
+    const params = [_.isBoolean(update.value) ? _.toInteger(update.value) : update.value];
     const whereList = []
     update.primaryKeys.forEach(({ column, value }) => {
       whereList.push(`${wrapIdentifier(column)} = ?`);
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -749,7 +749,9 @@ export default Vue.extend({
           editorParams: {
             verticalNavigation: useVerticalNavigation ? 'editor' : undefined,
             search: true,
-            values: column.dataType === 'bool' ? [true, false] : undefined,
+            values: /^(bool|boolean)$/i.test(column.dataType)
+              ? [true, false]
+              : undefined,
             allowEmpty: true,
             // elementAttributes: {
             //   maxLength: column.columnLength // TODO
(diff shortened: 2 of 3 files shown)


apps/studio/src/lib/db/clients/sqlite.js
```diff
@@ -252,7 +252,7 @@ export async function applyChanges(conn, changes) {
 
 export async function updateValues(cli, updates) {
   const commands = updates.map(update => {
-    const params = [update.value];
+    const params = [_.isBoolean(update.value) ? _.toInteger(update.value) : update.value];
```
rathboma, line 255: Could you add an integration test for this to tests/integration/lib/db/sqlite? You should verify that the test FAILS without this fix, then verify that this patch fixes the test.

rathboma (changes requested):
Test needed for the SQLite client

apps/studio/src/lib/db/clients/sqlite.js
rathboma: Make sure to also test setting the value to `null` to make sure no one breaks this in the future, for sure we want to make sure if someone sets the value to `null` it doesn't turn to `false` by mistake.

azmy60: Hi @rathboma , I've just added the tests for this. Also it's probably worth to mention that if you create a table that contains `boolean` and `bool` data types:

```sql
CREATE TABLE boolean_table (
	id integer PRIMARY KEY AUTOINCREMENT,
   	flag1 bool,
	flag2 boolean
);
```

Beekeeper would should show true/false autocomplete on the `bool` column, but not on the `boolean` column, which would allow users to insert any data type other than 0 and 1.

rathboma: oh, that seems like a mistake, can you fix that also?

azmy60: No problem, I'll add that in this PR soon :)

azmy60: done

rathboma (approved):

PR #1701, opened 2023-09-13
Add comment column and update some styling to the table builder

### LINKED ISSUE #1644: BUG: Can't see comment column when modifying a table
**Details:**
How to show comment colume

if i dont use auto_increment, here will show comment colume.
but if i use it, the comment colume will disappear

ver 3.9.17

image

azmy60:
- add comment column in the table edit screen
- ux: also freeze the name column to stay in position so we can see it when scrolling

Animation4

Closes #1644

### DIFF
--- a/apps/studio/src/components/tableinfo/TableSchema.vue
+++ b/apps/studio/src/components/tableinfo/TableSchema.vue
@@ -269,7 +269,7 @@ export default Vue.extend({
           editor: vueEditor(NullableInputEditorVue),
           minWidth: 90,
         }),
-        {
+        (this.disabledFeatures?.comments ? null : {
           title: 'Comment',
           field: 'comment',
           tooltip: true,
@@ -279,7 +279,7 @@ export default Vue.extend({
           cellEdited: this.cellEdited,
           editor: vueEditor(NullableInputEditorVue),
           minWidth: 90,
-        },
+        }),
         {
           title: 'Primary',
           field: 'primary',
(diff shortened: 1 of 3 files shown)


apps/studio/src/components/tableinfo/TableSchema.vue
```diff
@@ -262,8 +266,20 @@ export default Vue.extend({
           editable: this.isCellEditable.bind(this, 'alterColumn'),
           formatter: this.cellFormatter,
           cellEdited: this.cellEdited,
-          editor: vueEditor(NullableInputEditorVue)
+          editor: vueEditor(NullableInputEditorVue),
+          minWidth: 90,
         }),
+        {
+          title: 'Comment',
+          field: 'comment',
```
rathboma, line 274: @azmy60 this is great! One thing we need to change though -- not all databases support `comment` on columns. You'll need to check `this.disabledFeatures?.comments`, similar for how we check for `extra` on line 261 above.

If you want to see where this is specified, take a look at src/lib/dialects/sqlite.ts <- you can see comments are not supported by Sqlite.

shared/src/components/SchemaBuilder.vue
```diff
@@ -444,6 +462,22 @@ export default Vue.extend({
       }
     }
 
+    // Apply when not dragging rows so it can be transparent
+    .tabulator:not(.tabulator-block-select) .tabulator-row .tabulator-frozen {
```
rathboma, line 466: just double check this all looks good on `light` mode also, but seems to be great!

rathboma (changes requested):
You just need to check if the database being edited supports `comments`, other than that, looks great!

rathboma (approved):
Merge away!

azmy60:
@rathboma  Oh wait I missed one more check in the other file 😆. I'll merge it after.

PR #1706, opened 2023-09-22
Should trigger initial filters on first time load and other bugs

azmy60:
Animation4

Bugs:
- Does not run `initialFilters` initially
- After going to a table from a fk link, try adding some new filters, it will show empty AND/OR buttons
- Cannot focus the filter input with ctrl+f
- Converting to raw filter when the filter value is not a string gives an error. Steps to reproduce:
  1) Make a filter that uses number, e.g. `id` => `equals` => `1`
  2) Disconnect from db and reconnect
  3) Try switching to raw filter

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -224,7 +224,7 @@ import data_converter from "../../mixins/data_converter";
 import DataMutators, { escapeHtml } from '../../mixins/data_mutators'
 import { FkLinkMixin } from '@/mixins/fk_click'
 import Statusbar from '../common/StatusBar.vue'
-import RowFilterBuilder from './RowFilterBuilder.vue'
+import RowFilterBuilder, { normalizeFilters } from './RowFilterBuilder.vue'
 import ColumnFilterModal from './ColumnFilterModal.vue'
 import rawLog from 'electron-log'
 import _ from 'lodash'
@@ -924,6 +924,7 @@ export default Vue.extend({
       this.rawTableKeys = await this.connection.getTableKeys(this.table.name, this.table.schema)
       const rawPrimaryKeys = await this.connection.getPrimaryKeys(this.table.name, this.table.schema);
       this.primaryKeys = rawPrimaryKeys.map((key) => key.columnName);
+      this.filters = normalizeFilters(this.initialFilters || [])
 
       this.tabulator = new TabulatorFull(this.$refs.table, {
         height: this.actualTableHeight,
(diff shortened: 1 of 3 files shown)


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -224,7 +224,7 @@ import data_converter from "../../mixins/data_converter";
 import DataMutators, { escapeHtml } from '../../mixins/data_mutators'
 import { FkLinkMixin } from '@/mixins/fk_click'
 import Statusbar from '../common/StatusBar.vue'
-import RowFilterBuilder from './RowFilterBuilder.vue'
+import RowFilterBuilder, { normalizeFilters } from './RowFilterBuilder.vue'
```
rathboma, line 227: You should move this function somewhere else, Vue files should really only contain the vue component. Maybe utils?

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -924,6 +924,7 @@ export default Vue.extend({
       this.rawTableKeys = await this.connection.getTableKeys(this.table.name, this.table.schema)
       const rawPrimaryKeys = await this.connection.getPrimaryKeys(this.table.name, this.table.schema);
       this.primaryKeys = rawPrimaryKeys.map((key) => key.columnName);
+      this.filters = normalizeFilters(this.initialFilters || [])
```
rathboma, line 927: Did we need to do this previously? I don't think we did, shouldn't `initialFilter` in the tabulator constructor do this?

rathboma (changes requested):
export from a different file & investigate initialFilter

apps/studio/src/components/tableview/TableTable.vue
azmy60: Actually, we [had](https://github.com/beekeeper-studio/beekeeper-studio/blob/e4c8e713296b91045b361da076e4426a294932a9/apps/studio/src/components/tableview/TableTable.vue#L1124-L1126) this before but in a little different way. `this.filter` was used in [filterForTabulator()](https://github.com/beekeeper-studio/beekeeper-studio/blob/e4c8e713296b91045b361da076e4426a294932a9/apps/studio/src/components/tableview/TableTable.vue#L858) which then used in many different parts including the tabulator [dataFetch](https://github.com/beekeeper-studio/beekeeper-studio/blob/e4c8e713296b91045b361da076e4426a294932a9/apps/studio/src/components/tableview/TableTable.vue#L1656).

In this case, `this.filters` is empty and was not attached to the `initialFilters` initially, which is why when we initialize the tabulator, the data is not filtered at all.

azmy60: I accidentally removed that part and didn't know it's actually gonna affect the `initialFilter`.

apps/studio/src/components/tableview/TableTable.vue
azmy60: done :)

rathboma (approved):
Great stuff! Excited to launch this 🚀

PR #1708, opened 2023-09-25
Fix invalid sql array format on copy

### LINKED ISSUE #1596: BUG: Invalid SQL for "Copy insert" action of dropdown in table for Postgres

azmy60:
Instead of mutating the array/object into string for the whole table, we can convert them into string when viewing and editing, so we don't lose the real data type when calling `tabulator.getData()`. Preserving the real data type so knex can generate the query for us correctly.

Animation4

Changes:

- copy to SQL.
before => `insert into "public"."untitled_table" ("column_3") values ('[1,2,3,4,5]')`
after => `insert into "public"."untitled_table" ("column_3") values ('{1,2,3,4,5}')`

- copy to JSON
before => `["column_1":"[1,2,3]"}]`
after => `["column_1":[1,2,3]}]`

- copy to TSV for excel
before => `"1"	"2023-09-21 01:57:29.227987"	"[1,2,3]"`
after => `"1"	"2023-09-21 01:57:29.227987"	"1,2,3"`

- copy to Markdown
before => 
```md
| id | created_at                 | column_3 |
| -- | -------------------------- | -------- |
| 1  | 2023-09-21 01:57:29.227987 | [1,2,3]  |
```

after => 
```md
| id | created_at                 | column_3 |
| -- | -------------------------- | -------- |
| 1  | 2023-09-21 01:57:29.227987 | 1,2,3    |
```

Closes #1596

todo:
- [x] check result table
- [x] exporting sql using the exporter window is still wrong
- [x] check mysql json
- [x] bigquery array

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -640,6 +640,11 @@ export default Vue.extend({
               ? [true, false]
               : undefined,
             allowEmpty: true,
+            preserveObject: column.dataType.startsWith('_'),
+            onPreserveObjectFail: () => {
+              this.$noty.error("Value is not valid")
+              return true
+            }
             // elementAttributes: {
             //   maxLength: column.columnLength // TODO
             // }
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -644,7 +644,7 @@ export default Vue.extend({
             onPreserveObjectFail: () => {
               this.$noty.error("Value is not valid")
               return true
-            }
+            },
             // elementAttributes: {
             //   maxLength: column.columnLength // TODO
             // }
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -648,8 +648,8 @@ export default Vue.extend({
             search: true,
             allowEmpty: true,
             preserveObject: column.dataType.startsWith('_'),
-            onPreserveObjectFail: () => {
-              this.$noty.error("Value is not valid")
+            onPreserveObjectFail: (value: unknown) => {
+              log.error('Failed to preserve object for', value)
               return true
             },
             // elementAttributes: {
--- a/apps/studio/src/lib/data/tools.ts
+++ b/apps/studio/src/lib/data/tools.ts
@@ -10,16 +10,19 @@ function dec28bits(num: any): string {
 export const Mutators = {
 
   resolveTabulatorMutator(dataType?: string, dialect?: Dialect): (v: any) => JsonFriendly {
-    const mutator = this.resolveDataMutator(dataType, dialect)
+    const mutator = this.resolveDataMutator(dataType, dialect, true)
     return (v: any) => mutator(v, true) // this cleans off the additional params
   },
 
-  resolveDataMutator(dataType?: string, dialect?: Dialect): (v: any, p?: boolean) => JsonFriendly {
+  resolveDataMutator(dataType?: string, dialect?: Dialect, mutateJSON = false): (v: any, p?: boolean) => JsonFriendly {
     if (dataType && dataType === 'bit(1)') {
       return this.bit1Mutator.bind(this)
     }
     if (dataType && dataType.startsWith('bit')) {
       return this.bitMutator.bind(this, dialect)
+    }
+    if (dataType && dataType.startsWith('json') && mutateJSON) {
+      return this.jsonMutator.bind(this)
     }
     return this.genericMutator.bind(this)
   },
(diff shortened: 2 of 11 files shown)


azmy60:
@rathboma This PR would remove all the JSON centric array and objects format. So in md, tsv, etc.. we no longer stick with the json array brackets.

If we would do this change of copy styles, we will need to update the exporter too. Or just do this to the sql copy, it's up to you.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -640,6 +640,11 @@ export default Vue.extend({
               ? [true, false]
               : undefined,
             allowEmpty: true,
+            preserveObject: column.dataType.startsWith('_'),
```
rathboma, line 643: This looks like a good solution!

This will only fix this for Postgres though, correct? How do other database types identify arrays?

rathboma (changes requested):
This looks perfect. One question: How are arrays identified in other dialects?

apps/studio/src/components/tableview/TableTable.vue
rathboma: Isn't the value an array to begin with? If so, can't it preserve object by default for all arrays?

azmy60: Yup! Since postgres is the only database that support arrays, so this is safe IMO. However, mysql support json. So I'm gonna need to look at that to finish up the PR.

not-night-but: @azmy60 I don't know about other databases, but BigQuery does support arrays so maybe test that

rathboma: @azmy60 Hmm, if the mutator is converting from an array, that's probably not correct. I think we have a `preserveComplex` flag that is supposed to help fix that?

Good catch on the nullable editor turning the value into a string -- that definitely needs fixing. If the value is an array, or a json object, it should just keep it that way.

For background, Tabulator has 2 ways to manipulate values:
- Mutators - changes the ACTUAL value when set
- Formatters - only changes how the values are DISPLAYED

So mutators should def keep them as complex types when appropriate. See `genericMutator`. This has a `preserveComplex` argument that can be set to true. See also `resolveDataMutator`

`niceString` is only used in a formatter, so don't worry about that one.

rathboma: Sorry, this ended up way more complex than I thought it would be :-D

azmy60: Sorry, the `preserveObject` flag is a parameter for the nullable editor. It's the fix for turning the value back into it's original type when changing some values from cells.

As of `preserveComplex` in `genericMutator`, I had switched that to [true](https://github.com/beekeeper-studio/beekeeper-studio/pull/1708/files#diff-c43a89ad7180e5dcda2d74247d1a2823e3399ffd622b3fd5eee1af3b3ebd7b87L12-R15).

azmy60:
@not-night-but I still can't connect to bigquery for some reason. It tries to connect to an invalid url 🤔 ?

image

apps/studio/src/components/tableview/TableTable.vue
wmontgomery: This all look super helpful for an issue we're seeing with import where arrays are causing DB insert errors.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -647,6 +647,11 @@ export default Vue.extend({
             verticalNavigation: useVerticalNavigation ? 'editor' : undefined,
             search: true,
             allowEmpty: true,
+            preserveObject: column.dataType.startsWith('_'),
+            onPreserveObjectFail: () => {
```
rathboma, line 651: Probably this should just log to the console or something?

If there are 500 rows, this will trigger 500 times

rathboma (changes requested):
One final thing -- not spamming the user with `$noty` error messages :-D. Then good to go

apps/studio/src/components/tableview/TableTable.vue
azmy60: Oh yeah, that's really bad 😆

rathboma (approved):

PR #1712, opened 2023-09-29
Fix broken display after refreshing table list

azmy60:
to reproduce:
1. connect to any database
2. open a table or view
3. click refresh button on the table list

Animation4


rathboma:
Let me look into this.

wmontgomery (approved):

rathboma:
Your change does solve this error, but creates another side effect. I'm going to submit a better patch in a minute. I'll add you so you can see.

Problem with this PR:

- User has a table open
- Someone (elsewhere) runs a script to add columns to a table
- User clicks refresh so they can see the changes
- Columns don't update. The user is confused
- User would have to go into the tabletable and click refresh there

Issue #1716, opened 2023-09-29
BUG: Esc not exiting insert mode when using vim keybindings


austinwilcox:
[PR](https://github.com/beekeeper-studio/beekeeper-studio/pull/1717)

**Describe the bug**
Vim works to navigate around in normal mode and visual mode, but as soon as you enter insert mode, you cannot use the esc key to get back into normal mode. You can however do CTRL+[ and that will get you back into insert mode.

**Expected behavior**
When pressing escape in insert mode I expect it to kick me back out to normal mode.

**Screenshots / Gifs**
![Peek 2023-09-29 08-27](https://github.com/beekeeper-studio/beekeeper-studio/assets/43357911/72944bdb-9d44-4d9c-8b9b-8aa59bed7e7f)

**Version Information (please complete the following information):**
 - OS: Ubuntu
 - App Version: 4.0
 - Database type: Sql Server

I will check right now on a quick fix for this, also any software recommendations to show my keypresses for these gifs?

austinwilcox:
Merged with [PR 1717](https://github.com/beekeeper-studio/beekeeper-studio/pull/1717)

PR #1717, opened 2023-09-29
Removed the default keybind that bound esc to cancel query

austinwilcox:
This contains the fix for [Issue 1716](https://github.com/beekeeper-studio/beekeeper-studio/issues/1716). Esc was by default set to still cancel query instead of being handled later by checking to see if the user was using vim keybindings. This is a one liner change.

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -691,7 +691,6 @@
             "Shift-Cmd-F": this.formatSql,
             "Ctrl-/": this.toggleComment,
             "Cmd-/": this.toggleComment,
-            "Esc": this.cancelQuery,
             "F5": this.submitTabQuery,
             "Shift-F5": this.submitCurrentQuery,
             "Ctrl+I": this.submitQueryToFile,


rathboma:
In your PR - does Esc still cancel the query, but it's bound elsewhere?

austinwilcox:
```javascript
          if(this.userKeymap === "vim") {
            runQueryKeyMap["Ctrl-Esc"] = this.cancelQuery
          } else {
            runQueryKeyMap.Esc = this.cancelQuery
          }
```
This is the code that handles adding esc to cancel the query for regular users, it's just a couple lines below the longer list of keybindings. I will run a couple of quick tests and make sure it will still cancel the query for users not using vim.

austinwilcox:
Peek 2023-09-29 08-27
Here is a quick gif showing that, I still need to get something like screenkey running on my box so you can see the keypresses.

rathboma (approved):

rathboma:
Thanks Austin!

austinwilcox:
You're welcome. Happy to support the project where I can.

PR #1722, opened 2023-09-30
rollback sql-hint and rewrite schema autocomplete

### LINKED ISSUE #1720: BUG: Autocomplete wiping whole line
**Describe the bug**
When using the autocomplete for table columns on a left join, it wipes the line for the left join

**To Reproduce**
1. Start Typing out a left join
2. give table an alias
3. when typing the on clause, use the alias to autocomplete the column
4. select the autocompleted selection

**Expected behavior**
It should finish typing the selected column name

**Screenshots / Gifs**
I am attaching a video of it happening

 - OS: [e.g. Windows/Mac/Ubuntu]
        Windows
 - App Version [e.g. help -> About Beekeeper Studio]
        4.0.0
 - Database type and version [eg Posgresql 9.3]
        Mariadb and mysql

**Additional context**
Nope

https://github.com/beekeeper-studio/beekeeper-studio/assets/121575096/a5fb6714-edd6-4855-a7f9-f8194959d302

### LINKED ISSUE #1721: 4.0 Autocomplete regressions
There are a few regressions with autocomplete, I want to collect them all in one place for easy review and fixing.

- [ ]  #1720
- [ ] Autocomplete is quoting table aliases, which breaks the query - example `LEFT JOIN film as flm on a.id = flm.` <- autocomplete successfully finds the alias `flm` and the column names, but adds quotes around it `"flm"`, which breaks the query as this is not a formal identifier. See vid below

## Testing

Because our autocomplete system is not a real language server it is very brittle, and easy to break. That means we need to have a test suite. V1 test suite should be tackled as part of this issue.
- Tests should NOT require the DOM or CodeMirror
- We should include both of the above issues as explicit test cases
- Make it really easy to add a new test case for when we experience new regressions 

### Quoting table aliases

[Screencast from 09-29-2023 01:42:53 PM.webm](https://github.com/beekeeper-studio/beekeeper-studio/assets/279769/6078e51d-4991-43b6-b0b3-e47242c4566e)

azmy60:
I didn't realize we could autocomplete the schema with the original sql-hint file.

sql-hint is now reverted to this [commit](https://github.com/beekeeper-studio/beekeeper-studio/commit/f6f54a4e7ab0996a4394cf08957e38534e20b301).

todos:
- [x] fork [tests](https://github.com/codemirror/codemirror5/blob/master/test/sql-hint-test.js) from cm5 repo, add additional tests such as:
  - [x] must be able to autocomplete alias when using quotes on the table (e.g. `select t.<column> from "MyTable" as t`)
- [x] fix `text/x-pgsql` mode for postgres to allow " (double quotes) instead of \` (backticks) on identifiers or use `test/x-sqlite` mode temporarily (look at [codemirror5/mode/sql/sql.js](https://github.com/codemirror/codemirror5/blob/master/mode/sql/sql.js))
- [ ]  in postgres, alias shouldn't be quoted. The fix is not from sql-hint, but it's possible?, also look at codemirror5/mode/sql/sql.js

closes #1720


azmy60:
This was tested using postgres. I'm gonna need to test other databases and some edge cases too. I'm expecting there is not gonna be a lot of other changes after this.

rathboma:
@azmy60 what's the status of this PR, is it ready for review again? (only 3/4 tasks are marked as 'done')

azmy60:
Oh sorry, I forgot to tell you I couldn't find the solution to the 4th one yet. It is not from the autocomplete.

But it's ready for review if we want to ignore that for now 😆.

PR #1740, opened 2023-10-14
Fix: Leave aliases to be unquoted in autocomplete

azmy60:
Testing CodeMirror outside of bks just little bit. It turns out that this behavior was not from the CodeMirror itself😅 .

Fixes point 2 in #1721

### DIFF
--- a/apps/studio/src/lib/codemirror.ts
+++ b/apps/studio/src/lib/codemirror.ts
@@ -1,13 +1,52 @@
 import CodeMirror from "codemirror";
 
-CodeMirror.defineMIME("text/x-pgsql", {
+export function autoquoteHandler(
+  instance: CodeMirror.Editor,
+  changeObj: CodeMirror.EditorChangeCancellable
+) {
+  if (!instance.getOption("mode").includes("pgsql")) return;
+
   // eslint-disable-next-line
   // @ts-ignore
-  ...CodeMirror.resolveMode("text/x-pgsql"),
-  identifierQuote: '"',
-  hooks: {
+  const { to, from, origin, text } = changeObj;
+
+  // eslint-disable-next-line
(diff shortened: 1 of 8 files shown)


not-night-but (changes requested):
@azmy60 Nice! Glad we got this one figured out. 

Code looks good, but do you think you could add some regression tests for this?

azmy60:
Oh yeah, forgot that one. Let me see if I can do it.

apps/studio/src/lib/codemirror.ts
```diff
@@ -1,13 +1,52 @@
 import CodeMirror from "codemirror";
 
-CodeMirror.defineMIME("text/x-pgsql", {
+export function autoquoteHandler(
+  instance: CodeMirror.Editor,
+  changeObj: CodeMirror.EditorChangeCancellable
+) {
+  if (!instance.getOption("mode").includes("pgsql")) return;
```
rathboma, line 7: Quoting aliases isn't a problem in the other languages?

rathboma (commented):
Looks good, question: is this a problem in other dialects?

rathboma:
Ok for you to merge, but if also a problem in other dialects, they'll need sorting also. eg SQL Server has schemas + tables also

apps/studio/src/lib/codemirror.ts
rathboma: @azmy60 quick check-in for this question. If this isn't a problem, feel free to merge plz

rathboma (approved):
Feel free to merge this, but check my question first!

azmy60:
Redshift needs the autoquote too so that's added. Merging now :)

azmy60:
hmm.. I can't merge it somehow 🤔

PR #1745, opened 2023-10-21
Add comment to column tooltip if exists

p-Jimenez:
If the column has a comment, display it on the tooltip

image

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -614,6 +614,11 @@ export default Vue.extend({
           headerTooltip += ' [Primary Key]'
         }
 
+        // if column has a comment, add it to the tooltip
+        if (column.comment) {
+          headerTooltip += `<br/> ${column.comment}`
+        }
+
         const result = {
           title: column.columnName,
           field: column.columnName,


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -614,6 +614,11 @@ export default Vue.extend({
           headerTooltip += ' [Primary Key]'
         }
 
+        // if column has a comment, add it to the tooltip
+        if (column.comment) {
+          headerTooltip += `<br/> ${column.comment}`
```
rathboma, line 619: This needs to be made HTML safe, but otherwise, love this! So:

```
headerTooltip += `<br/> ${escapeHtml(column.comment)}`

```

Once you've made that change I will merge!

p-Jimenez:
Pushed already! Haven't thought about sanitizing HTML. [image](https://github.com/beekeeper-studio/beekeeper-studio/assets/55885320/ad3f6e76-75cc-4dfb-b80a-43e75433b4bd)

Also tried it with large comments, don't know if it should be truncated but I've actually found it way less intrussive than I thought it would be.

image

rathboma (approved):

rathboma:
Amazing, thanks!

PR #1760, opened 2023-10-26
Fix illegible connection type

### LINKED ISSUE #1756: BUG: Connection type dropdown has a PURE WHITE selection rectangle in dark mode

azmy60:
closes #1756 

A hack to hide the placeholder option because it's very tricky to style with CSS.

Animation4


Z3rio:
How does this look after the change?

Issue #1759, opened 2023-10-26
FEAT: Beautify and Format JSON Data Enhanced Readability


khad1r:
**Is your feature request related to a problem? Please describe.**
Yes, it is. When working with JSON data in MySQL using Beekeeper Studio, the data is displayed as a single line in a cell. This makes it difficult to read and understand the data structure, especially when dealing with complex JSON objects.

**Describe the solution you'd like**
I would like Beekeeper Studio to have a feature that formats or beautifies the JSON data in MySQL. Instead of showing the entire JSON data as a single line in a cell, it would be great if the data could be formatted in a readable structure with proper indentation.

**Describe alternatives you've considered**
An alternative could be to provide an option to view the formatted JSON data in a separate window or panel when clicked on the cell. This would avoid cluttering the main data view but still provide easy access to a more readable format of the JSON data.

Here's VSCode Database Client Example
![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/60246846/14126a5c-98c4-4b64-8d61-0114655325d7)
![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/60246846/61664625-a15d-4222-abad-d5d43a23b111)
![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/60246846/bf5f27ea-8820-4b51-8d88-8a88d02d615d)

**Additional context**
This feature would greatly improve the readability and accessibility of JSON data in MySQL, making it easier for users to understand and work with complex data structures.

**Database type, if appropriate**
MySQL/MariaDB databases.

Z3rio:
As much as I hate storing JSON in SQL, I do agree that this is a great idea.
I'll see if I have some free time to have a look at this.

khad1r:
Hi Z3rio, it’s great that you’re open to this idea
Can’t wait to see how it turns out!

Z3rio:
I've started working on this now, you guys can assign me to this issue if you wish :)

Issue #1698, opened 2023-09-10
FEAT: Rich-editing of large data in a modal


rathboma:
Problem:
It is common for a database table to contain fields of large values, for example the content of a blog post, a large json document, xml document, html page, css rules, etc, etc.

Editing these in Beekeeper Studio is terrible, because we only support in-line editing.

Solution:
We need a way to 'pop out' the editor for complex data in some way.
We've previously discussed a right-side sidebar, but this could also be a modal, or maybe even something like the gmail bottom-right pop up email composer window. We'd need to experiment with what feels natural.

Solutions we've spitballed in the past:
We've discussed having a right-side sidebar that displays the entire highlighted record for editing. This could be nice for other use cases, but doesn't intrinsically solve this use case, which is editing a single large value, in fact it could have many of the same problems.

Implementation notes:
- Beekeeper should try to auto-detect the type of content being edited, maybe using https://github.com/microsoft/vscode-languagedetection

not-night-but:
I think the Gmail composer is a good solution for this (we all know I despise modals for anything beyond a confirmation of action). In this case I think the modal could be frustrating, particularly if you're using other data in the app to edit the field. 

Another thing to consider with that is whether it's a popout that's local to the tab and will only show when that tab is active, or if it's global (and then can be minimised/expanded like how Gmail handles it)

rathboma:
Yeah I'm not sure honestly. I'll have to play with the different options.

RonnieSan:
PGAdmin does something similar. I would suggest something not unlike the [primevue overlay panel](https://primevue.org/overlaypanel/) which is basically an attached modal so it has a clear association with the exact field you are editing.

rathboma:
Something like that could work, but might still not be big enough.

What about a right sidebar?

This won't be in 4.0, but want to include it for 4.1

rathboma:
Maybe an overlay with an expand button that makes it a large modal? Still not sure. Probably need to try a few

RonnieSan:
> Maybe an overlay with an expand button that makes it a large modal? Still not sure. Probably need to try a few

Yeah, I think some kind of expand button would work great for larger objects. I think my main ask for this would be a keyboard shortcut for submitting whatever data is changed in the field (something like ctrl+enter). It's really time-consuming when you have to click a button if you have to update multiple fields.

rathboma:
Ok, I think for v1 of this:

[...]

PR #1769, opened 2023-11-04
Language specific cell editing

Z3rio:
As much as I hate having JSON in SQL databases, I did agree with this Issue/Suggestion.
This PR aims to make the experience of using JSON quite a lot better.

This PR also aims to make the implementation as universal as possible, so other languages, such as HTML, etc, can be added alongside JSON.

Most of what this aims to solve is already summarized in issue: #1759 
(This PR closes that issue)

This video also quickly summarizes the changes:
https://streamable.com/vk9env

Please note:
This is my first time working in this environment/repository, so please have a look through my code.


khad1r:
Hi Z3rio,
thank you for working on the feature i've issues for

I gave your build a spin and it’s coming along nicely! I did spot a small issue with the modal window when dealing with long JSON data. It looks like it could use an ‘overflow-y’ property. Just thought I’d share. Thanks for all your hard work!

here's the video about my finding

https://github.com/beekeeper-studio/beekeeper-studio/assets/60246846/3dabf0f3-a108-4394-b37f-ef6d4dd12ee7

Z3rio:
> Hi Z3rio, thank you for working on the feature i've issues for
> 
> I gave your build a spin and it’s coming along nicely! I did spot a small issue with the modal window when dealing with long JSON data. It looks like it could use an ‘overflow-y’ property. Just thought I’d share. Thanks for all your hard work!
> 
> here's the video about my finding
> 
>  Sibg.Live.-.Beekeeper.Studio.2023-11-06.12-35-22-1.mp4

Ah, my bad. I forgot to re-add that after moving it from using highlights to being an actual editor/textarea via codemirror.
I'll have a look at it, thank you for telling me.

not-night-but:
Hey @Z3rio !! Thank you so much for your contribution!! I'm just about to dive into the code here, but this is something we've been thinking about for a while (see #1698). If you're interested in expanding the functionality here to allow what was described [here](https://github.com/beekeeper-studio/beekeeper-studio/issues/1698) (ie editing long plaintext, xml, html, etc) that would be great, but we can always extend it later if you're not interested

Z3rio:
> Hey @Z3rio !! Thank you so much for your contribution!! I'm just about to dive into the code here, but this is something we've been thinking about for a while (see #1698). If you're interested in expanding the functionality here to allow what was described [here](https://github.com/beekeeper-studio/beekeeper-studio/issues/1698) (ie editing long plaintext, xml, html, etc) that would be great, but we can always extend it later if you're not interested

I'll see if I can have a look at at least setting up the structure to support multiple languages and such.

not-night-but:
That would be fantastic! I'll finish my review tomorrow

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -488,6 +565,44 @@ export default Vue.extend({
             this.copyCell(cell)
           }
         },
+        {
+          label: `<x-menuitem><x-label>Edit row as JSON</x-label></x-menuitem>`,
```
not-night-but, line 569: I really don't know what we should say here, something like 'Edit in Editor' but I don't like how that sounds :sweat_smile: The main thing is you are editing a cell, not a row, and I don't think it should be language specific here. The modal editor will ideally autodetect the type of content being edited.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -488,6 +565,44 @@ export default Vue.extend({
...
+              const row = cell.getRow()
+              const valueCell = row.getCell(column);
+              const value = valueCell.getValue();
+              let parsed: null | Record<string,unknown> = null
+
+              try {
+                parsed = JSON.parse(value)
+
+              } catch (e) {
+                console.log("Invalid JSON", e)
```
not-night-but, line 583: If you think it's important that this is logged, then use `log.debug` or `log.error` instead of `console.log`. That way it won't pollute the browser console. Otherwise just remove it

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -270,6 +327,12 @@ export default Vue.extend({
       actualTableHeight: "100%",
       loading: false,
 
+      // modal
+      modalName: "viewJsonModel",
```
not-night-but, line 331: There is a common issue with modals in our app, if we have multiple tabs open this could display the wrong content, save to the wrong tab, etc. Every data tab will have the same modal name and they can interfere with eachother. You can check TabQueryEditor.vue for the save modal and the parameter modal for how this was fixed. It's ugly but it works :joy:

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -806,6 +921,135 @@ export default Vue.extend({
...
+    },
+
+    saveCurrentJson() {
+      if (this.currentJsonCell) {
+        let parsed: Record<string, unknown> | null = null;
+
+        try {
+          parsed = JSON.parse(this.modalJsonContent)
+        } catch (e) {
+          console.log("Invalid JSON", e)
```
not-night-but, line 1006: Same deal as above with the other log

apps/studio/src/main.ts
```diff
@@ -22,6 +22,7 @@ import './assets/styles/app.scss'
 import $ from 'jquery'
 
 import 'codemirror/mode/sql/sql'
+import 'codemirror/mode/javascript/javascript'
```
not-night-but, line 25: Was this causing errors or something?

not-night-but (changes requested):
It's looking good!!! Most of my comments are very trivial. Wording, conventions that we often use in the app, etc. One thing I would suggest as you go about refactoring for multiple languages: it might be worth it to split this into its own Vue component like RichCellEditor.vue or something like that for code organisation/future maintenance.

Thanks again for such a great contribution! Can't wait to see what you do with it so we can get it merged!

apps/studio/src/components/tableview/TableTable.vue
rathboma: Yes Day is right, you should name this something like `view-json-modal-${tab.id}`

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -4,6 +4,52 @@
...
+      <modal
+        :name="modalName"
+        class="beekeeper-modal vue-dialog"
+        v-show="active"
+        @closed="onJsonModalClose"
+        @opened="onJsonModalOpen"
+      >
+        <div class="dialog-content">
+          <div class="dialog-c-title">
+            Viewing JSON
```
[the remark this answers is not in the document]
rathboma: NBD, we can *fix it in post*

apps/studio/src/components/tableview/TableTable.vue
rathboma: Hey! We can fix copy once it's merged. That's easy for us to do, mostly let's just check functionality, which looks like it's doing the right thing.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -488,6 +565,44 @@ export default Vue.extend({
...
+              const value = valueCell.getValue();
+              let parsed: null | Record<string,unknown> = null
+
+              try {
+                parsed = JSON.parse(value)
+
+              } catch (e) {
+                console.log("Invalid JSON", e)
+
+                const notification = new Noty({
```
rathboma, line 585: We have a helper for this!

`this.$noty.error("Unable to open in the JSON editor, are you sure this value is valid JSON?")`

apps/studio/src/components/tableview/TableTable.vue
rathboma: I agree with `log.error`

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -806,6 +921,135 @@ export default Vue.extend({
     }
   },
   methods: {
+    copyCurrentJson() {
+      this.$copyText(this.modalJsonContent)
+
+      const notification = new Noty({
```
rathboma, line 927: I think you get a notification automatically when using `copyText`

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -253,6 +299,17 @@ import { dialectFor, FormatterDialect } from '@shared/lib/dialects/models'
 import { format } from 'sql-formatter';
 import { normalizeFilters, safeSqlFormat } from '@/common/utils'
 import { TableFilter } from '@/lib/db/models';
+import Noty from 'noty'
```
rathboma, line 302: I think it'd be good to break out this code into a separate component, maybe call it `JsonEditorModal.vue` or something literal like that.

rathboma (changes requested):
Hey @Z3rio, this is AWESOME. Really really cool. I tried it out, and it makes a big difference.

A few things I think need to be fixed before merge:
1. Can you break out the code into a new component? Something like `JsonEditorModal.vue` or something like that? We do this with other, similar modals
2. The modal name will have to contain the `tab.id`, otherwise we'll get weird bugs.
3. When I was using it, when I clicked 'save' I got an error and the whole table went blank. I was only able to do this once, but here is the error I got:

Screenshot 2023-11-08 at 10 16 39 PM


Let me know if you don't have time to make these changes, and we'll try to find some time to take it over.

Matthew

apps/studio/src/main.ts
Z3rio: Based on my digging, they dont have a specific language for JSON, although you can use the JS mode to act like JSON.
If this isnt true, then please enlighten me, but that is atleast the info I could find online.

not-night-but: No I'm sure that's good! I just wasn't sure if it was just a stray import or not

apps/studio/src/components/tableview/TableTable.vue
Z3rio: I do not believe that to be the case, as that function is handled by an outide dependency (`vue-clipboard2`)

apps/studio/src/components/tableview/TableTable.vue
Z3rio: Agreed

apps/studio/src/components/tableview/TableTable.vue
Z3rio: Thank you, I'll use that instead then.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -305,6 +368,20 @@ export default Vue.extend({
...
+    userKeymap: {
+      get() {
+        const value = this.settings?.keymap?.value;
+        return value && this.keymapTypes.map(k => k.value).includes(value) ? value : 'default';
+      },
+      set(value) {
+        if (value === this.keymap || !this.keymapTypes.map(k => k.value).includes(value)) return;
+        this.$store.dispatch('settings/save', { key: 'keymap', value: value });
+        this.initialize();
+      }
```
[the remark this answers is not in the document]
Z3rio: Thank you

Z3rio:
I've had a look at all conversations now and resolved them.
I'll start having a look at making it more universal and such.

Z3rio:
Here is a new quick preview of it, for those who are interested but perhaps don't care enough to go on and build the files themselves.
https://streamable.com/hwucyz

The main changes between this and previously would be that it now supports multiple languages, is componentized, hopefully more polished and I overall just tried to add everything I would personally want.

rathboma:
Hey, this looks awesome! I'll get it reviewed this week. Thanks so much for spending your time on open source  :-)

rathboma:
Hey! I'm going to merge this into an internal branch for us to work on a little. Not going to make many changes, just a few things here and there.

Thanks so much for the great feature!

PR #1771, opened 2023-11-06
Use virtual list when listing tables

### LINKED ISSUE #479: BUG: Large memory usage when caching a large number of tables causing crashing
**Description**
I have a Redshift based DB that has upwards of 18,000 tables (yes, 18K). When I open a connection to it, Beekeeper tries to load the entire tableset to memory, taking up well over 2.5GB of RAM on my system (8GB RAM total). While having auto-complete functionality is invaluable, having a client that doesn't crash after running a few queries is even more important.

**To Reproduce**
1. Open a Redshift connecting that has several hundred tables.
2. Watch RAM fill up in task manager, along with high CPU usage.

**Expected behavior**
Ideally, to avoid hogging up such a large amount of memory and keeping in line with Beekeeper's easy-to-use principle, I would expect some sort of local caching that can either get automatically updated in the background after a first time setup, and then get updated on the fly with no user interaction.

 - OS: Windows 10
 - App Version 1.8.11, portable version.

**Additional context**
Not an issue for other databases on the same server that have fewer tables.

### LINKED ISSUE #748: BUG: 10k tables crashes client (Redshift + SSL)
**Describe the bug**
I installed 1.13.2 on Windows 10, connected it to my Redshift database over mandatory SSL and, and the side bar said it was loading all 10k+ tables and crashed.

**To Reproduce**
A fresh install of 1.13.2 on Windows 10 and connect to a cluster that has 10,000+ objects spread across 100+ schemas and consumes a petabyte of storage.

**Expected behavior**
The interface to not crash and become unusable just after initial connection

**Screenshots / Gifs**

 - OS: Windows 10  Build 19042
 - App Version 1.13.2

**Additional context**
Fresh install and other SQL clients do not have an issue connecting. I also believe they don't index anything until you open a particular schema.

azmy60:
closes #748
closes #479

### DIFF
--- a/apps/studio/src/components/sidebar/core/HiddenEntitiesModal.vue
+++ b/apps/studio/src/components/sidebar/core/HiddenEntitiesModal.vue
@@ -121,6 +121,7 @@
 
 <script lang="ts">
   import TableIcon from '@/components/common/TableIcon.vue'
+  import { AppEvent } from "@/common/AppEvent"
 
   export default {
     props: ['hiddenEntities', 'hiddenSchemas'],
@@ -146,13 +147,13 @@
         this.entities = [...this.hiddenEntities]
         this.schemas = [...this.hiddenSchemas]
       },
-      unhideSchema(index: number) { 
+      unhideSchema(index: number) {
         const [schema] = this.schemas.splice(index, 1)
-        this.$store.dispatch('hideEntities/removeSchema', schema)
+        this.trigger(AppEvent.toggleHideSchema, schema, false)
       },
       unhideEntity(index: number) {
         const [entity] = this.entities.splice(index, 1)
-        this.$store.dispatch('hideEntities/removeEntity', entity)
+        this.trigger(AppEvent.toggleHideEntity, entity, false)
       },
       closeModal() {
         this.$modal.hide(this.modalName)
--- a/apps/studio/src/components/sidebar/core/table_list/Item.vue
+++ b/apps/studio/src/components/sidebar/core/table_list/Item.vue
@@ -20,6 +20,7 @@
     :expanded="source.expanded"
     :table="source.entity"
     :pinned="source.pinned"
+    :loadingColumns="source.loadingColumns"
     @expand="onExpand($event, source)"
     @pin="onPin($event, source)"
     @contextmenu.prevent.stop="
--- a/apps/studio/src/components/sidebar/core/table_list/Item.vue
+++ b/apps/studio/src/components/sidebar/core/table_list/Item.vue
@@ -15,6 +15,7 @@
   />
   <stateless-table-list-item
     v-else-if="source.type === 'table'"
+    :id="source.key"
     :level="source.level"
     :expanded="source.expanded"
     :table="source.entity"
--- a/apps/studio/src/components/sidebar/core/table_list/Item.vue
+++ b/apps/studio/src/components/sidebar/core/table_list/Item.vue
@@ -1,0 +1,63 @@
+<template>
+  <stateless-sidebar-folder
+    v-if="source.type === 'schema'"
+    v-show="!source.hidden"
+    :expanded="source.expanded"
+    :schema="source.entity"
+    @expand="onExpand($event, source)"
+    @contextmenu.prevent.stop="
+      $bks.openMenu({
+        event: $event,
+        item: source.entity,
+        options: source.contextMenu,
+      })
+    "
+  />
+  <stateless-table-list-item
+    v-else-if="source.type === 'table'"
+    :level="source.level"
+    :expanded="source.expanded"
+    :table="source.entity"
+    :pinned="source.pinned"
+    @expand="onExpand($event, source)"
+    @pin="onPin($event, source)"
+    @contextmenu.prevent.stop="
+      $bks.openMenu({
+        event: $event,
+        item: source.entity,
+        options: source.contextMenu,
+      })
+    "
+  />
+  <stateless-routine-list-item
+    v-else-if="source.type === 'routine'"
+    :level="source.level"
+    :expanded="source.expanded"
+    :routine="source.entity"
+    :pinned="source.pinned"
+    @expand="onExpand($event, source)"
+    @pin="onPin($event, source)"
+    @contextmenu.prevent.stop="
+      $bks.openMenu({
+        event: $event,
+        item: source.entity,
+        options: source.contextMenu,
+      })
+    "
+  />
+</template>
+
+<script lang="ts">
+import Vue from "vue";
+import StatelessTableListItem from "./StatelessTableListItem.vue";
+import StatelessRoutineListItem from "./StatelessRoutineListItem.vue";
+import StatelessSidebarFolder from "@/components/common/StatelessSidebarFolder.vue";
+export default Vue.extend({
+  components: {
+    StatelessTableListItem,
+    StatelessRoutineListItem,
+    StatelessSidebarFolder,
+  },
+  props: ["source", "onExpand", "onPin"],
+});
+</script>
--- a/apps/studio/src/components/sidebar/core/table_list/StatelessTableListItem.vue
+++ b/apps/studio/src/components/sidebar/core/table_list/StatelessTableListItem.vue
@@ -46,7 +46,7 @@
       </span>
     </a>
     <div v-if="expanded" class="sub-items">
-      <span class="sub-item" v-if="!columnsLoading && !table.columns?.length">
+      <span class="sub-item" v-if="!loadingColumns && !table.columns?.length">
         No Columns
       </span>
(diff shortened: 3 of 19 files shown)


apps/studio/src/components/sidebar/core/HiddenEntitiesModal.vue
```diff
@@ -146,13 +147,13 @@
         this.entities = [...this.hiddenEntities]
         this.schemas = [...this.hiddenSchemas]
       },
-      unhideSchema(index: number) { 
+      unhideSchema(index: number) {
         const [schema] = this.schemas.splice(index, 1)
-        this.$store.dispatch('hideEntities/removeSchema', schema)
+        this.trigger(AppEvent.toggleHideSchema, schema, false)
```
rathboma, line 152: Why fire an app event rather than just dispatching to the store?

apps/studio/src/components/sidebar/core/table_list/StatelessTableListItem.vue
```diff
@@ -0,0 +1,176 @@
+<template>
```
rathboma, line 1: Question for my understanding - what makes these 'stateless' vs the original components?

Could you add a nice comment to this file outlining why we need a 'stateless' table list item, maybe linking to the virtual scroll library and such?

apps/studio/src/components/sidebar/core/table_list/TableListItem.vue
```diff
@@ -252,11 +252,10 @@ export default {
       this.$root.$emit(AppEvent.openTableProperties, { table: this.table })
     },
     pin() {
-      this.$store.dispatch('pins/add', this.table)
-      this.$store.dispatch('updateTableColumns', this.table)
+      this.trigger(AppEvent.togglePinTableList, this.table, true);
     },
     unpin() {
-      this.$store.dispatch('pins/remove', this.table)
+      this.trigger(AppEvent.togglePinTableList, this.table, false);
```
rathboma, line 258: Why do we trigger this vs just dispatching to the store directly?

apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
```diff
@@ -0,0 +1,195 @@
...
+}
+
+interface TableItem extends BaseItem {
+  type: "table" | "routine";
+  entity: TableOrView | Routine;
+  parent: SchemaItem;
+  pinned: boolean;
+}
+
+export default Vue.extend({
```
rathboma, line 51: This looks awesome, good stuff! Can you add a nice verbose comment to this file explaining why we need it?

Pretend someone new is starting on the team in 12 months, and they're reading the code for the first time -- what information is useful for them to know?

apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
```diff
@@ -0,0 +1,195 @@
...
+          hidden: this.hiddenSchemas.includes(schema.schema),
+          contextMenu: this.schemaMenuOptions,
+          skipDisplay: noFolder,
+          level: 0,
+        };
+        items.push(schemaItem);
+        schema.tables.forEach((table: TableOrView) => {
+          items.push({
+            type: "table",
+            key: `${schema.schema}.${table.name}.table`,
```
rathboma, line 82: Feels like we should have an `entityId` helper in `utils` so we could call `entityId(schema, table, 'table')`

apps/studio/src/components/sidebar/core/table_list/Item.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+  <stateless-sidebar-folder
+    v-if="source.type === 'schema'"
+    v-show="!source.hidden"
+    :expanded="source.expanded"
```
rathboma, line 5: It's not in this PR, but does this new system properly hook up with column lazy loading? Somewhere presumably we're dispatching `loadTableColumns` to the store?

rathboma (changes requested):
This looks fantastic! Great work. It's really fast and responsive!

Some issues that need fixing before this is merged:

- Columns don't load in when you expand a table. Looks like this isn't hooked up to `this.$store.dispatch('loadTableColumns', this.table)`. This should be as easy as adding a `watch` to `expanded` in the StatelessTableListItem component maybe?
- Refreshing the table list resets all the expanded tables, and flashes the whole view. Does the view need totally refreshing here?
- Are the expanded table list items responsive if new columns get added in the store? Will they update? Just checking
- All schemas are open by default for PostgreSQL. Can we just have the default schema (Defined per connection-type) open?
- I can no longer select a table in the sidebar. If I click it selects the whole table list and there's no way to unselect it!


https://github.com/beekeeper-studio/beekeeper-studio/assets/279769/f62d4f29-bb38-4610-9fdf-8628445f8dfc

apps/studio/src/components/sidebar/core/HiddenEntitiesModal.vue
azmy60: I just thought we could re-use of what we had in the `AppEvent.ts`, because we used to have `hideSchema` Event which now I change it to `toggleHideSchema`. (We can hide or show a schema with `toggleHideSchema` by passing a boolean)

In [VirtualTableList.vue](https://github.com/beekeeper-studio/beekeeper-studio/pull/1771/files#diff-095b36270d569f0d9e05acb75359243770919245870dddd4249dc5ba0a725296R134-R142), it's easier to track which item should I hide/show by this event because we know which schema it is immediately. The same thing with `togglePinTableList`.

apps/studio/src/components/sidebar/core/table_list/StatelessTableListItem.vue
azmy60: Will do! :)

azmy60: The original components have states that vanish when it's out of sight or "unmounted" from the virtual list. These states are `expanded`, `pinned`, `draggable` (though I'm not sure we still use draggable?) .

apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
azmy60: Will do :)

apps/studio/src/components/sidebar/core/table_list/Item.vue
azmy60: Oh, I don't think so. Let me check.

apps/studio/src/components/sidebar/core/table_list/TableListItem.vue
azmy60: Ideally, it's the same as `toggleHideSchema`. I wanted to have an easier tracking which item is pinned or not. We could use a vue `watch` but I feel like we'd need to compare two arrays and then determine which item is missing from the other. With an event, it's known which item immediately. In anyway, I'm very open if you'd want to have a different approach. Just let me know :)

rathboma: Ok, thats fine!

apps/studio/src/components/sidebar/core/table_list/StatelessTableListItem.vue
rathboma: Got it. That's some really useful information! Please def add as a comment

apps/studio/src/components/sidebar/core/HiddenEntitiesModal.vue
rathboma: kk

rathboma (approved):
This is 99% finished :tada:. Great stuff.

Still broken:
- Expand all, then go to collapse a specific table, this doesn't work on first click (presumably because the flag is false), second click collapses the table
- Can we add some negative margin to the progress bar so it doesn't cause the table list to shift down when it is displayed?

These are both super minor. Looks fantastic, A+

rathboma:
Oh wait, I found another one -- if you `expand all` it doesn't load table columns for the expanded tables and views

azmy60:
Yes, that's an existing bug. I guess I'm going to load table columns that are exists on view first, and then the rest if the user scrolls.

azmy60:
The scrollbar is still not very accurate if you expand all. It's super minor tho. Do you want this to be fixed too in this PR? Not sure how long will it take, cause I'll prob need to experiment with some maths lol.

rathboma:
I was going to say not to worry about it, but you already fixed it :-)

azmy60:
Something is broken here. If you make an empty new database, and then add a new table. The table is not immediately added to the list. Refreshing the table list also doesn't do it.

I'm gonna make the PR later today.

PR #1773, opened 2023-11-08
Fix: Syncing table columns after updates

### LINKED ISSUE #1751: BUG: Adding a column to a table, then clicking apply doesn't show the column

### LINKED ISSUE #1752: BUG: If I add a column to a table, but have the table data view open in another tab, that doesn't update to show the new table

azmy60:
closes #1751 
closes #1752

### DIFF
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -206,6 +206,11 @@ import TabIcon from './tab/TabIcon.vue'
 import { DatabaseEntity } from "@/lib/db/models"
 import PendingChangesButton from './common/PendingChangesButton.vue'
 import { safeSqlFormat as safeFormat } from '@/common/utils';
+import { AlterTableSpec } from "@shared/lib/dialects/models";
+import { TableOrView } from "@/lib/db/models";
+import rawLog from 'electron-log'
+
+const log = rawLog.scope('CoreTabs')
 
 export default Vue.extend({
   props: ['connection'],
@@ -284,6 +289,7 @@ export default Vue.extend({
         { event: AppEvent.createTable, handler: this.openTableBuilder },
         { event: 'historyClick', handler: this.createQueryFromItem },
         { event: AppEvent.loadTable, handler: this.openTable },
+        { event: AppEvent.alterTable, handler: this.alterTable },
         { event: AppEvent.openTableProperties, handler: this.openTableProperties },
         { event: 'loadSettings', handler: this.openSettings },
         { event: 'loadTableCreate', handler: this.loadTableCreate },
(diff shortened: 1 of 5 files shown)


apps/studio/src/components/CoreTabs.vue
```diff
@@ -593,6 +599,17 @@ export default Vue.extend({
...
     },
+    async alterTable(options: { changes: AlterTableSpec, table: TableOrView, onError?: (e: any) => void, onSuccess?: () => void }) {
+      try {
+        await this.connection.alterTable(options.changes)
+        await this.$store.dispatch('updateTableColumns', options.table)
+        options.onSuccess?.()
+        this.trigger(AppEvent.tableAltered, options.table)
+      } catch (e) {
+        log.error(e);
+        options.onError?.(e)
```
rathboma, line 610: It's not a good idea to pass functions around on the event bus. It's really just for data.

I don't really see the reason for `alterTable` to be pushed up to CoreTabs, why can't this stay in `TableSchema`?

apps/studio/src/components/tableinfo/TableSchema.vue
```diff
@@ -366,23 +356,21 @@ export default Vue.extend({
...
-        this.error = null
-        const changes = this.collectChanges()
-        await this.connection.alterTable(changes)
-
-        this.clearChanges()
-        await this.$store.dispatch('updateTableColumns', this.table)
-        this.$nextTick(() => {
+    async submitApply() {
+      this.error = null
+      this.trigger(AppEvent.alterTable, {
```
rathboma, line 361: Why did you move the alterTable function to CoreTabs? I don't see why it can't stay in here?

We should not be passing functions through the event bus

apps/studio/src/components/CoreTabs.vue
```diff
@@ -593,6 +599,17 @@ export default Vue.extend({
       this.tableExportOptions = options
       this.showExportModal = true
     },
+    async alterTable(options: { changes: AlterTableSpec, table: TableOrView, onError?: (e: any) => void, onSuccess?: () => void }) {
+      try {
+        await this.connection.alterTable(options.changes)
+        await this.$store.dispatch('updateTableColumns', options.table)
```
rathboma, line 605: won't updating the table columns just force an update to the TableTable without an extra event? I think there's a watcher on the columns already no?

rathboma (changes requested):
Bunch of questions, but notably -- let's not pass functions on the event bus!

apps/studio/src/components/CoreTabs.vue
azmy60: Yeah but for some reason it didn't detect the changes automatically, even after calling the updateColumnsTable

azmy60: Wait, let me find why the watcher didn't work.

azmy60:
Ok, fixed the root cause!

rathboma (approved):

Issue #1761, opened 2023-10-27
FEAT:  Omit auto-increment columns when copying/exporting row(s) as SQL.


ningzio:
**Is your feature request related to a problem? Please describe.**
when copying rows as an INSERT statement, it always contains `id` which is auto-increment. I have to delete this key the corresponding values in the VALUES.

**Describe the solution you'd like**
It's nice to omit all auto-increment columns when copying/exporting row(s) as SQL

**Describe alternatives you've considered**
as alternative is to add an option when we copy rows as SQL, let users decide if beekeeper should omit auto-increment columns.

**Additional context**
![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/45959659/e4e3a59c-8a98-46c6-8deb-b59c5a7a67bc)
when I copy as SQL, i got
```sql
insert into `test` (`col1`, `id`) values ('test', 1)
```
what I want is 
```sql
insert into `test` (`col1`) values ('test')
```

**Database type, if appropriate**

Z3rio:
Looks fair, I'll see if I can perhaps have a look at this sometime.

rathboma:
Ideas for implementation:
- include the auto-increment column, but have it commented out (Same with the value). This would be the best

If that is too hard, just exclude it and add a comment that this did not include the auto-increment column

PR #1778, opened 2023-11-11
Omit auto-increment columns when copying/exporting row(s) as SQL

Z3rio:
This PR aims to resolve issue: #1761
Most of what this changes is explained in that issue.


rathboma:
Hey! Would this deal with all types of auto increment columns across all the different database vendors, or just those that are explicitly `auto_increment`? For example `SERIAL` is used

Z3rio:
> Hey! Would this deal with all types of auto increment columns across all the different database vendors, or just those that are explicitly `auto_increment`? For example `SERIAL` is used

I'll have to test this further to be able to answer that, I've to be honest only tried it with normal MySQL at the moment.
The aim of it would probably be to support all database types.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -453,13 +453,12 @@ export default Vue.extend({
...
-
             const fixed = this.getCleanSelectedRowData(cell)
 
             const tableInsert = {
               table: this.table.name,
               schema: this.table.schema,
-              data: fixed,
+              data: this.parseSqlCopyData(fixed),
             }
             const query = await this.connection.getInsertQuery(tableInsert)
```
rathboma, line 463: Probably this logic should be moved to `getInsertQuery` so it can be database-specific. Maybe an additional parameter like `stripAutoIncrementColumns` or something

rathboma:
@Z3rio let me know if you have time to work on this change to make it multi-database capable, even if you just implement it for MySQL, moving the logic to the driver makes a lot of sense I think.

Z3rio:
> Hey! Would this deal with all types of auto increment columns across all the different database vendors, or just those that are explicitly `auto_increment`? For example `SERIAL` is used

Serial is mapped to `BIGINT UNSIGNED NOT NULL AUTO_INCREMENT UNIQUE`, so that would be included by this as well

I will be having a look at adding support for other SQL wrappers such as psql

Personally, I cant come up with any more MariaDB fields that might want to be ignored, if you do come up with any, then please tell me about them.

rathboma:
Hey! I think we need to add better metadata to columns to do this in a way that won't break. Labeling a column as `autoIncrement` at the driver level seems like the best way to do that.

I really appreciate you working on this, and we'll use your code as a template in the future, but I think until we add more metadata it probably doesn't make sense to merge it. We've hit this issue in other places too, so we have to address it soon.

Issue #1505, opened 2023-02-17
Date/datetime/time picker in TableTable for datetime, date, and time columns


rathboma:

rathboma:
- Date
- Time
- Date Time
- Timezone (should default to UTC probably)

rathboma:
Worth note that some columns won't let TZ be stored, so it'll need converting to UTC probably (and we'll need to tell users that)

rathboma:
We had to revert this due to several issues, but let's aim to do-over and get this in 5.3

PR #1779, opened 2023-11-11
Feature/data picker

wmontgomery:
[#1505](https://github.com/beekeeper-studio/beekeeper-studio/issues/1505)
datepicker


rathboma:
Hey Will, haven't had time to fully review yet, few questions first:

- Does it behave appropriately if the row is at the bottom of the window?
- Can you make it so that hitting `esc` cancels out of it, and moves back to text input?
- Can you have it highlight the 'current' datetime when first opening (in a different color maybe), so folks know what is currently selected?

wmontgomery:
> * Does it behave appropriately if the row is at the bottom of the window?
> 
> * Can you make it so that hitting `esc` cancels out of it, and moves back to text input?
> 
> * Can you have it highlight the 'current' datetime when first opening (in a different color maybe), so folks know what is currently selected?

* Haven't checked but will do so. I would assume it just hovers over but will verify
* I'll see what I can do. I don't remember exactly what kind of controls there were but should be able to find something
* Crap, forgot about that. Yeah can do that. I think it was one of those "I'll get back to this" and just never did

shared/src/components/tabulator/NullableInputEditor.vue
```diff
@@ -1,6 +1,20 @@
 <template>
   <div>
+    <date-picker
```
rathboma, line 3: I don't think we want to pollute the `NullableInputEditor` with the date picker.

Can we do this in a different way:

1. Add a named slot to the `NullableInputEditor` for the custom button
2. Pass in the calendar icon button into the slot
3. Then when that button is clicked it can show the date modal, this can then live a whole level UP from the nullableinputeditor.


Something like this: `DatePickerEditor.vue`

```vue
<NullableInputEditor ...>
  <template v-slot:buttons>
    <DatePicker ... />
  </template>
</NullableInputEditor

```

shared/src/components/tabulator/NullableInputEditor.vue
```diff
@@ -10,27 +24,47 @@
...
     }
   },
   computed: {
+    typeEditorTitle() {
+      if (this.typeEditorActive) return 'Use standard input'
+      if (helpers.isDateTime(this.params.dataType)) return 'Open date/time picker' 
+      return ''
+    },
+    typeEditorIcon() {
+      if (this.typeEditorActive) return 'edit'
```
rathboma, line 64: Yep, then all this stuff can move out of the NullableInputEditor and into a DatePickerEditor

yarn.lock
```diff
@@ -15994,6 +15999,13 @@ vue-template-es2015-compiler@^1.6.0, vue-template-es2015-compiler@^1.9.0:
   resolved "https://registry.yarnpkg.com/vue-template-es2015-compiler/-/vue-template-es2015-compiler-1.9.1.tgz#1ee3bc9a16ecbf5118be334bb15f9c46f82f5825"
   integrity sha512-4gDntzrifFnCEvyoO8PqyJDmguXgVPxKiIxrBKjIowvL9l+N66196+72XVYR8BBf1Uv1Fgt3bGevJ+sEmxfZzw==
 
+vue2-datepicker@^3.11.1:
```
rathboma, line 16002: This looks pretty abandoned.... plus it will break with vue3. Are there any javascript-only implementations?

apps/studio/src/assets/styles/app/vendor/datepicker/var.scss
```diff
@@ -0,0 +1,35 @@
+$namespace: 'mx' !default;
+
+$default-color: #73879c !default;
+$primary-color: #1284e7 !default;
```
rathboma, line 4: Does this all work in light mode, solarized, and solarized dark?

rathboma (changes requested):
Great work! It'll be cool to have this in the app.

Questions:
- vue2-datepicker looks abandoned. Are there any pure-js alternatives?
- Let's invert how we do the component - so we don't materially change the `NullableInputEditor`.

apps/studio/src/assets/styles/app/vendor/datepicker/var.scss
wmontgomery: I tried it in light and dark mode and it looked pretty good.

yarn.lock
wmontgomery: I can look for a pure JS version, was hoping to use a vue specific one but can look into another one. Had a couple people recommend this one to me.

shared/src/components/tabulator/NullableInputEditor.vue
wmontgomery: Yeah that was all part of the plan for a later PR. Was taking this approach as a get this out there and see how people like it first. I'll make the changes

yarn.lock
Z3rio: Have you had a look at Vuetify's date pickers?
Might be nice, and you should be able to only use the date pickers, so it wont mess up the bundle size.

shared/src/components/tabulator/DateTimePickerEditor.vue
```diff
@@ -0,0 +1,218 @@
...
+      :clearable="false"
+      :confirm="true"
+      :open.sync="typeEditorActive"
+      ref="datepicker"
+      v-model="datePickerValue"
+      prefix-class="bkdates"
+      confirm-text="click to confirm new selection"
+      @confirm="submit"
+    />
+    <input
```
rathboma, line 15: So we don't want to re-implement the whole nullable editor inside the DateTimePickerEditor.

Have you used Vue's slots before?  We should add a slot to `NullableInput`, like this:

```vue
<div>
    <input
      class="nullible-input"
      :placeholder="smartPlaceholder"
      ref="input"
      type="text"
      v-model="value"
      @blur.prevent="submit"
      @change.prevent="submit"
      @keydown="keydown"
    >
    <span class="nullable-input-buttons">
      <slot name="buttons" />
      <i
        class="material-icons clear"
        @mousedown.prevent.stop="clear"
        title="Nullify Value"
      >cancel</i>
    </span>
  </div>
```

Then your datetime picker can look like this:
```vue
<template>
  <NullableInputEditor>
    <template v-slot:buttons>
      <i  class="material-icons" @click.prevent="pickDate">calendar</i>
    </template>
  </NullableInputEditor>
</template>
```

rathboma (changes requested):
Def should not be re-implementing the nullableinputeditor, let's just add a slot for the button!

shared/src/components/tabulator/DateTimePickerEditor.vue
wmontgomery: Slot has been added to handle this. Should be decently extendable.

rathboma:
@wmontgomery You'll need to merge master into this

shared/src/components/tabulator/DateTimePickerEditor.vue
```diff
@@ -0,0 +1,135 @@
...
+    },
+    typeEditorActive() {
+      if (this.typeEditorActive) {
+        const refValue = this.cell.getValue()
+        const dataType = this.params.dataType || ''
+        let dataValue = refValue == null ? refValue : helpers.niceString(refValue)
+
+        if (this.isTimeType(dataType) && dataValue !== null) {
+          dataValue = dataValue.search(/(\+|-)/i) > -1 && !isNaN(dataValue.slice(-1)) ? `${dataValue}:00`: dataValue  
+          this.datePickerValue = new Date(`2023-03-31T${dataValue}`)
```
rathboma, line 108: Where did this date come from? Should this be today's date?

rathboma: If not, or it doesn't matter, please add a comment or something

shared/src/components/tabulator/NullableInputEditor.vue
```diff
@@ -83,13 +86,16 @@ export default Vue.extend({
...
+      // nullifyInput is listened to by any slots which will then nullify the value and go through the whole submission process. Not having the below line caused sadness
+      this.$emit('nullifyInput')
       this.$emit('value', null)
     }
   },
   watch: {
     rendered() {
       if (this.rendered) {
-        this.value = helpers.niceString(this.cell.getValue())
+        this.value = this.cell.getValue() == null ? null : helpers.niceString(this.cell.getValue())
```
rathboma, line 98: Maybe use `_.isNil(this.cell.getValue())`

shared/src/lib/tabulator.ts
```diff
@@ -50,6 +50,9 @@ export default {
     const result = `<pre>${cellValue}</pre>`
     return result;
   },
+  isDateTime(dataType: string|null) {
```
rathboma, line 53: I think this is fine for now, but I think we probably need to add `bksType` to our `ColumnDefinition` interface at some point to identify stuff like this at the driver level, rather than in the GUI

rathboma (approved):
Looks good overall!

Had a couple of minor issues, please look at before merging!

shared/src/components/tabulator/DateTimePickerEditor.vue
wmontgomery: It shouldn't matter what the date is since we are really just creating a date object to grab the time, it was a Parks and Recreations reference. There are no meetings on Mar 31st because it's not a real day.

PR #1784, opened 2023-11-15
fix: get columns for a table when setting the foreign keys

### LINKED ISSUE #1750: BUG: Relations drop downs do not populate when a table's columns are not yet loaded in the vue store

wmontgomery:
[fix #1750](https://github.com/beekeeper-studio/beekeeper-studio/issues/1750) 

bugfix-1750


apps/studio/src/components/tableinfo/TableRelations.vue
```diff
@@ -219,7 +219,27 @@ export default Vue.extend({
...
+          cellEdited: async (cell) => { 
+            const data = cell.getRow().getData()
+            const schema = data['toSchema']
+            const table = data['toTable']
+            let tableData
+            
+            if (table !== null) {
+              if (!schema) {
+                tableData = this.tables.find((t: TableOrView) => escapeHtml(t.name) === table)
+              } else {
```
rathboma, line 231: You should just add the async await to the `getColumns` method. Tabulator supports the `valuesLookup` parameter being a promise. See https://github.com/olifolkerd/tabulator/blob/0675efd4b7532d91415b9b5ac3bd11b602ba57cf/src/js/modules/Edit/List.js#L525

it calls `valuesLookup` and if the result is a promise, it awaits.

rathboma (changes requested):
we should just await in the `getColumns` method.

apps/studio/src/components/tableinfo/TableRelations.vue
rathboma: @wmontgomery if you make this change we can merge this.

wmontgomery: Seeing this now, will be making the change today

PR #1785, opened 2023-11-16
Replace deprecated apt-key w/signed-by install instructions

jc00ke:
`apt-key` is deprecated, so it's advised to install the GPG key and explicitly set the `signed-by` attribute.

I'm not 100% sure if this docs change should go here or on the DuckDocs platform. I'm not familiar with DuckDocs (looks great!) so if this needs to go elsewhere, I'd be happy to help make that happen.

### DIFF
--- a/apps/studio/docs/installation/README.md
+++ b/apps/studio/docs/installation/README.md
@@ -29,16 +29,12 @@ A repo is provided for Debian and Ubuntu 16.04+. To install the repo:
 
 
 ```bash
-# Install our GPG key
-wget --quiet -O - https://deb.beekeeperstudio.io/beekeeper.key | sudo apt-key add -
+curl -fsSL https://deb.beekeeperstudio.io/beekeeper.key | sudo gpg --dearmor --output /usr/share/keyrings/beekeeper.gpg \
+  && sudo chmod go+r /usr/share/keyrings/beekeeper.gpg \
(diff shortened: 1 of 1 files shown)


rathboma:
Hey! Thanks for doing the work to figure this out. Honestly futzing with this stuff really worries me because I could break updating for so many people (100s of thousands of installs!)

I didn't realize initially that you're only updating the readme, we also do this automatically in the [deb-postinstall script](https://github.com/beekeeper-studio/beekeeper-studio/blob/c51e75f6cef46ec0cffd8e2747e02bdff0f87dd0/apps/studio/build/deb-postinstall#L4). Would you feel comfortable updating that script also?

Questions:
- For systems that have gpg1 vs gpg2, will this work the same?
- If you added this to the deb-postinstall script, would you also be able to add a way for it to remove the existing list file?

jc00ke:
Hey @rathboma sorry I didn't get a chance to address your questions last week.

I can tackle updating the deb-postinstall script. As for gpg1... it's a relatively basic command so I would assume it would work, but I don't know for sure.

By old file, do you mean [OLD_FILE](https://github.com/beekeeper-studio/beekeeper-studio/blob/c51e75f6cef46ec0cffd8e2747e02bdff0f87dd0/apps/studio/build/deb-postinstall#L3C1-L3C1) or the [NEW_FILE](https://github.com/beekeeper-studio/beekeeper-studio/blob/c51e75f6cef46ec0cffd8e2747e02bdff0f87dd0/apps/studio/build/deb-postinstall#L4) that doesn't have the `signed-by` attribute?

Issue #1737, opened 2023-10-12
FEAT: CTRL + P multiple select


mattnSwimd:
**Is your feature request related to a problem? Please describe.**
It is not at all a problem. I would like to request the ability to select multiple tables when using the CTRL + P menu to select tables.

**Describe the solution you'd like**
I do not know how complicated it would be to implement, but the right arrow key opening a table without closing the CTRL + P menu would be perfect. If you need more reference, using the CTRL + P shortcut in vscode and use the right arrow on multiple files in the list. 

**Describe alternatives you've considered**
Potentially, maybe a modifier like holding shift to once the menu is open and pressing enter on the tables you want, then upon releasing shift it would open all selected tables.

**Additional context**
This exact feature is in vscode currently for opening files and it is fantastic, If you haven't used it 10/10 I would highly recommend.

**Database type, if appropriate**
I use mariaDB and MySQL mainly, and am unsure how the other databases function in beekeeper, but I'd imagine it would apply to all of them.

rathboma:
Huh, I had no idea you could do that in VSCode, neat! Yeah we can probably have the exact same behavior

rathboma:
Note for implementation, pressing right arrow on an item in the list opens it in a tab in the background. Ctrl+right arrow should work the same as ctrl+enter

mattnSwimd:
The feature is awesome, thanks!

PR #1794, opened 2023-11-19
feat: add multi select to quick search

Christof-P:
Attempt to help with feat #1737 

Any feedback is welcome

### DIFF
--- a/apps/studio/src/components/quicksearch/QuickSearch.vue
+++ b/apps/studio/src/components/quicksearch/QuickSearch.vue
@@ -46,6 +46,24 @@
           </div>
           <span class="hint">(tables only)</span>
         </div>
+        <div class="shortcut-item">
+          <div>Open In Background</div>
+          <div class="shortcut">
+            <!-- <span v-if="this.$config.isMac">Cmd</span>
+            <span v-if="!this.$config.isMac">Ctrl</span> -->
+            <span>Right Arrow</span>
+          </div>
+          <span class="hint">(tables only)</span>
+        </div>
+        <div class="shortcut-item">
+          <div>Alt In Background</div>
+          <div class="shortcut">
+            <span v-if="this.$config.isMac">Cmd</span>
+            <span v-if="!this.$config.isMac">Ctrl</span>
+            <span>Right Arrow</span>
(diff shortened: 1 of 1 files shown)


apps/studio/src/components/quicksearch/QuickSearch.vue
```diff
@@ -146,6 +146,7 @@ export default Vue.extend({
       // /me *evil laugh*
       result['ctrl+p'] = this.selectUp
       result['ctrl+n'] = this.selectDown
+      result['right'] = this.persistentSearchEnter
```
rathboma, line 149: how do we signal to users that pressing right will keep the quick search open?

rathboma (commented):
This is great. Happy to merge it. Quick question -- how can we advertise this feature?

apps/studio/src/components/quicksearch/QuickSearch.vue
Christof-P: Good question, I hadn't really thought about it. VS code doesn't really show this well, so I couldn't really get inspiration from there. 
We could add it to the initial quick search popup (I'll add a picture below of what that could look like), but I don't know if that is the right place to do this.
Screenshot from 2023-11-30 09-49-28

rathboma:
I checked VSCode, it looks like it does your first suggestion - detect if at the end of the input, and only open then.

I like that idea, makes sense. I'd like to stay consistent with VSCode on this

Christof-P:
Found a way to resolve the bug.
It works now like in VSCode.

apps/studio/src/components/quicksearch/QuickSearch.vue
rathboma: You could have an entry for each:
- Open in background: [Right Arrow]
- Alt in background: [Meta + Right Arrow]

Christof-P:
I added the labels as requested, I'll add a screenshot of them below.
Thank you for letting me work on this!

search change

rathboma:
So good. Thanks so much!

rathboma (approved):
:rocket: :rocket:

PR #1795, opened 2023-11-20
Support updating rows with internal primary keys

### LINKED ISSUE #1532: FEAT: Support editing tables without primary keys where the database provides a `rowid`
Some databases (like SQLite) provide a pseudo-primary key for every row in a table - the rowid.

We can use this to allow data editing for tables without primary keys:

```sql
UPDATE no_pk
	SET details='Baz'
	WHERE no_pk.rowid=2
```

## TODO
1. Identify all the databases that provide a rowid, or something similar
2. Enhance Beekeeper Studio to start reading this data from tables in the `TableTable` view
3. Allow editing & deleting if the table has a rowid.

azmy60:
fix #1532 

NOTE: need a follow-up PR in ultimate for cassandra


azmy60:
I just tested oracle in ultimate. Unfortunately the `getInternalPrimaryKey` approach is kind of not suitable to oracle. But instead, I'm thinking about adding `rowid` option to the `selectTop` function. And then put the `rowid` name of each database to the dialects (many databases have different `rowid` name).

azmy60:
I used his [branch](https://github.com/beekeeper-studio/beekeeper-studio/compare/master...davidkaufman:beekeeper-studio:issue-issue-1532-editing-tables-with-rowids) at first but there were a lot of stuff to change so I ended up starting over to keep it a bit cleaner.

rathboma (changes requested):
Hey, I think you've done a good job here, but I'm not sure I actually want to merge this feature for postgresql.

From the PostgreSQL docs: [ctid isn't always stable](https://www.postgresql.org/docs/current/ddl-system-columns.html) - `a row's ctid will change if it is updated or moved by VACUUM FULL`.

I'm OK with the SQLite implementation, but I think we should remove the PostgreSQL one.

rathboma:
Note from the Oracle docs on ROWID:

```
You should **not** use ROWID as the primary key of a table. If you delete and reinsert a row with the Import and Export utilities, for example, then its rowid may change. If you delete a row, then Oracle may reassign its rowid to a new row inserted later. 
```
from https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ROWID-Pseudocolumn.html#GUID-F6E0FBD2-983C-495D-9856-5E113A17FAF1

davidkaufman:
Yes Oracle, PG and SQL Server all say similar things such as "this internal row id is not to be relied upon not tot change" as PG says, after a VACUUM, Oracle and SQL Server say it may change after a backup & restore, but all of the discussions I saw of these on StackOverflow, etc. users reported that they were stable and reliable enough for exactly our use case here: being able to update a row in a table without a primary key, in a table you just queried.

rathboma:
I think they are reliable enough 99.9% of the time, but the 0.1% of the time when someone is updating a production database and it overwrites the wrong row -- not worth it to me. I don't want Beekeeper Studio depending on `good enough` semantics with something so important.

apps/studio/src/lib/db/client.ts
```diff
@@ -76,6 +76,7 @@ export interface DatabaseClient {
   listMaterializedViews: (filter?: FilterOptions) => Promise<TableOrView[]>,
   getPrimaryKey: (db: string, table: string, schema?: string) => Promise<string | null>,
   getPrimaryKeys: (db: string, table: string, schema?: string) => Promise<PrimaryKeyColumn[]>,
+  getInternalPrimaryKey: (db: string, table: string, schema?: string) => Promise<InternalPrimaryKey>,
```
rathboma, line 79: This should be `Promise<InternalPrimaryKey | null>`

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1085,14 +1095,51 @@ export async function getPrimaryKeys(conn: HasPool, _database: string, table: st
...
+      // Exclude internal primary key
+      const idx = columns.findIndex((col) => col.columnName === internalPrimaryKey.select)
+      if (idx >= 0) columns.splice(idx, 1)
+    }
+  }
+
+  return columns
+}
+
+export async function getInternalPrimaryKey(conn: HasPool, database: string, table: string, schema: string): Promise<InternalPrimaryKey | null> {
```
rathboma, line 1121: We need to kill this, no?

rathboma (changes requested):
let's remove the psql implementation, then we can merge for sqlite

rathboma:
@azmy60 question. Instead of making all these changes to the UI for a single DB client, what if SQLite just returned the secret primary key as the primary key for tables that don't have an explicit one? Then we'd need no UI changes right?

azmy60:
@rathboma Sounds like this is a special behavior for SQLite only. Which is good and it's possible to do that I think. I guess we don't have to show the secret primary key to the table too? Or show it in a way that tells it's a "secret primary key".

rathboma:
I think if we just show it as a normal PK as `rowid` with a special column header tooltip explaining it's automatically provided by sqlite.

Only show it where a primary key hasn't been defined otherwise.

Yeah special behavior just for SQLite.

azmy60:
Ok. That's much simpler than the current implementation. Love that lol.

rathboma:
@azmy60 going to close this in favor of just making the SQLite `rowid` visible, that can be a separate PR

PR #1799, opened 2023-11-23
fix: TabTableBuilder and VirtualTableList

azmy60:
bugs from recent PRs:
- click the + icon to make a new table, then it shows blank screen.
  cause: undefined `defaultSchema` method in sqlite.ts
- when the database is empty, creating a new a table does not show the table in the table list, even after refreshing.
  cause: `NaN` value (zero divided by zero) in VirtualTableList
- dislocated empty state when database is empty

### DIFF
--- a/apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
+++ b/apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
@@ -2,6 +2,7 @@
   <virtual-list
     ref="vList"
     class="list-body"
+    :class="{ 'list-body-empty': displayItems.length === 0 }"
     :data-key="'key'"
     :data-sources="displayItems"
     :data-component="itemComponent"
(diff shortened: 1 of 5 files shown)


apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
```diff
@@ -74,6 +75,8 @@ interface RoutineItem extends BaseItem {
   parent: BaseItem;
 }
 
+const INITIAL_ITEM_HEIGHT = 22.8;
```
rathboma, line 78: put this in globals.ts?

apps/studio/src/lib/db/clients/sqlite.ts
```diff
@@ -69,6 +69,10 @@ export class SqliteClient extends BasicDatabaseClient<SqliteResult> {
     return this.version?.data[0]["sqlite_version()"];  
   }
 
+  defaultSchema(): string | null {
```
rathboma, line 72: Can this just be implemented like this in the `BasicDatabaseClient` instead? I think null is a sensible default.

rathboma (approved):
See minor changes - then good to merge.

PR #1800, opened 2023-11-23
Removing variableHeight that causes lags when resizing columns in TableTable

azmy60:
Resizing columns is super laggy if the table is big. This is not the case in the ResultTable, but in the TableTable. So I tried the devtool profiler, and found this:

image

image

image

There's a function `checkCellHeights` that takes very long (pun intended). This is an internal method from tabulator. It disappears if I remove the `variableHeight` from the code. I don't know if that's ok, but It seems like it's safe if we could remove it. I'm not seeing any use case for that yet. Tested a cell that uses textarea editor works just fine too.


rathboma:
I think that was there so that text columns would get more height during editing, but with the new pop out editor, we don't really need that anymore as much.

rathboma (approved):

PR #1803, opened 2023-11-30
Language editor tweaks

rathboma:
Minor tweaks to the excellent PR #1769 

TODO:

- [x] Instead of showing errors when clicking 'done'. Debounce the error check and have it be at the bottom right of the editor all the time
- [x] Editing an array does not work. It should open it as a JSON array
- [x] There are two scrollbars (both x and y). I don't know why
- [x] If it can't open a column for some reason (eg binary data), it should not just open a blank editor window
- [x] Make sure the editor modal is always a good vertical size, even when resizing


rathboma:
@azmy60 Looking nice!  Great style update.

Some stuff that still needs fixing:

- [ ] I accidentally killed my code to add a `Wrap Text` checkbox to the header. This should be checked by default for `text` and `markdown` types, but unchecked for others. I implemented it by adding a `wrapTextByDefault` to the `LanguageData`, Codemirror has a setting to Wrap Text. Can you add this?

- [ ] When editing a long string, it goes off the right side of the editor modal, but you can't scroll to the end
image

- [ ] Is there a way to make the height dynamic to the content but give it a max height? For example this JSON object isn't very big, so we could display the whole thing easily if the modal were bigger initially (rather than having to manually resize it every time:
image
image

- [ ] There is no horizontal scrollbar

PR #1811, opened 2023-12-04
feat: firebird support

### LINKED ISSUE #959: FEAT: Firebird support
Is it possible to add [Firebird](https://firebirdsql.org/) support?

azmy60:
closes #959 

Several things to note from [node-firebird](https://github.com/hgourvest/node-firebird?tab=readme-ov-file#charset-for-database-connection-is-always-utf-8):
- Charset for database connection is always UTF-8. This is why you should use Firebird 2.5 server at least.
- Firebird 3 and 4 wire protocol are not supported yet. For now, the only way to access a firebird database is by enabling the legacy authentication from the server config so we can login via user/password auth:
  - https://github.com/hgourvest/node-firebird/issues/42. 
  - https://firebirdsql.org/manual/migration-mssql-client-access.html#migration-mssql-client-access-developers


apps/studio/tests/docker/firebird.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
+services:
+  firebird:
```
rathboma, line 3: we should avoid using a docker-compose file for database tests, becuase depending on the version of Docker used, the name of the entry is different. It's really stupid, but it's broken things for us before.

rathboma: See how we specify stuff like environment variables and such in the sql server and oracle test harnesses

azmy60:
I changed some tests to make firebird work though, such as:
- not naming tables with keywords (`group` -> `group_table`), this is an error in firebird.
- no multiple inserts.
- map column names to lower case, because querying table columns always result in UPPERCASE.

yarn.lock
```diff
@@ -10281,6 +10400,15 @@ klona@^2.0.4:
   resolved "https://registry.yarnpkg.com/klona/-/klona-2.0.5.tgz#d166574d90076395d9963aa7a928fabb8d76afbc"
   integrity sha512-pJiBpiXMbt7dkzXe8Ghj/u4FfXOOa98fPW+bihOJ4SjnoijweJrNThJfd3ifXpXhREjpoF2mZVH1GfS9LV3kHQ==
 
+knex-firebird-dialect@^2.0.0:
```
rathboma, line 10403: This knex dialect depends on `node-firebird-driver-native`, which depends on the native client libraries (libfb).

We want to use `node-firebird` only, because it is pure JavaScript and does not depend on client libraries.

I've filed an issue with the knex dialect repo - https://github.com/Tomas2D/knex-firebird-dialect/issues/262,

but we'll probably just have to fork it and use `node-firebird` instead

rathboma:
@azmy60 doesn't this merged PR implement some of the wire protocol? https://github.com/hgourvest/node-firebird/pull/242

rathboma (approved):

PR #1814, opened 2023-12-05
fix: stop tabulator from updating an array to a string when it shouldn't

### LINKED ISSUE #1807: BUG: Array editing is broken
Editing an array converts the array to CSV, but then even if you don't make any changes, it will mark it as changed.

How it should work:
- Arrays should be edited as JSON arrays, not csv strings

wmontgomery:
fix #1807 
bugfix-1807

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -954,16 +954,22 @@ export default Vue.extend({
     },
     cellEdited(cell) {
       const pkCells = cell.getRow().getCells().filter(c => this.isPrimaryKey(c.getField()))
+      const oldValue = cell.getOldValue()
+      const currentValue = cell.getValue()
+      const isArray = _.isArray(oldValue)
 
       // some number fields were being converted to strings so were triggered the cellEdited event because tabulator probably `===` stuff
       // If the cell value does fall into this, we don't want anything edited.
-      if (cell.getOldValue() == cell.getValue()) {
+      if (oldValue == currentValue) {
+        // here's another one: If you are editing an array, tabulator is turning it into a comma-delimited string so if it was an array, keep it as such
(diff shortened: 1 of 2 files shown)


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -954,16 +954,22 @@ export default Vue.extend({
...
+      const oldValue = cell.getOldValue()
+      const currentValue = cell.getValue()
+      const isArray = _.isArray(oldValue)
 
       // some number fields were being converted to strings so were triggered the cellEdited event because tabulator probably `===` stuff
       // If the cell value does fall into this, we don't want anything edited.
-      if (cell.getOldValue() == cell.getValue()) {
+      if (oldValue == currentValue) {
+        // here's another one: If you are editing an array, tabulator is turning it into a comma-delimited string so if it was an array, keep it as such
+        if (_.isArray(oldValue) && !_.isArray(currentValue)) {
```
rathboma, line 965: I think we should solve this in the `NullableInputEditor`. Shouldn't the editor just return an Array rather than a string?

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -993,6 +999,15 @@ export default Vue.extend({
           value: cell.getValue()
         }
       })
+
+      let valueToSave = cell.getValue(0)
+
+      if (isArray) {
+        if (!_.isArray(valueToSave)) {
```
rathboma, line 1006: These exceptions make my spidey sense tingle, I think we need to solve this in the editor

rathboma (changes requested):
This is good, but I think it's brittle, and will be easy to break. Also it'll screw us when we add even more complex types (what about jsonb, bytes, etc?). I think we should solve this in the NullableInputEditor

PR #1820, opened 2023-12-09
fix(TableSchema): column name should be green if it's new row

azmy60:
The easiest way to fix this is by not making the column name `frozen` (which is what I did lol).  See https://github.com/beekeeper-studio/beekeeper-studio/pull/1701 for why column name needs to be `frozen`.


azmy60:
OK, keep the cells frozen, and match the colors.

Issue #810, opened 2021-09-21
FEAT: Quick search through databases aswell


GijsGoudzwaard:
Would be nice if the new quick search method would also support searching through databases

rathboma:
Good idea! Do you need to switch databases a lot?

GijsGoudzwaard:
Not that much but I have a lot of databases and it would be simpler to search through them instead of using the dropdown. Maybe use another keybinding like `command + shift + p` for searching through your databases?

DKhalil:
> Good idea! Do you need to switch databases a lot?

For a multi instance setup this is definitely the case (as it is for me), so having a shortcut to switch between databases would be very helpful!

TomSchillemans:
At my work place I do have to change databases regularly. This would be a really handy to have.

PR #1821, opened 2023-12-10
feat: Enable quick switch for connections and databases

marco-lavagnino:
This PR allows quick switching to connections and databases, as discussed in issue https://github.com/beekeeper-studio/beekeeper-studio/issues/810 .

See it in action:
image

To achieve this, some of the internal state of `DatabaseDropdown` was moved to the main Vuex store, so it can be queried from both the dropdown as well as the QuickSearch module.


rathboma:
Sorry for the late response on this PR. This is great!

rathboma (approved):

PR #1824, opened 2023-12-14
Wayland native support for Beekeeper Studio!

rathboma:
- Yes, we beat VSCode to a fully supported and documented way of enabling Wayland support
- Works with Wayland fractional scaling with no blurriness
- Tested and working on Ubuntu 22.04


rathboma:
TODO: needs a docs page

PR #1833, opened 2023-12-23
Example monorepo using turborepo

rathboma:
Ok, this works well.

```
# Runs electron:serve and rebuilds the tsc project on changes, webpack properly hot reloads
turbo dev
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.turbo/
 .idea/
 .cache/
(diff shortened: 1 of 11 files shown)


Z3rio:
Cool!
I've personally played around a bit with both Lerna and Turborepo, and it is actually fairly nice
so would be great to see this added to Beekeeper Studio

great work so far mate

PR #1834, opened 2023-12-26
Fix: One table only has one tab and show warning dialogs when closing unsaved tabs

### LINKED ISSUE #1700: BUG: Closing all tabs does not give warning of unsaved changes
**To Reproduce**
1. Make some changes in many tabs (writing queries, creating a table, etc.)
2. Right click the tab and `Close All`/`Close Other`/`Close Tabs to The Right`

**Expected behavior**
It should give a warning like when closing a single tab.

**Additional context**
It should be able to handle unsaved changes in multiple tabs too.

azmy60:
closes #1700

### DIFF
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -207,6 +207,7 @@ import { DatabaseEntity } from "@/lib/db/models"
 import PendingChangesButton from './common/PendingChangesButton.vue'
 
 import { safeSqlFormat as safeFormat } from '@/common/utils';
+import pluralize from 'pluralize'
 
 export default Vue.extend({
   props: ['connection'],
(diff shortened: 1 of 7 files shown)


apps/studio/src/components/CoreTabs.vue
```diff
@@ -635,7 +613,15 @@ export default Vue.extend({
         }
       }
     },
-    async close(tab) {
+    async close(tab: OpenTab) {
+      if (tab.unsavedchanges) {
+        const confirmed = await this.$confirm(
+          `really close [icon here] ${tab.title}?`,
```
rathboma, line 619: looks like this has a placeholder? `[icon here]`?

rathboma (changes requested):
looks good, but seems like there's some placeholder text?

apps/studio/src/components/CoreTabs.vue
azmy60: Lol, I forgot. My bad.

rathboma: Once this is done, I'll get it merged it. Probably for the next update

azmy60: image

done :)

rathboma (approved):

Issue #1738, opened 2023-10-13
FEAT:  Automatically convert unixtime to a human-readable DateTime


rathboma:
When a user is hovering over an obvious timestamp, show them the date in a human readable form.

Questions to figure out

- Is this integer really a timestamp, or just an integer?

For future:
- Let the users choose this manually somehow in the columns modal

PR #1835, opened 2023-12-26
Feature/1738 readable date time

wmontgomery:
#1738 
image


rathboma:
Hey Will,

Can we remove the icon and just make the title something like `${datetime} in unixtime`?

wmontgomery:
You betcha.

wmontgomery:
image

apps/studio/src/mixins/data_mutators.ts
```diff
@@ -33,6 +33,14 @@ function emptyResult(value: any) {
   return null
 }
 
+function buildFormatterWithTooltip(cellValue: string, tooltip: string, icon?: string) {
+  if (!icon) {
+    return `<div class="cell-link-wrapper" title="${tooltip}">${cellValue}</div>`
```
rathboma, line 38: These dynamic values need HTML escaping

apps/studio/src/mixins/data_mutators.ts
```diff
@@ -53,22 +61,30 @@ export default {
...
           fkLink.onclick = (e) => params.fkOnClick(e, cell);
         })
-      }
+      } else if (
+          params?.isPK != null &&
+          !params.isPK &&
+          _.isInteger(Number(cellValue)) &&
+          Number(cellValue) <= 8640000000000000 // the date object doesn't go higher than said number https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#description
+        ) {
+        tooltip = `${new Date(Number(cellValue)).toISOString()} in unixtime`
```
rathboma, line 85: What if this fails? We should error catch here and set a default tooltip

rathboma (changes requested):
Error catching, html escaping, then done

apps/studio/src/mixins/data_mutators.ts
```diff
@@ -79,11 +78,14 @@ export default {
...
-          Number(cellValue) <= 8640000000000000 // the date object doesn't go higher than said number https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#description
+          _.isInteger(Number(cellValue))
         ) {
-        tooltip = `${new Date(Number(cellValue)).toISOString()} in unixtime`
-        result = buildFormatterWithTooltip(cellValue, tooltip)
+        try {
+          tooltip = `${new Date(Number(cellValue)).toISOString()} in unixtime`
+          result = buildFormatterWithTooltip(cellValue, tooltip)
+        } catch (e) {
+          console.error(`${cellValue} cannot be converted to a date`)
```
rathboma, line 87: You should log the error here too

rathboma (approved):

PR #1837, opened 2023-12-27
TBD: Get routine SQL fails for aggregate functions

rathboma:
This is simply a test replicating the issue. Needs fixing


not-night-but:
@rathboma the tests for this aren't failing anymore, is this not an issue?

PR #1843, opened 2023-12-29
Add option to show password

### LINKED ISSUE #1828: FEAT: View saved passwords
I have multiple saved connections, and thought that because it is saved, I would also be able to see the passwords. It is frustrating when there are instances where I need to copy the password but I'm not able to provide them to other people, or when I need it or something else. I'm certain that this is my fault for not storing passwords in a password manager but it'd be a nice QOL if this was added.

Like one of those typical login form in the password field where users can toggle if they want to view the password they've typed.

A copy button to immediately copy the password would be okay, too.

Christof-P:
fix #1828
showPassword

Let me know if this is what you had in mind

### DIFF
--- a/apps/studio/src/assets/styles/app/connection-interface.scss
+++ b/apps/studio/src/assets/styles/app/connection-interface.scss
@@ -181,3 +181,12 @@
     margin-bottom: $gutter-h;
   }
 }
+
+.password {
+  display: inline-block;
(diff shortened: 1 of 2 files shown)


Z3rio (approved):
LGTM 🚀

rathboma (approved):

PR #1847, opened 2024-01-04
add table limiter to tabletable

### LINKED ISSUE #941: FEAT: Configurable amount of rows returned per page. (currently seems to be 100 rows per page)
when using Beekeeper-studio with a large table, it appears to limit the number of returned rows to 100.

Would be excellent if there were a configuration that could be set for the number of returned rows per page.

azmy60:
Closes #941 

Animation4

### DIFF
--- a/apps/studio/src/assets/styles/app/tabs/tabletable.scss
+++ b/apps/studio/src/assets/styles/app/tabs/tabletable.scss
@@ -61,4 +61,11 @@
   }
 
 
+  .table-limiter-container {
+    display: flex;
+    min-width: 130px;
+    & > span {
+      white-space: nowrap;
+    }
+  }
 }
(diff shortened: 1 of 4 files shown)


azmy60:
Not sure if having a button like this is great since we have a lot of things in status bar:

image

The other option is to have a modal that's solely for the limiter, but that feels a little bit too much? Having a button in status bar (or somewhere near the table) is more aligned with design patterns.

rathboma:
@azmy60 can you add it to the :gear: menu? as a submenu?

azmy60:
@rathboma like this?

image

I'm also thinking of adding a floating element at the bottom of screen like when you zoom in/out in chrome? so we can customize the limit numbers.

image

rathboma:
What if we abandon this in favor of #1911. I like the idea of centralizing all the configuration in one place. We can also (eventually) add a settings screen

rathboma:
Actually, yeah I really like the ini file.

azmy60:
Yup that's a very nice idea.

rathboma:
We can still allow users to change these settings in the app, but I think using an ini file as the core storage place is going to make that really flexible and nice to work with.

Issue #1435, opened 2022-11-18
FEAT: Notifications for long-running queries


asugden:
**Is your feature request related to a problem? Please describe.**
When testing queries, some of them run for a long time and I keep switching back to the app to check the status of the query.

**Describe the solution you'd like**
It would be amazing if one could do any of the following:
1. Have an option to trigger a system notification for all queries that run longer than some unit of time (10s? 30s? Optional)
2. Have a toggle on a particular query "notify when completed" that would trigger a system notification just for that query.

**Describe alternatives you've considered**
Non-system notifications might work as well (perhaps capturing focus as VS Code does when it hits an error, but that is more intrusive). I imagine that system notifications are easier to trigger.

rathboma:
Great idea :-).

I'm soon looking to add support for BigQuery and Snowflake, so this will definitely become useful when we do that.

not-night-but:
Looks like this was fixed in #1855

PR #1855, opened 2024-01-05
feat: notify query execute successfully when app is not focused.

invisal:
https://github.com/beekeeper-studio/beekeeper-studio/issues/1435

We will push notification only when the app is not in focused. We can make the notification look better if we have some images.

**For Windows Screenshot**
image

**For Mac Screenshot**
image

**How to test**
- Connect to MySQL server.
- Run this query `SELECT SLEEP(10);`
- Open other app besides Beekeeper and wait
- The notification will show after 10 seconds when it is complete.


rathboma:
This is awesome! Does this create a native notification on different OSes, or a weird electron-looking one?

invisal:
I believe it should be native. In Window, it appears in the bottom right side. I will try to test on Mac to see how it looks like, but because my Mac is M1, it seem to have some problem running the project. Let me try again and provide you with Mac screenshot.

invisal:
@rathboma I have updated the screenshot for Mac. The thing that need to do is wording. My english is terrible and maybe you can give suggestion.

rathboma (approved):

PR #1859, opened 2024-01-05
Format numbers in TableLength to improve readability

luca1197:
The component which displays the number of rows in a table currently displays the number without any formatting, which can make high numbers (6+ figures) difficult to read.

It seems this was already supposed to be partly implemented in line 11 using `totalRecords.toLocaleString()`, but this did not seem to have any effect. Using `Number(totalRecords).toLocaleString()` fixed this. In addition to that, the number in the tooltip is formatted aswell now.

**Before:**
Beekeeper_Studio_XORIzxtfli

**After:**
image


rathboma (approved):

rathboma:
Thanks!

PR #1861, opened 2024-01-09
Move `mysql` to BasicDatabaseClient

### LINKED ISSUE #1849: Move MySQL to the `BasicDatabaseClient` abstract class implementation
This will help us simplify everything.

We should have two final classes:
- MySqlClient
- MariaDbClient extends MySqlClient

There are basically 0 differences between MariaDB and MySQL, but this is a good convention we should follow.

azmy60:
closes #1849


rathboma:
TODO: Check if removed functions are used anywhere

rathboma:
@not-night-but once you've approved it tag @azmy60 so he can merge it!

apps/studio/src/lib/db/client.ts
```diff
@@ -156,7 +156,7 @@ export interface IDbConnectionDatabase {
   connecting: boolean,
 }
 
-export class DBConnection {
+export class DBConnection implements DatabaseClient {
```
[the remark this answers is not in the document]
not-night-but: Yeah I think this is good!

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -0,0 +1,1340 @@
...
+    } else {
+      config.ssl.rejectUnauthorized = server.config.sslRejectUnauthorized;
+    }
+  }
+
+  return config;
+}
+
+function resolveDefault(
+  version: MysqlClient["versionInfo"],
```
not-night-but, line 167: Did not know this was possible lol, but I think you should just have a class for versionInfo.

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -0,0 +1,1340 @@
...
+    defaultValue.length >= 2
+  ) {
+    // MariaDb escapes all single quotes with two single quotes in default value strings, even if they are
+    // escaped with backslashes in the original `CREATE TABLE` statement.
+    return defaultValue
+      .substring(1, defaultValue.length - 1)
+      .replaceAll("''", "'");
+  }
+
+  return defaultValue;
```
not-night-but, line 191: We don't actually need to branch here. This function could just return `defaultValue` here, and then override it in the MariaDB class to do all that extra logic

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -0,0 +1,1340 @@
...
+}
+
+export class MysqlClient extends BasicDatabaseClient<ResultType> {
+  server: IDbConnectionServer;
+  database: IDbConnectionDatabase;
+  versionInfo: {
+    versionString: string;
+    version: number;
+    isMySql: boolean;
+    isMariaDb: boolean;
```
not-night-but, line 278: If at all possible I think we should get rid of these. Any logic that isn't the same between the db's should be overridden in the child

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -0,0 +1,1340 @@
...
+    table: string,
+    orderBy: OrderBy[],
+    filters: string | TableFilter[],
+    chunkSize: number,
+    _schema?: string
+  ): Promise<StreamResults> {
+    const qs = buildSelectTopQuery(table, null, null, orderBy, filters);
+    const columns = await this.listTableColumns(db, table);
+    const rowCount = await this.getTableLength(table);
+    // TODO: DEBUG HERE
```
not-night-but, line 618: Is this your todo?

not-night-but (changes requested):
Looking good!! A couple things:
- Rather than branching on whether or not it is mysql, we should just override the function in the child
- Functions that need access to instance members, and don't need to be statically accessed can just be private/protected helper functions.

Fantastic job as always. Common Azmy W

apps/studio/src/lib/db/clients/mysql.ts
azmy60: Yeah, that makes total sense.

apps/studio/src/lib/db/clients/mysql.ts
azmy60: I don't think so.

wmontgomery (approved):

not-night-but (approved):
Looks good to me!

PR #1873, opened 2024-01-10
using sass is a spiderweb of problems

### LINKED ISSUE #1874: REFACTOR: Move from node-sass to sass
Node-sass is deprecated, long live sass!

node-sass allows some stuff that the `sass` package doesn't like. In particular a lot of `@extends` are not valid in `sass`, but are in `node-sass`.

There is a prototype PR for this #1873.

It should work fine with `sass-loader`.

TODO:

- [ ] Solve all the `@extends` things that sass doesn't like
- [ ] Fix all the warnings from building with sass
- [ ] Make sure we didn't break any styles

rathboma:
fix #1874


azmy60:
I think this is good. Some changes I made:

- I changed the way we import vendor css files a little bit, cause the old one uses `~` in the beginning of the path that doesn't work with the new config.
- For some reason, some styles we made couldn't override the vendor css after the new config. I went through some basic flows and fixed all that I've seen but I wouldn't say I've covered all the breaking styles.

PR #1875, opened 2024-01-10
add folders to saved connection list

### LINKED ISSUE #65: [Enhancement] Ability to create folder in saved connection
It would be awesome to have the chance to create folders to better arrange the saved connections

### LINKED ISSUE #799: FEAT: folders for saved connections
To improve the organization for saved connections, it would be great to have the option to create folders.
For example you could have folders for every project / customer which could consist of multiple saved connections.

Already described above.

Since this should be optional, it wouldn't break existing connections and no one is forced to use this feature.
Having more than one level of depth would be great, but at least for me not needed at the moment.

Well you can prefix you saved connections which will become messy in the future.

azmy60:
close #65, close #799

Here we have a folder tree component that's reusable for other things in sidebar. It should be capable of:
- Moving items to different expanded/collapsed folders
- Moving items to outside folders (which is a root folder)

It can't make folders inside folders yet, but I add `folder_id` to the model for parent folders just in case.

https://github.com/beekeeper-studio/beekeeper-studio/assets/38707148/5a562be3-0a03-4b40-a52f-c9bdd8f01412

(i'm running this from WSL which is why the item looks white when moving it)

### DIFF
--- a/apps/studio/src/App.vue
+++ b/apps/studio/src/App.vue
@@ -25,6 +25,7 @@
     />
     <data-manager />
     <confirmation-modal />
+    <prompt-modal :name="this.$promptModalId" />
   </div>
 </template>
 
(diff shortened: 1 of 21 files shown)


rathboma:
@azmy60 does this work with workspaces?

azmy60:
Do you mean in ultimate? Not yet. This is just for the saved connection in the main screen.

rathboma:
@azmy60 let's move this feature to Ultimate. That way we don't have to implement twice. Ultimate only

PR #1876, opened 2024-01-10
unfreeze column name and fix nullable editor onblur

azmy60:
in previous version, we always call the `submit` event on blur or on input in `NullableInputEditor`. If we blur the input, and didn't call `submit` or `cancel`, it would cause some weird "unfocusable" cells. It looks like tabulator blocks some click events or something if we didn't call one of those two events.


rathboma (approved):

rathboma:
@azmy60 is this problem on any of the other tabulator tables in the properties screen? Eg indexes or relations?

azmy60:
Yes. Any tables that use `NullableInputEditor`.

PR #1879, opened 2024-01-11
update vimrc parsing

austinwilcox:
Commands like: vmap K :m \'<-2<CR>gv=gv are now parsed correctly (note, codemirror does not support the :m command, so this code sample is not a great one) but now if you have commands that involve a space in the command part, they should now be registered. @not-night-but do you have any commands in a vimrc that would benefit from this? I have the above command set up with neovim and it will move the visually selected line up one line, I have the same thing for J and it moves it down, but the above command does not work with codemirror.


rathboma:
@austinwilcox This is awesome, thanks so much! Is there any way you could add a test for this to make sure we're parsing vimrc files properly?

Feel like this could be easy to break if we aren't careful!

rathboma (changes requested):
If there's any way we can add a parsing unit test for some examples it'd be great.

austinwilcox:
@rathboma great thought! I'll get to work on some tests, and create some example vimrcs to add into the test suite to ensure things don't break when/if we make additional parsing changes. Sometime this week I'll get into it.

austinwilcox:
@rathboma I have some tests in here, I added some additional logic to that will be nice to have in a vimrc. If you were to supply duplicate remappings in the vimrc file, the last one found in the file is what will be kept and that's the natural behavior of vim/neovim which I just mimicked. I am running a build really quick to make sure that everything still works as expected in the editor. Let me know if there's anything else you would like in this PR. It's a relatively small change, but nice to have.

Would you like me to write docs to go somewhere in the public facing site or somewhere here in the repository for instructions in using vim and the vimrc?

apps/studio/tests/unit/mixins/vimrc.spec.js
```diff
@@ -0,0 +1,85 @@
...
+    const vimrcSampleContent = [
+      "nmap gl $",
+      "nmap gh ^",
+      "nmap Y y$",
+      "nmap J :tabp",
+      "nmap K :tabn",
+      "vmap K :m '<-2gv=gv"
+    ];
+
+    const expected = {
```
rathboma, line 14: This is awesome!

rathboma (approved):
Fantastic, thanks! I'm going to merge this, but would love for you to add some of this to the documentation -- the docs are part of the repository again:

https://github.com/beekeeper-studio/beekeeper-studio/blob/master/docs/user_guide/using-the-sql-editor.md#vim-mode

PR #1888, opened 2024-01-15
add copy as plain text

### LINKED ISSUE #1867: Can't copy only cell contents any more
**Details:**
Since the update to v4.1 (and maybe 4.0) I can no longer right-click and copy the highlighted cell contents.  There used to be an option to copy the cell contents - now the option is to copy as json, copy as tsv, etc. but the problem is this now copies the field name as well and also add quotes or escape sequences.

My use-case is that I have a field that contains a large json document - when I copy it I now have to remove the field name and additional nesting that is present and I also have to remove backslashes that have been added to the cell content.

I'm not sure of any use-case where I would want that content nested inside an object including the field name.

Is this an intentional change?  Or will the previous behaviour be restored?  Even an additional option to copy just the field content without the extra stuff would work.

Operating System: Linux
App Version [help -> about]: 4.1.0
Database type and version [eg Posgresql 9.3]: MariaDB 10.6

azmy60:
Bringing back previous behaviors that are missing:

1. Added a "copy as plain text" option to the context menu. This won't add quotes or escapes the data, and is available when right clicking a range of cells, column headers, and row headers.

2. Based on the previous version, copying a single cell should not add quotes, but copying multiple cells should. This behavior is now added to the keyboard shortcut.

image

fix #1867

### DIFF
--- a/apps/studio/src/components/editor/ResultTable.vue
+++ b/apps/studio/src/components/editor/ResultTable.vue
@@ -162,12 +162,7 @@
       },
       copySelection() {
         if (!document.activeElement.classList.contains('tabulator-tableholder')) return
-        const range = this.tabulator.getActiveRange()
-        if (range.getCells().length === 1) {
-          copyRange({ range, type: 'plain' })
-        } else {
-          copyRange({ range, type: 'tsv' })
-        }
+        copyRange({ range: this.tabulator.getActiveRange(), type: 'plain' })
       },
       dataToJson(rawData, firstObjectOnly) {
         const rows = _.isArray(rawData) ? rawData : [rawData]
--- a/apps/studio/src/components/editor/ResultTable.vue
+++ b/apps/studio/src/components/editor/ResultTable.vue
@@ -162,7 +162,12 @@
       },
       copySelection() {
         if (!document.activeElement.classList.contains('tabulator-tableholder')) return
-        copyRange({ range: this.tabulator.getActiveRange(), type: 'tsv' })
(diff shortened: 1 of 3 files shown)


rathboma:
What happens when someone uses the keyboard copy shortcut?

rathboma:
I think tab delimited is fine, but it shouldn't quote numbers, and it shouldn't include the header

azmy60:
Using keyboard shortcut will copy as plain text using `cell.getValue()` if it's only one cell, so no formatting.

This was how it worked before spreadsheet:
1. If we select one cell, we copy with `cell.getValue()`. So no formatting.
2. If we select an entire row, we copy in TSV format with quotes.

TBH, this PR is somehow not consistent in terms of copying with quotes or not. @rathboma do you have any solution? Should we just remove the quotes from copy TSV? and then use that for keyboard shortcut? Though I'm not sure why did we end up quoting TSV values in the beginning.

azmy60:
Just realized that google sheet does not quote the data for copying, so at least we have one reason to not use quotes lol.

rathboma:
Ok, let's just call this `Copy`, no format. I agree with your settings. Let's make sure this is what happens when someone presses `Ctrl/CMD-C`

azmy60:
Done :) One cell or multiple cells, pressing `ctrl+c` will copy with no quotes and escapes (still compatible with google sheet). You can still have quotes and escapes by `right click` > `Copy as TSV for Excel`.
image

rathboma (approved):

PR #1904, opened 2024-01-17
Bring back sort buttons

### LINKED ISSUE #1887: BUG: Table sort order isn't indicated in the column header at all
Prior to 4.1, if you sorted a table by a specific column it would show a little up or down arrow to indicate the direction of the sort. This is missing from the table in 4.1.

### LINKED ISSUE #1900: Simple sort no longer available. Must use context menu or column header ellipses.
**Describe the bug**
Before updating to 4.1.10, clicking a column header toggle between ascending/descending sort

**To Reproduce**
1. Open a Table
2. Click a column header. Nothing happens. Clicking a column header no longer toggles between ascending sort or descending sort. You have to right click the column header then select a sort order, which is much less convenient than the original feature.

**Expected behavior**
Clicking the header toggles between ascending sort or descending sort without having to open a context menu or click the vertical ellipses in the header. 

**Version Information:**
- MacOS Ventura 13.5.2
- BeeKeeper Studio 4.1.10
- Postgres 14.2

azmy60:
fix #1900 fix #1887

Animation4

TODO
- [x] Initial width can be set by tabulator. No need to do this manually.
- [ ] Reset column widths when someone install the new update? (Beware to just reset the __width__)

### DIFF
--- a/apps/studio/src/plugins/CustomTabulatorModule.ts
+++ b/apps/studio/src/plugins/CustomTabulatorModule.ts
@@ -1,0 +1,19 @@
+import { Module } from "tabulator-tables";
+
+export class CustomTabulatorModule extends Module {
+  static moduleInitOrder = 100;
+
+  initialize() {
+    // @ts-expect-error Module class not fully typed
+    this.subscribe("column-layout", this.moveSorterToTitleElement.bind(this));
+  }
+
+  moveSorterToTitleElement(column) {
+    const arrowEl = column.titleHolderElement.querySelector(
+      ".tabulator-col-sorter"
+    );
+    if (arrowEl) {
+      column.titleElement.insertBefore(arrowEl, column.titleElement.firstChild);
+    }
+  }
+}
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -70,7 +70,7 @@
     "sql-formatter": "~10.7.2",
     "sql-query-identifier": "^2.6.0",
     "ssh2": "^1.14.0",
-    "tabulator-tables": "beekeeper-studio/tabulator#f4217f7be7d0fe4880e49f311a5076ff01e4f981",
+    "tabulator-tables": "beekeeper-studio/tabulator#93b922ff4cc9bec5485fa546527eda702ad17f1b",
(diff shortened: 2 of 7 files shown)


azmy60:
To be honest, I'm not sure if I like clicking a sort button should also select a column. I did that because It's what happens if you click a column menu. And I think it's possible if we could just select the sort button without selecting the column without working on the tabulator side. But I'll leave it here to see anyone's opinion on this.

rathboma:
This is great! A really simple and obvious solution. What happens when you make the column really narrow, or have a really long column name?

azmy60:
Oh yeah... I made a function to set an initial column width that adjusts to the column contents. But that works if the table was never opened because we're using the tabulator persistence options.

The next thing to look into is probably how to refresh the data when next time someone has the new UI.

azmy60:
@rathboma I misunderstood the yesterday's question lol. I updated the gif, if the text is too long, the icons should overlap the text and it has the triple dots just like before.

rathboma:
Great!

apps/studio/src/plugins/CustomTabulatorModule.ts
```diff
@@ -0,0 +1,19 @@
+import { Module } from "tabulator-tables";
+
+export class CustomTabulatorModule extends Module {
```
rathboma, line 3: Hmm, I think this probably needs  better name? Maybe `HeaderSortTabulatorModule`, or something?

rathboma (approved):
LGTM, merge away, but think about renaming that module

apps/studio/src/plugins/CustomTabulatorModule.ts
azmy60: I've still got one very little issue. This is how it looks if someone open a table that they have opened before:

image

It's using the previous width sizes, and we're adding a new element to the header. It's easy to adjust though, just right click > `resize all column to match contents`.

azmy60: It's good if someone never opened the table before. It looks like the gif I showed you.

PR #1907, opened 2024-01-18
Encapsulate codemirror

### LINKED ISSUE #1883: REFACTOR: Make a 'texteditor' component to encapsulate CodeMirror
We need to standardize how we use codemirror in all the different places.

Some stuff we need:
- Codemirror ALWAYS inserts `tab` when tab is pressed, this should be configurable (eg yaml hates tabs)
- Multi-language support
- Sensible events and properties

azmy60:
fix #1883


apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -744,51 +736,11 @@
...
-              closeOnBlur: false
-            },
-            // eslint-disable-next-line
-            // @ts-ignore
-            hint: CodeMirror.hint.sql,
-            hintOptions: this.hintOptions,
-            keyMap: this.userKeymap,
-            getColumns: this.getColumnsForAutocomplete
-          } as CodeMirror.EditorConfiguration)
+          this.editor = this.$refs.textEditor.initialize()
```
rathboma, line 739: We shouldn't be calling methods on child components. 

We should initialize inside of the text-component itself

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -744,51 +736,11 @@
...
           this.editor.setValue(startingValue)
           this.editor.addKeyMap(runQueryKeyMap)
-          this.editor.on("keydown", (_cm, e) => {
-            if (this.$store.state.menuActive) {
-              e.preventDefault()
-            }
-          })
-
+          this.editor.setOption('hintOptions', this.hintOptions)
```
rathboma, line 742: Again, we should not be calling methods on child components, this breaks the Vue.js model of component communication.

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -807,13 +759,6 @@
...
-
-
-            if(!codeMirrorVimInstance) {
-              console.error("Could not find code mirror vim instance");
-            } else {
-              setKeybindingsFromVimrc(codeMirrorVimInstance);
-            }
           }
 
           this.editor.on("paste", (_cm, e) => {
```
rathboma, line 764: Again, all this stuff should no longer be in `TabQueryEditor`, and we shouldn't be calling methods on the child component

rathboma (changes requested):
This is a great start, but we're still 'leaking' too much of codemirror into the parent components. More of that work needs pushing down into text-editor.

Also, while I know we do this in a couple of places because it is unavoidable, we should not call methods directly on child components.

apps/studio/src/components/TabQueryEditor.vue
azmy60: Oh that's a great idea!

I initially put `hintOptions` as a prop but it's only used for codemirror so the separate component probably would look something like this:

```vue
<sql-text-editor :tables="..." @getColumns="..." />
```

apps/studio/src/components/common/texteditor/SQLTextEditor.vue
```diff
@@ -0,0 +1,179 @@
...
+        ...this.extraKeybindings,
+      };
+    },
+  },
+  methods: {
+    formatSql() {
+      const formatted = format(this.value, {
+        language: FormatterDialect(dialectFor(this.lang)),
+      });
+      this.editorInterface.setValue(formatted);
```
rathboma, line 83: same thing here -- we don't wan `editorInterface` leaking up from below. Probably `TextEditor` can handle formatting on it's own, it knows the mime type

apps/studio/src/components/common/texteditor/SQLTextEditor.vue
```diff
@@ -0,0 +1,179 @@
...
+      if (!tableToFind) return null;
+      // Only refresh columns if we don't have them cached.
+      if (!tableToFind.columns?.length) {
+        await this.$store.dispatch("updateTableColumns", tableToFind);
+      }
+
+      return tableToFind?.columns.map((c) => c.columnName);
+    },
+    handleInterface(editorInterface: any) {
+      this.editorInterface = editorInterface;
```
rathboma, line 99: Let's remove this plz

apps/studio/src/components/common/texteditor/SQLTextEditor.vue
```diff
@@ -0,0 +1,179 @@
...
+      this.$emit("interface", editorInterface);
+    },
+    handleInitialized(cm: CodeMirror.Editor) {
```
rathboma, line 102: Same here, we don't want `Codemirror` leaking into parent components.

apps/studio/src/components/common/texteditor/SQLTextEditor.vue
```diff
@@ -0,0 +1,179 @@
...
+      registerAutoquote(cm);
+      this.$emit("initialized", ...arguments);
+    },
+    handlePaste(cm: CodeMirror.Editor, e) {
```
rathboma, line 106: `TextEditor` can handle paste, no?

apps/studio/src/components/common/texteditor/SQLTextEditor.vue
```diff
@@ -0,0 +1,179 @@
...
+      let clipboard = (e.clipboardData.getData("text") as string).trim();
+      clipboard = removeQueryQuotes(clipboard, this.identifyDialect);
+      if (this.hasSelectedText) {
+        cm.replaceSelection(clipboard, "around");
+      } else {
+        const cursor = cm.getCursor();
+        cm.replaceRange(clipboard, cursor);
+      }
+    },
+    handleKeyup(editor: CodeMirror.Editor, e) {
```
rathboma, line 117: don't like leaking the editor up here either.

apps/studio/src/components/common/texteditor/TextEditor.vue
```diff
@@ -0,0 +1,335 @@
...
+        this.$bks.openMenu({
+          ...menu,
+          options: customOptions,
+        });
+      }
+    },
+  },
+  mounted() {
+    this.initialize();
+    this.$emit("interface", {
```
rathboma, line 315: We shouldn't be leaking objects with methods in events like this. It needs to only be data we send upstream with props. This really breaks the communication model of Vue.

apps/studio/src/components/tableview/EditorModal.vue
```diff
@@ -41,7 +41,13 @@
         </div>
 
         <div class="editor-container">
-          <textarea name="editor" ref="editorRef" />
+          <text-editor
```
rathboma, line 44: love it.

apps/studio/src/lib/editor/languageData.ts
```diff
@@ -122,3 +125,86 @@ export function getLanguageByContent(content: string): LanguageData | undefined
...
 }
+
+type Language = ConnectionType | "sql" | "text" | "html" | "json";
+
+interface CodeMirrorLanguage {
+  mode: string | Record<string, unknown>;
+  hint?: unknown;
+}
+
+export function resolveLanguage(lang: Language): CodeMirrorLanguage {
```
rathboma, line 136: This is probably where we can put some of the autocomplete logic that is currently in SQLTextEditor

rathboma (changes requested):
This looks really really good. I think this simplifies the codebase a lot.

That said, this PR still does some things that aren't really in the spirit of how `Vue` expects components to communicate with each other (editorInterface).

If we can remove editorInterface I think we'll be good to merge.

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -31,13 +30,16 @@
...
+      <sql-text-editor
+        v-model="unsavedText"
+        :lang="connectionType"
+        :extra-keybindings="keybindings"
+        :vim-config="vimConfig"
+        :height="editorHeight"
+        :read-only="editorReadOnly"
+        @initialized="handleEditorInitialized"
+        @cursorActivity="handleCursorActivity"
+        @interface="editorInterface = $event"
```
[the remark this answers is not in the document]
azmy60: No, it's not CodeMirror specific. It's an interface for the `TextEditor` component. The idea is to avoid calling child functions directly, but the child provides an interface to the parent so later the parent should call anything that's available on that interface instead. I got this idea from this [answer](https://stackoverflow.com/a/70723343/10012118).

I'm trying to avoid `props/watcher` method, cause this can get so messy IMO. I think we can do event bus? `this.$emit()` and then `this.registerHandlers`? We don't have to put the listener names to the `AppEvent` though.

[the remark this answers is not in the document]
azmy60: I think I agree `editorInterface` is not the best. Maybe just do event bus.

azmy60:
@rathboma I removed the `editorInterface`. I'm finally able to update stuff just by props. But some [parts](https://github.com/beekeeper-studio/beekeeper-studio/pull/1907/commits/74aaf5baa43438c318cb2837550870bd590a08c8) really depends on the CodeMirror instance, like `removeQuotesFromQuery`, `autoComplete`, etc.. So I put these ones as plugins. Don't know if you'd like it or not, but let me know if it needs some changes.

rathboma (approved):
Fantastic work here. This was a LOT of refactoring.

rathboma:
@azmy60 Just need to resolve the merge conflicts, then merge away

PR #1909, opened 2024-01-19
Postgres inherited client

not-night-but:


rathboma:
@azmy60 I'll have you review the PostgreSQL version of this. Once it's finished, and you've reviewed it, tag @not-night-but so they can merge it!

rathboma:
@not-night-but how's this going? It's looking pretty complete aside from failing tests. Need help on the tests, or are you all good?

not-night-but:
@rathboma it is mostly done! The tests should be fine once I get some of azmy's changes in from the MySQL PR

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -39,1347 +38,1419 @@ const pgErrors = {
...
-    })
-    const item: TableIndex = {
-      table, schema,
-      id,
-      name: indexName,
-      unique,
-      primary,
-      columns
+  async listMaterializedViews(filter?: FilterOptions): Promise<TableOrView[]> {
+    if (!this.version.isPostgres || this.version.number < 90003) {
```
azmy60, line 780: Is it possible to avoid checking `version.isPostgres` here?

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -39,1347 +38,1419 @@ const pgErrors = {
...
-    FROM   pg_index i
-    JOIN   pg_attribute a ON a.attrelid = i.indrelid
-                        AND a.attnum = ANY(i.indkey)
-    WHERE  i.indrelid = ${tablename}::regclass
-    AND    i.indisprimary
-    ORDER BY a.attnum
-  `
+    // postgres 9 seems to freak out if the charset isn't wrapped in single quotes and also requires the template https://www.postgresql.org/docs/9.3/sql-createdatabase.html
+    // later version don't seem to care
+    if (isPostgres && versionAsInteger < 100000) {
```
azmy60, line 958: Can we remove `isPostgres`? Other classes should override this if needed.

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -39,1347 +38,1419 @@ const pgErrors = {
...
+   */
+  private async getVersion(): Promise<VersionInfo> {
+    const { version } = (await this.driverExecuteSingle("select version()")).rows[0]
+
+    if (!version) {
+      return {
+        version: '',
+        isPostgres: false,
+        isCockroach: false,
+        isRedshift: false,
```
azmy60, line 1257: I think we probably don't need these since we have different classes for each database.

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -39,1347 +38,1419 @@ const pgErrors = {
...
+      oid = '${wrapIdentifier(options.schema)}.${wrapIdentifier(options.table)}'::regclass
+  `
 
-  return buildInsertQuery(knex, tableInsert, columns, _.toString)
-}
+  // if we're not filtering data we want the optimized approximation of row count
+  // rather than a legit row count.
+  let countQuery = options.version.isPostgres && !filters && !options.forceSlow ? tuplesQuery : `SELECT count(*) as total ${baseSQL}`
+  if (options.version.isRedshift && !filters) {
+    countQuery = `SELECT COUNT(*) as total ${baseSQL}`
```
azmy60, line 1440: Probably make a method for `countQuery` and then override `countQuery` in other postgres classes, or anything to avoid using `version.isPostgres`.

apps/studio/src/lib/db/clients/cockroach.ts
```diff
@@ -0,0 +1,143 @@
...
+import pg, { PoolConfig } from "pg";
+import { IDbConnectionDatabase, IDbConnectionServer, IDbConnectionServerConfig } from "../types";
+import { SupportedFeatures, TableIndex, TablePartition, TableProperties, TableTrigger } from "../models";
+import { PostgresClient } from "./postgresql";
+import _ from 'lodash';
+
+
+export class CockroachClient extends PostgresClient {
+
+  constructor(server: any, database: IDbConnectionDatabase) {
```
azmy60, line 11: I think we can skip writing `constructor` if it's the same as the parent, no?

apps/studio/src/lib/db/clients/cockroach.ts
```diff
@@ -0,0 +1,143 @@
...
+      relations,
+      triggers,
+      partitions,
+      owner
+    };
+  }
+
+  protected async configDatabase(server: { sshTunnel: boolean, config: IDbConnectionServerConfig }, database: { database: string }) {
+    let optionsString = undefined;
+    const cluster = server.config.options?.cluser || undefined;
```
azmy60, line 97: is this typo? `cluser` -> `cluster`

package.json
```diff
@@ -39,5 +39,9 @@
     "node-abi": "^3.34.0",
     "cpu-features": "./.yarn/packages/empty-package",
     "electron-builder": "23.0.2"
+  },
+  "volta": {
+    "node": "16.19.1",
+    "yarn": "1.22.21"
```
azmy60, line 45: Is this intended? It's all good if it is. Maybe I'm missing something.

azmy60 (changes requested):
Great stuff! Just a couple of fixes before we can merge this.

apps/studio/src/lib/db/clients/postgresql.ts
not-night-but: Yup!

apps/studio/src/lib/db/clients/postgresql.ts
not-night-but: Yeah I'm a moron, completely forgot about these

apps/studio/src/lib/db/clients/cockroach.ts
not-night-but: Good catch

apps/studio/src/lib/db/clients/cockroach.ts
not-night-but: Yeah I originally had some extra configuration in there and just forgot to remove it after I refactored. 🤦

package.json
not-night-but: Didn't actually mean to check this in. It doesn't make a difference for building or anything, it's just for local machines if you happen to use Volta

azmy60: Oh ok. That's cool.

azmy60 (approved):
LGTM!

PR #1913, opened 2024-01-22
initial import/export sql files

### LINKED ISSUE #1857: FEAT: Import & export saved queries from/to .sql files.
Right now the only way to 'import' a SQL file is to drag/drop it into the editor window. If you drag and drop onto any other screen in the app, it doesn't work. This is just weird tbh.

In the future we need a way to work with a directory of SQL files more natively, but for now being able to import and export them makes sense.

## Import
- File -> Import Saved Queries (also add an import button to the saved queries tab)
  - lets you select multiple files on the file system, clicking 'import' adds them to the database as saved queries with the name of the file as the name of the query.
- Drag & Drop any .sql file into the app opens it in a new query tab. Dragging and dropping into an existing query editor works as it does now, opening that code in the current query editor

## Export
- Right click a saved query -> export - select a directory and filename (default is a snake_case version of the query  name with a date time extension - eg "My Awesome Query" becomes my_awesome_query_{yyyy-mm-dd-hh-ss}.sql. Default the directory to Documents. Remember the directory between exports.
- File -> Export Saved Queries - choose the directory, plus a searchable list of queries to export. Adds them to the directory using the naming scheme above.

azmy60:
fix #1857

### DIFF
--- a/apps/studio/src/common/utils.ts
+++ b/apps/studio/src/common/utils.ts
@@ -9,6 +9,7 @@ import _ from 'lodash';
 import platformInfo from './platform_info';
 import { format } from 'sql-formatter';
 import { TableFilter, TableOrView, Routine } from '@/lib/db/models';
+import { SettingsPlugin } from '@/plugins/SettingsPlugin';
 
 export function having<T, U>(item: T | undefined | null, f: (T) => U, errorOnNone?: string): U | null {
   if (item) return f(item)
@@ -85,6 +86,26 @@ export function readJSONFileSync(filename: string): any {
   const filePath = resolveHomePathToAbsolute(filename);
   const data = fs.readFileSync(path.resolve(filePath), 'utf-8');
   return JSON.parse(data);
+}
+
+export function readWebFile(file: File) {
+  const reader = new FileReader()
+  const result = new Promise<string>((resolve, reject) => {
+    reader.onload = () => {
+      resolve(reader.result as string)
+    }
+    reader.onerror = () => {
+      reject(reader.error)
+    }
+    reader.onabort = () => {
+      reject(new Error('File reading aborted'))
+    }
+  })
+  reader.readAsText(file)
+  return {
+    result,
+    abort: reader.abort,
+  }
 }
 
 export function createParentDirectory(filename: string): Promise<string> {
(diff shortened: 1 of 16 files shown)


apps/studio/src/plugins/NotyPlugin.ts
```diff
@@ -21,10 +21,10 @@ const VueNoty = {
     return new Noty(params)
   },
 
-  show(text, type = 'alert', opts = {}) {
+  show(text, type = 'alert', opts: { escapeHtml?: boolean } = {}) {
```
rathboma, line 24: let's make this `allowRawHtml`, rathar than `escapeHtml`, then the check is `if (opts.allowRawHtml)`, which is better

apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -253,7 +252,12 @@ export default Vue.extend({
     },
     determineLabelColor() {
       return this.config.labelColor == "default" ? '' : `connection-label-color-${this.config.labelColor}`
-    }
+    },
+    rootBindings() {
+      return [
+        { event: AppEvent.dropzoneDrop, handler: this.maybeLoadSqlite },
```
rathboma, line 258: Great integration of the SQLite dropzone

apps/studio/src/common/utils.ts
```diff
@@ -178,3 +199,31 @@ export function entityId(schema: string, entity?: TableOrView | Routine) {
...
+    for (let i = 0; i < dt.types.length; i++) {
+        if (dt.types[i] === "Files") {
+            return true;
+        }
+    }
+    return false;
+}
+
+export async function getLastExportPath(filename?: string) {
+  const lastExportPath = await SettingsPlugin.get('lastExportPath')
```
rathboma, line 214: God I hate our settings system. Good re-use of the plugin though.

apps/studio/src/common/utils.ts
```diff
@@ -178,3 +199,31 @@ export function entityId(schema: string, entity?: TableOrView | Routine) {
...
+
+  if (lastExportPath) {
+    return path.join(lastExportPath, filename)
+  }
+
+  if (platformInfo.isMac || platformInfo.isWindows) {
+    return path.join(homedir(), 'Documents', filename)
```
rathboma, line 221: I don't think you need to use `app.getPath("documents")` from the electron API as users can move their Documents folder.

rathboma: I actually think it's better to just use `homedir` anyway, as many software engineers don't put sql in the `Documents` folder.

rathboma (changes requested):
This looks awesome.

1 minor fix.
Please add information about this feature to the docs -- see docs/

rathboma (approved):

Issue #1925, opened 2024-01-30
BUG: Creating a new table triggers the unsaved changes dialog


azmy60:
**To Reproduce**
1. Create a new table
2. Click on `create table`
3. A new table is created, but this modal shows up

![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/38707148/915837c7-2b49-4f3f-85eb-d7d6cfe3e1ef)

**App Version**
master

**Solution**
Maybe mark the table builder as "not dirty" or "saved" after successfully creating a table so it doesn't trigger the modal.

PR #1935, opened 2024-02-03
Don't show unsaved changes warning modal after table creation

therealrinku:
Fixes: #1925

### DIFF
--- a/apps/studio/src/common/appdb/models/CloseTab.ts
+++ b/apps/studio/src/common/appdb/models/CloseTab.ts
@@ -1,0 +1,3 @@
+export interface CloseTabOptions {
+  ignoreUnsavedChanges?: boolean;
+}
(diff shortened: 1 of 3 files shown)


rathboma (changes requested):
Hey! Thanks for submitting this code. I think this is a great addition, but I think we should make one change to the API you are introducing.

Instead of passing an `action` to the close function, I think it'd be good if instead it was an optional object. This way it's a general purpose feature that any tab can use.

Something like:

```js
this.$root.$emit(AppEvent.closeTab, this.tabId, { ignoreUnsavedChanges: true })
```

Then the close signature is something like:

```ts
interface CloseTabOptions {
  ignoreUnsavedChanges?: boolean
}
async close(tab: OpenTab, options?: CloseTabOptions) {...}
```

therealrinku:
Hi @rathboma !
Thanks for the review. Yeah that make sense. I have made some changes for that.

I was also thinking about maybe adding the tabId property also to that options object 
but for now I just added **ignoreUnsavedChanges**  property only in there 😄

rathboma (approved):
Thanks so much!

Issue #1920, opened 2024-01-25
BUG: Copying JSON result in `[object Object]`


fauh45:
**Describe the bug**
When the row result from query is in JSON format (such as from `EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON)`), then copying it (by right click -> copy) it result with the clipboard containing `[object Object]` and not the result itself.

**To Reproduce**
1. Create a new query in a new tab
2. Do the following query

```sql
EXPLAIN (ANALYZE, COSTS, VERBOSE, BUFFERS, FORMAT JSON)
SELECT 1
````

3. Right click on the resulting row (there should be only one row)
4. Click copy then paste it somewhere
5. The clipboard content should only be `[object Object]`

**Expected behavior**
The result of the action of copying the row should just be copying the raw result of the query to the clipboard.

**Screenshots / Gifs**
The action done in the screenshot below,
![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/7534613/04b4c440-1123-4c81-910b-9d69131b40be)

Result in the text shown below,
![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/7534613/38d98a11-98d6-4087-a022-bc02943b0dd1)

**Version Information (please complete the following information):**
 - OS: Pop!_OS 22.04 LTS x86_64
 - App Version: 4.1.13
 - Database type and version: PostgreSQL 14

PR #1936, opened 2024-02-04
Copying single row resulting in object object fixed

therealrinku:
fixes: #1920

### DIFF
--- a/apps/studio/src/common/utils.ts
+++ b/apps/studio/src/common/utils.ts
@@ -220,3 +220,22 @@ export async function getLastExportPath(filename?: string) {
 export async function setLastExportPath(exportPath: string) {
   await SettingsPlugin.set('lastExportPath', exportPath)
 }
+
+// Stringify all the arrays and objects in range data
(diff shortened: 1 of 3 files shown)


azmy60:
@therealrinku This is great! Thanks for submitting the PR! :) Would you mind taking a look if this also happens with the other formats? (JSON/TSV/etc.)

therealrinku:
Hey @azmy60 ! 
sure, I just checked and turns out there is same issue with TSV and Markdown as well.
I will push some changes for that too as soon as I can 😄

therealrinku:
@azmy60 I was looking into fixing the same for tsv and markdown as well.
I'm thinking about maybe converting all the values from options.range.getData() to string before going to switch statement.

Let me know what do you think about that or if you have any other ideas on how to handle this better ?

azmy60:
Yep. I think that sounds good! :)

azmy60 (approved):
@therealrinku This is awesome! Love the test! I also added a little change to your PR so we could prevent regression issues. :)

PR #1953, opened 2024-02-13
Remove DBConnection and such, Move Bigquery to new class

### LINKED ISSUE #1852: Move BigQuery to the `BasicDatabaseClient` abstract class
Let's move this to the client

TODO:
- Also set up Will with a keyfile for BigQuery

not-night-but:
Close #1852


apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
```diff
@@ -1,9 +1,13 @@
...
 import { AlterPartitionsSpec, AlterTableSpec, IndexAlterations, RelationAlterations, TableKey } from '@shared/lib/dialects/models';
 import { buildInsertQuery } from './utils';
 import { Knex } from 'knex';
-import { DatabaseClient, DatabaseElement } from '../client';
 import { ChangeBuilderBase } from '@shared/lib/sql/change_builder/ChangeBuilderBase';
+import { ConnectionType, DatabaseElement, IDbConnectionDatabase, IDbConnectionServer } from '../types';
+import createLogger from '@/lib/logger';
+import connectTunnel from '../tunnel';
+
+const logger = createLogger('db');
```
rathboma, line 10: Oh, we shouldn't be using this. Let's remove it. We should use `electron-log` instead

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
```diff
@@ -30,42 +34,78 @@ export interface AppContextProvider {
...
     this.knex = knex;
     this.contextProvider = contextProvider
-  }
-  listTablePartitions(_table: string, _schema: string): Promise<TablePartition[]> {
-    return Promise.resolve([])
-  }
-  alterPartitionSql(_changes: AlterPartitionsSpec): string {
-    return ''
+    this.server = server;
+    this.database = database;
```
rathboma, line 49: How do we get rid of this `server` and `database` nonsense? Like what is the `database`?

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
```diff
@@ -30,42 +34,78 @@ export interface AppContextProvider {
...
   defaultSchema(): string | null {
     return null
   }
+  // ****************************************************************************
 
-  abstract getBuilder(table: string, schema?: string): ChangeBuilderBase
+  // Connection *****************************************************************
+  async connect(): Promise<void> {
+    /* eslint no-param-reassign: 0 */
+    if (this.database.connecting) {
```
rathboma, line 66: On the topic of the `database` variable

Don't we just keep the state internally? I'm not sure the value of the `database` object, it just seems confusing?

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
```diff
@@ -30,42 +34,78 @@ export interface AppContextProvider {
...
+    try {
+      this.database.connecting = true;
+
+      // terminate any previous lost connection for this DB
+      if (this.database.connected) {
+        this.disconnect();
+      }
+
+      // reuse existing tunnel
+      if (this.server.config.ssh && !this.server.sshTunnel) {
```
rathboma, line 79: So here we only use `config` from `server`. Calling it `server` has always been confusing to me, shouldn't we just accept the `config` portion of it?

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
rathboma: Is it to provide state to the front-end? Can't we just have a `get status()` function in the BasicDatabase client to handle that?

apps/studio/src/lib/db/client.ts
```diff
@@ -24,581 +30,10 @@ export class ClientError extends Error {
...
-export interface IDbConnectionDatabase {
-  database: string,
-  connection: Nullable<DatabaseClient>,
-  connecting: boolean,
-}
-
-export class DBConnection implements DatabaseClient {
-  connectionType = this.server.config.client
-  constructor (private server: IDbConnectionServer, private database: IDbConnectionDatabase) {}
-  supportedFeatures = supportedFeatures.bind(null, this.server, this.database)
```
rathboma, line 162: DIE DIE DIE double method binding :heart:

rathboma (approved):
This looks great. Lots of work.

Not a criticism of your work, but more general question:
- What are 'server' and 'database' and why do we need them?

While I talk about this in my review, I don't think this is a topic for this PR, but definitely we should also kill them with :fire: in a follow-up.

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
rathboma: @not-night-but This is the one thing to change before merging, be sure to use `electron-log` and `rawLog.scope('db')`, then kill this old file.

rathboma:
(I know this is still draft, but changes look good so far)

not-night-but:
@rathboma server and database are just used to configure the connection. There are a lot of changes I'd like to make in the future in general, as this is now a class I think we can definitely add some more state to it. For instance, bigquery uses datasets way more than we use databases in other DBs, so we have to use `this.database.database` to actually get a dataset. Ideally I'd like to just have the connection keep track of which database we're looking at. There's a bunch of little things like that that I think would make it much easier to work with

apps/studio/src/components/export/ExportManager.vue
```diff
@@ -22,13 +22,13 @@ import Vue from 'vue'
...
 import { mapMutations, mapGetters } from 'vuex'
 import { AppEvent, RootBinding } from '../../common/AppEvent'
-import { DBConnection } from '../../lib/db/client'
 import { TableFilter, TableOrView } from '../../lib/db/models'
 import ExportNotification from './ExportNotification.vue'
 import ExportModal from './ExportModal.vue'
 import { CsvExporter, JsonExporter, JsonLineExporter, SqlExporter } from '../../lib/export'
 import { ExportProgress, ExportStatus } from '../../lib/export/models'
 import globals from '@/common/globals'
+import { BasicDatabaseClient } from '@/lib/db/clients/BasicDatabaseClient'
```
rathboma, line 31: Don't think this is used

apps/studio/src/lib/db/clients/oracle.ts
```diff
@@ -59,9 +59,7 @@ export class OracleClient extends BasicDatabaseClient<DriverResult> {
   readOnlyMode: boolean
 
   constructor(server: IDbConnectionServer, database: IDbConnectionDatabase) {
-    super(knexLib({ client: 'oracledb'}))
-    this.server = server
-    this.database = database
+    super(knexLib({ client: 'oracledb'}), NoOpContextProvider, server, database);
```
rathboma, line 62: NIT: You should put `ContextProvider` at the end of the constructor arguments, then it can just have a default of `NoOpContextProvider` and we don't need to explicitly provide it in all the drivers.

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -62,22 +61,26 @@ interface STQResults {
 
 }
 
+const postgresContext = {
```
rathboma, line 64: NIT: You should use the `NoOpContextProvider`

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -62,22 +61,26 @@ interface STQResults {
...
+  
   constructor(server: IDbConnectionServer, database: IDbConnectionDatabase) {
-    super(knex);
-
+    super(knex, postgresContext, server, database);
     this.dialect = 'psql';
-    this.dbReadOnlyMode = server?.config?.readOnlyMode || false;
-    this.server = server;
-    this.database = database;
+    this.readOnlyMode = server?.config?.readOnlyMode || false;
```
rathboma, line 83: FUN: you can use double-bang: `!!server?.config?.readOnlyMode`

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -643,6 +652,7 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult> {
     } catch (ex) {
       log.error("query exception: ", ex)
       await this.driverExecuteSingle('ROLLBACK');
+      this.releaseCachedConnection();
```
rathboma, line 655: should this be in `catch` or in `finally`?

apps/studio/src/lib/db/clients/sqlite.ts
```diff
@@ -60,14 +58,14 @@ const SD = SqliteData;
...
+  databasePath: string;
 
   constructor(server: IDbConnectionServer, database: IDbConnectionDatabase) {
-    super(knex, sqliteContext);
+    super(knex, sqliteContext, server, database);
 
     this.dialect = 'sqlite';
-    this.dbReadOnlyMode = server?.config?.readOnlyMode || false;
-    this.database = database?.database;
+    this.readOnlyMode = server?.config?.readOnlyMode || false;
```
rathboma, line 67: NIT: I see this repeated a lot, probably means it should be in `BasicDatabaseClient`

apps/studio/src/lib/db/types.ts
```diff
@@ -1,32 +1,8 @@
...
-import type { default as cockroach } from './clients/cockroach'
-import type { default as redshift } from './clients/redshift'
-import type { default as sqlserver } from './clients/sqlserver'
-import type { default as sqlite } from './clients/sqlite'
-import type { default as cassandra } from './clients/cassandra'
-import type { default as mariadb } from './clients/mariadb'
-import type { default as bigquery } from './clients/bigquery'
-import type { default as firebird } from "./clients/firebird";
-import type { default as oracle } from "./clients/oracle"
+import type { RedshiftOptions, BigQueryOptions, CassandraOptions } from '@/common/appdb/models/saved_connection';
```
rathboma, line 2: We can't use `models` inside of `lib`. We should add these interfaces to `shared` if we need them in both places

rathboma: (like alongside the dialect interfaces)

apps/studio/src/lib/export/formats/json.ts
```diff
@@ -16,7 +16,7 @@ export class JsonExporter extends Export {
 
   constructor(
     filePath: string,
-    connection: DBConnection,
+    connection: BasicDatabaseClient<any>,
```
rathboma, line 19: I think you mean to use `DatabaseClient` here, `BasicDatabaseClient` is an abstract implementation, but there could be others.

apps/studio/src/lib/export/formats/jsonline.ts
```diff
@@ -10,7 +10,7 @@ export class JsonLineExporter extends Export {
 
   constructor(
     filePath: string,
-    connection: DBConnection,
+    connection: BasicDatabaseClient<any>,
```
rathboma, line 13: ditto

apps/studio/tests/integration/lib/db/clients/postgres.spec.ts
```diff
@@ -400,20 +400,20 @@ function testWith(dockerTag, socket = false, readonly = false) {
...
       }
+
       const query = util.connection.query(`
-          CREATE TABLE public.withquestionmark (
-            "approved?" boolean NULL DEFAULT false,
-            str_col character varying(255) NOT NULL,
-            another_str_col character varying(255) NOT NULL PRIMARY KEY
-          );
-        `);
+        CREATE TABLE IF NOT EXISTS public.withquestionmark (
```
rathboma, line 405: NIT: Shouldn't need to do this, if the table already exists, then this was run twice by mistake

rathboma (approved):
Great work, this was a big effort!

I have some nitpicks that need fixing, the biggest being not importing stuff from `model` in the `lib` directory, but feel free to merge this PR then fix that in a new PR.

Also this just got merged: https://github.com/beekeeper-studio/beekeeper-studio/pull/1953,
so you might need to make this change to the PostgreSQL client too somewhere.

apps/studio/src/lib/export/formats/json.ts
not-night-but: We don't have DatabaseClient anymore

Issue #1768, opened 2023-11-01
FEAT: Keybind for reopening recently closed tab


Z3rio:
**Is your feature request related to a problem? Please describe.**
No, this is not related to an issue, it is simply an idea/feature request

**Describe the solution you'd like**
**_Ctrl + Shift + T_** should reopen the most recently closed tab.
Just like it does in most web browsers, code editors, and such.
It has become quite a universal key bind, so why not include it here?

**Why should this be added?**
Often, we can accidentally close important tabs, especially when using Ctrl + W quickly.
And being able to quickly reopen the accidentally closed tab would be great.

**Describe alternatives you've considered**
As I am a bit new to using Beekeeper Studio, I'm unaware of other solutions.
If there are any, then please tell me.

**Database type, if appropriate**
Universal (aka not appropriate)

Z3rio:
@rathboma Would love to get your input on this issue, as I'm thinking about starting to work on it and it would be a bit wasteful to work on it if it isn't wanted/needed.

azmy60:
Hi @Z3rio ! I think it's reasonable to support `Ctrl+Shift+T` keybinding. Feel free to submit a PR! : )

Z3rio:
Thank you, I'll see if I can make some time for this :)

PR #1963, opened 2024-02-20
feat: shortcut added for reopening last closed tab

therealrinku:
addressing: #1768 

added Ctrl+Shift+T shortcut to re-open recently closed tab


apps/studio/src/components/CoreTabs.vue
```diff
@@ -344,6 +345,7 @@ export default Vue.extend({
     },
     keymap() {
       const result = {
+        'ctrl+shift+T': this.reopenLastClosedTab,
```
rathboma, line 348: I think this is probably better to store this value in the TabModule of the vue store -- that way we don't need to remember to set `lastClosedTab` anywhere we close a tab.

Some other things we'll need to do here:
- We'll want to create a new `OpenTab` with a clone of the values in the specific tab because this is an ORM model and will include an `id`, but when `$store.dispatch('tabs/remove', tab)` is called we delete it from the database. So we'll need to add an optional constructor argument to copy values from.
- We should store array of `recently closed tabs` in the store, that way calling `reopenClosedTab` can be clicked multiple times until the stack runs out!

With this new setup, to restore a tab we could just call `this.$store.dispatch('reopenTab')` and it will re-add the tab most recently closed

rathboma (changes requested):
Thank you so much for contributing this! I think your logic is sound, but I've added a few pointers to help us do this in the right place.

apps/studio/src/components/CoreTabs.vue
therealrinku: Hi @rathboma 
thanks for the review! yeah that all make sense.. I'll take a look 😄

apps/studio/src/store/modules/TabModule.ts
```diff
@@ -53,6 +55,16 @@ export const TabModule: Module<State, RootState> = {
     },
     setActive(state, tab?: OpenTab) {
       state.active = tab
+    },
+    addLastClosedTab(state, tabs: OpenTab[]){
```
rathboma, line 59: don't think you need this, it can just be done in the `remove` mutation.

so remove would do:

```js
    remove(state, tab: OpenTab) {
      state.tabs = _.without(state.tabs, tab)
      state.lastClosedTabs.push(tab)
    },
```

apps/studio/src/store/modules/TabModule.ts
```diff
@@ -96,6 +111,7 @@ export const TabModule: Module<State, RootState> = {
     },
     async remove(context, rawItems: OpenTab | OpenTab[]) {
       const items = _.isArray(rawItems) ? rawItems : [rawItems]
+      context.commit("addLastClosedTab", items)
```
rathboma, line 114: Then you don't need this!

rathboma (changes requested):
Almost perfect!

apps/studio/src/store/modules/TabModule.ts
```diff
@@ -50,9 +52,17 @@ export const TabModule: Module<State, RootState> = {
...
       state.tabs = _.without(state.tabs, tab)
+      state.lastClosedTabs.push(tab)
     },
     setActive(state, tab?: OpenTab) {
       state.active = tab
+    },
+    async reopenLastClosedTab(state){
+      const lastClosedTab = state.lastClosedTabs.pop()
+      if(lastClosedTab){
+        state.tabs.push(lastClosedTab)
```
rathboma, line 63: Hey! Unfortunately once the tab has been removed it has lost it's ID, and has been deleted from the database -- see actions -> remove.

Probably you'll need to make a copy of it and send it through actions/add to make sure it's saved to the database

apps/studio/src/store/modules/TabModule.ts
```diff
@@ -78,6 +88,9 @@ export const TabModule: Module<State, RootState> = {
       context.commit('set', [])
       context.commit('setActive', null)
     },
+    async reopenLastClosedTab(context) {
+      context.commit('reopenLastClosedTab')
```
rathboma, line 92: Here it should probably get the last closed tab, then call context.dispatch('add', clonedTab)

rathboma (changes requested):
Hey! This still needs one more change to account for the fact that the original tab was deleted from the database.

therealrinku:
Hi @rathboma 
thanks for the review. I think it's already been added to the database when we call reopenLastClosedTab

here is the video. Let me know if it's okay or I'm missing something here



https://github.com/beekeeper-studio/beekeeper-studio/assets/76877078/00ccf3ab-c93d-491e-bb2d-ccfa224be9bb

rathboma:
Hey, it will work, but when you close and re-open Beekeeper Studio those tabs will be missing, because they were deleted from the database. That's why the tabs need to go through the sequence I described.

rathboma:
@azmy60 this doesn't require much to fix for merging, see the comments above. You should be able to commit directly to this branch.

apps/studio/src/store/modules/TabModule.ts
```diff
@@ -51,8 +53,14 @@ export const TabModule: Module<State, RootState> = {
     remove(state, tab: OpenTab) {
       state.tabs = _.without(state.tabs, tab)
     },
+    addLastClosedTab(state, tab: OpenTab) {
```
rathboma, line 56: this can just happen in `remove`

apps/studio/src/store/modules/TabModule.ts
```diff
@@ -78,6 +86,18 @@ export const TabModule: Module<State, RootState> = {
       context.commit('set', [])
       context.commit('setActive', null)
     },
+    async reopenLastClosedTab(context) {
+      for (let i = context.state.lastClosedTabs.length - 1; i >= 0; i--) {
+        const tab = context.state.lastClosedTabs[i]
+        // Does this tab belong to the current connection?
```
rathboma, line 92: Do you also have to check workspace id?

rathboma: The `OpenTab` here has still been removed from the database and doesn't have an ID, so it still needs saving.

Also -- what happens with the position of the tab? Does it go back to the right place?

apps/studio/src/store/modules/TabModule.ts
```diff
@@ -96,7 +123,10 @@ export const TabModule: Module<State, RootState> = {
     },
     async remove(context, rawItems: OpenTab | OpenTab[]) {
       const items = _.isArray(rawItems) ? rawItems : [rawItems]
-      items.forEach((i) => context.commit('remove', i))
+      items.forEach((tab) => {
+        context.commit('addLastClosedTab', tab.duplicate())
```
rathboma, line 127: no need for this, it can happen in remove

apps/studio/src/store/modules/TabModule.ts
```diff
@@ -86,6 +106,13 @@ export const TabModule: Module<State, RootState> = {
         item.connectionId = usedConfig.id
         await item.save()
       }
+
+      // Prevent multiple tabs per table
+      const existingTabInClosedTabs = context.state.lastClosedTabs.find((tab) => tab.matches(item))
+      if (existingTabInClosedTabs) {
```
rathboma, line 112: this can just happen in `add`, no need to put it here else we'll forget about it.

apps/studio/src/store/modules/TabModule.ts
```diff
@@ -78,6 +86,18 @@ export const TabModule: Module<State, RootState> = {
       context.commit('set', [])
       context.commit('setActive', null)
     },
+    async reopenLastClosedTab(context) {
+      for (let i = context.state.lastClosedTabs.length - 1; i >= 0; i--) {
```
rathboma, line 90: Why are we looping here? We only want the very last item in the lastClosedTabs array.

rathboma (changes requested):
def still needs some work

apps/studio/src/store/modules/TabModule.ts
azmy60: When the `lastClosedTabs` is filled, and then we disconnect from current connection to connect to a new connection, we need to search for tabs that belong to the right connection. Using `array.pop` would get the last item, but that item could be from a different `connectionId`. Another solution to this is I was thinking if we should just wipe out the stack so it's empty whenever we connect to a new connection, but maybe it's better to do the first one?

apps/studio/src/store/modules/TabModule.ts
azmy60: I'll look into the workspace id and the position of the tab.

I was under the impression that we want this to behave like vscode, so no need to save `lastClosedTabs` to database. But I just realized that chrome persists the closed tabs even though I exit the program. It works too.

azmy60: Should we even not remove tabs from database and just mark it as "closed"?

rathboma:
@azmy60 if this is good, approve and merge

azmy60 (approved):

azmy60:
@rathboma I can't merge

PR #1964, opened 2024-02-20
feat: add a hasDefault to the extendedTableColumn objects

### LINKED ISSUE #1959: MySQL Should properly identify default values
Moved from the ultimate repo. Sometimes only 'extra' has information about a default value, this is wrong!

This is related to an import bug, please link here when you find it!

wmontgomery:
This will help to fix a MySQL bug on import that was not properly determining whether a column really was an autogenerated value

resolves #1959 (still has a fix in ultimate to make)


apps/studio/src/lib/db/clients/firebird.ts
```diff
@@ -366,6 +366,7 @@ export class FirebirdClient extends BasicDatabaseClient<FirebirdResult> {
           defaultValue,
           nullable,
           primaryKey,
+          hasDefault: Boolean(defaultValue),
```
rathboma, line 369: This won't work if the defaultValue is 0. You should use `!_.isNil`

rathboma: image

rathboma: image

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -333,6 +333,7 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult> {
       nullable: row.is_nullable === 'YES',
       defaultValue: row.column_default,
       ordinalPosition: Number(row.ordinal_position),
+      hasDefault: Boolean(row.column_default)
```
rathboma, line 336: see above

rathboma (changes requested):
0 and "" are valid default values

apps/studio/src/lib/db/clients/firebird.ts
wmontgomery: Didn't think about those being valid default values. Good catch!

rathboma (changes requested):
Few things:
- This needs reviewing after all of Day's merges to make sure it is in every driver
- This needs tests! For each driver we should check `hasDefault` is set, for mysql and mariadb we should check the specific bug we are fixing

rathboma:
@wmontgomery Hey Will, this still needs some tests -- can you add them and fix the merge conflicts so we can merge this?

wmontgomery:
No problemo. I'll get that knocked out tomorrow night.

rathboma (approved):

PR #1966, opened 2024-02-20
make the app aware of generated columns

### LINKED ISSUE #1804: BUG: SQLite generated columns not displayed
SQLite generated columns are not displayed in the UI. 

- Generated columns are displayed and there is some UI indication that the column is generated.
- When using the `Structure` UI, generated columns are annotated as such

Using another tool or sqlite3 cli

**Additional context**
It looks like this is because `PRAGMA table_info(...);` is used. This doesn't include generated columns.

`PRAGMA table_xinfo(...);` includes generated columns.

SQLite

azmy60:
fix #1804


apps/studio/src/lib/db/clients/sqlite.ts
```diff
@@ -135,7 +135,7 @@ export class SqliteClient extends BasicDatabaseClient<SqliteResult> {
 
   async listTableColumns(db: string, table?: string, _schema?: string): Promise<ExtendedTableColumn[]> {
     if (table) {
-      const sql = `PRAGMA table_info(${SD.escapeString(table, true)})`;
+      const sql = `PRAGMA table_xinfo(${SD.escapeString(table, true)})`;
```
rathboma, line 138: One big change we'll need to go along with this:

When someone is editing table data, Beekeeper would let them edit this column. That's wrong as you can't update a generated column.

## This PR will need:
- Some way for the app to understand that a column is read only and disable editing of data in that column. 
- Some visual indicator in the tabletable column header to indicate that the column is generated (maybe edit_off from materialize)
- Some sort of addition to the table structure view so that folks can see it is generated. We don't need support for creating generated_columns as part of this PR
- Additions to the `TableColumn` interface to indicate the column is read only. Maybe something like `readOnly`, or `generated`.

rathboma (changes requested):
See large comment below

psychedelicious:
Thanks for working on this, it's a repeated pain point for me. Feel free to @ me for user testing if that's useful (happy to build the app from PR branch).

apps/studio/src/lib/db/clients/sqlite.ts
azmy60: Right, totally missed that.

azmy60: Many databases we support can have generated columns. So I'm probably gonna add this for those databases as well, as long as it's not too complicated.

rathboma (approved):
This looks :fire:. Pretty big merge conflict with all of Day's changes though. Once you've resolved the conflicts, feel free to merge.

PR #1981, opened 2024-02-26
add row as array flag to sqlite `executeQuery`

### LINKED ISSUE #1976: BUG: SQLite3 result columns overwrite previous columns with the same name
**Describe the bug**

When SELECTing columns that end up with the same name, the result table overwrites the first column.

**To Reproduce**

1. Open a SQLite3 database to the query screen
2. Run any command that results in a table with duplicate column names - smallest example is probably `SELECT 1 as id, 2 as id;`
3. Check the results window to show that only one column with that name appears

**Expected behavior**

I expect to see both columns end up in the result window.

**Screenshots / Gifs**
If applicable, add screenshots or GIFS to help explain your problem.

### Screenshot of overwritten column in SQLite

image

### This does not occur in the SQLite3 shell

```
$ sqlite3
-- Loading resources from /Users/bbkane/.sqliterc
SQLite version 3.45.1 2024-01-30 16:01:20
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> SELECT 1 as id, 2 as id;
┌────┬────┐
│ id │ id │
├────┼────┤
│ 1  │ 2  │
└────┴────┘
```

 - OS: macOS
 - App Version: 4.1.13
 - Database type and version: SQlite 3.45.1 (from the sqlite shell)

**Additional context**

This does not occur for a Beekeeper MySQL connection, so I have a hunch its the SQLite3 library at fault

image

### LINKED ISSUE #1978: FEAT: Use .raw instead of .all for better-sqlite3
Using .all causes queries with multiple columns having the same name to conflict. See #1977 .

The solution is to use .raw and .columns, not .all:

https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md#rawtogglestate---this

azmy60:
fix #1978, fix #1976


rathboma (changes requested):
This should have a new integration test to make sure queries that return duplicate column names work right!

azmy60:
Doesn't `queryTests` cover the duplicate columns? I removed the filter to run `queryTests` from `sqlite`.

rathboma:
Oh, does it? If so then good to merge

rathboma (approved):

Issue #1994, opened 2024-03-04
Pre-release regressions that need fixing


rathboma:
- [x] Opening a new tab doesn't focus the codemirror editor
- [x] Switching to an editor tab doesn't focus codemirror unless it has previously been focused
- [x] Remove the kebab menu (three dots) from table column headers -- right click is good enough and this gives more space for the column headers
- [x] Foreign key clicks don't navigate to the correct record when a tab for that table is already open [Slack Link to gif](https://beekeeperstudio.slack.com/archives/C01T11HCK16/p1709526809294989)
- [x] Table tabs show [all] when a filter is applied, should show [filtered] or [id]
- [x] Codemirror left gutter is a different width vs the result table gutter (row headers), looks funny
- [x] If you have a saved firebird connection, double clicking in the sidebar will connect you, even though these are labelled as ultimate-only
- [x] Exporting to CSV adds an empty column at the beginning
- [ ] Query Magics mess up alignment - https://github.com/beekeeper-studio/beekeeper-studio/issues/1986
- [ ] Other one is the index view in structures sometimes gets all whacked out. I haven't successfully recreated with steps just randomly which is great [Slack link to image](https://beekeeperstudio.slack.com/archives/C01T11HCK16/p1709526333579559)
- [x] adding/renaming a column in the structure view isn't working. By adding/renaming I mean you can't change the name of the column. [Slack link for gif](https://beekeeperstudio.slack.com/archives/C01T11HCK16/p1709526246847369?thread_ts=1709526213.450079&cid=C01T11HCK16)
- [x] Closing a tab modal has broken flex style 
![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/279769/be3bc1c1-5ac4-481e-9ff2-dbf3b0119a65)

rathboma:
@azmy60 I marked some of these as complete, did I get them all?

azmy60:
Some of them are complete. I can mark them for you.

PR #1998, opened 2024-03-04
fix focusing text editor each time opening the query tab

azmy60:
Focusing a new tab and a previous tab should focus the text editor.

ref #1994 

fixing number 1 and 2

>  - Opening a new tab doesn't focus the codemirror editor
>  - Switching to an editor tab doesn't focus codemirror unless it has previously been focused


rathboma (approved):
we should at some point figure out what is stealing the focus

PR #2000, opened 2024-03-05
hide tabulator header icon menu

azmy60:
ref #1994 

fix number 3

> Remove the kebab menu (three dots) from table column headers -- right click is good enough and this gives more space for the column headers

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -532,11 +532,11 @@ export default Vue.extend({
           headerTooltip += ' [Primary Key]'
         }
 
-        let cssClass: string;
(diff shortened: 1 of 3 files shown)


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -532,11 +532,11 @@ export default Vue.extend({
           headerTooltip += ' [Primary Key]'
         }
 
-        let cssClass: string;
+        let cssClass = 'hide-header-menu-icon';
```
rathboma, line 535: Going to accept this, but a clean way to do this in future is using an array:

cssClasses = ['x']
cssClasses.join(" ")

rathboma (approved):

PR #2001, opened 2024-03-05
fix table filter builder that partially works

azmy60:
ref #1994 

fix number 4

> Foreign key clicks don't navigate to the correct record when a tab for that table is already open.


rathboma (approved):
lgtm

PR #2004, opened 2024-03-06
Fix codemirror gutters & ultimate connection interface

azmy60:
ref #1994 

fixing 2 issues:

- Codemirror left gutter is a different width vs the result table gutter (row headers), looks funny:
  - the fix: Moving settings button to status bar image
  - increase left gutter size and align the horizontal scrollbar to table image

- If you have a saved firebird connection, double clicking in the sidebar will connect you, even though these are labelled as ultimate-only
  - Note: It was also possible to connect to ultimate connection with right clicking > and then clicking `Connect`. This PR hides the `connect` option if it's ultimate.


apps/studio/src/assets/styles/app/query-editor.scss
```diff
@@ -112,6 +102,24 @@
 
   }
 
+  .CodeMirror {
+    padding-bottom: 0.25rem;
+  }
+
+  .CodeMirror-linenumbers {
+    width: 22px;
```
rathboma, line 110: Can these be `rem` based, or are we trying to match Tabulator's widths?

rathboma (approved):

rathboma:
@azmy60 sorry merge conflict now. Fix that then merge is good

apps/studio/src/assets/styles/app/query-editor.scss
azmy60: Yes, it can. I was trying to match the tabulator, but it should be easy to convert that.

PR #2023, opened 2024-03-12
add hidden column indicator

### LINKED ISSUE #1946: FEAT: Show some indication when columns are hidden in TableTable
There are two columns hidden in the table screenshot below, but there is no visual indication of this:

image

Previously we would make the :eye: button yellow, but now we've removed this button we need some other visual indicator.

## Implementation

- Let's add the 'hidden items' indicator to the statusbar to the left of the buttons.
- 
Hidden columns

azmy60:
fix #1946 

image

- add hidden column indicator in statusbar
- it should have a tooltip that has a link to open the column filter modal
image
- add `hide column` option to column header menu so we're able to hide them right away instead of opening the modal first
image

Note: I ended up using plain css to make the tooltip just like the hidden entities. `<v-popover>` (from v-tooltip), and `<x-tooltip>` (xel) didn't work. `<v-popover>` threw a bunch of weird errors for some reason.

### DIFF
--- a/apps/studio/src/assets/styles/app/_all.scss
+++ b/apps/studio/src/assets/styles/app/_all.scss
@@ -17,4 +17,5 @@
 @import './tabs/tabletable.scss';
 @import './tabs/table-builder.scss';
 @import './modals.scss';
+@import './tooltips.scss';
 @import './ultimate-only.scss';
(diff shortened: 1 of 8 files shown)


azmy60:
@rathboma since we renamed the `show and hide columns` in the cog menu to have better consistency. I feel like we should do that to the `Open Column Filter` in column header too, no?

image

rathboma (approved):

Issue #1985, opened 2024-02-28
BUG: column name drop down do not show while writing query in sql


mahimaholla:
While writing the select query in sql server the column dropdown do not show column names when a letter is typed.

![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/48175865/6b2f5328-fd02-4ae4-90e6-bcdc2477c830)
it shows before typing a letter but when a letter is typed then 
![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/48175865/f1d9f6f0-7064-46ba-aa1d-2a8115ecb9be)

 - OS: [ Windows]
 - App Version 4.1.13 (the same bug existed in previous versions as well)
 - Microsoft SQL Server 2016 (RTM-GDR) (KB4019088) - 13.0.1742.0 (X64) 
	Jul  5 2017 23:41:17 
	Copyright (c) Microsoft Corporation
	Web Edition (64-bit) on Windows Server 2016 Datacenter 6.3 <X64> (Build 14393: ) (Hypervisor)

azmy60:
That's really bad. Thanks for submitting this!

PR #2024, opened 2024-03-14
Fix columns autocomplete in SQLTextEditor

not-night-but:
This PR fixes the columns auto complete, in the case that we had not loaded the columns for a table yet, we would load the columns into the store, but not reload the selected table we were returning for the autocomplete.


azmy60:
@not-night-but does this fix this one? #1985

not-night-but:
@azmy60 I think it might? I haven't seen that behaviour before, this was to fix an issue that was emailed in.

azmy60:
Alright. I did a quick test, and it's a separate issue.

rathboma (approved):

PR #2064, opened 2024-03-29
Debugging possible pool client leaks in MySQL and PostgreSQL

rathboma:
- added some pool listeners
- Fixed some pool race conditions

STILL TODO:

- [ ] SQL Server should create the connection pool on connect, and use it properly in runWithConnection - https://github.com/beekeeper-studio/beekeeper-studio/blob/master/apps/studio/src/lib/db/clients/sqlserver.ts#L929
- [ ] Audit cassandra.ts for race conditions & mis-use of client pools
- [ ] Audit BigQuery for race conditions and mis-use of client pools


apps/studio/src/lib/db/clients/sqlserver.ts
```diff
@@ -914,23 +915,18 @@ export class SQLServerClient extends BasicDatabaseClient<SQLServerResult> {
...
-      : this.runWithConnection(runQuery)
+    return this.runWithConnection(runQuery)
   }
 
   private async runWithConnection(run) {
-    const connection = await new ConnectionPool(this.dbConfig).connect()
-    this.connection = connection
-    this.connection.dbConfig = this.dbConfig
-    return run(this.connection)
+    const connection = this.pool.request();
```
rathboma, line 928: @not-night-but don't you need to `release` the connection after calling `run`?

apps/studio/src/lib/db/clients/sqlserver.ts
```diff
@@ -692,7 +693,7 @@ export class SQLServerClient extends BasicDatabaseClient<SQLServerResult> {
     return {
       totalRows: undefined,
       columns: undefined,
-      cursor: new SqlServerCursor(this.connection, query, chunkSize),
+      cursor: new SqlServerCursor({ dbConfig: this.dbConfig }, query, chunkSize),
```
rathboma, line 696: @not-night-but The cursor shouldn't make it's own new connection, it should take the existing `pool` and use that.

apps/studio/src/lib/db/clients/sqlserver.ts
not-night-but: I don't think so?

image

rathboma: Huh, that's really weird

rathboma: Is `Request` the right thing here?

not-night-but: I believe that's what we were using before, but I'll see if there's something more idiomatic

Issue #1457, opened 2022-12-19
FEAT: Make the sidebar maintain its width


silverberrylog:
**Is your feature request related to a problem? Please describe.**
Whenever I open the application, connect or disconnect from a database, the sidebar returns to its original width.

**Describe the solution you'd like**
Since the user is able to change the width of the sidebar, the application should remember the user's preferences.

**Describe alternatives you've considered**

**Additional context**
Add any other context or screenshots about the feature request here. (we love screenshots with red circles if the request is for something visual)

**Database type, if appropriate**
Does this feature only relate to a specific type of database? (eg Postgresql)

rathboma:
This needs fixing.

The connection screen sidebar also resizes itself when connecting to a DB, then disconnecting. It's really frustrating

PR #2067, opened 2024-04-01
fix: remember last sidebar size

therealrinku:
Remeber last size of the sidebar
Related Issue: #1457 

[Screencast from 2024-04-01 09-16-38.webm](https://github.com/beekeeper-studio/beekeeper-studio/assets/76877078/3decd844-0ffd-4509-b531-3e343a7d1ab6)

### DIFF
--- a/apps/studio/src/common/LocalStorage.ts
+++ b/apps/studio/src/common/LocalStorage.ts
@@ -1,6 +1,13 @@
 
 
 export const SmartLocalStorage = {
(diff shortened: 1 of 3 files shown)


rathboma:
Awesome, thanks for the PR! I'm going to merge this, but we should also probably do the same thing on window resize (maybe that's already covered, not sure)

rathboma (approved):

PR #2072, opened 2024-04-02
fix migration order and ultimate connections

not-night-but:
fix migrations and upsell for ult dbs

### DIFF
--- a/apps/studio/src/migration/index.js
+++ b/apps/studio/src/migration/index.js
@@ -46,7 +46,7 @@ const setupSQL = `
  )
 `
 const realMigrations = [
-  a, b, c, d, domains, createSettings, addZoom,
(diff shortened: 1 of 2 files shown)


apps/studio/src/migration/index.js
```diff
@@ -102,7 +102,11 @@ export default class {
       }
       const hasRun = await Manager.checkExists(runner, migration.name)
       if (!hasRun) {
-        await migration.run(runner, this.env)
+        try {
+          await migration.run(runner, this.env)
+        } catch (err) {
+          console.log(`Migration ${migration.name} failed with`, err.name, err.message)
+        }
         await Manager.markExists(runner, migration.name)
```
rathboma, line 110: @not-night-but move this inside the `try`?

not-night-but: Doh

PR #2079, opened 2024-04-04
`IS NULL` and `IS NOT NULL` filters

### LINKED ISSUE #1957: FEAT: Table filters should support `IS NULL` and `IS NOT NULL`
Right now there's no way to search a table for fields that are null or not null explicitly. We should let users do this.

How to implement:
1. The database being queried must support the operation, otherwise hide it
2. In the drop down have `is NULL` and `is NOT NULL` options
3. When either of these options is selected, BKS should clear the `value` input, and disable it, have an appropriate tooltip like `you cannot provide a comparison value when checking for NULL or NOT NULL`

not-night-but:
close #1957 

Adds is null and is not null filters to table table view.

### TODO
- [x] check what dbs don't support this


rathboma:
This done?

apps/studio/src/components/tableview/RowFilterBuilder.vue
```diff
@@ -243,6 +249,9 @@ export default Vue.extend({
         "greater than": ">",
         "greater than or equal": ">=",
         in: "in",
+        // TODO (@day): check which databases don't support these
```
rathboma, line 252: this is done right? Remove todo if so!

apps/studio/src/components/tableview/RowFilterBuilder.vue
```diff
@@ -125,12 +126,17 @@
...
+              v-tooltip="isNullFilter(filter) ? 
+                'You cannot provide a comparison value when checking for NULL or NOT NULL' : 
+                ''"
+              >
               <div class="filter-wrap">
                 <input
                   class="form-control filter-value"
                   type="text"
                   v-model="filter.value"
+                  :disabled="isNullFilter(filter)"
```
rathboma, line 139: Maybe add a `:title` if it's disabled because of the null filter?

apps/studio/src/lib/db/clients/utils.ts
```diff
@@ -98,12 +98,14 @@ export function buildFilterString(filters: TableFilter[], columns = []) {
           : '?'
 
         return `${field} ${item.type.toUpperCase()} (${questionMarks})`
+      } else if (item.type.includes('is')) {
```
rathboma, line 101: This needs integration tests for the drivers to make sure they process and query themselves correctly

rathboma (changes requested):
Integration tests please!

rathboma:
Killing it :fire:

rathboma (approved):

PR #2086, opened 2024-04-05
Fix cloning rows with bit fields in MSSQL

### LINKED ISSUE #1945: BUG: Bit field = "false" okay for in-line edit but not for cloned row
**Describe the bug**
When I edit an individual cell that is a bit type, I can type true or false and apply the change. When I clone a row, I need to change the bit fields from true and false to 1 and 0 respectively before I can save.

**To Reproduce**
1. Select "Clone Row" on a table with one or more bit fields
2. Click Apply
3. Error Message "Invalid column name: false"

**Expected behavior**
Save the new record/row

**Screenshots / Gifs**
Selection_036

 - OS: Ubuntu 23.10
 - App Version SNAP 4.1.13 Beta
 - Database type and version MS SQL 12

**Additional context**
This has been an issue for many versions

not-night-but:
fix #1945 

Knex doesn't quote bit values in insert queries for some reason, this isn't an issue in update queries, so for inserts we're now manually converting them to strings

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -1197,7 +1197,7 @@ export default Vue.extend({
 
       })
     },
-    cellCloneRow(_, cell: Tabulator.CellComponent) {
+    cellCloneRow(_e, cell: Tabulator.CellComponent) {
       this.cloneSelection(_.last(cell.getRanges()))
     },
     cellAddRow() {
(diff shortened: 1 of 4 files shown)


rathboma (changes requested):
Needs an integration test for Mysql and SQL Server.

Please also verify that the test fails without the patch

rathboma (approved):

PR #2111, opened 2024-04-10
Fix vim mode yank/put with sys clipboard

### LINKED ISSUE #2062: BUG: Cannot copy to the clipboard in Vim mode using the Yank command
I'm unable to copy to the clipboard when I'm in Vim mode by using the Vim Yank command.
I have to use <CMD>+C to copy to the system clipboard.

I have tried to configure the *~/Library/Application Support/beekeeper-studio/.beekeeper.vimrc* by setting `clipboard` to unnamedplus and unnamed and they do not work.

Beekeeper Studio: Version 4.2.5 (4.2.5)
Mac: 13.4.1 (c)

not-night-but:
Fix #2062 

Add a custom register to codemirror to allow for yanking/putting using the * register.

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -300,17 +300,6 @@
 <script lang="ts">
 
   import _ from 'lodash'
-  import CodeMirror from 'codemirror'
-  import 'codemirror/addon/comment/comment'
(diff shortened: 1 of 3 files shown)


rathboma (approved):
I'm just going to trust you know what you're doing :upside_down_face: 

Merge away

PR #2124, opened 2024-04-13
Use the local node-gyp, should stop windows failures

rathboma:


rathboma:
Ref: https://github.com/vercel/hyper/pull/7270

PR #2126, opened 2024-04-13
Fix tabulator bit mutator

azmy60:
two fixes:
1. adding a row with `null` bit (except `bit(1)` ) from TableTable doesn't work:

```mysql
CREATE TABLE
  `with_bits` (
    `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
    `created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
    `bitties` bit(6) DEFAULT NULL
    PRIMARY KEY (`id`)
  )
```

It'll cause an error because our bit mutator doesn't check for `undefined` or `null`. And the `buildInsertQuery` didn't check for null data as well.

2.  The bug that day found last night was because our `bitMutator` expects the given data to be a `binary object`, not a `string/number`.

When data goes in for the first time, it comes from the driver in `binary object` and goes to tabulator. Later, tabulator converts the `binary` to `string/number` using our `bitMutator`, so it's not binary anymore.

When we clone the data, we use `.getData()` which return the data that tabulator has converted to (`string/number`) and then we give it to tabulator, and  tabulator converts it to `string/number` again using `bitMutator`, but the data is already a string, `bitMutator` didn't expect the input to be string.


rathboma:
If not - can you review and merge if good?

PR #2128, opened 2024-04-14
fix: mariadb has defaultvalue null as a string, so have it return obj…

wmontgomery:
…ect null instead of string

### DIFF
--- a/apps/studio/tests/lib/db.ts
+++ b/apps/studio/tests/lib/db.ts
@@ -336,12 +336,9 @@ export class DBTestUtil {
     const columns = await this.connection.listTableColumns(null, this.defaultSchema)
     const mixedCaseColumns = await this.connection.listTableColumns('MixedCase', this.defaultSchema)
     const defaultValues = mixedCaseColumns.map(r => r.hasDefault)
+    const trueFalseDBs = ['mariadb', 'mysql', 'tidb', 'postregresql', 'cockroachdb']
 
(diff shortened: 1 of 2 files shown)


apps/studio/tests/lib/db.ts
```diff
@@ -336,12 +336,9 @@ export class DBTestUtil {
...
     const mixedCaseColumns = await this.connection.listTableColumns('MixedCase', this.defaultSchema)
     const defaultValues = mixedCaseColumns.map(r => r.hasDefault)
+    const trueFalseDBs = ['mariadb', 'mysql', 'tidb', 'postregresql', 'cockroachdb']
 
-    if (this.dbType === 'mariadb') expect(defaultValues).toEqual([true,  true])
-    else if (this.dbType === 'mysql') expect(defaultValues).toEqual([true,  false])
-    else if (this.dbType === 'tidb') expect(defaultValues).toEqual([true,  false])
-    else if (this.dbType === 'postgresql') expect(defaultValues).toEqual([true,  false])
-    else if (this.dbType === 'cockroachdb') expect(defaultValues).toEqual([true,  false])
+    if (trueFalseDBs.indexOf(this.dbType) !== -1) expect(defaultValues).toEqual([true,  false])
```
rathboma, line 341: Question -- can we not do this test for all databases now?

rathboma (approved):

PR #2135, opened 2024-04-18
Add Window menu on macOS

### LINKED ISSUE #1767: FEAT: Window menu in the toolbar that shows open windows (and lets users switch between them)
This is a pretty standard feature in many desktop apps, especially on MacOS, but we're missing it from BKS!

### LINKED ISSUE #2018: FEAT: Support a Window menu item
When multiple connection windows are open, it's difficult to find a particular window as they tend to overlap each other. For example, I might have four connections open - one for local, one for dev, one for QA, and one for production. Without the Window menu item , I have to drag the connections around to look for the one I want.

Implement a standard Window menu item that is pretty common on most software. See the screenshot for an example from the Preview app.

No alternatives other than dragging connection windows around looking for the one I want

**Additional context**
The window name should be the connection name. It's possible there will be duplicates if the same connection is opened several times.

Applies to all databases

image

### LINKED ISSUE #2057: BUG: Missing window menu
On Mac, having the `window` menu is pretty standard and allows better OS interaction. We should provide this for ALL OSes when we're running in `native` window mode. That's ALWAYS for mac, and sometimes for Windows and Linux.

This is provided by electron as a built-in option, see `window`

https://www.electronjs.org/docs/latest/api/menu-item#roles

Arnie97:
The "Window" system menu is particularly useful when moving windows between monitors in multi-monitor setups.

Beekeeper Studio with Window Menu

Fix #2018 
Fix #2057 
Fix #1767

### DIFF
--- a/apps/studio/src/common/menus/BaseMenuBuilder.ts
+++ b/apps/studio/src/common/menus/BaseMenuBuilder.ts
@@ -9,7 +9,7 @@ export default class BaseMenuBuilder {
 
   app = "Beekeeper Studio"
 
-  constructor(private settings: IGroupedUserSettings, private actionHandlers: IMenuActionHandler) {}
+  constructor(protected settings: IGroupedUserSettings, private actionHandlers: IMenuActionHandler) {}
 
   buildTemplate(): any {
     throw new Error("Must implement buildTemplate for your platform")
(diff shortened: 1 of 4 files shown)


Arnie97:
This patch resolves #2018, however I missed comments in #2057 before creating the patch.
Should I remove the `platformInfo.isMac` condition to enable the "Window" menu for Windows and Linux too?

Arnie97:
Yet another duplicate #1767

rathboma:
No, I think the isMac check is good!

rathboma:
Thanks for this. I'll get this merged.

rathboma (approved):

PR #2136, opened 2024-04-18
fix postgres array check

### LINKED ISSUE #2130: BUG: Unexpected token g in JSON at position 0
**Describe the bug**
When editing or adding a row in a table, sometimes this error shows up. When copying the output to SQL and execute it there it works fine. 

**To Reproduce**
1. Open a table and view the data
2. Editing a column
3. Apply chanes

**Screenshots / Gifs**

image

 - OS:  Arch Linux
 - App Version 4.2.9
 - Database type and version PostgreSQL 12.16

azmy60:
fix #2130 

Not many databases that support array column like `postgres`. We have `sqlite` with `json` column but it doesn't name the column type with `_`.

DuckDB also support array btw.

Oracle has a similar type called `VARRAY` but the insert/update query is not generic like Postgres or DuckDB.


rathboma:
@azmy60 please add a postgresql test to confirm that this is interpreting column types correctly for both ENUMS and ARRAYs

rathboma (approved):

PR #2146, opened 2024-04-23
A couple of tabulator navigation small issues

### LINKED ISSUE #2078: BUG: TableTable ctrl+arrow movement breaks cell selection
How to replicate:

1. Open a table table and click a cell on the table
2. Moving the arrow keys should correctly navigate the table
3. Press ctrl+down-arrow

Expected:
- Cell 1,100 should be selected
- Pressing arrow keys continues to navigate the table

Actual:
- Cell 1,100 is selected
- Pressing the arrow keys does nothing

The same happens if you press ctrl+up from anywhere below the scrollable area. It seems causing the table to scroll breaks keyboard navigation.

Also: ctrl-arrow is also bound to changing the page number, this should be removed in favor of table navigation

azmy60:
I initially searched for an issue but ended up seeing a couple of small issues related to the Select Range (spreadsheet) navigation.

1. If we navigate on a big table using `Ctrl+Arrow` (try `film` in postgres), it can stop in any position unpredictably. https://github.com/olifolkerd/tabulator/pull/4473 <-- this also fixes #2078

2. Tabulator doesn't completely trap the keyboard events, so if we have a keybinding that's conflicting with tabulator, our keybinding can be triggered https://github.com/olifolkerd/tabulator/pull/4474

3. Pressing an arrow key after selecting a range of multiple cells should break the selection and continue navigating. This sometimes doesn't happen.  https://github.com/olifolkerd/tabulator/pull/4478

4. Autoscrolling to a big column is more sensible now. https://github.com/olifolkerd/tabulator/pull/4479

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -74,7 +74,7 @@
     "sql-formatter": "~10.7.2",
     "sql-query-identifier": "^2.7.0",
     "ssh2": "^1.14.0",
-    "tabulator-tables": "beekeeper-studio/tabulator#76c253a7437bcab5fcf67d601f09e67aea12f99e",
+    "tabulator-tables": "beekeeper-studio/tabulator#32107a671f00847d888f89ae0640cd4cc655229d",
     "tinyduration": "^3.2.4",
     "typeface-roboto": "^0.0.75",
     "typeface-source-code-pro": "^1.1.3",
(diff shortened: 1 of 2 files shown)


rathboma (approved):
Great finds! Love these types of continual tweaks

PR #2161, opened 2024-04-25
FEAT: Filter saved queries and connections

### LINKED ISSUE #2033: FEAT: Ability to filter connections and saved queries
It's possible to filter tables and schemas by typing in the filter box. This is useful when there are a lot of tables. But what if a user has a lot of saved queries, or saved connections?

This should support:
- Saved connection filtering (just filter the items in the store, like we do tables)
- Saved query filtering (same)

not-night-but:
resolve #2033

This allows the user to filter saved connections and saved queries:

Peek 2024-04-25 11-12
Peek 2024-04-25 11-13


rathboma:
Looks great! Will merge after the next release - can you make sure this works in ultimate with workspaces?

not-night-but:
@rathboma will do!

apps/studio/src/store/modules/data/query/LocalQueryModule.ts
```diff
@@ -9,9 +9,35 @@ export const LocalQueryModule: DataStore<FavoriteQuery, DataState<FavoriteQuery>
     loading: false,
     error: null,
     pollError: null,
+    filter: undefined // maybe this can be more advanced? date filter?
   },
   mutations: mutationsFor<FavoriteQuery>({
     // more mutations go here
```
rathboma, line 15: So looks like this isn't implemented for queries and connections that come from the cloud?

rathboma (commented):
Is this workspace enabled?

rathboma:
@not-night-but - does this work in ultimate with workspaces?

If so, then merge after fixing conflicts.

rathboma (approved):

PR #2162, opened 2024-04-26
refactor resolveLanguage

### LINKED ISSUE #2110: CHORE: Generalize LanguageData#resolveLanguage so we don't have to add a line when we add a new driver
If we forget to add a line to this file for new drivers then autocomplete won't work (boo).

Generalize it so that it 'just works' when adding a new driver.

For example: #2105

azmy60:
close #2110


apps/studio/src/lib/editor/languageData.ts
```diff
@@ -126,16 +126,49 @@ export function getLanguageByName(name: string): LanguageData | undefined {
...
-
 interface CodeMirrorLanguage {
   mode: string | Record<string, unknown>;
   hint?: unknown;
 }
 
-export function resolveLanguage(lang: Language): CodeMirrorLanguage {
+export function resolveLanguage(lang: "sql", type: ConnectionType): CodeMirrorLanguage;
+export function resolveLanguage(lang: "text" | "html" | "json"): CodeMirrorLanguage;
+export function resolveLanguage(lang: "text" | "html" | "json" | "sql" , type?: ConnectionType): CodeMirrorLanguage {
```
rathboma, line 136: No let's get rid of this all together for sql dialects -- the issue is:

- We don't want to have to change this file when we add a new database type

The change you've made doesn't solve this, we still have `resolveSqlLanguage`. Instead, can we move this to the `DialectData` for each language, or someplace else we're editing anyway when we add a new language?

rathboma (changes requested):
We need to kill `resolveSqlLanguage` all together

rathboma (approved):

PR #2171, opened 2024-04-30
Allow renaming database elements from table list

### LINKED ISSUE #2114: FEAT: Table renaming on right-click menu
Let users right click a table and rename it.

azmy60:
Add an option to table list context menu to rename database elements. This includes:

- renaming tables/views/schemas
- disabling `rename` if the database doesn't support it
- show errors if exists
- confirmation modal using `await this.$confirm` (can't help but this needed improvements, even though it's not on the ticket lol)
- copy to sql

close #2114 

renaming_table

image

image

TODO:
- [ ] ~Rename routines~ (blocked: automation tests don't like it. I'm only able to test this manually.)
- [x] Test cassandra
- [x] Renaming tables can be **destructive**. Put a warning notice.
- [x] refactor: we can put `setElementName()` just in BasicDatabaseClient


rathboma:
@azmy60 Is this good for review? Looks like you still have some TODOs undone?

azmy60:
@rathboma Yes. I was looking into the tests on renaming routines but somehow they don't work except when I tested it in app manually.

apps/studio/src/components/common/modals/RenameDatabaseElementModal.vue
```diff
@@ -0,0 +1,150 @@
...
+            Rename
+          </x-button>
+          <x-button
+            class="btn btn-primary"
+            menu
+          >
+            <i class="material-icons">arrow_drop_down</i>
+            <x-menu style="--align: end;">
+              <x-menuitem @click.prevent="rename">
+                <x-label>Rename</x-label>
```
rathboma, line 49: This should be consistent with the app and be: `Apply`

apps/studio/src/components/common/modals/RenameDatabaseElementModal.vue
```diff
@@ -0,0 +1,150 @@
...
+          @click.prevent="close"
+        >
+          Cancel
+        </button>
+        <x-buttons :disabled="loading">
+          <x-button
+            class="btn btn-primary"
+            @click.prevent="rename"
+          >
+            Rename
```
rathboma, line 40: `Apply`

rathboma (approved):
Looks good. Just change the button text, make sure Firebird is covered, then merge

azmy60:
Firebird doesn't support renaming tables/views.

PR #2185, opened 2024-05-07
fix building primary key with autoincrement

### LINKED ISSUE #2180: BUG: Adding an enum to a table disables editing
Video: https://www.loom.com/share/458422e9c98a4340915a0a2620cd0c26?sid=5be77852-400c-4457-b882-e9fd24a26f62

This is either:
1. Changing the id column when it shouldn't be
2. Confusing something with the table table

azmy60:
Based on this issue #2180 , the video shows that editing is disabled due to the lack of primary key, not after adding an `enum` column. It's disabled even before adding an `enum` column.

But after testing the table builder, I found that creating a column as a pk with `autoincrement` doesn't make the column pk. Thus the editing was disabled in the first place.

Note: There's no auto increment functionality for a PK in BigQuery and Cassandra DBs.

close #2180

### DIFF
--- a/apps/studio/tests/lib/db.ts
+++ b/apps/studio/tests/lib/db.ts
@@ -5,7 +5,7 @@ import { createServer } from '../../src/lib/db/index'
 import log from 'electron-log'
 import platformInfo from '../../src/common/platform_info'
 import { IDbConnectionPublicServer } from '../../src/lib/db/server'
-import { AlterTableSpec, Dialect, DialectData, FormatterDialect, SchemaItemChange } from '../../../../shared/src/lib/dialects/models'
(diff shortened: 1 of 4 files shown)


apps/studio/tests/lib/db.ts
```diff
@@ -770,15 +775,38 @@ export class DBTestUtil {
...
+      name: 'test_table',
+      columns: [{
+        columnName: 'id',
+        dataType: 'autoincrement',
+        primaryKey: true,
+        nullable: false,
+      }],
+    }
+    const query = generator.buildSql(schema)
+    const expectedQueries = {
```
rathboma, line 793: I think if we're expecting something different per-dialect, these should just go directly in the driver tests, but it's fine for now

rathboma (approved):

PR #2206, opened 2024-05-15
Add minimal mode

azmy60:
image

Enabling minimal mode will do these:
- Hiding row filter builder from table view. In result, we don't need the title scope so it's hidden now.
- Tab title and column title are not bold
- Hiding routines, paginator, time estimation, table length, and global sidebar.
- Optimizing the column auto width by hiding the column type.

Also there was some weird little styling issues:
- Database selector is not aligned to the tab. Looks kinda weird to me.
image
- The table list doesn't look "visually" aligned to the database selector at the top, and it has little space on the left. This becomes more obvious when enabling the minimal mode.
image

### DIFF
--- a/apps/studio/src/App.vue
+++ b/apps/studio/src/App.vue
@@ -1,6 +1,9 @@
 <template>
   <div class="style-wrapper">
-    <div class="beekeeper-studio-wrapper">
+    <div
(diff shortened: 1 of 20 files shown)


rathboma (approved):
Looks great. Excited to see where we go with minimal mode!

PR #2209, opened 2024-05-17
[Fix] Add exit confirmation modal before quitting

### LINKED ISSUE #2201: BUG: Losing pending changes after closing the app
**To Reproduce**
1. Go to table view
2. Edit some data
3. exit the app
4. BKS doesn't show a warning for losing changes
5. After re-opening BKS, the changes in the table view is gone

**Expected behavior**
It should show a warning modal like when closing a tab.

H0onnn:
## Issue

- close #2201 

## Work Detail

I modified it so that you can check the confirmation alert when the application is terminated using electron's default dialog.

If I need to use a custom modal in Vuejs, I need some advice on how to do that.

Thanks !

## Preview

image

### DIFF
--- a/apps/studio/src/background.ts
+++ b/apps/studio/src/background.ts
@@ -1,6 +1,6 @@
 'use strict'
 import * as fs from 'fs'
-import { app, protocol } from 'electron'
+import { app, protocol, dialog } from 'electron'
 import log from 'electron-log'
 import * as electron from 'electron'
 import { ipcMain } from 'electron'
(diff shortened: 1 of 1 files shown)


rathboma:
Hey, thanks for this! I think we need a way to disable this warning, as many folks might find it frustrating. Is that possible with the electron API?

H0onnn:
Hello, @rathboma :)

Are you talking about the toggle? I'll give it a try.

rathboma (changes requested):
Ok, here's what I think:

- Adding a warning is a good idea, but only if users have UNSAVED changes.

I don't want it to always warn the user regardless, as this is super annoying.

Is this possible for you to work on?

rathboma:
Closing this, don't think we want to do this at all.

PR #2212, opened 2024-05-18
fix: disabled state styling for filter input

### LINKED ISSUE #2160: BUG: Disabled state for the filter input looks the same as non-disabled state
image

vs

image

therealrinku:
fixes #2160 

Untitled-2024-01-09-1526


rathboma (approved):
Thanks!

Issue #1923, opened 2024-01-29
FEAT: Add keyboard shortcut to open editor modal


EmanNo1:
**Is your feature request related to a problem? Please describe.**
When selecting a cell to edit, the cursor is placed on the end of the content. When I need to replace the content, extra keystrokes or clicks are needed

**Describe the solution you'd like**
I would like a new keyboard shortcut/pop-up menu option (eg. Alt + Enter) that would enter edit mode with the entire contents selected.

**Describe alternatives you've considered**
Default the <<Enter>> key to enter edit mode with the contents selected instead of placing the cursor at the end.

**Additional context**
The modal mode is also useful, please add alt+m (or another) keyboard shortcut

**Database type, if appropriate**
Using with SQLite and MS SQL (2012)

PR #2213, opened 2024-05-19
feat: keyboard shortcut added for opening editor modal

therealrinku:
resolves: #1923 
_**shift + enter**_ keyboard shortcut added for opening editor modal

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -411,7 +411,7 @@ export default Vue.extend({
       //   const focusingTable = this.tabulator.element.contains(document.activeElement)
       //   if (!focusingTable) this.page--
       // }
-      result['alt+m'] = this.openEditorMenuByShortcut.bind(this)
+      result['shift+enter'] = this.openEditorMenuByShortcut.bind(this)
       result[this.ctrlOrCmd('r')] = this.refreshTable.bind(this)
       result[this.ctrlOrCmd('n')] = this.cellAddRow.bind(this)
       result[this.ctrlOrCmd('s')] = this.saveChanges.bind(this)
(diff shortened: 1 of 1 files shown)


rathboma:
@azmy60 wanna review this? If good, approve and merge!

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -974,14 +975,20 @@ export default Vue.extend({
...
     },
+    isEditorMenuDisabled (cell: CellComponent) {
+      if (this.isPrimaryKey(cell.getField())) return true
+      return !this.editable && !this.insertionCellCheck(cell)
+    },
+    openEditorMenuByShortcut() {
+      const range: RangeComponent = _.last(this.tabulator.getRanges())
+      const cell = range.getCells().flat()[0];
+      if (this.isEditorMenuDisabled(cell)) return
+      this.$refs.editorModal.openModal(cell.getValue(), undefined, cell)
```
azmy60, line 986: Would you be able to add a FIXME note here so we can avoid calling child methods directly like this? Normally, it should be done by calling an event using `this.$modal.show(modalName)` or `this.$trigger(AppEvent.something)`.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -411,6 +411,7 @@ export default Vue.extend({
       //   const focusingTable = this.tabulator.element.contains(document.activeElement)
       //   if (!focusingTable) this.page--
       // }
+      result['alt+m'] = this.openEditorMenuByShortcut.bind(this)
```
azmy60, line 414: Should we press `shift+enter` (DBeaver uses this)? Or anything similar with pressing `enter`? I usually look at other tools to decide. I'm just not too familiar with `alt+m` though I feel like it might works or I might be too nitpicky. Feel free to use anything that's best for you.

azmy60 (changes requested):
Just a couple of notes, and we're good :)

therealrinku:
hi @azmy60 ! thanks for the review!
there was not any specific reason to add **alt + m** shortcut. I agree, that it makes more sense to use **shift + enter* so I updated it to use **shift + enter** and added FIXME note as well :)

azmy60 (approved):
Perfect! Thanks so much.

PR #2215, opened 2024-05-21
fix focus trap in modals

azmy60:
- trap focus on all modals when they're opened.
- an element inside a modal should be focused initially to make the focus trap work
- some modals have a close button that can't be focused. Added `href="#"` to make them focusable.


rathboma (approved):

rathboma:
@azmy60 merge conflict, then this is good

rathboma:
@azmy60 feel free to merge this (once approved, you can merge!)

PR #2216, opened 2024-05-23
Fix TableTable auto resize when enabling minimal mode

azmy60:
Enabling minimal mode would resize the columns automatically. But doing that while the tab is hidden isn't correct.


rathboma (approved):
thanks!

PR #2218, opened 2024-05-28
Utility proc

not-night-but:
Starting to migrate to a client server architecture. Still needs a lot of work.

### TODO
- [x] Monitor status of the utility process, if it crashes we need a way to restart it cleanly
- [x] Multiple windows of beekeeper will definitely not work with the current architecture. Every client will probably have to have a unique id that is passed with every request, or their own ports? Actually that last one is more viable, as you can't clone ports. handlerState will have to be instantiated for every client as well.

### Notes
- I don't like how the handlers are situated right now. Think it would be nicer if they were just classes and have `Handlers` be a child of all the handler classes. Very similar to how it is now, just the implementation of the handlers would be the type definition as well. 
- Need to work on getting some of the types rebuilt on the backend (eg. `SavedConnection` and `UsedConnection`). If the object has functions associated we lose those, this is so far just a problem in `conn/create` and maybe `conn/test`
- Need to get type checking in the frontend working. May have to be separate from the backend types, as some of those will be importing stuff that can't be in the frontend.


apps/studio/src/handlers/exportHandlers.ts
```diff
@@ -0,0 +1,120 @@
...
+  'export/error': ({ id }: { id: string }) => Promise<string>,
+  'export/name': ({ id }: { id: string }) => Promise<string>,
+  'export/start': ({ id }: { id: string }) => Promise<void>,
+  'export/cancel': ({ id }: { id: string }) => Promise<void>
+}
+
+
+export let exportHandlers = {} as unknown as IExportHandlers;
+
+exportHandlers['export/add'] = async function({ options }: { options: StartExportOptions }) {
```
rathboma, line 56: I understand why you did this, but it would be better for it to be a Const (also Uppercase the first word):

```ts
export const exportHandlers extends IExportHandlers = {
  'export/add': ...
}
```
Works just fine, then this can be a const!

image

apps/studio/src/handlers/exportHandlers.ts
```diff
@@ -0,0 +1,120 @@
...
+  const exporter = state.exports.get(id);
+
+  if (!exporter) {
+    throw new Error(errorMessages.noExport);
+  }
+
+  return exporter;
+}
+
+export interface IExportHandlers {
```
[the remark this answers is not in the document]
rathboma: Very smart to use an interface for this.

apps/studio/src/utility.ts
```diff
@@ -0,0 +1,68 @@
+import rawLog from 'electron-log'
+import ORMConnection from './common/appdb/Connection'
+import platformInfo from './common/platform_info';
+import { connHandlers } from './handlers/connHandlers';
+import { exportHandlers } from './handlers/exportHandlers';
+import { generatorHandlers } from './handlers/generatorHandlers';
+import { Handlers } from './handlers/handlers';
+import { state } from './handlers/handlerState';
```
rathboma, line 8: I don't think you should export this constant `state` object, but rather instantiate it here in the utility file.

I don't see any other imports, so I think it's safe to just import the class, and create a new instance here in Utility.ts

apps/studio/src/utility.ts
```diff
@@ -0,0 +1,68 @@
...
+process.parentPort.on('message', ({ ports }) => {
+  if (ports && ports.length > 0) {
+    state.port = ports[0]
+    log.info('RECEIVED PORT: ', state.port);
+    init();
+    state.port.start();
+  }
+})
+
+async function runHandler(id: string, name: string, args: any) {
```
rathboma, line 38: Great. Simple and clear

apps/studio/tsconfig.background.json
```diff
@@ -0,0 +1,36 @@
+{
+  "files": ["src/background.ts"],
+  "compilerOptions": {
+    "target": "es2020",
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "outDir": "./dist/main",
+    "experimentalDecorators": true,
+    "esModuleInterop": true,
+    // "strict": true,
```
rathboma, line 10: kill if you don't want it

rathboma: Not valid json technically

apps/studio/vite.config.mjs
```diff
@@ -0,0 +1,54 @@
+import { defineConfig } from 'vite';
```
rathboma, line 1: Let's just kill this file seeing as it doesn't work

rathboma (changes requested):
This is awesome! Glad you got it all working.

Some changes I think would be beneficial:
- We can reduce the blast radius of Vue changes by making a `BasicConnection` instance that sends to the `server`. That way when we add new Connection methods, we'll get prompted to update the implementation, and it'll be transparent to any other devs, they can just get the connection from the store and go nuts.
- Let's change the Export Handler to just be a const.

Other questions:
1. What if the utility process dies? Can you add code to restart it?
2. If you're developing the utility process, in dev does it have to restart the whole app? We should be able to add a new esbuild hook to send SIGUSR1, and have that just restart the utility process possibly?

Let's get these changes wrapped up and merged. Very excited for this! Great work.

rathboma:
Oh - also don't worry about `SavedConnection` and `UsedConnection` right now, we can do that in a separate PR. Let's have this change get merged and settle for a while first.

apps/studio/src/components/export/ExportNotification.vue
```diff
@@ -28,47 +31,53 @@ export default {
...
-          this.notification.setText(this.notificationText);
+          this.notification?.setText(this.notificationText);
         }
       },
     },
   },
-  mounted() {
-    this.exporter.onProgress(this.updateProgress)
-    this.notification.show();
+  async mounted() {
```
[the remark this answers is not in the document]
not-night-but: @rathboma Okay so, this part actually doesn't handle errors, it handles progress, was annoying to get working lol.

For errors, just like any handler, if the underlying function throws, it will be sent back to the client as an 'error' (see utility.ts `runHandler`). Once the client receives a reply, if it's an error it will reject the promise it returned with the error, which is then handled through whatever channels we want

apps/studio/src/utility.ts
not-night-but: We do actually have to use it elsewhere. It is used in the connection handlers, the queryHandlers, the exportHandlers, and the generatorHandlers

rathboma:
Passing build!

Issue #2242, opened 2024-06-08
BUG: arrow keys not working while editing cell


mcanepa:
Double click to edit cell and cursor goes to the end of content. Then press arrow key to go left but the cursor doesn't move. I have manually move the mouse to desired position. (right arrow key doesn't work either)

rathboma:
@azmy60 is the fix for this in master? I know you did a lot of work on the keyboard behavior

azmy60:
No, this is new. Those were for trapping focus while navigating buttons in modals.

PR #2247, opened 2024-06-11
fix: arrow left/right not working on table

therealrinku:
fixes: #2242 
adding navLeft and navRight true in tabulator keybindings options was enough for this, 
not sure if this has any other side effects 🤔


azmy60:
Unfortunately, this would also disable navigating the table. This is a bug from [tabulator](https://github.com/olifolkerd/tabulator) and I have some outstanding PRs there to fix this. So I'll close this in favor of that :)

PR #2251, opened 2024-06-12
Fix SQLite and LibSQL in-memory connection

azmy60:
We keep creating and closing a new connection when connecting to an in-memory database, which shouldn't happen. The data in an in-memory database is destroyed when the connection is closed, so we need to maintain the same connection until we close it explicitly.

Note: there are two kinds of in-memory databases (https://www.sqlite.org/inmemorydb.html):
- In Memory database => stores data in memory, use `:memory:` as URL
- Temporary Database => stores data in a temporary file, use empty string as URL


azmy60:
@rathboma Our SQLite form only allows Temporary Database (pressing `connect` without selecting a file). I don't really know in what scenario some folks will need to choose `:memory:` and not the other, but it's probably a nice thing to add a checkbox or something that says "connect to `:memory:` or Temporary Database?".

rathboma (approved):

PR #2261, opened 2024-06-18
Feature/import scripts

### LINKED ISSUE #2259: BUG: Program freezes when importing large SQL files
**Describe the bug**
When I want to import an SQL file that has +20K lines of inserts it makes the program stucks for ~30 seconds.

The reason of this is because the import button instead of executing the file it opens it on the editor, making it hangs for a lot of time until all the lines cached on memory.

**Possibles solutions:**
- An option to execute directly the sql file instead of open it on the editor.
- Only render X lines of the imported file until you scroll to the bottom.
- Show "Show more" button like VSCode does to resolve the same problem:
image

**To Reproduce**
Import a .sql file that has a lot of lines.

**Version Information:**
Tried on MacOS

**Additional context**
Other programs like SequelAce stuck but for ~2 seconds and it does not open it on the editor.

wmontgomery:
resolves #2259

Part 1 of 2, the rest on ultimate


rathboma:
@wmontgomery looks like a broken test?

wmontgomery:
Yeah, have to figure out what's up. They pass locally but on GitHub they
choose violence.

rathboma (approved):

Issue #2267, opened 2024-06-19
BUG: Copying value of a cell in the results using keyboard shortcut copies wrong value


SamuelBolduc:
**Describe the bug**
When using a keyboard shortcut to copy a cell in the results of a query, it copies to clipboard the value of the selected cell in the last tab that ran a query. It does not happen, however, when using right click -> copy.

**To Reproduce**
1. Open a new tab, run a simple SELECT query
2. Open another tab, run another simple SELECT query but of different data (to illustrate)
3. Go back to first tab, click on the first cell in the results, copy using keyboard shortcut
4. Paste value anywhere - it will be the value of the selected cell in the other tab

**Expected behavior**
Should be the same behavior as the context menu copy

**Version Information (please complete the following information):**
 - OS: Mac
 - App Version: 4.3.4
 - Database type and version: PostgreSQL 14

rathboma:
Uck, good find. Thank you!

PR #2273, opened 2024-06-20
fix: ctrl + c on result table copying wrong values

therealrinku:
fixes: #2267 

problem is when doing ctrl + c, copySelection functions runs for each opened query's ResultTable not just the currently active query's ResultTable which overrides the last copied value, not sure if that's expected behaviour or it's a bug itself

for now, I have just added a check to know if it's a active query tab before copying the value on ctrl + c


rathboma:
Thank you!

rathboma (approved):

PR #2281, opened 2024-06-23
fix: add columns and total count the queryStream in order to get head…

wmontgomery:
…ers on csvs and whatnot

image

### DIFF
--- a/apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
+++ b/apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
@@ -26,6 +26,11 @@ export interface QueryLogOptions {
     error?: string
 }
 
+interface ColumnsAndTotalRows {
+  columns: TableColumn[]
+  totalRows: number
+}
+
 // this provides the ability to get the current tab information, plus provides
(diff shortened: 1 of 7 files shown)


rathboma (changes requested):
Is this query executed inside the same transaction as the export? If not, the total rows (and even column names) can be different. Is there an easy way to fix that?

wmontgomery:
It's done prior to calling the cursor steps I believe, 

> Is this query executed inside the same transaction as the export? If not, the total rows (and even column names) can be different. Is there an easy way to fix that?

It's in this else statement which is prior to beginning the stream https://github.com/beekeeper-studio/beekeeper-studio/blob/8799ecbd8f183b38720bcbe5f0631ae03a884f04/apps/studio/src/lib/export/export.ts#L151, matching the selectTopStream call above. 

Since the query is done prior to streaming it (the export transaction), the count and columns should be the same. What sucks is the query being done 2x, but there are also multiple preliminary queries being done on the full table export as well so at least matches the pattern. Not sure a great way to optimize it without some heavy refactoring, but definitely up to talk it out

PR #2291, opened 2024-06-27
Appdb handlers

not-night-but:
This PR finishes off writing all of the handlers for the util process, so no native deps are in the frontend anymore, thus context isolation has been turned on! Still loads of work to do:

- [ ] Write some docs for the new preload script
- [ ] Thorough review of the features in the app to make sure they are functioning
- [x] Fix tabs (some tabs don't seem to want to display their content at all)
- [x] Fix switching to newly opened tab
- [x] Fix editing cells in TableTable
- [x] Maximizing window with client menuStyle doesn't work
- [x] Fix the placeholder text for the connection type dropdown


apps/studio/src/lib/utility/appdb/SavedConnection.ts
```diff
@@ -0,0 +1,118 @@
+import { ConnectionType, IConnection, SshMode } from "@/common/interfaces/IConnection";
+import { AzureAuthOptions, BigQueryOptions, CassandraOptions, LibSQLOptions, RedshiftOptions } from "@/common/appdb/models/saved_connection";
+import { BaseUtilityAppDbEntity } from "./BaseUtilityAppDbEntity";
+import { ConnectionString } from "connection-string";
+import rawLog from 'electron-log';
+import { baseFind, baseFindOne } from "./Util";
+
+const log = rawLog.scope('SavedConnection');
+
+export class SavedConnection extends BaseUtilityAppDbEntity implements IConnection {
```
rathboma, line 10: So the idea here is right, but we already have an abstraction for this because we needed one for Workspaces. It's done at the store level, rather than the model level.

I think you need to extend those interfaces, rather than introducing a second competing abstraction.

See for example `CloudClient`, `CloudConnectionModule`, and `DataModuleBase`

apps/studio/src/components/tableinfo/TableIndexes.vue
```diff
@@ -359,6 +359,7 @@ export default Vue.extend({
   },
   mounted() {
     // this.initializeTabulator()
+    ipcRenderer.invoke('hi')
```
wmontgomery, line 362: Hello 👋

apps/studio/esbuild.mjs
```diff
@@ -94,98 +92,126 @@ const electronUtilityPlugin = {
...
+  }
+}
+
+const env = isWatching ? '"development"' : '"production"';
+const commonArgs = {
+  platform: 'node',
+  publicPath: '.',
+  outdir: 'dist',
+  bundle: true,
+  external: [...externals, '*.woff', '*.woff2', '*.ttf', '*.svg', '*.png'],
```
azmy60, line 117: There's a similar `externals` variable in `vue.config.js` and I feel like this one isn't meant to be added here, no?

apps/studio/src/store/modules/data/DataModuleBase.ts
```diff
@@ -103,14 +104,67 @@ const buildBasicMutations = <T extends HasId>(sortBy?: SortSpec) => ({
...
+    async load(context) {
+      context.commit("error", null);
+      await safely(context, async () => {
+        const items = await Vue.prototype.$util.send(`appdb/${type}/find`, { options: loadOptions });
+        if (context.rootState.workspaceId === LocalWorkspace.id) {
+          context.commit('upsert', items);
+        }
+      })
+    },
+    async poll() {
```
wmontgomery, line 125: SInce this doesn't actually do anything, do we need it?

shared/src/lib/dialects/firebird.ts
```diff
@@ -21,16 +20,33 @@ function wrapIdentifier(id: string) {
...
+      case 'boolean':
+            return value ? 'true' : 'false';
+      case 'number':
+          return value.toString();
+      case 'string':
+          return "'" + value.replace(/'/g, "''").replace(/\\/g, '\\\\') + "'";
+  }
+
+  if (value instanceof Date)
+      return "'" + value.getFullYear() + '-' + (value.getMonth()+1).toString().padStart(2, '0') + '-' + value.getDate().toString().padStart(2, '0') + ' ' + value.getHours().toString().padStart(2, '0') + ':' + value.getMinutes().toString().padStart(2, '0') + ':' + value.getSeconds().toString().padStart(2, '0') + '.' + value.getMilliseconds().toString().padStart(3, '0') + "'";
```
wmontgomery, line 37: Minor thing, a template string might be easier to read and if nothing else is more the pattern for putting a string together like this

apps/studio/src/preload.ts
```diff
@@ -0,0 +1,209 @@
...
+  }
+}
+
+
+export const api = {
+  isReady: () => {
+    ipcRenderer.send('ready');
+  },
+  send: (event: AppEvent, name: string, arg?: any) => {
+    if (!AppEvent[event]) return;
```
azmy60, line 60: This function would always stop here if we call it like `window.main.send(AppEvent.click, 'me')`.

Instead of `if(!AppEvent[event])`, maybe we can do `if(!_.hasValue(AppEvent, event))` or something like that.

azmy60 (changes requested):

apps/studio/src/App.vue
```diff
@@ -78,21 +77,21 @@ export default Vue.extend({
       log.info('database changed', this.database)
     },
     themeValue() {
+      console.log("THEME VALUE: ", this.themeValue)
```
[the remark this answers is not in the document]
not-night-but: I thought I had taken this out 🙃🙃

apps/studio/esbuild.mjs
not-night-but: I believe vue.config.js is actually useless at this point. That's for vuecli and we use esbuild now, I think I removed it in master and it just hasn't made its way over here yet

apps/studio/src/preload.ts
not-night-but: Good catch! I just wanna make sure we're at least doing some sort of whitelisting of events. Good suggestion!

apps/studio/esbuild.mjs
```diff
@@ -94,98 +92,126 @@ const electronUtilityPlugin = {
...
+  setup(build) {
+    if (!isWatching) return;
+    build.onStart(() => console.log("ESBUILD: Building Preload 🏗"));
+    build.onEnd(() => {
+      console.log("ESBUILD: Built Preload ✅");
+      if (electron) {
+        process.kill(electron.pid, 'SIGINT');
+      }
+      //start electron again
+      electron = spawn(path.join('../../node_modules/electron/dist/electron'), ['.'], { stdio: 'inherit' });
```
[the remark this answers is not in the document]
not-night-but: Yeah I haven't brought master in in a bit, I've been living in my own world

apps/studio/src/components/tableinfo/TableIndexes.vue
not-night-but: F*CK

apps/studio/src/store/modules/data/DataModuleBase.ts
not-night-but: It does stuff in ultimate, this is just to make merging stuff easy I believe

shared/src/lib/dialects/firebird.ts
not-night-but: Lol so uh. Azmy and I kinda stole this directly from firebord's source code 😂 I'm down to change it, just wanna be careful I don't muck it up

apps/studio/esbuild.mjs
azmy60: Yeah I still see the file in master. That's cool we don't need it anymore!

apps/studio/src/preload.ts
azmy60: That's great! In that case, maybe adding a warning log in case we're calling a non-existing event will also be great.

not-night-but: so `_.hasValue` doesn't seem to exist. Did some more digging online, and the best solution to me seems to be `if (!Object.values<string>(AppEvent).includes(event))`

azmy60: Yeah, I made that one up 😆 . Good find!

PR #2309, opened 2024-07-08
Adding DuckDB driver

### LINKED ISSUE #1500: FEAT: DuckDB Support
No

DuckDB support similar SQLite

Implementing basic DuckDB support would allow us to provide the following:
1. Native SQL support for Parquet files (query your parquet)
2. Native SQL support for CSV files (query your csv)
3. Reading Parquet and CSV data from S3
5. Easily generate data and send it to parquet/csv files

Notes on implementation:
- Create table can be expanded to also allow `create view from external file`, where the user can select a parquet or csv file to use as the base for the view, in line with [DuckDBs docs on views](https://duckdb.org/docs/data/parquet/overview.html#inserts-and-views). This would make it super easy to query a bunch of parquet files
  - We need to make sure we support [hive partitioning](https://duckdb.org/docs/data/partitioning/hive_partitioning) when implementing this. We can even scan the filesystem to tell them what partitions it would use

azmy60:
resolves #1500
also mentioned in #1486


apps/studio/src/lib/db/clients/duckdb.ts
```diff
@@ -0,0 +1,944 @@
...
+      `,
+      { params: [table, schema] }
+    );
+
+    return data.map((row) => ({
+      id: row.index_oid,
+      table: row.table_name,
+      schema: row.schema_name,
+      name: row.index_name,
+      columns: [], // TODO
```
not-night-but, line 462: Is this not supported by duckdb yet? Or is this a forgotten todo?

apps/studio/src/lib/db/clients/duckdb.ts
```diff
@@ -0,0 +1,944 @@
...
+    // toColumn: string;
+    // fromTable: string;
+    // fromSchema: string;
+    // fromColumn: string;
+    // constraintName?: string;
+    // onUpdate?: string;
+    // onDelete?: string;
+    // const { data } = await this.driverExecuteSingle(``);
+    return [];
+  }
```
not-night-but, line 495: Did you check this? If it's not supported I'd just remove all the commented stuff and add a note.

apps/studio/src/lib/db/clients/duckdb.ts
```diff
@@ -0,0 +1,944 @@
...
+
+  applyChangesSql(changes: TableChanges): string {
+    return applyChangesSql(changes, this.knex);
+  }
+
+  async applyChanges(changes: TableChanges): Promise<TableUpdateResult[]> {
+    let results = [];
+
+    await this.driverExecuteSingle("BEGIN TRANSACTION");
```
not-night-but, line 647: Does duckdb only have one constant connection? If not, there's a danger of these queries (`begin transaction`, the update stuff, etc) being run on different connections. If it is truly just like sqlite though, then we don't have to worry about it.

apps/studio/src/lib/db/clients/duckdb.ts
```diff
@@ -0,0 +1,944 @@
...
+
+    let sql = "";
+
+    if (typeOfElement === DatabaseElement.TABLE) {
+      sql = `ALTER TABLE ${elementName} RENAME TO ${newElementName}`;
+    } else if (typeOfElement === DatabaseElement.VIEW) {
+      sql = `ALTER VIEW ${elementName} RENAME TO ${newElementName}`;
+    } else if (typeOfElement === DatabaseElement.SCHEMA) {
+      sql = `ALTER SCHEMA ${elementName} RENAME TO ${newElementName}`;
+    }
```
not-night-but, line 711: Can't this just be 
```ts
sql = `ALTER ${typeOfElement} ${elementName} RENAME TO ${newElementName}`
```
?

azmy60: I think DuckDB only supports those 3 types.

apps/studio/src/lib/db/clients/duckdb.ts
azmy60: DuckDB can have multiple connections, but it won't allow write actions if there are multiple connections. So it's impossible to do transactions in that situation. We need to only have one connection to be able to both read and write.

apps/studio/src/lib/db/clients/duckdb.ts
azmy60: Turns out it supports foreign keys lol.

rathboma:
@azmy60 Hey! So I actually think this should go against master instead of community-4, would that be hard to change?

rathboma:
I know I said the opposite a few weeks ago, but I think working against master is the right way to do it.

azmy60:
Yeah I can do that. I've worked adding clickhouse on v5 and haven't seen any major change on db client so it should be relatively easy.

azmy60:
Added DuckDB for the new electron and tested! :) @rathboma 
https://github.com/beekeeper-studio/duckdb-node/releases/tag/v1.0.0-bks18

florentfgrs:
Can't wait to discover this feature! I'm a DuckDB user and look forward to DuckDB support in Beekeper.

apps/studio/src/lib/db/clients/duckdb.ts
```diff
@@ -0,0 +1,1037 @@
...
+  defaultSchema(): string {
+    return "main";
+  }
+
+  protected async rawExecuteQuery(
+    q: string,
+    options: any
+  ): Promise<DuckDBResult | DuckDBResult[]> {
+    const queries = identify(q, { strict: false });
+    const params = options.params;
```
rathboma, line 597: can't options be null? Should se check `options?.params || []` or something like that?

rathboma (changes requested):
Overall, I'm ready to merge this.
Some changes needed:

- The Cursor is Cursed
- Can't we use existing query builders in util?
- Move driver code to src-commercial
- Merge in master to make sure it's all working with the new architecture.

apps/studio/src/lib/db/clients/duckdb/DuckDBCursor.ts
```diff
@@ -0,0 +1,50 @@
...
+  ) {
+    super(options.chunkSize);
+  }
+
+  async start() {
+    log.info("Starting cursor");
+    this.cursorPos = 0;
+  }
+
+  async read() {
```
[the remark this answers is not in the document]
azmy60: lol yep!

rathboma:
@azmy60 SQL Server looks broken on this branch still. It's not broken in master, so something in here is breaking it.

rathboma (approved):

PR #2335, opened 2024-08-01
Adding azure SSO modal and sign out button in master

azmy60:
~NOTE: this also upgrade the node version to 18 because I added a new package that requires node 18. And It also seems to work fine..?? 🤯~


rathboma:
It won't work fine -- native modules might not work properly when bundled. It works because we compile the JS to a earlier target. We def need to keep the version of NodeJS in sync with the version of electron we're using

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -181,9 +172,11 @@ export class AzureAuthService {
     const authUrl = await this.pca.getAuthCodeUrl(authCodeUrlParams);
 
     log.debug('Getting auth code')
-    window.location.href = authUrl;
 
-    this.start = Date.now();
+    // FIXME (azmi): we don't need this after appdb handlers PR is merged.
+    // we can just use window.location.href instead.
+    open(authUrl);
```
rathboma, line 178: We shouldn't need this at all. We have a way to do this:

Example:
```
window.main.openExternally("https://beekeeperstudio.io/get");
```

.nvmrc
```diff
@@ -1 +1 @@
-v16.17.1
+v18.16.0
```
rathboma, line 1: Please revert this

apps/studio/esbuild.mjs
```diff
@@ -145,7 +145,12 @@ const electronRendererPlugin = {
           to: './dist/'
         },
       ]
-    })]
+    })],
+    // import.meta.url is undefined in esbuild
```
rathboma, line 149: What is this used for?

apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -454,7 +468,24 @@ export default Vue.extend({
       } else {
         this.errors = null
       }
-    }
+    },
+    // Before running connect/test method
+    _beforeConnect() {
```
rathboma, line 473: This should just be `beforeConnect`, no need for the underscores

apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -454,7 +468,24 @@ export default Vue.extend({
...
+      if (
+        this.config.connectionType === 'sqlserver' &&
+        this.config.azureAuthOptions.azureAuthEnabled &&
+        this.config.azureAuthOptions.azureAuthType === AzureAuthType.AccessToken
+      ) {
+        this.loadingSSOModalOpened = true
+      }
+    },
+    // After running connect/test method, success or fail
+    _afterConnect() {
```
rathboma, line 483: same

apps/studio/src/store/index.ts
```diff
@@ -422,6 +422,9 @@ const store = new Vuex.Store<State>({
         await context.state.connection.connect();
       }
     },
+    async cancelConnect() {
```
rathboma, line 425: I'd prefer to keep all connection behavior on the connection shim. Why can't this be `context.state.connection.cancel()` or even `.disconnect()`?

See `ElectronUtilityConnectionClient`

rathboma (changes requested):
Looks good, but some changes needed.

rathboma:
Video would be helpful!

apps/studio/esbuild.mjs
azmy60: This was used by the new package I added to `open` a externals url. I can revert this since we don't need the package anyway.

apps/studio/src/store/index.ts
azmy60: `disconnect` is probably the best place to do that. I'll take a look.

azmy60: Using `context.state.connection.cancel()` or `.disconnect()` doesn't work in this case. `this.connection` only exists after we are successfully connected. `cancelConnect` is essentially for stopping the `checkStatus()` loop and it happens before this.connection exists. I guess `conn/invoke` is the place that makes more sense for this? I don't know how you feel about `conn/invoke` though.

azmy60:
It's pretty much the same with the one that's merged in `community-4`. Here's the demo:

https://github.com/user-attachments/assets/8736b605-304b-404e-bd75-95b9b5bae879

rathboma (approved):

PR #2345, opened 2024-08-07
implement clickhouse driver

azmy60:


apps/studio/src/shared/lib/knex-clickhouse/index.ts
```diff
@@ -0,0 +1,369 @@
+// "forked" from  https://github.com/cleverbrush/framework/blob/e1dbee24d205ddb84807ae5327c8fe9592cd6064/libs/knex-clickhouse/src/index.ts
```
rathboma, line 1: This can't be included unless they make it available under the MIT/Apache2 license. Even then you need to replicate the copyright notice at the top.

Otherwise you'll have to clean-room reimplement this file I'm afraid. I filed an issue with them to ask them to include a license.

apps/studio/src/shared/lib/sql/change_builder/ClickHouseChangeBuilder.ts
```diff
@@ -0,0 +1,176 @@
...
+    console.log(columnInfo)
+    if (columnInfo.nullable) {
+      const re = /^Nullable\((.*)\)$/
+      const match = re.exec(dataType)
+      dataType = match?.[1]
+    }
+    return `ALTER COLUMN ${this.wrapIdentifier(column)} TYPE ${nullable ? `Nullable(${dataType})` : dataType}`
+  }
+
+  // existingColumns: SchemaItem[]
```
rathboma, line 58: Can we resolve this commented out code? 
Need it? Uncomment, 
Don't need it? Let's delete it.

apps/studio/tests/integration/lib/db/clients/all.js
```diff
@@ -705,9 +718,14 @@ export const itShouldApplyAllTypesOfChangesCompositePK = async function(util) {
     ]
   }
 
+  function wait(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms))
```
rathboma, line 722: see shared/lib/wait

rathboma (changes requested):
Please check the license of the knex lib, but after that, looks good to merge.

apps/studio/src/shared/lib/knex-clickhouse/index.ts
rathboma: https://github.com/cleverbrush/framework/issues/53

rathboma: Actually, looks like they're using BSD 3-clause. So you'll need a copyright statement on the files you've forked.

rathboma:
@azmy60 Oh, also -- please add a docs page for ClickHouse!

apps/studio/src/shared/lib/knex-clickhouse/index.ts
azmy60: Seems like we need them to add their license file first so we can copy it over to our forked files?

apps/studio/package.json
```diff
@@ -29,6 +29,8 @@
     "@aws-sdk/client-redshift": "^3.145.0",
     "@azure/msal-node": "^2.12.0",
     "@electron/remote": "^2.0.10",
+    "@cleverbrush/knex-clickhouse": "^1.1.10",
```
rathboma, line 32: This adds knex-clickhouse, but you also have `knex-clickhouse` in the source tree. Do we need both?

apps/studio/src/lib/db/clients/clickhouse/ClickHouseCursor.ts
```diff
@@ -0,0 +1,60 @@
...
+    const offset = this.cursorPos * this.chunkSize;
+    const limit = this.chunkSize;
+    const query = ClickHouse.buildSelectTopQuery(
+      this.options.table,
+      offset,
+      limit,
+      this.options.orderBy,
+      this.options.filters
+    );
+    const result = await this.client.query({
```
rathboma, line 44: This isn't running inside of a transaction, so this cursor isn't really generating a consistent view right? Are there no streaming APIs for node-clickhouse?

Also according to the docs, the query_id must be different on each call:
```
If you are overriding the query_id parameter, you need to ensure its uniqueness for **every call**. A random UUID is a good choice.
```

rathboma: Their github has some examples of using streams. I recommend using `JSONCOMPACTEACHROW` -https://clickhouse.com/docs/en/interfaces/formats#jsoncompacteachrow

Original discussion:
https://github.com/ClickHouse/clickhouse-js/issues/243

Example code:
https://github.com/ClickHouse/clickhouse-js/blob/a351508eda15c27e47ffcaa4ec96b63016d2e87b/examples/node/select_streaming_for_await.ts

rathboma (changes requested):
Two things:
1. Two knex libs?
2. Clickhouse cursor needs to do actual data streaming

apps/studio/src/lib/db/clients/clickhouse/ClickHouseCursor.ts
azmy60: There's a similar code in firebird cursor so I'll fix that as well 😅

apps/studio/src/common/interfaces/IConnection.ts
```diff
@@ -11,6 +9,7 @@ export function isUltimateType(s: ConnectionType) {
     'firebird',
     'cassandra',
     'libsql',
+    // 'clickhouse',
```
rathboma, line 12: Intended?

rathboma: Seems like it should be an ultimate type

apps/studio/src/lib/db/clients/clickhouse.ts
```diff
@@ -0,0 +1,1167 @@
...
+        result.resultType === "stream"
+          ? await streamToString(result.data)
+          : result.data;
+
+      const isEmptyResult = !data;
+
+      if (isEmptyResult) {
+        ret.push({
+          fields: [],
+          affectedRows: undefined, // TODO (azmi): implement affectedRows
```
rathboma, line 712: Maybe just make this 0?

apps/studio/src/lib/db/clients/clickhouse/ClickHouseCursor.ts
```diff
@@ -0,0 +1,66 @@
...
+      query_params: this.options.params,
+      query_id: this.queryId,
+      format: "JSONCompactEachRow",
+    });
+
+    this.stream = result.stream();
+
+    return new Promise((resolve, reject) => {
+      this.stream.on("data", (rows) => {
+        rows.forEach((row) => {
```
rathboma, line 43: You need to `pause` the stream when you have `chunkSize` rows, otherwise you're not controlling how many rows get pushed to you from the client. What if it pushes a billion rows before `read` is called?

The ReadableStream implements `Stream` from the stream node library:

https://github.com/ClickHouse/clickhouse-js/blob/43751b0012e0a3e26523eed6ca333df7d81f7f4d/packages/client-node/src/result_set.ts#L23

apps/studio/src/lib/db/clients/clickhouse/ClickHouseCursor.ts
```diff
@@ -0,0 +1,66 @@
...
+          this.rows.push(row.json());
+        });
+      });
+      this.stream.on("end", () => {
+        resolve();
```
rathboma, line 48: This should resolve when the stream is READABLE, not when it's finished.

apps/studio/src/lib/db/clients/clickhouse/ClickHouseCursor.ts
rathboma: Ok, you really need to look at the `MysqlCursor`, as it implements this type of cursor exactly.

As it stands yours has a lot of bugs:

1. Your `start` promise doesn't resolve until the whole stream is empty, rather than when the cursor is ready for `read` to be called.
2. Your `read` method doesn't wait for the rows array to be `chunkSize` big
3. If the row buffer exceeds `chunkSize`, but `read` hasn't been called, you're not pausing the stream, you need to otherwise you get a massive memory leak

There are probably other problems, but that's def enough, lol. Def check the mysql implementation

rathboma (changes requested):
Cursor is broken, some minor issues in the main driver.

apps/studio/src/lib/db/clients/clickhouse.ts
```diff
@@ -0,0 +1,1167 @@
...
+    _schema?: string
+  ): Promise<string> {
+    tableName = `${ClickHouseData.wrapIdentifier(
+      this.database.database
+    )}.${ClickHouseData.wrapIdentifier(tableName)}`;
+    duplicateTableName = `${ClickHouseData.wrapIdentifier(
+      this.database.database
+    )}.${ClickHouseData.wrapIdentifier(duplicateTableName)}`;
+    return `
+      CREATE TABLE ${duplicateTableName} AS ${tableName};
```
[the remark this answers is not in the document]
azmy60: Aren't they already escaped with `wrapIdentifier`?

apps/studio/src/lib/db/clients/clickhouse.ts
```diff
@@ -0,0 +1,1167 @@
...
+
+  private async runWithTransactionIfSupported(fn: () => Promise<any>) {
+    let supportsTransaction = false;
+
+    try {
+      log.debug("Checking if transaction is supported");
+      await this.driverExecuteSingle("BEGIN TRANSACTION");
+      supportsTransaction = true;
+    } catch (e) {
+      log.error(e);
```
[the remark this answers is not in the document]
azmy60: There are no clickhouse versions that support it by default. It's under an experimental flag, so you need to enable it from the server to use transactions. I can add a comment about that though.

apps/studio/src/lib/db/clients/clickhouse.ts
```diff
@@ -0,0 +1,1167 @@
...
+      }
+      throw e;
+    }
+  }
+
+  private async updateValues(updates: TableUpdate[]) {
+    log.info("Applying updates", updates);
+    let results: TableUpdateResult[] = [];
+
+    const updateQueries = buildUpdateQueries(this.knex, updates);
```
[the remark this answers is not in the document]
azmy60: Yep, that method is called inside a transaction method. You can see it in `applyChanges()`.

apps/studio/src/lib/db/clients/clickhouse.ts
```diff
@@ -0,0 +1,1167 @@
...
+      params: { table, ...valueParams },
+      returnType: "RowAsObject",
+    });
+    const json = result.data as ResponseJSON;
+    results.push(json.data);
+    return results;
+  }
+
+  private async deleteValues(deletes: TableDelete[]) {
+    log.info("Applying deletes", deletes);
```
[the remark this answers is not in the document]
azmy60: `deleteValues` is called inside a transaction. You can see it in `applyChanges()`.

apps/studio/src/lib/db/clients/clickhouse/ClickHouseCursor.ts
```diff
@@ -0,0 +1,100 @@
...
+    });
+
+    this.stream = result.stream();
+
+    this.stream.on("data", this.handleRows.bind(this));
+    this.stream.on("end", this.handleEnd.bind(this));
+    this.stream.on("error", this.handleError.bind(this));
+  }
+
+  private handleRows(rows: Row[]) {
```
azmy60, line 51: @rathboma This is probably the closest implementation of the cursor. I used the exact codes from `MysqlCursor`. But the only different thing is `ClickHouse` stream returns the data in "chunks" which can contain a size of rows that may vary instead of "a single row". Unfortunately, I couldn't find a way to control it.

I tried [max_block_size](https://clickhouse.com/docs/en/operations/settings/settings#setting-max_block_size) and [limit](https://clickhouse.com/docs/en/operations/settings/settings#limit) to set the chunk size but those don't work like how we want.

I can fake the `chunkSize` by making a new array or something else but are we OK to ignore `chunkSize` actually? Maybe it's possible to make a flag that says "this driver doesn't support chunkSize stream"?

rathboma:
@azmy60 fix conflicts, then merge away!

not-night-but:
@azmy60 @rathboma this should also be a commercial driver, no?

azmy60:
Yeah, it should be ultimate only.

rathboma:
@azmy60 I'm going to have Day move this to src-commercial, then merge

PR #2351, opened 2024-08-09
Remove platformInfo import from the frontend

not-night-but:
This PR removes any imports of `platform_info.ts`, and instead uses an object that is received over IPC. This is all to comply with vite's wishes. 

From now on, in the renderer, we should be using `this.$config` in vue components, and if it's needed, it is also available on `window.platformInfo`. If `platform_info.ts` is imported in the renderer, an error is thrown.


rathboma (approved):
Dope

PR #2352, opened 2024-08-09
add missing `setElementName` to `connHandlers`

azmy60:


not-night-but (approved):
Just make sure you add that function to the `IBasicDatabaseClient` interface, and then you're good to go!

PR #2372, opened 2024-08-21
One app to rule them all

rathboma:
TODO:
- [x] Backup, import, and multi-export BACKEND code goes in src-commercial/
- [x] All front-end code stays in src/
- [x] Entrypoints (utility, renderer, main, preload) move to src-commercial/entrypoints
- [x] Move paid drivers to src-commercial/lib/db (libsql, oracle, cassandra, ...other new ones)
- [x] Initialization code for drivers will need to be in src-commercial/
- [x] esbuild and vite config needs changing for the new entrypoints
- [x] Github workflow script needed to make sure nothing in src-commercial is ever imported into src


not-night-but:
Code has all been moved to the appropriate places, commercial code is never included in the open-source code, and there is a workflow in place to ensure this now (it's not separate, I just added it to the non-terrible tests, can be added as its own check if we want). There are still a litany of bugs, so those will need to be squashed, but the main goal of this was to just get the code put together.

PR #2381, opened 2024-08-25
Fix typeorm

not-night-but:
Fix some issues with the typeorm upgrade

### DIFF
--- a/apps/studio/src/common/appdb/models/OpenTab.ts
+++ b/apps/studio/src/common/appdb/models/OpenTab.ts
@@ -18,7 +18,7 @@ export class OpenTab extends ApplicationEntity {
       this.tabType = init
       return this;
     }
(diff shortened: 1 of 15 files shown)


apps/studio/src/common/appdb/models/OpenTab.ts
```diff
@@ -13,15 +13,13 @@ const pickable = ['title', 'tabType', 'unsavedChanges', 'unsavedQueryText', 'tab
...
-
-  constructor(init: TabType | TransportOpenTab) {
-    super()
+  withProps(init: TabType | TransportOpenTab): OpenTab {
     if (_.isString(init)) {
       this.tabType = init
-      return;
+      return this;
     }
     _.merge(this, init)
```
rathboma, line 21: restore this to `OpenTab.merge`

PR #2382, opened 2024-08-25
feat: focus query editor on window focus

### LINKED ISSUE #2369: BUG: Editor focus is lost after app switch
**Describe the bug**
When focus is in the Query editor and you switch away from the app and then return, the editor has lost focus. So you can't, for example, run the query again.

**To Reproduce**
1. Open a database
2. Create a query (e.g. "SELECT * FROM my_table;")
3. Execute the query (e.g. <kbd>Cmd</kbd>-<kbd>Shift</kbd>-<kbd>Return</kbd>)
4. Switch away from the app (<kbd>Cmd</kbd>-Cmd)
5. Return to the app and try to run the query again with Cmd-Shift-Return.

**Expected behavior**
The app should remember its focused element. But it does not, so the query is not executed again. You have to click in the query view again.

**Screenshots / Gifs**

https://github.com/user-attachments/assets/34681568-bffb-4b4b-8305-087766dad1eb

 - OS: macOS Sonoma 14.5
 - App Version: 4.6.4
 - Database type and version: SQLite3 3.43.2

therealrinku:
fixes #2369


apps/studio/src/components/common/texteditor/TextEditor.vue
```diff
@@ -129,6 +130,11 @@ export default {
     },
   },
   methods: {
+    focusEditor() {
+      if(this.editor && this.autoFocus){
```
rathboma, line 134: I think this might need to be a level up for the cases where there are multiple editor tabs.

@azmy60 Are there changes that need to be made here to account for `tab.active`?

rathboma:
This looks good, but I need @azmy60 to check if we also need to see if the editor is active (the tab is selected)

apps/studio/src/components/common/texteditor/TextEditor.vue
azmy60: Yes, this will steal focus from other components even though the text editor wasn't focused.

Right before the window is blurred, we should check if the editor is on focus. Once we reopen the window, the text editor knows if it should focus or not. These are the steps:

1. Before the window is blurred, we can use `document.activeElement` to see if text editor was focused? put this in a state (called `focusAfterWindowFocus` or smth like that 🤷)
2. After the window gains focus, we should call focus() if the state is true, and then switch the state to false

azmy60 (changes requested):

therealrinku:
hey @azmy60 
I have updated the code to check if textarea was focused, I'm also checking for `activeElement.className === "tabulator-tableholder"` because when user clicks run button on query editor, tabulator table will be focused actually. I believe in that case as well query editor should be focused when user is back, since it was the exact issue mentioned on issue 2369.

you can check it out if it's good to go!

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -389,7 +389,7 @@
       ...mapState('tabs', { 'activeTab': 'active' }),
       userKeymap: {
         get() {
-          const value = this.settings?.keymap.value;
+          const value = this.settings?.keymap?.value;
```
azmy60, line 392: I have a PR that will fix the `keymap` from returning null or undefined so it should be fine without the `?`. But It's fine with it too regardless.

azmy60 (changes requested):
This looks awesome! Just a couple of minor fixes and we're good!

azmy60 (approved):

PR #2386, opened 2024-08-27
add sidebar to TableTable

### LINKED ISSUE #1002: FEAT: detailed view for a row
For tables that have a large number of columns, it can sometimes be clunky to view specific data, especially when you need to view data from multiple columns at the same time

It would be nice to be able to select a column and view it in some sort of details mode, where the columns could actually be displayed as rows, in a vertical manner so it's easy to see all the content at once.

The alternative would be to write a query only for the fields you want, but then you'd have to re-write the query every time your needs change so it's not really a good solution.

**Additional context**
DBeaver has a feature that let's you select a single record and get a detailed view for it:
image
image

It would be great if Beekeeper Studio had a similar feature :slightly_smiling_face: 

All DBs

### LINKED ISSUE #2277: FEAT: Right-sidebar detail view for any row in a data table
We have some problems:

1. Viewing rows with JSON/JSONB data is clunky
2. Wide rows are hard to search and pull information out of when the column is alllll the way to the right ------>
3. It'd be nice to offer some sort of foreign key data preview without having to open another tab
4. The app would be TERRIBLE for NoSQL databases with multi-level data columns.

What's the solution? A JSON sidebar with a detailed view of all data in the row. Simply arrow up and down the row to easily scan for data.

Right Sidebar (detail view)

Some notes:
1. Visually this sidebar lives WITHIN a tab, not outside of it
2. If someone opens the sidebar, it should be open on all tabs that support it (tabletable, querytab)
3. Each tab maintains its own state, don't use a store for all the tabs.

Some great things about this feature:
- Can be implemented without any driver changes (!)
- Codemirror supports a read-only json view with collapsible fields already (quick to implement)
- Can be extended to support other views pretty easily

azmy60:
resolves #1002, resolves #2277

detail-view


rathboma (approved):
Looks great Azmi! Going to merge this in, there will probably be a lot of design changes needed before the release, but this is great for now.

PR #2392, opened 2024-08-31
Fix Postgres malformed array literal errors on indexes

### LINKED ISSUE #1334: BUG: Structure view breaks with tstzrange exclusion constraint
**Describe the bug**
The Structure view of a table with an timestamptz exclusion constraint is broken, returning a `malformed array literal: "{<constraint goes here>}"` error message.

**To Reproduce**
1. A table has the following exclusion constraint, defined with Typeorm @Exclusion directive (copied from the codebase I am on, with column names replaced)
```TypeScript
@Exclusion(
  'UQ_NON_OVERLAPPING',
  `USING gist (
  "id" WITH =,
  TSTZRANGE("startAt", "endAt") WITH &&
)`
)
```
`id` is the primary key of that table, with `startAt` and `endAt` being `timestamptz` columns.
2. Right click on the table in the sidebar
3. Click on View Structure
4. See error `malformed array literal: "{tstzrange("startAt", "endAt")}"`

**Expected behavior**
The structure of the table, with information on all the columns and their types should have been displayed.

**Screenshots / Gifs**
image

 - OS:
Edition         Windows 11 Home
Version	        21H2
OS build	22000.832
Experience	Windows Feature Experience Pack 1000.22000.832.0
 - App Version 3.6.2
 - Database type and version: Postgres version 12.

Kratious:
fix #1334 

The main issue is the expression text to array conversion happening in `(('{' || pg_get_expr(i.indexprs, i.indrelid)|| '}')::text[])[k.i]`. Commas in an expression become delimiters and split the expression up prematurely. Double quotes and other special characters may cause the malformed array literal errors.

The postgres function `pg_get_indexdef` does what the snippet above is trying to do, but safely.

Here's a minimal test case I ran through:

Table
```
CREATE TABLE
  public.test_indexes (
    first_name text NOT NULL,
    last_name text NOT NULL,
    data jsonb NOT NULL DEFAULT '{"a": {"b": ["foo", "bar"]}}'::jsonb
  );
```

Indexes
```
CREATE INDEX single_column ON test_indexes (first_name);
CREATE INDEX multi_column ON test_indexes (first_name, last_name);

CREATE INDEX single_expression ON test_indexes (lower(first_name));
CREATE INDEX multi_expression ON test_indexes (lower(first_name), lower(last_name));

CREATE INDEX expression_with_comma ON test_indexes ((lower(first_name) || ', ' || lower(last_name)));
CREATE INDEX expression_with_double_quote ON test_indexes (('"' || first_name));
CREATE INDEX expression_with_jsonb_operator ON test_indexes ((data #>> '{a,b,1}'));
```

In `View Structure -> Indexes`:
image


rathboma (changes requested):
This is wonderful, thank you!

Would you please add a test for this, so that we don't change something in the future and accidentally break your fix?

You can add code to `apps/studio/tests/integration/lib/db/clients/postgres.spec.ts`

It's fairly simple to add a table in the `beforeAll`, and then add a test to make sure the method works.

Kratious:
I wasn't able to get the integration tests running on my machine, so I can't attest to the test's correctness 🙂

rathboma:
Thanks so much for this and sorry for the late review.

It looks good, but I think it will probably break `redshift.ts` and `cockroach.ts` - both are downstream clients that depend on the Postgres client implementation of this function.

Also looks like the test is broken.

You should be able to run the tests locally using:

`yarn test:integration postgres`

rathboma (changes requested):
Broken test and likely this will break redshift & cockroachdb?

AykutCevik:
Came into this exact issue while creating an index on Postgres and viewing the table structure in Beekeeper:
```sql
CREATE INDEX table_idx ON table USING GIN (
    (
        setweight(to_tsvector('english', coalesce(col1, '')), 'A') ||
        setweight(to_tsvector('english', coalesce(col2, '')), 'B')
    )
);
```

Now, Beekeeper is not able to view the structure of the table anymore. Works in DBeaver though.

bqrkhn:
I'm also facing this issue.  `data` is a `jsonb` column.

Capture-2024-12-24 at 11 31 04

rathboma:
@not-night-but Hoping you can step in to take over this bug fix?

not-night-but:
@rathboma I've fixed the tests, updated the branch, and tested on cockroach and redshift. Cockroach works like a dream, and we don't actually support indexes on redshift so this seems like it should be good to go!

rathboma:
@not-night-but doesn't look like the test checks the output of indexes with array literals? Please fix the test then good to merge

rathboma:
@not-night-but can you wrap this up ready for the 5.2 release?

azmy60:
@rathboma @not-night-but I just fixed the merge conflicts and added tests that check the output of indexes with array literals, though I couldn't make it very specific with the content of the array literals because it returns limited info, but it should be good enough I think.

not-night-but (approved):

Issue #2401, opened 2024-09-03
BUG: shift+cmd+bracket doesn't work for switching tabs on Mac


rathboma:
**Describe the bug**
`shift+cmd+[` and `shift+cmd+]` should be bound to previous/next tab. This is a MacOS default. I thought I fixed this at somepoint, but probably broke it with a bad merge.

PR #2403, opened 2024-09-04
fix: missing shortcuts added for tab switching on macos

therealrinku:
fixes: #2401


apps/studio/src/components/CoreTabs.vue
```diff
@@ -408,6 +408,8 @@ import { TransportOpenTab, setFilters, matches, duplicate } from '@/common/trans
         'alt+7': this.handleAltNumberKeyPress,
         'alt+8': this.handleAltNumberKeyPress,
         'alt+9': this.handleAltNumberKeyPress,
+        'shift+meta+[':this.previousTab,
```
rathboma, line 411: We'll need to check for MacOS here!

```
if (this.$config.isMac) {
  results["Cmd+["] = this.previousTab
  results["Cmd+]"] = this.nextTab
}
```

rathboma (changes requested):
Thanks! Needs a MacOS filter, but then it's good

rathboma (approved):

Issue #1980, opened 2024-02-26
BUG: SQL: Create - misses PRIMARY KEY, does not show all constraints


lowang:
**Describe the bug**
When I use SQL: Create I've got following:
```
CREATE TABLE
  transactions (
    source varchar(50) NOT NULL,
    key varchar(32) NOT NULL,
  )
ALTER TABLE
  transactions
ADD
  CONSTRAINT PK__dbo_t__3213E83FEE0B4DDE PRIMARY KEY (id)
``` 
However when I'm using *Azure Data Studio* I also get UNIQUE constraint on 2 columns *source*, *key*:
```
CREATE TABLE [dbo].[transactions](
	[id] [int] IDENTITY(1,1) NOT NULL,
	[source] [varchar](50) NOT NULL,
	[key] [varchar](32) NOT NULL,
PRIMARY KEY CLUSTERED 
(
	[id] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY],
UNIQUE NONCLUSTERED 
(
	[source] ASC,
	[key] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON) ON [PRIMARY]
) ON [PRIMARY] TEXTIMAGE_ON [PRIMARY]
```

**To Reproduce**
1. create following MSSQL table:
```
CREATE TABLE transactions (
        id INTEGER NOT NULL IDENTITY,
        source VARCHAR(50) NOT NULL,
        [key] VARCHAR(32) NOT NULL,
        PRIMARY KEY (id),
        UNIQUE (source, [key])
)
```
1. Right click table in Entities browser
3. Click on 'SQL: Create'

**Expected behavior**
I would expect to get following `id` as a first column and constraint on `(source, [key])`

**Version Information (please complete the following information):**
 - OS: Mac Apple M1
 - App Version: Beekeeper Studio 4.1.13
 - Database mcr.microsoft.com/mssql/server:2022-latest

**Additional context**
Running mssql from docker image `mcr.microsoft.com/mssql/server:2022-latest`

mattgen88:
[...]

PR #2417, opened 2024-09-11
Fix SQL Script Generation for Multiple Primary Keys in `PostgreSQL`

luizfelipemacedo:
This PR addresses an issue where SQL scripts generated for tables in PostgreSQL with multiple primary keys were incorrect. Previously, the script would return only the first primary key, resulting in incomplete tables and incorrect behavior in the database.

This issue was reported in [Beekeeper Studio GitHub issue #1980](https://github.com/beekeeper-studio/beekeeper-studio/issues/1980#issuecomment-2136460070).

#### **Changes:**
- Updated the `getTableCreateScript` method to correctly identify and combine all primary keys returned by the database.
- Previously, the result returned two instances of `ALTER TABLE`, each defining a separate primary key, which was invalid.
- Now, the primary keys are combined into a single `PRIMARY KEY` declaration, fixing the issue.
- Some formatting was done using Prettier in the file `apps/studio/src/lib/db/clients/postgresql.ts`.

#### **Expected Outcome:**
Now, the SQL script generated for a table with multiple primary keys will be formatted correctly, as shown below:

```sql
CREATE TABLE public.test_pk_script (
  first_key character varying(255) NOT NULL,
  second_key character varying(255) NOT NULL
);

ALTER TABLE
  public.test_pk_script
ADD
  CONSTRAINT test_pk_script_pkey PRIMARY KEY (first_key, second_key)
```

#### **Impact:**
- Improves accuracy and reliability when generating creation scripts for tables with multiple primary keys.
- Fixes issues when creating tables in PostgreSQL and enhances data integrity.

#### **Tests:**
- Added test to ensure expected behavior

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -4,7 +4,7 @@ import { readFileSync } from 'fs';
 
 import pg, { QueryResult, PoolConfig, PoolClient } from 'pg';
 import { identify } from 'sql-query-identifier';
-import _  from 'lodash'
+import _ from 'lodash'
 import knexlib from 'knex'
 import logRaw from 'electron-log'
 
@@ -31,7 +31,7 @@ const PD = PostgresData
 const log = logRaw.scope('postgresql')
 const logger = () => log
 
-const knex = knexlib({ client: 'pg'})
+const knex = knexlib({ client: 'pg' })
 
 const pgErrors = {
   CANCELED: '57014',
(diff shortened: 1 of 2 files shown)


rathboma (changes requested):
This is awesome! Could you add a test for this to tests/integration/lib/db/clients/postgresql.spec.ts?

That way we can make sure we never break this by mistake.

luizfelipemacedo:
> This is awesome! Could you add a test for this to tests/integration/lib/db/clients/postgresql.spec.ts?
> 
> That way we can make sure we never break this by mistake.

Thank you very much for the feedback! A test has been added to ensure the expected behavior. Special thanks to my friend @queirozlc who helped me 🤝

rathboma (approved):

PR #2427, opened 2024-09-15
Quick filters ⚡

### LINKED ISSUE #2389: FEAT: Quick filter tables based on current cell value + right click
Allow users to easy filter a table using the value in a specific cell.

Right click a cell -> Quick filter -> options

Table quick filter

style guidelines:

1. the operators (=, !=, >=, etc) should be the same width (even though some are 2 characters and some are 1 character), so it is easy to scan without too much thought.
2. Overflow - the value should be truncated to 80 characters and should overflow if too long.
3. If tabulator doesn't allow easy dynamic menus like this (IE including the column name and value), then do what works rather than spending a week custom building something.

rathboma:
- Right click any value to filter the table by that value
- Disabled for null values
- Added some testing for filter filtering
- Fixed a bug where voiding out a filter doesn't reset the table filter

Fix #2389
Peek 2024-09-14 22-07

### DIFF
--- a/apps/studio/src/common/utils.ts
+++ b/apps/studio/src/common/utils.ts
@@ -131,15 +131,21 @@ export function createTableFilter(field: string): TableFilter {
   return { op: "AND", field, type: "=", value: "" }
 }
 
+// isEmpty(1) returns true, we don't want that!
+// https://stackoverflow.com/questions/36691125/lodash-isblank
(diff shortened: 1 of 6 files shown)


rathboma:
only annoying thing is having to click the 'quick filter' menu item to get it to open. Not sure if tabulator supports hover activation

PR #2428, opened 2024-09-15
Support fetching comment in column in PostgreSQL

bgb10:
Screenshot 2024-09-14 at 9 49 36 PM

* Even though there was `comment` column in `Columns` Tab, fetching comment in PostgreSQL didn't work. (To check bug, you need to refresh the page because changes are kept in local state)
* Found that there was no comment fetching logic in `postgresql.ts/listTableColumns`, but `mysql.ts/listTableColumns` was.
* Modified query and return object for supporting `comment` in postgresql.
* By the way, adding comment worked well because it was handled in different function.

### Previous
https://github.com/user-attachments/assets/70a397b6-327c-4681-a8d4-1ed7f34e775a

### Current
https://github.com/user-attachments/assets/a92d6921-ab7a-4eb5-a8a8-f04c992b09f0

### DIFF
--- a/apps/studio/src/lib/db/clients/cockroach.ts
+++ b/apps/studio/src/lib/db/clients/cockroach.ts
@@ -1,6 +1,6 @@
 import globals from "@/common/globals";
 import pg, { PoolConfig } from "pg";
-import { FilterOptions, SupportedFeatures, TableIndex, TableOrView, TablePartition, TableProperties, TableTrigger } from "../models";
+import { FilterOptions, SupportedFeatures, TableIndex, TableOrView, TablePartition, TableProperties, TableTrigger, ExtendedTableColumn } from "../models";
 import { PostgresClient, STQOptions } from "./postgresql";
 import _ from 'lodash';
 import { defaultCreateScript } from "./postgresql/scripts";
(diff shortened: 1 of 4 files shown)


apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -348,7 +348,8 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult> {
         CASE
           WHEN data_type = 'ARRAY' THEN 'YES'
           ELSE 'NO'
-        END as is_array
+        END as is_array,
+        pg_catalog.col_description(format('%s.%s', table_schema, table_name)::regclass::oid, ordinal_position) as column_comment
```
rathboma, line 352: Thanks!
Couple of questions on this function:
- Does it work for Redshift and CockroachDB? They are subclasses of the postgreSQL driver. If the tests fail, you will have to add a `listTableColumns` for their drivers specifically.
- Can you add a test for this in `tests/integration/lib/db/client/postgres.spec.ts`? It should confirm that comments can be added and returned.

rathboma (changes requested):
Amazing change thank you! The SQL looks good, I just have a couple of comments about DB compatibility and testing

rathboma:
Also :wave: hi!

bgb10:
By the way, does it possible to run integration test in local? I tried to, but it failed with so many connection fails. Also, even though my unit test (`yarn run test:unit`) passed in local, it seems to fail during CI. 
Screenshot 2024-09-15 at 10 18 24 AM

rathboma:
Hey! Yes it's possible to just run the postgres ones too:

yarn test:integration postgres cockroach # this will run both postgres and cockroach integration tests, but nothing else

apps/studio/src/lib/db/clients/postgresql.ts
bgb10: > Does it work for Redshift and CockroachDB? They are subclasses of the postgreSQL driver. If the tests fail, you will have to add a listTableColumns for their drivers specifically.
- `postgreSQL` and `cockroachDB` both supports `pg_catalog.col_description` (because cockroachDB uses postgreSQL driver)
- However, old version of cockroachDB(v21) [[does not support](https://www.cockroachlabs.com/docs/v21.1/functions-and-operators)](https://www.cockroachlabs.com/docs/v21.1/functions-and-operators) `format` syntax, so unit test error happens(beekeeper use v21 when doing unit test). But, latest version of cockroachDB(v24) [[supports](https://www.cockroachlabs.com/docs/v24.1/functions-and-operators)](https://www.cockroachlabs.com/docs/v24.1/functions-and-operators)  `format` syntax, as it is kind of basic SQL syntax.
- `cockroachDB` supports separate [`[column_comment](https://www.cockroachlabs.com/docs/v21.1/information-schema#columns)`](https://www.cockroachlabs.com/docs/v21.1/information-schema#columns) column in `information_schema.columns`, which `postgreSQL` doesn’t have.
- _In conclusion, I’ll keep `listTableColumns` in `postgresql.ts`, but I’ll write separate function for `cockroach.ts` by using `column_comment`. This can support lower versions of `cockroachDB` and also can use more clear syntax `column_comment` which `cockroachDB` supports._

bgb10: > Can you add a test for this in tests/integration/lib/db/client/postgres.spec.ts? It should confirm that comments can be added and returned.
Done :)

bgb10:
@rathboma Done. Can I get approval to run CI? Thanks for making this awesome library ☺️

apps/studio/src/lib/db/clients/cockroach.ts
```diff
@@ -26,6 +26,61 @@ export class CockroachClient extends PostgresClient {
     return [];
   }
 
+  async listTableColumns(table?: string, schema: string = this._defaultSchema): Promise<ExtendedTableColumn[]> {
```
rathboma, line 29: I think you'll have to do the same thing for `redshift.ts` also.

rathboma (changes requested):
Exception also needed for Redshift (although the tests won't pick that up)

apps/studio/src/lib/db/clients/cockroach.ts
bgb10: Done. I think we need to add integration test at a separate PR(not in this task).

bgb10:
@rathboma I think `sqlserver.spec.ts` error is not caused by this changes. Is there any other things that I can help?

apps/studio/src/lib/db/clients/cockroach.ts
rathboma: Yeah, we don't really have a way to run Redshift integration tests right now, but that'll come in the future

rathboma (approved):

PR #2436, opened 2024-09-16
single app upgrade

azmy60:


apps/studio/src/components/LicenseExpiredModal.vue
```diff
@@ -0,0 +1,91 @@
...
+    purchase() {
+      this.close();
+      this.$root.$emit(AppEvent.enterLicense);
+    },
+    async onLicenseExpired(_license) {
+      if (!this.isOpeningModalAllowed()) return;
+      this.$modal.show(this.modalName);
+    },
+    async isOpeningModalAllowed() {
+      const openExpiredLicenseModal = await this.$util.send(
```
rathboma, line 69: This is something we store to determine if we've already shown the modal right?

But doesn't this modal work for both `trial expiry` and `license expiry`? If so - don't we need two checks?

Also -- We should probably just use `localStorage` for this. It's not a 'setting' in the traditional sense, it's just a state that needs remembering.

See `LocalStorage.ts` for `SmartLocalStorage`

rathboma: I think this should just be two files, not one. No reason to double up the modals. We'll also want to customize each modal separately.

rathboma (changes requested):
Some minor changes, then this looks good.

Question to think about:
I plan on having features gated behind different versions of the app soon. Eg you can only use Snowflake if you're on 'professional'. Does your design make that easy?

I'd extend it by having the cloud service return a list of things you're allowed access to, so we'd have to not only check `isUltimate`, but also `license.featureGrants.snowflake` for example.

Probably it would look something like:

```
featureGrants: {
  snowflake: true,
  sso: false,
```

apps/studio/src/common/platform_info.ts
```diff
@@ -66,6 +66,14 @@ if (isRenderer()) {
       sessionType === 'wayland' && !isWindows && !isMac
   }
 
+  const appVersion = testMode ? 'test-mode' : e?.app.getVersion() ?? p.env.version
+  const parsedAppVersion = appVersion.split('.').reduce((acc, v, i) => {
```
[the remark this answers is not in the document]
azmy60: Oh yeah, I didn't see that

apps/studio/src/handlers/devHandlers.ts
```diff
@@ -0,0 +1,41 @@
...
+import { LicenseKey } from "@/common/appdb/models/LicenseKey";
+
+export interface IDevHandlers {
+  'dev/switchLicenseState': ({ label, sId }: { label: string, sId: string }) => Promise<void>
+}
+
+export const DevHandlers: IDevHandlers = {
+  'dev/switchLicenseState': async function({ label }: { label: string, sId: string }) {
+    switch(label) {
+      case "First time install, no license, no trial.":
```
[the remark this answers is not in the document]
azmy60: Yes, I can hack it to do that. It's a little bit messy with how it works now.

apps/studio/src/handlers/devHandlers.ts
```diff
@@ -0,0 +1,41 @@
+import { getActiveWindows } from "@/background/WindowBuilder";
+import { LicenseKey } from "@/common/appdb/models/LicenseKey";
+
+export interface IDevHandlers {
+  'dev/switchLicenseState': ({ label, sId }: { label: string, sId: string }) => Promise<void>
+}
+
+export const DevHandlers: IDevHandlers = {
+  'dev/switchLicenseState': async function({ label }: { label: string, sId: string }) {
+    switch(label) {
```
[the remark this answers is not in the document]
azmy60: Yes I can add that. Though we already have that check in the `utility.ts`.

not-night-but:
`isUltimate` is still used in `BasicDatabaseClient.ts`: https://github.com/beekeeper-studio/beekeeper-studio/blob/69f207cb3e40ab1f7d1a779c6e7cc7bdf918008b/apps/studio/src/lib/db/clients/BasicDatabaseClient.ts#L77

rathboma:
image

Error

apps/studio/src/common/platform_info.ts
```diff
@@ -17,7 +20,18 @@ function isUtility() {
...
+let platformInfo: IPlatformInfo & { isUltimate: boolean, isCommunity: boolean };
+
+async function updateLicenseStatus() {
+  if (platformInfo.testMode) return
+  const status = getLicenseStatus({
+    licenses: await LicenseKey.find(),
+    currentDate: new Date(),
+    currentVersion: platformInfo.parsedAppVersion,
+  })
+  platformInfo.isUltimate = status.edition === 'ultimate'
```
rathboma, line 32: We can't use platformInfo for this anymore, because it's dynamic, not static.

apps/studio/src/common/platform_info.ts
```diff
@@ -94,13 +112,19 @@ if (isRenderer()) {
...
     // cloudUrl: isDevEnv ? 'https://staging.beekeeperstudio.io' : 'https://app.beekeeperstudio.io',
     cloudUrl: 'https://app.beekeeperstudio.io',
     locale,
+    isUltimate: false,
+    isCommunity: true,
 
     // cloudUrl: isDevEnv ? 'http://localhost:3000' : 'https://app.beekeeperstudio.io'
   }
+
+  updateLicenseStatus()
```
rathboma, line 126: 🐞🐞 This is not ok -- we can't have files run code on import like this -- firstly it doesn't work (crashes the app), but also we need all code like this behind an initializer of some sort.

Instead of `platformInfo.isUltimate`, I suggest something like:
```
const license = await LicenseKey.findBest()
license.isUltimate
```

apps/studio/src/common/platform_info.ts
```diff
@@ -94,13 +112,19 @@ if (isRenderer()) {
...
+  setInterval(updateLicenseStatus, globals.licenseCheckInterval)
```
rathboma, line 127: this can't be here. Needs to be inside the utility initializer

rathboma (changes requested):
We have licenses in two places now - utility and renderer, they can get out of sync. This isn't good. For example -- if someone adds a new license.

I think the easiest way to handle this is to keep the utility process STATELESS:
- Whenever the Utility process needs to check the license, it should just fetch the license from SQLite.
- There should be an interval on the utility process to refresh the sqlite data regularly from the API. This has to happen during utility initialization, not in platformInfo. This can be every 5 minutes or so.
- The front end can keep it's refresh code, but instead of getting license information from the API, it needs to get it from the utility process. This should happen every minute (now that the data is just in sqlite)
- remove platformInfo.isCommunity and platformInfo.isUltimate -- have to get that directly.

Also -- no need for all these other libs to return license information, just add new getters / methods to the `LicenseKey` model, and return a fully fleshed out json object to the front-end.

eg: `LicenseKey.findBestAvailable` -> can fetch all license keys and return the one with the most access. If no license key, it can return a special static `community` license key. This moves ALL license key logic to the model, which simplifies a lot.

You can add getters for `status`, rather than having it in a separate file. You can even add getters for `major`, `minor` and `patch`, and also for anything else that's parsed from licenses.

Users won't have more than one active license at a time, this can entirely be scoped to the model.

Also -- status of a license can be easy: `active`, `lifetime`, `expired`, `trial`, `community`

azmy60:
@rathboma Don't we need the libs for tests?

rathboma (approved):
BOOM

PR #2437, opened 2024-09-16
Fix/ultimate bugs

not-night-but:
These are just a bunch of small fixes for some bugs I've found since the merge, gonna move to some bigger ones so I wanted to get these in.

 - Fixed misuse of typeorm, since they deprecated a bunch of stuff in the last update
 - Removed all references to `platformInfo.isUltimate` and `platformInfo.isCommunity`. This will probably be overwritten by Azmy's changes
 - Fixed the context menu
 - Fixed used connections
 - Fixed backup/restore
 - Fixed ultimate themes
 - Fixed refreshing tables with pinned entities
 - Fixed saving used connections as saved connections
 - Fixed importing .sql files as saved queries

### DIFF
--- a/apps/studio/src/components/common/ContextMenu.vue
+++ b/apps/studio/src/components/common/ContextMenu.vue
@@ -25,8 +25,9 @@
               class="material-icons menu-icon"
               v-if="option.icon"
             >{{ option.icon }}</i>
+            <!-- NOTE (@day): this is supposed to only appear when you don't have an ult license, but this component can't use the store -->
             <i
-              v-if="option.ultimate && !hasActiveLicense"
+              v-if="option.ultimate"
               class="material-icons menu-icon"
             >stars</i>
           </span>
@@ -38,9 +39,9 @@
 
 <script lang="ts">
 
+// NOTE (@day): we can't use the store here for some reason
 import { ContextOption } from '@/plugins/BeekeeperPlugin'
 import Vue from 'vue'
-import { mapGetters } from 'vuex'
 
 export default Vue.extend({
   name: 'ContextMenu',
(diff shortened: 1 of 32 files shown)


apps/studio/src/components/common/ContextMenu.vue
```diff
@@ -38,6 +39,7 @@
 
 <script lang="ts">
 
+// NOTE (@day): we can't use the store here for some reason
```
rathboma, line 42: I think because context menus are dynamically mounted? There's probably a way to do it

apps/studio/src/components/common/ContextMenu.vue.orig
```diff
@@ -0,0 +1,207 @@
+<template>
+  <!-- Original file souce copyright John Datserakis https://github.com/johndatserakis/vue-simple-context-menu -->
```
rathboma, line 2: What is this file and why do we need it?

apps/studio/src/components/common/ContextMenu.vue.orig
```diff
@@ -0,0 +1,207 @@
...
+          <span v-html="option.name" />
+          <div class="expand" />
+          <span>
+            <span
+              class="shortcut"
+              v-if="option.shortcut"
+              v-html="option.shortcut"
+            />
+            <i
+<<<<<<< HEAD
```
rathboma, line 25: Also has merge conflicts in it

rathboma (approved):
One or two questions, plz resolve, then feel free to merge

apps/studio/src/components/common/ContextMenu.vue.orig
not-night-but: We don't need it uggghhh. meld likes to create files when I do a merge and they got auto staged for some reason. Will remove

PR #2454, opened 2024-09-23
disable minimal mode in v5

azmy60:


rathboma:
@azmy60 don't you have to remove it from the menus?

azmy60:
Oh yeah, I forgot that

azmy60:
@rathboma It's already [removed](https://github.com/beekeeper-studio/beekeeper-studio/blob/b0425a95438eb7bbb36063f0c939196e2c534c52/apps/studio/src/common/menus/MenuBuilder.ts#L21-L22) from the menus.

PR #2463, opened 2024-09-26
Feature/2176 retain last workspace

### LINKED ISSUE #2176: BUG: App doesn't remember the last used workspace when quitting and reloading the app
When starting Beekeeper Studio the app should remember the last workspace used, and select it by default.

Why?
- If you always want to save data to the cloud, it's annoying for the app to always start on the local workspace

wmontgomery:
resolves #2176


rathboma (approved):
Hey! Can we change this to use Local

PR #2483, opened 2024-10-06
feat: have the "buy beekeeper" update show up every 3 hours when using community

wmontgomery:
purchase reminder
The timer is set to every 5 minutes, this was every 30 seconds to show what it looked like


rathboma (changes requested):
Love this. Let's do a couple of things:
- 5 minutes would be super annoying, so let's change this so it's ~twice per day, so every 3 hours?
- Timeout of 10s means it goes away after 10s right? Make that 1m?
- Can you double check that this only triggers if the user is on the free version?

apps/studio/src/components/NotificationManager.vue
```diff
@@ -20,21 +22,45 @@ export default Vue.extend({
...
+  methods: {
+    initNotifyInterval() {
+      const intervalTime = 1000 * 60 * 60 * 3
+      if (this.notificationInterval) {
+        this.notificationInterval = null
+        clearInterval(this.notificationInterval)
+      }
+      if (this.timeoutID) {
+        this.timeoutID = null
+        clearTimeout(this.timeoutID)
```
not-night-but, line 47: Could be wrong here, but if you're setting the `notificationInterval` and `timeoutID` before clearing them, then clearing them does nothing, right?

not-night-but (changes requested):

apps/studio/src/components/NotificationManager.vue
wmontgomery: I think you may be right here. Oops. I switched the logic to clear the interval/timeout first, then set the data to null so hopefully the madness has stopped

rathboma (approved):

PR #2490, opened 2024-10-10
Remove Native Titlebar and add option to use beta releases

not-night-but:
There is no longer an option to use a native titlebar (as this is borked on the new version of electron :upside_down_face: ), I left it enabled for MacOS, but Mac also forces the client titlebar to show.

Added an option to toggle the beta channel in the app's updater. This hasn't been tested yet.
image

### DIFF
--- a/apps/studio/src/components/menu/NewAppMenu.vue
+++ b/apps/studio/src/components/menu/NewAppMenu.vue
@@ -35,7 +35,13 @@
               @mouseover.prevent="setHover(item)"
               :class="hoverClass(item)"
             >
-              <span class="label">{{ item.label }}</span>
+              <span class="label">
+                <span 
+                  class="material-icons" 
+                  v-if="item.checked"
+                >done</span>
+                <span>{{ item.label }}</span>
+              </span>
               <span class="shortcut">{{ shortcutText(item) }}</span>
             </a>
             <!-- Second Level Menu, eg Dark Theme, Light Theme -->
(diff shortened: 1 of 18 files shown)


.github/workflows/studio-publish.yml
```diff
@@ -153,30 +154,30 @@ jobs:
...
           ARCH: ${{ matrix.os.arch }}
         run: |
-          brew install python@3.8
+          brew install python@3.12
           if [ "$ARCH" = "arm64" ]; then
-              # Add Python 3.8 to PATH for arm64
-              echo "/opt/homebrew/opt/python@3.8/libexec/bin" >> $GITHUB_PATH
-              export PATH="/opt/homebrew/opt/python@3.8/libexec/bin:$PATH"
-              export PYTHON_PATH="/opt/homebrew/opt/python@3.8/libexec/bin/python3"
+              # Add Python 3.12 to PATH for arm64
```
rathboma, line 171: doesn't this need to be 3.11 to match the test build?

apps/studio/src/components/menu/NewAppMenu.vue
```diff
@@ -255,9 +261,11 @@ export default {
       // Empty on purpose
     }
   },
-  mounted() {
-    this.menuBuilder = new MenuBuilder(this.$store.state.settings.settings, this.actionHandler, this.$config)
-    this.menus = this.menuBuilder.buildTemplate()
+  async mounted() {
+    if (this.settings && this.settings.length > 0) {
```
rathboma, line 265: Doesn't this mean that it's possible for the app to start without a menu? Seems like something else should happen here.

rathboma (changes requested):
Couple of minor changes needed

rathboma (approved):

PR #2527, opened 2024-10-23
Fix Logging

not-night-but:
Biggest change here is that we will no longer import from `electron-log`. There is now an alias `@bksLogger` which will import the correct file. This also means that the proper settings for logging will be applied for every logger, no matter what. 

Utility logs will no longer appear in the devConsole, only in the terminal in dev mode. They also now go to their own file called `utility.log` in the logs folder.

I've also modified the format (mostly just to know that I could, but also to make it clear in the terminal which process the logs are coming from). Unfortunately there doesn't seem to be an easy way to format renderer logs when they appear in the terminal, I can also turn those logs off if we want (ie renderer logs would only be in the dev console, and only main and utility would be in the terminal).


apps/studio/src-commercial/entrypoints/main.ts
```diff
@@ -2,17 +2,13 @@
 import * as fs from 'fs'
 import path from 'path'
 import { app, protocol } from 'electron'
-import log from 'electron-log/main'
 import * as electron from 'electron'
 import { ipcMain } from 'electron'
 import _ from 'lodash'
+import log from '@/lib/log/bksLogger'
```
rathboma, line 8: you don't want this to be @bksLogger?

rathboma (changes requested):
This looks good, but a couple of open questions:
1. How do you prevent folks from importing electron-log (think about someone new to the project in 12 months time)?
2. How do you prevent renderer / utility folks from importing the wrong version of bksLogger?

apps/studio/src-commercial/entrypoints/main.ts
not-night-but: Oops, missed this

not-night-but:
@rathboma In regards to your questions, I think both of them could be solved with an eslint rule (although I'm not sure how much we actually use eslint), or with a github workflow to ensure that electron-log is not imported, and that only `@bksLogger` is imported.

azmy60 (commented):
This looks great! We probably just need to have something that checks if someone use the right logger or not. I could see that's something I can probably forget 😆

rathboma:
@not-night-but fix merge conflicts and merge it!

PR #2535, opened 2024-10-27
Feature/upsert for import take 2

### LINKED ISSUE #2158: FEAT: Import should be able to selectively UPDATE existing rows

wmontgomery:
resolves #2158


apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
```diff
@@ -379,14 +381,13 @@ export abstract class BasicDatabaseClient<RawResultType extends BaseQueryResult>
...
     const columns = await this.listTableColumns(tableInsert.table, tableInsert.schema);
-    tableInsert.data.forEach((row) => {
-      Object.keys(row).forEach((key) => {
-        row[key] = this.deserializeValue(row[key]);
-      })
-    })
-    return buildInsertQuery(this.knex, tableInsert, columns);
+    const primaryKeysPromise = await this.getPrimaryKeys(tableInsert.table, tableInsert.schema)
+    const primaryKeys = primaryKeysPromise.map(v => v.columnName)
+    return buildInsertQuery(this.knex, tableInsert, { columns, runAsUpsert, primaryKeys, createUpsertFunc: this.createUpsertFunc });
```
azmy60, line 390: It looks like we are missing a deserialization. `tableInsert` needs to be deserialized so knex can generate the right query (especially with buffer types). I missed adding the query test for buffers but you can also test this from the table view by right-clicking a row -> `copy as SQL`. This is a sample SQLite table with buffers:

```sql
CREATE TABLE test_table (id INTEGER PRIMARY KEY AUTOINCREMENT, data BLOB NOT NULL);
INSERT INTO test_table (data) VALUES (X'DEADBEEF'), (X'CAFEBABE');
```

P.S. I hate that the deserialization does nested loop like that :). There must be a better way probably.

azmy60 (changes requested):
Just found a regression on generating query with `copy as SQL`. Overall, looks awesome!

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
wmontgomery: @azmy60 , looks like a snippet got lost probably in one of the million merge conflicts that happened since starting this :D I added the snippet back and got the expected behavior.

azmy60: 😆 thanks!

azmy60 (approved):

rathboma:
@wmontgomery how to folks choose to upsert / insert? Is there an option in the import flow?

PR #2554, opened 2024-11-04
Fix titlebar listeners

not-night-but:
Since we turned on contextIsolation, we can no longer directly access the window in the titlebar. Now we attach listeners through preload to the BeekeeperWindow instance, which will fire off `maximize`, `unmaximize`, etc events when they are received, along with the sId so the event is sent to the correct renderer. This also seems to have fixed the double clicking issue (I think, I didn't notice any issues after making this change)

I tested this on the latest Ubuntu LTS, but let me know if it doesn't work on your system

### DIFF
--- a/apps/studio/src-commercial/entrypoints/main.ts
+++ b/apps/studio/src-commercial/entrypoints/main.ts
@@ -43,7 +43,6 @@ function initUserDirectory(d: string) {
 }
 
 let utilityProcess: Electron.UtilityProcess
-// don't need this
 let newWindows: number[] = [];
 
 async function createUtilityProcess() {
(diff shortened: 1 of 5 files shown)


azmy60:
@not-night-but  Double clicking the titlebar works on my windows machine, but it doesn't seem to update the icon state. Does this PR aim to fix that as well?

https://github.com/user-attachments/assets/45d0bfa9-05d2-4b15-b88a-3158aa77b1d1

not-night-but:
@azmy60 it should update the icon state 😡😡. I shall take a look

azmy60:
Nvm it actually works lol. I was on a different branch, my bad :(

azmy60 (approved):
billy approved

krasava

PR #2555, opened 2024-11-05
Binary serialization

azmy60:
- Fix apply changes execution and sql string generation with binary or buffer values
- Implement serialization
- Show binary values correctly
- Auto "hex string to buffer" conversion in table view

https://github.com/user-attachments/assets/aab0411b-cc8e-49db-bb51-3646fe708ebc


azmy60:
Yes. Clickhouse is prioritized as it seems that we only tested the latest version and not the older ones that folks are still using.

azmy60:
Ok this is ready for review.

azmy60:
This is how it works now:

Regardless of the column type, back-end will send UInt8Array to front-end if the the data is buffer.

The front-end will receive UInt8Array and keep that without mutating to string.

When applying changes, the front-end will send Uint8Array to back-end if the data's column is binary or the previous value in the same cell was a binary.

The data is sent as Uint8Array back and forth.

rathboma:
@azmy60 broken test?

azmy60:
@rathboma Just fixed it. Let me know what you think!

rathboma (approved):
Great stuff, ended up being a large PR

PR #2590, opened 2024-11-13
fix clickhouse materialized views query

azmy60:
Using strings to compare enum columns was not supported until ClickHouse 24.3 https://clickhouse.com/docs/en/whats-new/changelog#improvement-7

Tested this via integration tests on version 23.9 to 24.10 (latest). I wanted to test more old versions but so far I don't see any new issues other than the string to UInt8 error, which is what this PR addresses.


rathboma:
@azmy60 does listing databases work? I think that was another error he saw. Are there tests for that? If not, can you add some?

azmy60:
I believe we have that test. I'll take a look.

azmy60:
We have tests for listing databases but not switching database. It works fine when I test it manually but if we want to add the test script, we'll need to refactor the `switchDatabase`. Other than that it's good.

PR #2617, opened 2024-11-26
added function to get Mysql and Server statistics

zeeshan1212:


rathboma:
@zeeshan1212 I know this was a long time ago, but if you're willing to do a bit more work on it, would love to potentially add this functionality. It will need to be moved to `src/lib/db/clients/mysql.ts`, and an optional method adding to `BasicDatabaseClient`. There will also need to be a way to display the data, but that can be a separate PR for sure

PR #2619, opened 2024-11-27
Implement clickhouse import commands

azmy60:
This PR attempts to support importing data via Import From File in ClickHouse. 

To avoid potential issues or inefficiency that we can't predict when processing large files, like server cost, it'll pass a file directly to a ClickHouse client instead of running multiple of `INSERT INTO ... VALUES ...`. It'll also use [async_insert](https://clickhouse.com/docs/en/optimize/asynchronous-inserts) to make sure that we follow the best practice.

But, this PR only do that on CSV files. Other formats like JSON/JSONL/XLSX will use the default processing pipeline.

TODO:
- [ ] import json/jsonl/xlsx by using the default importer

### DIFF
--- a/apps/studio/src-commercial/backend/lib/db/clients/clickhouse.ts
+++ b/apps/studio/src-commercial/backend/lib/db/clients/clickhouse.ts
@@ -12,6 +12,8 @@ import {
   InsertParams,
   ResponseJSON,
   ClickHouseClient as NodeClickHouseClient,
+  ClickHouseSettings,
+  DataFormat,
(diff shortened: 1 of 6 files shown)


rathboma:
@azmy60 This PR dead?

azmy60:
@rathboma I forgot to pick this up again after working on other stuff. This should be good enough for handling csv imports so other file formats can be submitted on different PRs? Should we make new issues? Not sure I can continue it now so probably somebody else need to work on it.

PR #2628, opened 2024-12-09
Feature: Support manual commit mode for Postgres

### LINKED ISSUE #1036: FEAT: Manual commits and rollbacks
It makes me nervous every time I execute sql statements.
I wish there was a manual commit and rollback feature, like PLSQL Developer and DBeaver.
image

nykoh2001:
### Issue
- resolve #1036 
- Added toggle to change commit mode (Auto/Manual)
- Two options are available in Manual commit mode (Rollback/Commit)
  - Rollback and Commit are only enabled when manual query is running
- Added shortcut
- If error happened, executed query in manual mode is rollbacked
- If the user changes commit mode to auto before manual query run is committed, executed query in manual mode is rollbacked

### Demo
[Run in manual mode - Rollback]

https://github.com/user-attachments/assets/b0074e54-b4fa-46d2-b57c-7135eddbe8e3

[Run in manual mode - Commit]

https://github.com/user-attachments/assets/e9e9389d-1777-4beb-ab85-dc46258e2143

[Run selected query in manual mode - Commit]

https://github.com/user-attachments/assets/b3b881be-76fc-4faa-b561-f8375a2163e3

### DIFF
--- a/apps/studio/src/assets/styles/app/_layout.scss
+++ b/apps/studio/src/assets/styles/app/_layout.scss
@@ -164,6 +164,9 @@ x-buttons {
     border-radius: 0;
   }
 }
+x-buttons > x-button.btn-rounded {
+  border-radius: 6px !important;
+}
 .btn-block {
   display: flex;
   justify-content: center;
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -78,7 +78,7 @@
             <x-button
               class="btn btn-primary btn-small"
               v-tooltip="'Ctrl+Enter'"
-              @click.prevent="submitTabQuery"
+              @click.prevent="handleSubmitTabQuery"
             >
               <x-label>{{ hasSelectedText ? 'Run Selection' : 'Run' }}</x-label>
             </x-button>
@@ -88,7 +88,7 @@
             >
               <i class="material-icons">arrow_drop_down</i>
               <x-menu>
-                <x-menuitem @click.prevent="submitTabQuery">
+                <x-menuitem @click.prevent="handleSubmitTabQuery">
                   <x-label>{{ hasSelectedText ? 'Run Selection' : 'Run' }}</x-label>
                   <x-shortcut value="Control+Enter" />
                 </x-menuitem>
@@ -114,6 +114,34 @@
                   >
                     stars
                   </i>
+                </x-menuitem>
+              </x-menu>
+            </x-button>
+          </x-buttons>
+          <x-buttons class="">
+            <x-button
+              @click.prevent="toggleCommitMode"
+              class="btn btn-flat btn-small"
+              :class="{ 'btn-rounded': !isManualCommit }"
+            >
+              <x-label>{{ commitModeLabel }}</x-label>
+            </x-button>
+            <x-button v-if="isManualCommit" class="btn btn-flat btn-small" menu>
+              <i class="material-icons">arrow_drop_down</i>
+              <x-menu>
+                <x-menuitem 
+                  :disabled="!this.isManualQueryRunning"
+                  @click.prevent="manualCommit"
+                >
+                  <x-label>Commit</x-label>
+                  <x-shortcut value="Shift+Cmd+C" />
+                </x-menuitem>
+                <x-menuitem 
+                  :disabled="!this.isManualQueryRunning"
+                  @click.prevent="manualRollback"
+                >
+                  <x-label>Rollback</x-label>
+                  <x-shortcut value="Shift+Cmd+R" />
                 </x-menuitem>
               </x-menu>
             </x-button>
@@ -383,6 +411,9 @@
          */
         focusElement: 'none',
         focusingElement: 'none',
+
+        isManualCommit: false,
+        isManualQueryRunning: false,
       }
     },
     computed: {
@@ -554,8 +585,8 @@
         const keybindings: any = {
           "Shift-Ctrl-Enter": this.submitCurrentQuery,
           "Shift-Cmd-Enter": this.submitCurrentQuery,
-          "Ctrl-Enter": this.submitTabQuery,
-          "Cmd-Enter": this.submitTabQuery,
+          "Ctrl-Enter": this.handleSubmitTabQuery,
+          "Cmd-Enter": this.handleSubmitTabQuery,
           "Ctrl-S": this.triggerSave,
           "Cmd-S": this.triggerSave,
           "F5": this.submitTabQuery,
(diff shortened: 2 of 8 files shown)


nykoh2001:
Hi @rathboma, I'm looking forward to applying this change to the release 👀 Can anyone take a look at this?

rathboma:
Hey! I had a look at the feature. Super cool! I like the idea of making this available.

Couple of things we'll need to resolve:
1. When someone has enabled manual-commit mode, and run a query we need a way to signal this to the user. Like maybe the button bar of the query editor can change color, sort of how we do it for applying changes in the table tab?
2. I think there needs to be a `Start manual commit mode` button, and when 'manual commit mode' is active, the state of the tab changes to indicate it's in this mode somehow.

@azmy60 any ideas on how to do these visual touches?

Also -- we'd need to properly disable this for other clients so it doesn't show in the UI

azmy60:
@rathboma I have two ideas:

1. We can add an `M` icon to the tab icon, if we hover it, it will say "You're in manual commit mode". But this one is too subtle I think.
2. I'm not sure if this is the good, but having to know that there are Commit/Rollback buttons appears in the UI is enough probably to know that we are in manual transaction mode. Consider this UI:
image
We know that auto mode does not have the ability to commit and rollback, so we can hide those in UI by default. When the manual mode is enabled, we should see these buttons in the UI. I feel like just by seeing these two buttons exist or not, we should know which mode we are in? Plus, if we implement the above design, we can reduce the amount of clicks to commit/rollback.

nykoh2001:
@rathboma Actually I already implemented button that describes whether it's manual commit mode or actual commit mode, is it not enough for the signal? It's the toggle that changes the mode and also shows current commit mode. (cc @azmy60)
image
- Auto commit mode
<br/>

image
- Manual commit mode

nykoh2001:
Attached the video that shows the commit mode toggle

https://github.com/user-attachments/assets/8bc5644b-38b8-44a1-8479-4bee820e3755

rathboma:
So looks like this keeps the connection open if the commit mode is manual -- forgive the stupid question -- but how do you make sure you get the same connection for subsequent queries, or does any query through the pool use the open connection?

Also -- what I mean by 'signal to the user' -- if they switch tabs, the manual commit **is still active**, no? Anything else they do in the app will be inside of that transaction until they return to the original tab and manually commit.

nykoh2001:
@rathboma Great point—thanks for catching that! I haven’t had the time to address it yet, but I’ll come back to it and dive in soon. 🙏

rathboma:
@nykoh2001 Would still love to have this feature if you get more time to work on it!

nykoh2001:
@rathboma Thanks so much! I’m definitely happy to work on it :D
> what I mean by 'signal to the user'

We can think about 1) changing the tab background color — I'll be glad if you have any color preference for this — or 2) changing the text color theme — e.g. black-hot pink(highlighted) to black-blue(highlighted). 
Otherwise we might want to show a message in somewhere that lets the users know currently it's in manual mode — e.g. You are in manual commit mode now, please make sure to commit or rollback to complete this transaction.

Let me know if any of these sound good to you, or if you have other ideas in mind — I’m totally open to new suggestions :)

not-night-but:
@nykoh2001 We definitely need good messaging here, but I'm not totally sure how we should go about it. Maybe we override the statusbar theme? Or change the theme of the buttons at least to like a danger red?

azmy60:
What about a new button beside `download` on status bar? The button can switch between manual transaction mode and auto transaction mode.

If it's manual, show the "commit" and "rollback" buttons in the query editor.

If it's auto, don't show the buttons. Keep the UI simple.

The idea is you should know if it's manual if "commit" and "rollback" buttons appear on screen. Also, the statusbar button tells us what mode is active. Similar to my screenshot above.

Question - why do we hide the commit and rollback actions inside the drop-down menu if we can just show them as individual buttons? It makes the UI look simpler, but increase the amount of clicks. (suppose you want to commit often)

rathboma:
Before continuing, someone needs to have reasonable answers to the following questions:

1. A way to isolate a connection to a specific tab when manual commit mode is started -- like maybe the tab has to pass through some sort of ID to identify the connection it's using
2. The manual transaction connection cannot be accidentally shared with other tabs - otherwise they'd see different things depending on which connection they used
3. A way to visualize the manual commit mode in the query editor. Some color change or something. Just get this working and we can figure out what it looks like afterwards
4. Answer to the question: When enabled does the user have to run `BEGIN`, or is that part automatic when they run their first query? Is this just a state the happens when the user runs `BEGIN`, or are they clicking a button to BEGIN?
5. Does it even need a toggle? Like if someone runs `BEGIN`, shouldn't that tab now be in transaction mode?

azmy60:
Yeah, we could do the design afterward. We should do more research for number 5 though, like how this is done by other apps. I think it's best to practice design decision with common patterns. I mean my solution is not the common pattern, so idk lol. (Maybe follow common patterns if it still looks good for our app)

not-night-but:
Okay so:

1. This should be relatively simple. I think we can just pass the tabId along with the queries, and have a map of ids to connections in the driver. One issue is we are probably going to have to limit the amount of active manual commit mode tabs there, otherwise the user could allocate the whole pool and then the rest of the app would cease to function.
2. I think this would be solved with the above
3. I do think that the existing button that @nykoh2001 put in may work for this? it says what mode you are in right next to the run button, but there are other things we can do. As long as we are confining this feature to the query editor I think it should be pretty easy.
4. Personally I think we could do both? I like having the UI button to toggle manual commit (especially if we have a keybind for it), but if the user begins a transaction and doesn't commit, that should be pretty easy to detect.
5. I think my opinion on this is clear above. I don't see the harm in having both. Some people like controlling things with scripts, others like having the UI to do it for them. And I don't really see this as being extra clutter.

nykoh2001:
>Answer to the question: When enabled does the user have to run BEGIN, or is that part automatic when they run their first query? Is this just a state the happens when the user runs BEGIN, or are they clicking a button to BEGIN?

@rathboma 
In the current implementation, the `BEGIN` clause is automatically run when the user executes their first query in manual commit mode.

@not-night-but 
I agree that we could support both cases:
	1.	Automatically run BEGIN when the user is in manual commit mode.
	2.	Detect a manual BEGIN statement and enable manual commit mode based on that.

But I might want to **clarify the actual blocker of this PR and future enhancement** to keep things focused—especially since this PR has already stale. So many things are already going around here, and things could be getting bigger and bigger. I suggest we move forward **only with option 1 within this PR**, which is already implemented and we can then create a **follow-up issue for option 2 as a future enhancement**.

>Does it even need a toggle? Like if someone runs BEGIN, shouldn't that tab now be in transaction mode?

I also think we should **keep the toggle for enabling or exiting manual transaction mode**, as it’s the most straightforward and simple way to support manual commit functionality.

@rathboma 
As I think, connection isolation is the critical blocker of this feature. Allowing multiple manual connections at the same time could be too complicated and can cause unexpected side effects that could be critical to whole app — seems like it could be too risky. My recommendation is to **disable tab switching when the user is in manual commit mode**, and if the user hovers on the tab that are disabled, we can show hover tooltip that says 
> _"Tab switching is not allowed in manual commit mode"_

What do you think? Happy to hear you guys' thoughts :) 

Summary:
- Want to keep the toggle for enabling/exiting manual commit mode
- The actual blocker of this PR is connection isolations and we can create separate follow-up issues for non-blockers, future enhancements
- To prevent potential issues, suggest disabling tab switching when it is manual commit mode for now
- The very next action items would be;
  - Manage multiple transactions in a more clear way (depends on my suggestion and further discussion)
  - Keep discussing for visual alerts — if this could be done before I starting further implementation, I will be able to merge this PR much faster

nykoh2001:
@rathboma Would love to hear your thoughts!

PR #2631, opened 2024-12-10
fix postgres connection forces ssl mode

### LINKED ISSUE #2630: BUG: SSL error message in postgresql localhost after 5.0.5 update
**Describe the bug**

Beekeeper was just updated to version 5.0.5 (with the debian apt package). Before that, my connection to localhost worked perfectly. But after the release, I am always getting the following error when I am trying to connect to my local postgres instance: `There was a problem. The server does not support SSL connections`. I am not using SSL or SSH (as it is a localhost connection), as seen in the screenshot below.

I am certain it is an issue in beekeeper, because I can connect to the database through other means.

**To Reproduce**

Create a connection to a local postgres database, and then try to connect.

**Expected behavior**

The connection works as expected, and does not try to use SSL unless I want to.

**Screenshot**

beekeeper-screenshot

 - OS: Linux (debian-based)
 - App Version: 5.0.5
 - Database type and version: Postgresql

**Additional context**

Remote connections seem to be working fine. Localhost connection was also fine before the update.

azmy60:
fix #2630 

SSL is enforced in Redshift connections, but that logic is inside the `PostgresClient` class. It checks if `redshiftOptions` is "truthy", but the default value is an object and coerced to `true` which forces us to always use SSL, even though we don't enable it.

Moving that logic to `RedshiftClient` should fix the issue, but I saw other redshift codes in the class so I guess why not move them too. 🤷

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -18,7 +18,6 @@ import {
   buildSelectQueriesFromUpdates,
   buildUpdateQueries,
   escapeString,
-  refreshTokenIfNeeded
 } from './utils';
 import { createCancelablePromise, joinFilters } from '../../../common/utils';
 import { errors } from '../../errors';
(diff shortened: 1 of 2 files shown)


rathboma:
@azmy60 I fixed this in another simpler PR

PR #2641, opened 2024-12-12
Feature/2379 tab history

### LINKED ISSUE #2379: FEAT: Ctrl+P Quicksearch defaults to the last tabs opened in reverse order
This is the same as VSCode.

Why?
- When navigating between tabs, this allows back/forward behavior that is different to next-tab/previous-tab. I use this all the time to switch to the last tab I had open. This would be useful for BKS also

image

wmontgomery:
resolves #2379 
tabs-history


sharandeepak1211:
Shall we have a key binding to move between tables like "CMD + ]" for moving forward and "CMD + [" for moving backward ?
Because this is a universal shortcut in MAC for these purpose. It will be super handy to navigate : )

@rathboma @wmontgomery

apps/studio/src/common/appdb/models/TabHistory.ts
```diff
@@ -0,0 +1,162 @@
...
+import { ApplicationEntity } from "./application_entity";
+import _ from 'lodash'
+import { TransportTabHistory } from "@/common/transport/TransportTabHistory";
+import { TransportOpenTab } from "@/common/transport/TransportOpenTab";
+import { OpenTab } from "./OpenTab";
+
+type TabType = 'query' | 'table' | 'table-properties' | 'settings' | 'table-builder' | 'backup' | 'import-export-database' | 'restore' | 'import-table';
+
+@Entity({ name: 'tab_history'})
+export class TabHistory extends ApplicationEntity {
```
rathboma, line 11: Hey, I think the changes here look pretty good, and it's a slick experience using this.

Only question I have: Do we really need a new model to handle this, or can we just add a few more details to the `OpenTab` model, or even do it without changing OpenTab much at all?

Is this a stupid question, or could we do this by:

1. Adding `lastActiveAt` to OpenTab (time the user last clicked it)
2. Add `deletedAt` to OpenTab (time the user closed it). Then we set this instead of deleting the opentab
3. Adding an index to [deletedAt, workspaceId, connectionId]
4. Almost everything can then be done in the front-end. we just need a way to load recentlyClosed tabs at app start, then they can just be added to the recentlyClosedTabs stack when someone closes them.
5. On start, delete any `OpenTab` models with `deletedAt` < 3 days ago

Would this save a lot of duplication? Am I being naive and missing something big?

rathboma (changes requested):
This looks awesome! So excited to get this into Beekeeper Studio.

Big question is around -- can we do this without duplicating tab data to another model?

Also -- can you implement a `Reopen Closed Tab` function in the right click menu for tabs using this data?

apps/studio/src/common/appdb/models/TabHistory.ts
wmontgomery: I wanted to keep Open Tabs exactly what they are: Tabs that are open and active. I didn't want to add any logic to the open tabs model for either maintaining a certain history (open and closed to be reopened). Having the additional model does have some duplicated pieces, no doubt there, but it's largely for historic reasons and keeping the open tabs model pure.

I had hoped to keep as much out of the front end as possible to not muddy the store up as well.

apps/studio/src/store/modules/TabModule.ts
```diff
@@ -63,22 +63,28 @@ export const TabModule: Module<State, RootState> = {
...
+        let tabs = await Vue.prototype.$util.send('appdb/tabs/find', {
           options: {
             where: {
               connectionId: usedConfig.id,
               workspaceId: context.rootState.workspaceId
             }
           }
-        });
+        })
+        tabs = tabs.filter(tab => !tab.deletedAt)
```
azmy60, line 87: I know we can't really pass `deletedAt` to options from here so this is good for now. In the future, we might wanna filter deleted rows by default like using [@deleteDateColumn](https://typeorm.io/select-query-builder#querying-deleted-rows) if that doesn't break other parts of the app. Or something like `appdb/tabs/findExceptDeleted`.

azmy60 (changes requested):
This is awesome! One little issue when reopening the closed tab, the opened tab is not focused and I can't close it by pressing `Ctrl+w`.

history-tab

apps/studio/src/store/modules/TabModule.ts
wmontgomery: oh good catch. I didn't even know about this.

azmy60 (approved):

apps/studio/src/migration/20251116_tab_history_updates.js
```diff
@@ -0,0 +1,12 @@
+export default {
+  name: '20251116-tab-history-updates',
+  async run(runner) {
+    const queries = [
+      `ALTER TABLE tabs ADD COLUMN lastActive datetime`,
```
rathboma, line 5: We'll need to add a reverse index to `lastActive`, seeing as you're using it in a DESC search.

rathboma: Probably also need an index on the deletedAt column

rathboma (changes requested):
let's add a couple of indexes to make sure OpenTab queries remain fast, then I'm good to merge!

apps/studio/src/migration/20251116_tab_history_updates.js
wmontgomery: sure thing

apps/studio/src/common/appdb/models/OpenTab.ts
```diff
@@ -156,5 +166,57 @@ export class OpenTab extends ApplicationEntity {
...
+          },
+          order: {
+            deletedAt: 'DESC'
+          },
+          withDeleted: true
+        })
+  }
 }
 
+export const TabHistoryHandlers = {
```
rathboma, line 212: @not-night-but is this in the right place? Where should will put this object?

wmontgomery: I've seen them in their own file before. I want to say I've seen them in the same file as the model but I could have also done that when I had the tab history model set up. I can move them if it's the wrong spot. Was going to keep them close in file structure to each other since they expand nowhere else.

not-night-but: @wmontgomery Yeah these should be in the handlers directory. Ideally in their own file called `tabHistoryHandlers.ts` in `src/handlers/`

rathboma (approved):

PR #2659, opened 2024-12-17
Fix JSON column display in JSON row viewer

azmy60:


azmy60:
What could be the best place to fix this rather than this PR is in the `apps/studio/src/lib/data/tools.ts` file, and the culprit is the json mutator. Looks like it's not only that tabulator mutates its own data, but it also mutates our data in `TableTable`.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1698,9 +1698,20 @@ export default Vue.extend({
       if (selectedRowIndex === this.selectedRowIndex) return
 
       const position = (this.limit * (this.page - 1)) + (row.getPosition() || 0)
+      const cleanData = { ...this.$bks.cleanData(data, this.tableColumns) }
+
+      // un-stringify json fields that have been stringified by tabulator mutator
+      Object.keys(cleanData).forEach((key) => {
```
rathboma, line 1704: @azmy60 you should use the LanguageData#getLanguageByContent to figure it out if it's a text field that's actually JSON.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1698,9 +1698,20 @@ export default Vue.extend({
...
+        const dataType = this.table.columns.find((c) => c.columnName === key)?.dataType
+        // see apps/studio/src/lib/data/tools.ts#L25
+        if(dataType?.startsWith('json') && typeof cleanData[key] === 'string') {
+          cleanData[key] = JSON.parse(cleanData[key])
```
rathboma, line 1708: What if JSON parsing fails? What's the fallback? Needs a default here

rathboma (changes requested):
When we interpret column data as JSON, we should be consistent and use the LanguageData

apps/studio/src/components/tableview/TableTable.vue
azmy60: You're right.

apps/studio/src/components/sidebar/DetailViewSidebar.vue
```diff
@@ -157,18 +158,31 @@ export default Vue.extend({
...
+          _.set(clonedValue, path, (value as string).slice(0, globals.maxDetailViewTextLength))
         }
       })
       return clonedValue
     },
+    jsonFields() {
+      const fields = []
+      Object.keys(this.value).forEach((key) => {
+        const val = this.value[key]
+        if (typeof val === 'string' && getLanguageByContent(val).name === "json") {
```
rathboma, line 174: Is val ever already an object or array? If so `getLanguageByContent` would fail, as it expects a string.

Can we test this somehow?

apps/studio/src/components/sidebar/DetailViewSidebar.vue
```diff
@@ -157,18 +158,31 @@ export default Vue.extend({
     },
     processedValue() {
       const clonedValue = _.cloneDeep(this.value)
-      eachPaths(clonedValue, (path, value: string) => {
-        if (this.truncatedPaths.includes(path)) {
-          _.set(clonedValue, path, value.slice(0, globals.maxDetailViewTextLength))
+      eachPaths(clonedValue, (path, value) => {
+        if (this.jsonFields.includes(path)) {
+          _.set(clonedValue, path, JSON.parse(value as string))
```
rathboma, line 163: is value always a string? Can it ever be an object / array at this point?

rathboma: What if this fails?

rathboma (changes requested):
Questions mostly.

apps/studio/src/components/sidebar/DetailViewSidebar.vue
azmy60: `val` can be a number or an ArrayBuffer, and it seems like `getLanguageByContent` only accepts strings.

Like a unit test for `getLanguageByContent`?

apps/studio/src/components/sidebar/DetailViewSidebar.vue
azmy60: It shouldn't fail since all values in `jsonFields` are filtered with `getLanguageByContent` which uses `JSON.parse` to check whether it is a valid json string or not. But I agree that even though that's true, this still looks vulnerable.

apps/studio/src/components/sidebar/DetailViewSidebar.vue
rathboma: Why aren't these values already objects when they're in the front-end? I'm super confused why we'd need to inflate them from strings.

rathboma: Ok, the problem isn't the front-end, it's the backend. It should be sending these back as objects.

rathboma: See for example `_text` fields are sent back correctly as arrays

rathboma (commented):
The real change we need to make is on the driver side

rathboma:
I'm going to merge this, but we should figure out why JSONB data is being returned as a string.

azmy60:
@rathboma They are indeed objects from utility, and we receive the objects in the front end too. Until it goes through tabulator, it gets mutated by tabulator mutator, and some how tabulator mutator mutates our data.

azmy60:
This is the folllow up PR https://github.com/beekeeper-studio/beekeeper-studio/pull/2672

PR #2664, opened 2024-12-18
filter used connections by workspace id

not-night-but:


rathboma:
@not-night-but is this done or still draft?

not-night-but:
@rathboma I just haven't tested it yet, if it works, it's ready to merge

rathboma:
@not-night-but we should only be pulling used configs back from the Utility process for the correct workspace id.

FYI - can you check we're also doing this correctly for used queries?

not-night-but:
@rathboma We are now properly loading used connections when the workspaceId changes. There was also an issue with workspaces as used connections are not saved to the cloud, so there was no cloud query module, I just made it so they both are the util query module.

Used queries also work as expected

rathboma:
Great! I'll take a look tomorrow

rathboma (approved):

PR #2671, opened 2024-12-19
Offline License Checking

rathboma:
Business users can download an offline license file.

This is a JSON file that contains license information + a signature generated with a private key.
Beekeeper Studio uses the public key to validate the license information in the file, and enables paid features.


apps/studio/src/backend/lib/OfflineLicense.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+ */
+
+
+export interface LicenseOptions {
+  licensePath?: string
+  keyPath?: string
+}
+
+export class OfflineLicense {
```
azmy60, line 42: Should this be in the commercial folder?

azmy60 (approved):
Just have a picky comment. Otherwise, LGTM!

apps/studio/src/backend/lib/OfflineLicense.ts
rathboma: Good question, but I think it's fine -- you can only export business licenses anyway, so no way they can use it without a business license.

PR #2672, opened 2024-12-20
Prevent direct object-string mutation in tabulator mutator

azmy60:
This PR fixes json column values that gets stringified and displayed as strings instead of objects in JSON row viewer.

### DIFF
--- a/apps/studio/src/common/utils.ts
+++ b/apps/studio/src/common/utils.ts
@@ -292,3 +292,37 @@ export function friendlyUint8Array(bytes: string | Uint8Array, encoding?: 'hex')
   })
   return bytes as FriendlyUint8Array
 }
+
+/** Make `object.toString` look better :D */
+export function friendlyJsonObject<T extends object>(obj: T): T {
+  Object.defineProperties(obj, {
+    [Symbol.toPrimitive]() {
+      try {
+        return stringifyWithBigInt(obj);
+      } catch (ex) {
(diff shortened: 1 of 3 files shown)


rathboma:
I don't think we even need to mutate do we? Or do we because we're using the text cell and built in editor?

azmy60:
I think we do as we're using built-in cell editor for some types.

azmy60:
I feel like accessors and formatters should be enough for our case. NullableCellEditor should be able to replace the built-in editor, right?

rathboma:
You can test it, but wasn't really designed for big long values. 

Maybe we should just add an issue to fix this later. Everything's kind of working fine at the moment so I'm not sure if we need to do it right now

azmy60:
Yeah, for sure!

rathboma (approved):
I'm ok with this for now. Please add an issue to fix this properly in the future (no mutation).

Probably we should built a set of basic native tabulator editors without Vue that can handle native types.

Merge when ready

azmy60:
Yes, that makes a lot of sense. As to me, it looks like mutators were used because Tabulator would show `[Object object]` instead of the real value, so we use it so the data can be viewed and copied in JSON friendly format.

Another place to handle this is formatters and use the same formatters' logic in the editor. Ideally, I think the final result would be an extension module for tabulator. It won't be as simple as mutators, but doable.

PR #2673, opened 2024-12-20
Implement table length function

anilsenay:
`getTableLength` function implemented for Cassandra

### DIFF
--- a/apps/studio/src-commercial/backend/lib/db/clients/cassandra.ts
+++ b/apps/studio/src-commercial/backend/lib/db/clients/cassandra.ts
@@ -98,6 +98,7 @@ export class CassandraClient extends BasicDatabaseClient<CassandraResult> {
       backDirFormat: false,
       restore: false,
       indexNullsNotDistinct: false,
+      transactions: false,
     }
   }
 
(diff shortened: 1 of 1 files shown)


apps/studio/src-commercial/backend/lib/db/clients/cassandra.ts
```diff
@@ -462,8 +465,7 @@ export class CassandraClient extends BasicDatabaseClient<CassandraResult> {
   async selectTopStream(table: string, orderBy: OrderBy[], filters: string | TableFilter[], chunkSize: number, _schema?: string): Promise<StreamResults> {
     const qs = this.buildSelectTopQuery(table, null, null, orderBy, filters)
     const columns = await this.listTableColumns(table)
-    // const rowCount = await this.getTableLength(table, filters)
-    const rowCount = 0; // we don't have a table length implemented yet
+    const rowCount = await this.getTableLength(table)
```
rathboma, line 468: Hey! One of the reasons we didn't implement it (although we should have added this to the comments) is that the count(*) operation can be very expensive for Cassandra -- as it has to scan all the table rows. Is there a more efficient way to get this information?

rathboma (changes requested):
Is this expensive?

apps/studio/src-commercial/backend/lib/db/clients/cassandra.ts
anilsenay: @rathboma Hey, thank you for the information; I didn't know that.

As you said, it was expensive based on this post: https://www.datastax.com/blog/running-count-expensive-cassandra. They suggested some ways in post but they are not suitable.

Since there is no better way for now, i guess we can close the pull-request.

rathboma: Kk

PR #2706, opened 2025-01-16
verify sqlite database is valid when connecting

### LINKED ISSUE #2694: BUG: Testing an SQLite connection always succeeds
**Describe the bug**
No matter the kind of file selected, even if there's no file selected, testing the connection always succeeds.

**To Reproduce**
1. Open up the "New connection" dialog, and select the SQLite type.
2. Click "Test".
3. Observe that the "Connection looks good"
4. Select a file that's not a database.
5. Test, the connection's still good.

**Expected behavior**
I'd expect that if it isn't an empty file or actual database, for the connection to be invalid.

**Screenshots / Gifs**

https://github.com/user-attachments/assets/2fc56552-53de-4018-8b06-dc2b2f36013d

 - OS [e.g. Windows/Mac/Ubuntu]: NixOS 25.05.20250110.130595e (Warbler)
 - App Version [e.g. help -> About Beekeeper Studio]: 5.0.9
 - Database type and version [eg Posgresql 9.3]: SQLite3

**Additional context**
This toes the line of being intended behaviour (I'll look at the source later to confirm), after all, SQLite databases are just files, and as long as you can "see" the file, you can technically open the database. I do think that when a .png is selected, an error should come up at least...maybe it should try and parse the tables at the very minimum.

And maybe if there's no file selected, the "Test" button could be greyed out, and the "Connect" button could become a "Create" button, which'd be clearer to users about the intent.

(also password protected SQLite dbs are [apparently a thing](https://stackoverflow.com/a/24349415/21733670), which aren't supported by Beekeeper right now, but if they're ver going to be, then it makes sense for this to be laid as plumbing)

MasterOdin:
Fixes #2694 

PR makes it so that when trying to connect to an invalid sqlite database file, an error will be thrown that states that "the file is not a valid database". I utilize [`PRAGMA schema_version`](https://www.sqlite.org/pragma.html#pragma_schema_version) to accomplish this, where it should return a number to indicate the version (`0` for an empty file, otherwise some number > 0), or throw an exception if the schema version header wasn't found (due to it not being a valid sqlite file).


rathboma (approved):
Awesome, thanks!

PR #2724, opened 2025-01-29
Fix connecting to a new file-based database

### LINKED ISSUE #2703: BUG: Connect to SQLite database when created
**Describe the bug**
When connected to sqlite database, unable to connect to a new database that is created during the connection of the first database.

**To Reproduce**
1. Open SQLite database
2. Click on the "Add database" button 
3. Error comes up stating that the database cannot be connected but it is created 
4. Database is shown in the dropdown menu
5. When changing the dropdown menu to original or existing database, the option for the new database is gone.
6. When checking in the directory of the directory, the new/existing database does exist.

**Expected behavior**
Add database and it should switch to an empty database if the database does not exist or an existing database if it does.

**Screenshots / Gifs**
Visual evidence of the steps:

https://github.com/user-attachments/assets/d81684e6-21f0-4ff0-81d0-6469644cee32

Screenshot of the database that existed/was created:

Screenshot 2025-01-14 at 3 37 46 PM

 - OS: Mac
 - App Version: 3.41.2
 - Database type and version: SQLite 3.41.2

azmy60:
After creating a new sqlite db from the create modal, it tries to open the new db but it throws an error because the file path is wrong. This PR would fix that and a similar error in firebird.

fix #2703

TODO:
- [ ] For file-based dbs (SQLite, Firebird, and DuckDB), it should open the new database in the same window instead of creating a new window. This will make it easier for Firebird since it needs authentication info.
- [ ] Should work with SQLite
- [ ] Should work with Firebird
- [ ] Should work with DuckDB


wmontgomery (approved):

apps/studio/src/components/sidebar/core/DatabaseDropdown.vue
```diff
@@ -110,14 +108,16 @@
...
-          fileLocation.pop()
-          const url = this.connectionType === 'sqlite' ? `${fileLocation.join('/')}/${db}.db` : `${fileLocation.join('/')}/${db}`
+        if (this.dialectData.disabledFeatures?.multipleDatabase) {
+          const c = this.usedConfig
+          // We'll need the auth credentials for databases like firebird
+          const username = c.username ?? ''
+          const password = c.password ? `:${c.password}` : ''
+          const host = c.host ? `@${c.host}` : ''
+          const port = c.port ? `:${c.port}` : ''
+          const url = this.connectionType === 'sqlite' ? db : `${this.connectionType}://${username}${password}${host}${port}/${db}`;
```
rathboma, line 120: Woooah, we absolutely shouldn't be doing this in the UI -- switching off connectionType is bad, but this connection url creation is worse -- what if we have ssh enabled for the connection? What if a DB driver doesn't follow the standard conventions? This doesn't work for oracle for example.

I'm ok with the SQLite fix, but nee you to make sure we fix this new window function too.

Why doesn't it send the connection ID plus the database name? Something as simple as this will work: `bks://connect?connectionId=1&database=foo`

rathboma (changes requested):
Everything is fine except for the URL creation in the App UI

apps/studio/src/components/sidebar/core/DatabaseDropdown.vue
azmy60: `connectionId` will be `null` if we connect to a db without saving the connection. But I agree that this is the worst way to do this lol.

rathboma (approved):

rathboma:
@azmy60 fix conflicts and merge

PR #2728, opened 2025-01-30
fix positional params

### LINKED ISSUE #2725: BUG: Positional parameters are broken
**Describe the bug**
It seems like in some cases, positional parameters are not being properly parsed from the query (this would be a bug in the `sql-query-identifier` repo).

**To Reproduce**
1. Connect to a MySQL/MariaDB db
2. try to run `SELECT * FROM titles where emp_no = ?`
3. Notice that there is an error and no param popup

**Expected behavior**
The param modal should popup (ie, the param should be parsed by `sql-query-identifier`)

**Screenshots / Gifs**

Image

not-night-but:
Fix #2725 

I don't know how long these have been broken, but if there was a positional param as the first param, none of them were replaced. Beyond that, we weren't actually allowing them to be replaced properly. This is a tentative fix, as I'm not sure about the UI for it, but I solved it by using indexes of the positional params to store them, and since they're stored as numbers, we use a different regex to replace the instances (ie for each param, we just replace the first instance of the param, then the next, etc)


azmy60 (approved):

not-night-but:
Not sure I want to merge these changes. sql-formatter supports replacing parameters, so I think it might be a better idea to get rid of our regex, as I'm almost positive that this has some issues. sql-identifier doesn't identify parameter strings if they happen to be inside strings, but our regex doesn't take that into account, which means if a matching parameter string is inside a string (in the case of positional parameters, `?` surrounded by whitespace), that will also get replaced. sql-formatter actually parses the SQL being provided and won't have this issue. I'm sure this probably also happens with other param types, we just haven't run into it yet.

rathboma:
didn't we implement this in sql-query-identifier also? We should use either that, or the formatter lib for sure. I thought we were using query identifier for params?

rathboma:
@not-night-but what did you decide about this PR?

not-night-but:
@rathboma I'm working on a refactor, will flip this back to draft

apps/studio/src/lib/db/sql_tools.ts
```diff
@@ -11,6 +15,64 @@ export function splitQueries(queryText: string, dialect) {
...
+export function defaultParamTypesFor(dialect: string): ParamTypes {
+  switch (dialect) {
+    case 'psql':
+      return {
+        numbered: ['$'],
+      };
+    case 'mssql':
+      return {
+        named: [':'],
+      };
```
SemProvoost, line 27: Could I suggest adding the `named: [':']` for mysql as well here? 😁
```suggestion
      };
    case 'mysql':
      return {
        positional: true,
        named: [':'],
      };
```

apps/studio/src/lib/db/sql_tools.ts
```diff
@@ -11,6 +15,64 @@ export function splitQueries(queryText: string, dialect) {
   return result
```
SemProvoost, line 15: The identify above this line probably needs the new paramTypes as well here:
```diff
-  const result = identify(queryText, { strict: false, dialect })
+  const paramTypes = defaultParamTypesFor(dialect);
+  const result = identify(queryText, { strict: false, dialect, paramTypes })
```

SemProvoost (commented):
I'd like to bump this PR by adding a suggestion of `:named` params for mysql as well. I tested the functionality locally, but the `splitQueries` needed the use of `defaultParamTypesFor` as well to make it work.

not-night-but:
@SemProvoost what I'd like to do is actually have these as the defaults, and then allow users to override what param types they want through the Ini config system. Does that sound good to you?

SemProvoost:
@not-night-but For sure! Config file works as well.

rathboma:
@not-night-but can this be brought back to life? Also see #3162 I just created

not-night-but:
@rathboma yeah this can definitely be pushed forward. Should we enable everything by default though? I had thought we were going to use dialect defaults but allow overriding

rathboma:
Up to you, exercise best judgement

SemProvoost:
Hey, any chance this could get merged? Would love to use this feature. 😃

azmy60 (changes requested):
We might just need to fix the tests from failing. Otherwise, LGTM.

PR #2729, opened 2025-02-04
MongoDB V1

not-night-but:
Absolute bare minimum implementation of MongoDB for the app. You can browse the DB, but no edits/querying is possible.

image
image


not-night-but:
I have now also added an ObjectId transcoder: 
image

rathboma (approved):

PR #2744, opened 2025-02-07
Move license checking back to the frontend

not-night-but:
For some firewalled systems, the node networking defaults don't work properly, so license checking is broken when done from the backend. This doesn't happen with the chromium frontend, so we're moving license checking back to the frontend.

### DIFF
--- a/apps/studio/src/store/modules/LicenseModule.ts
+++ b/apps/studio/src/store/modules/LicenseModule.ts
@@ -106,7 +106,6 @@ export const LicenseModule: Module<State, RootState>  = {
       } else {
         const result = await CloudClient.getLicense(window.platformInfo.cloudUrl, email, key);
         // if we got here, license is good.
-        await Vue.prototype.$util.send('license/wipe');
         let license = {} as TransportLicenseKey;
         license.key = key;
         license.email = email;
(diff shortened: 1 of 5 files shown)


apps/studio/src/store/modules/LicenseModule.ts
```diff
@@ -103,14 +104,45 @@ export const LicenseModule: Module<State, RootState>  = {
         await Vue.prototype.$util.send('license/createTrialLicense')
         await Vue.prototype.$noty.info("Your 14 day free trial has started, enjoy!")
       } else {
-        await Vue.prototype.$util.send('license/add', { email, key })
+        const result = await CloudClient.getLicense(window.platformInfo.cloudUrl, email, key);
+        // if we got here, license is good.
+        await Vue.prototype.$util.send('license/wipe');
```
rathboma, line 109: Let's not wipe here. We can have multiple licenses saved in the DB

apps/studio/src/store/modules/LicenseModule.ts
```diff
@@ -103,14 +104,45 @@ export const LicenseModule: Module<State, RootState>  = {
...
+        license.licenseType = result.licenseType;
+        await Vue.prototype.$util.send('appdb/license/save', { obj: license });
       }
       // allow emitting expired license events next time
       SmartLocalStorage.setBool('expiredLicenseEventsEmitted', false)
       await context.dispatch('sync')
     },
-    async update() {
-      await Vue.prototype.$util.send('license/update')
+    async update(_context, license: TransportLicenseKey) {
```
rathboma, line 123: Once we've updated or a license we also need to call `sync`

apps/studio/src/store/modules/LicenseModule.ts
```diff
@@ -103,14 +104,45 @@ export const LicenseModule: Module<State, RootState>  = {
...
+      try {
+        const data = await CloudClient.getLicense(window.platformInfo.cloudUrl, license.email, license.key)
+        license.validUntil = new Date(data.validUntil)
+        license.supportUntil = new Date(data.supportUntil)
+        license.maxAllowedAppRelease = data.maxAllowedAppRelease
+        license = await Vue.prototype.$util.send('appdb/license/save', { obj: license });
+      } catch (error) {
+        if (error instanceof CloudError) {
```
rathboma, line 131: Do we actually always get a CloudError here? I don't think we throw a cloud error on a 404. You probably need to check for a 404 specifically.

rathboma (changes requested):
Couple of missing pieces, and a broken piece of logic

PR #2748, opened 2025-02-09
Fix showing postgres table with leftmost json column

### LINKED ISSUE #2734: Loading Error with Tables Having JSON Columns with NULL Values in Beekeeper Studio
### Description
When using Beekeeper Studio with PostgreSQL 14 (running on Docker), tables that contain JSON columns with all NULL values as the leftmost columns display an error message upon first load, followed by a false notification of 'no data'. Sorting the table on a non-JSON column resolves the issue and loads the data correctly.

### Steps to Reproduce
1. Use Beekeeper Studio with PostgreSQL 14.
2. Create a table with the leftmost columns being JSON columns containing only NULL values.
3. Load the table in Beekeeper Studio.
4. Note the error message and false 'no data' notification.
5. Sort the table by a non-JSON column and observe the data load correctly.

### Expected Behavior
Tables should load correctly without error messages or false 'no data' notifications, regardless of the column data types and values.

### Additional Notes
- The refresh button does not resolve the issue; manually sorting on a non-JSON column is required.
- A `CREATE TABLE` script is suggested to replicate the issue in a test suite.
***

MasterOdin:
Fixes #2734 

PR fixes an issue with postgres, where when opening a table where the leftmost column is json or jsonb, it would throw an error. This was due to it trying to sort the table on the json/jsonb column via SQL, which postgres expressly does not support.

To reproduce the bug, can use the following SQL:

```sql
CREATE TABLE bob
( a json
 , b jsonb
 , c text
 );
 
 INSERT INTO bob (a, b, c) VALUES (null, null, null);
 INSERT INTO bob (a,b,c) VALUES ('{"a": true}', '{"b": true}', 'hello');
```

My fix was to have the frontend pass the datatype for the `orderBy` columns, which then allows for filtering on the backend based on unsupported types.

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -1570,7 +1570,10 @@ export default Vue.extend({
       let filters = this.filters
 
       if (params.sort) {
-        orderBy = params.sort
+        orderBy = params.sort.map((obj) => ({
+          ...obj,
+          dataType: this.table.columns.find((col) => col.columnName === obj.field)?.dataType
+        }));
       }
 
       if (params.size) {
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -1201,13 +1201,20 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult> {
     let params: (string | string[])[] = []
 
     if (orderBy && orderBy.length > 0) {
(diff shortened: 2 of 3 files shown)


MasterOdin:
@rathboma Before I added tests for this, I wanted to make sure you were alright with the approach. I was also going to look at the other ways that the `selectTop` function is triggered to ensure they also properly pass down the `dataType`.

MasterOdin:
Closing this as will not be able to finish it up, someone else will need to pick it up.

rathboma:
Thanks for taking a look, we'll figure it out

PR #2760, opened 2025-02-11
fix text selection in query editor

### LINKED ISSUE #2742: BUG: Incorrect query selection
Selecting the first query doesn't really select it. It selects the one after it. To replicate this, you will need to place the cursor from top and then drag to bottom, not the opposite.

Image

azmy60:
fix #2742


not-night-but (approved):

rathboma:
@azmy60 merging this, but any changes that need to be made to do this in the UI-Kit?

azmy60:
@rathboma yes. I'll add a patch to the PR. Though, the implementation is gonna be a little bit different.

PR #2787, opened 2025-02-21
Fix: JSON Viewer Bottleneck

### LINKED ISSUE #2727: BUG: Degraded performance in TableTable
**Describe the bug**
Clicking around tabletable when there are large fields for the JSON row viewer has degraded performance (possibly due to loading the data into the row viewer)

azmy60:
fix #2727

This PR fixes two issues:
1. Clicking around the table is slow due to the `persistJsonFold` plugin that's O(n^2) or probably even worse
2. Pressing and holding arrow keys is slower too because it doesn't wait for us release the keys to set the value. When the data is big enough, changing value with `CodeMirror.setValue()` can be a little slower, or it could be something else like a combo of all stuff we are doing inside the JSON viewer. But debouncing the key press event seems good enough.

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -1000,7 +1000,13 @@ export default Vue.extend({
       })
       this.tabulator.on("cellMouseUp", this.updateDetailView);
       this.tabulator.on("headerMouseUp", this.updateDetailView);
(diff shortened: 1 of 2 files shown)


rathboma (approved):
@azmy60 this will cause a conflict with your other PR about non-reactive tabulator.

PR #2800, opened 2025-02-25
Fix: tabulator errors in console

### LINKED ISSUE #2563: BUG: Result table striping flickers
The result table 'flickers' as you scroll. In light mode the white areas go grey, then back to white. It's very distracting.
[Screencast from 11-07-2024 04:26:05 PM.webm](https://github.com/user-attachments/assets/99f12711-8d2a-4866-b98b-f1cb0e65a53d)

azmy60:
image

This PR fixes 3 errors related to tabulator:

- Fixes errors that shows in the console when Vue accesses `tabulator` for reactivity. Therefor, it's removed from reactivity. In exchange, I added a `tabulatorBuilt` flag if we want to watch whether `tabulator` is defined or not.
- Fixes a `verticalFillMode` error that comes from calling `tabulator.setData()` when the tabulator is not initialized completely.
- Fixes #2563


rathboma (approved):

rathboma (commented):
Wait, won't removing the reference to Tabulator cause problems? You're not storing the reference to tabulator anywhere now?

rathboma:
@azmy60 Looks like you should maybe use the $options api instead? See - https://stackoverflow.com/questions/45814507/how-to-set-a-component-non-reactive-data-in-vue-2

azmy60:
@rathboma Yeah that sounds good. The `$options` api would also expose the `tabulator` to the vue template in comparison with just assigning `this.tabulator`.

azmy60:
@rathboma wait actually using `this.tabulator` should be enough I think? I wrote that [inside the data()](https://github.com/beekeeper-studio/beekeeper-studio/pull/2800/files#diff-729fbb92b07a0f4dd64e507cc9ca3ad100cea6dc3df2fe26d8a1b284030452c2R349)

rathboma:
Maybe I'm not reading it right, but I don't see where you assign `this.tabulator`? I see `const tabulator` and `this.tabulatorBuilt`

azmy60:
Actually we just need to disable the debug flag from the tabulator. I don't know why I didn't think of this in the beginning :facepalm:

rathboma:
@azmy60 merge this whenever you like!

PR #2802, opened 2025-02-25
Feat: Support binary encoding configuration

### LINKED ISSUE #2460: Customization Required for Viewing Binary Columns Encoding in Latest Release
### Description

In the update from version 4.6.4 to 4.6.8, the default encoding for binary columns was changed to hex encoding to fix an issue with binary primary keys. This alteration has caused JWT encoded columns to display incorrectly, as hex rather than a string

### Steps to Reproduce
1. Update Beekeeper Studio from version 4.6.4 to 4.6.8.
2. View data in a column that stores JWT encoded values.

### Expected Behavior
- JWT encoded values stored as binary should look correct

### Actual Behavior
- JWT encoded values are hex

### Suggested Solution
- Add a feature in the front-end to allow users to change the encoding of binary columns.
- Indicate the specific encoding used in the column header.

### Temporary Workaround
- Revert to an older version 4.6.4 using the following link: [v4.6.4 Release](https://github.com/beekeeper-studio/ultimate-releases/releases/tag/v4.6.4)

### Additional Notes
- This issue was identified by user feedback and backed by screenshots showing the discrepancy across versions.
- Fixing this will improve usability for users needing specific encodings for their binary data.

### Environment
- Beekeeper Studio version: 4.6.8 (Issue not present in 4.6.4)
- Operating System: [Specify OS]
***

azmy60:
fix #2460 

Binary format can be changed to base64 or hex using the .ini file config.


rathboma (approved):
I think this doesn't make sense at the app-level, each table/database is likely different. Rather this should be in the right click menu of the column header for tabulator for binary columns.

I'm fine for you to merge this as the 'default' setting, but if you do merge this PR, I need you to add a new issue for 'per-column display options for binary data'

Issue #2696, opened 2025-01-13
BUG: No file restrictions for when inserting a new db file in a new connection


JeanPaiva42:
**Describe the bug**

When starting a new connection we give the users the option of inserting a new Database File. In this moment, users can add any type of file even those that are not related to Databases, such as pdfs, images, scripts or .zip files.

**To Reproduce**
1. Go to `New Connection`
2. Select a Connection Type
3. Select `SQLite`
4. Pick a random (non Db file from your computer)
5. Notice that you are able to select such file
6. Also Notice that you are able to test the connection (it passes) and also `start` the connection by clicking `Connect`

**Expected behavior**
- Users shouldn't be able to select non db files for the SQLite connection.

**Screenshots / Gifs**

- On the video I select a `.py` file as the db file and start a connection with it.

https://github.com/user-attachments/assets/8c6c5c64-4c6e-44df-9e24-58b8a3217457

**Version Information (please complete the following information):**
 - OS: Mac
 - App Version: 5.0.9
 - Database type and version: SQLite

**Additional context**

- None.

PR #2846, opened 2025-03-06
fix: add allowed file extensions for database file picker

therealrinku:
fixes: #2696 

closes: #2696

added allowed file extensions for database file picker for sqlite, libsql and ducked forms.

**Allowed extensions**
_for sqlite and libsql_ => .db, .sqlite and .sqlite3
_for duck db_ => .duckdb

https://github.com/user-attachments/assets/3a1b7944-a8e2-4733-9608-cf80b9022a51

### DIFF
--- a/apps/studio/src/components/common/form/FilePicker.vue
+++ b/apps/studio/src/components/common/form/FilePicker.vue
@@ -46,6 +46,11 @@
 /* options and all for the native file picker can be found here https://www.electronjs.org/docs/latest/api/dialog */
 export default {
   props: {
+    allowedFileExtensions:{
+      type: Array,
+      required: false,
+      default: () => ([])
+    },
     value: {
       required: true,
     },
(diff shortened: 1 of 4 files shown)


rathboma (changes requested):
This is great, thanks! Could you somehow unify this with the types in electron-builder-config.js? Maybe they could share a `.json` file or something?

therealrinku:
hey @rathboma sure! I can check that
also, I didn't get what you mean by sharing .json file, could you explain that a bit?

rathboma:
Hey, I've decided not to merge this PR -- the databases explicitly state that the extension can be anything, so this would be going against the specs.

For example, DuckDB says [this here](https://duckdb.org/docs/stable/operations_manual/footprint_of_duckdb/files_created_by_duckdb.html):

> The file can have **any extension** with typical extensions being .duckdb, .db, and .ddb.

Issue #2835, opened 2025-03-04
FEAT: Edit data in the JSON sidebar


rathboma:
If a table is editable, users should be able to do this in the JSON sidebar.

Limits:
- Shouldn't be able to edit foreign key data
- Shouldn't be able to edit non-editable tables
- Shouldn't be able to edit query results

PR #2849, opened 2025-03-06
Make JSON Row Viewer Editable

azmy60:
This PR is linked to #2843

- Added `editablePaths` to json viewer component. This will allow the values of each `editablePath` to be editable. It should be noticed that the text editor will highlight the parts that is editable when hovered or the cursor is on it.
- It will submit the changes while typing using `_.debounce`
- If there is a syntax error, it won't submit and will show the error
- The mouse cursor is changed to normal but you still can navigate around the editor normally
- Primary keys, foreign keys, and generated keys are not editable.
- If the table is not editable, and so the json viewer.

json-viewer-editable

We can definitely support foreign keys to be editable. But this raises questions like:
1. How do we edit the foreign key if it's already expanded?
2. Are the keys inside the expanded foreign key editable as well?


azmy60:
@not-night-but Want to see your reviews on this especially performance-wise when using this with mongodb. Or if you have ideas how the edit should work to make it easier for the UX by any means, just let me know!

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1860,6 +1893,10 @@ export default Vue.extend({
         this.split = null
       }
     },
+    handleJsonValueChange({key, value}) {
+      console.log('handleJsonValueChange', key, value)
+      this.tabulator.getRow(this.selectedRowIndex).getCell(key).setValue(value)
+    },
```
not-night-but, line 1899: When I was using this, getRow didn't actually return a row. I was able to get around the issue by just using `this.selectedRow`, but this needs to be updated in some way to make sure it can always find the row.

apps/studio/src/lib/editor/plugins/PartialReadOnlyPlugin.ts
```diff
@@ -0,0 +1,180 @@
+/**
+ * This plugin sets the whole editor to be read-only except for some defined
+ * ranges. **/
+
```
not-night-but, line 4: Looks like there's a bug here where you can continue typing outside of the value.

image

If I place my cursor at the closing quotation mark of the password value, I can continue typing afterwards and do stuff like this:
image

Not actually sure if this is something we can really fix? Thought I'd mention it at least

not-night-but (commented):
Looking good to me! I think there may be a couple issues, mentioned some in comments. There's also an issue where if you reset your changes, the detail view doesn't actually update.

In terms of how this relates to mongo, I think an add field button on hover in the gutter might be nice? Just place the field after wherever you click? Either that or just have a button near the top that appends the field to the end of the object (Although that would only allow for top level changes I believe, so the gutter might be a better options). You might still need to add a column just to show the changes, but we can figure all that out once my mongo-v2 PR is merged methinks.

azmy60:
@not-night-but I think a button in the gutter is perfect! We can also use that to show other actions too like `remove` (if mongo supports it I guess?). Check https://jsonformatter.org/json-editor. If you generate a sample, it'll show a JSON view with a button on the gutter.

apps/studio/src/lib/editor/plugins/PartialReadOnlyPlugin.ts
azmy60: Ughh we probably can't fix this. But I'll try to figure something out. If not then we'll need to initiate a new CodeMirror editor on clicking each value.

azmy60: @not-night-but Does that only happen on text that were truncated? Or any text in general?

azmy60: image

@not-night-but just updated the error markers and the text markers background. The visual cues should be able to tell that doing that is not the intended use.

Also it won't submit if the JSON.parse fails, so that should be good I guess?

rathboma:
@not-night-but this been sitting for 2 weeks, you good with the changes?

not-night-but (approved):
@rathboma Yeah sorry I'm happy with the changes. My main concerns have been addressed so I'm happy to merge this

rathboma:
Fix #2835

PR #2855, opened 2025-03-07
fix app version license checking

### LINKED ISSUE #2854: BUG: "Upgrade required" window shows when it shouldn't
A customer support email:

1. Their license is cancelled, but in lifetime mode
2. The app says they can use any version up to the latest (5.1.4)
3. They're using 5.0.9
4. The JSON viewer tells them they need to upgrade to use the feature (it shouldn't)

What should happen: They should be allowed to use all the features (except workspaces) in their version.

Image

Image

not-night-but:
fix #2854 

This was caused by improper version comparison.

This also fixes dev license checking so we can actually use the dev license switch again.

### DIFF
--- a/apps/studio/tests/unit/common/appdb/models/license.spec.ts
+++ b/apps/studio/tests/unit/common/appdb/models/license.spec.ts
@@ -166,11 +166,11 @@ describe("License", () => {
 
     // Regression tests for version comparison
     it("Should properly compare versions", async () => {
-      expect(isVersionLessThanOrEqual(parseVersion("v2.4.6"), parseVersion("v5.7.2")));
-      expect(isVersionLessThanOrEqual(parseVersion("v2.5.1-beta.4"), parseVersion("v5.0.0")));
-      expect(!isVersionLessThanOrEqual(parseVersion("v6.3.7"), parseVersion("v4.2.1")));
-      expect(!isVersionLessThanOrEqual(parseVersion("v3.1.3-beta.4"), parseVersion("v1.8.1")));
-      expect(isVersionLessThanOrEqual(parseVersion("v5.0.0"), parseVersion("v5.0.0")));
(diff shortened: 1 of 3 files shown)


apps/studio/tests/unit/common/appdb/models/license.spec.ts
```diff
@@ -163,5 +163,14 @@ describe("License", () => {
         condition: ["Expired support date", "App version not allowed"],
       });
     })
+
+    // Regression tests for version comparison
+    it("Should properly compare versions", async () => {
+      expect(isVersionLessThanOrEqual(parseVersion("v2.4.6"), parseVersion("v5.7.2")));
```
rathboma, line 169: Don't you have to do `expect(x).toBeTrue()`?

not-night-but: Nope. I think jest actually inserts that in the background as the errors will say that we do have `.toBe(true)`, but it works just with booleans

rathboma: I'm adding toBeTruthy to be explicit. I think that's a good habit, I think it's accidentally skip a test without it

rathboma: Yeah it doesn't work without, see - broken test:

image

not-night-but: Oh it's just a bad test in general, will fix

rathboma: Your test has 'v' before the version number, the others don't. That's breaking the test. How is the version gotten in the actual app?

rathboma (changes requested):
test broken - plz fix! Please also check if the version string should be with or without the `v`. Maybe it needs to handle both

PR #2866, opened 2025-03-11
Add secondary sidebar and move JSON Viewer to the secondary sidebar

### LINKED ISSUE #2834: FEAT: Add JSON Row viewer to query tab
When someone runs a query and gets a result table, we should allow them to see the data in the JSON sidebar.

The json sidebar should be full-height (same as in table table), not scoped to the result table.

azmy60:
resolves #2834 

primary-secondary-sidebar


apps/studio/src/components/CoreInterface.vue
```diff
@@ -55,30 +58,33 @@
   import ProgressBar from './editor/ProgressBar.vue'
   import LostConnectionModal from './LostConnectionModal.vue'
   import Vue from 'vue'
-  import { SmartLocalStorage } from '@/common/LocalStorage'
   import RenameDatabaseElementModal from './common/modals/RenameDatabaseElementModal.vue'
-  import { mapGetters } from 'vuex'
+  import { mapGetters, mapActions, mapState } from 'vuex'
+  import _ from "lodash"
+
+  const MAIN_CONTENT_MIN_WIDTH = 200 // in pixels
```
rathboma, line 65: Man, wish we had some sort of ini system where we could put stuff like this :eyes:

rathboma (approved):
Looks good, maybe magic numbers should move to the ini?

rathboma:
Merge when ready

apps/studio/src/components/CoreInterface.vue
azmy60: Lol. Should we? I mean we could..

PR #2879, opened 2025-03-13
Fix #2489: Cannot select / copy relation names in UI

### LINKED ISSUE #2489: BUG: Cannot select / copy relation names in UI
**Describe the bug**
In the table structure page > relations, we cannot select and copy relation names.

**To Reproduce**
1. Go to table structure (right click on table, "View Structure").
2. Go to "Relations" tab.
3. Cannot copy / select anything in the relations table.

**Expected behavior**
Should be able to select / copy anything from relations table.

**Screenshots / Gifs**

https://github.com/user-attachments/assets/0a16b605-4ca1-4fda-9282-595750c1a37b

 - OS: macOS 15.1 Public Beta 3
 - App Version: 4.6.8
 - Database type and version: Postgres 15

RodrigoPerestrelo:
In the table structure page > Indexes, Relations and Triggers, we cannot select the relation name to copy it.
I created a method (handleCellDoubleClick(cell)) in the Vue files TableRelations, TableIndexes, and TableTriggers to handle double-click selection. I added this method to the respective columns (only those that made sense to me, avoiding fields such as checkboxes).
Close #2489

### DIFF
--- a/apps/studio/src/components/tableinfo/TableTriggers.vue
+++ b/apps/studio/src/components/tableinfo/TableTriggers.vue
@@ -66,17 +66,17 @@ export default {
     },
     sqliteTableColumns() {
       return [
-        { field: 'name', title: 'Name', tooltip: true},
-        { field: 'sql', title: 'SQL', tooltip: true}
+        { field: 'name', title: 'Name', tooltip: true, cellDblClick: (e, cell) => this.handleCellDoubleClick(cell)},
+        { field: 'sql', title: 'SQL', tooltip: true, cellDblClick: (e, cell) => this.handleCellDoubleClick(cell)}
       ]
     },
     normalTableColumns() {
       return [
-        { field: 'name', title: "Name", tooltip: true},
-        { field: 'timing', title: "Timing"},
-        { field: 'manipulation', title: "Manipulation"},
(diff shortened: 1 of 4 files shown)


apps/studio/src/components/tableinfo/TableTriggers.vue
```diff
@@ -88,6 +88,32 @@ export default {
...
+      // Function to remove contenteditable when clicking anywhere
+      const removeEditable = (event) => {
+        element.removeAttribute("contenteditable");
+        document.removeEventListener("click", removeEditable);
+      };
+
+      // Attach a global event listener
+      document.addEventListener("click", removeEditable);
+    }
+  },
```
not-night-but, line 116: Since this same function is used in all three components, I'd prefer if we add it to a mixin (you can probably create a mixin specifically for this as I don't think it really fits with any of our others). That way we don't have code duplication, and we can add this functionality easily to any other tables that exist in the future. Let me know if you have any questions about how mixins work, but you can look at our fk_click.ts mixin for an example.

not-night-but (changes requested):
Good job! This functions well, I just think we need to not have as much code duplication, so we should probably add a mixin for the `handleCellDoubleClick` function.

not-night-but (approved):

PR #2881, opened 2025-03-13
Mongo v2

not-night-but:
This is introducing editing for mongo. You can now edit/insert/delete documents in a collection, add/drop/rename/duplicate collections, add databases, add/drop indexes (this might need some UI work in the future, but I think it works for now), and schema validation.

image
image

A mongo shell will follow in a separate PR.


rathboma:
@not-night-but is schema validation a per-collection thing, or a connection-wide thing?

not-night-but:
@rathboma It's per-collection. It's so if you want you can actually enforce a schema for documents in a given collection. If an insert doesn't match the schema provided, it will then reject the insert

rathboma (approved):
Great work! Please add the minor //FIXME comments through the codebase before merging

PR #2888, opened 2025-03-15
Parse JSON data in table detail view sidebar

### LINKED ISSUE #2887: BUG: JSON viewer not parsing JSON columns
**Describe the bug**
I have a sqlite table with a json column. When opening the sidebar I see a json representation of my row. However, the column which is actually JSON is only shown as string and not as parsed JSON.

```json
{
  "id": 1,
  "name": "Foo",
  "meta": "{\"ping\": \"pong\"}"
}
```

**Expected behavior**
If I have a table with the following columns: id (int), name (text), meta (json).
The expected sidebar view should be:

```json
{
  "id": 1,
  "name": "Foo",
  "meta": {
    "ping": "pong"
  }
}
```

**Screenshots / Gifs**
Image

 - OS: Ubuntu 24.04.2 LTS
 - App: 5.1.5
 - Database type: sqlite

**Additional context**
I've never seen this working in earlier versions

fightbulc:
- Added JSON parsing for columns with JSON/JSONB data types
- Added detection for JSON-like strings in non-JSON columns
- Improved handling of JSON data display in the detail view sidebar
- Added error handling for invalid JSON strings with console warnings

Fixes #2887

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -1726,7 +1726,41 @@ export default Vue.extend({
       this.detailViewTitle = `Row ${position}`
       this.selectedRow = row
       this.selectedRowIndex = position
-      this.selectedRowData = this.$bks.cleanData(data, this.tableColumns)
+      
+      // Clean the data first
+      let cleanedData = this.$bks.cleanData(data, this.tableColumns)
+      
+      // Parse JSON columns
+      this.tableColumns.forEach(column => {
+        const columnValue = cleanedData[column.field]
+
+        // Check if the column is a JSON column
+        let isJsonColumn = String(column.dataType).toUpperCase() === 'JSON' || String(column.dataType).toUpperCase() === 'JSONB'
+
+        // If the column is not a JSON column, check if it is a JSON string
+        if (!isJsonColumn) {
+          const isColumnHasStringAndNotEmpty = typeof columnValue === 'string' && columnValue.trim() !== ''
+
+          if (isColumnHasStringAndNotEmpty) {
+            const isJsonObjectString = columnValue.startsWith('{') && columnValue.endsWith('}')
+            const isJsonArrayString = columnValue.startsWith('[') && columnValue.endsWith(']')
+
+            if (isJsonObjectString || isJsonArrayString) {
+              isJsonColumn = true
+            }
+          }
+        }
+
+        if (isJsonColumn) {
+          try {
+            cleanedData[column.field] = JSON.parse(cleanedData[column.field])
+          } catch (e) {
+            console.warn(`Failed to parse JSON for column ${column.field}:`, e)
+          }
+        }
+      })
+      
+      this.selectedRowData = cleanedData
       this.expandablePaths = this.rawTableKeys
         .filter((key) => !row.hasForeignData([key.fromColumn]))
         .map((key) => ({


rathboma:
@azmy60 can you review this?

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1726,7 +1726,41 @@ export default Vue.extend({
...
+      
+      // Parse JSON columns
+      this.tableColumns.forEach(column => {
+        const columnValue = cleanedData[column.field]
+
+        // Check if the column is a JSON column
+        let isJsonColumn = String(column.dataType).toUpperCase() === 'JSON' || String(column.dataType).toUpperCase() === 'JSONB'
+
+        // If the column is not a JSON column, check if it is a JSON string
+        if (!isJsonColumn) {
```
azmy60, line 1741: If we end up checking the string if it's json or not, then why don't we just call `JSON.parse` all the time? That way it'd be a lot simpler :)

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1726,7 +1726,41 @@ export default Vue.extend({
...
+
+            if (isJsonObjectString || isJsonArrayString) {
+              isJsonColumn = true
+            }
+          }
+        }
+
+        if (isJsonColumn) {
+          try {
+            cleanedData[column.field] = JSON.parse(cleanedData[column.field])
```
azmy60, line 1756: Also, could you move this implementation to the detail view component instead of `TableTable`? Currently, we are adding detail view sidebar / json viewer to the query editor tab. So it'd make sense if this behavior would also exist there.

azmy60 (changes requested):
I have some feedbacks. Other than that, this is good stuff! Would love to see this in the app.

apps/studio/src/components/tableview/TableTable.vue
fightbulc: Totally. I tried a quick implementation but did not get it to run. I am lacking the time at the moment.

apps/studio/src/components/tableview/TableTable.vue
azmy60: Yeah, I can see that. Something like `"123"` on a text column would be converted to an actual number which is not good.

apps/studio/src/lib/data/jsonViewer.ts
```diff
@@ -137,3 +138,37 @@ export function eachPaths(
...
+
+    // Check if the column is a JSON column
+    let isJsonColumn = String(column.dataType).toUpperCase() === 'JSON' || String(column.dataType).toUpperCase() === 'JSONB'
+
+    // If the column is not a JSON column, check if it is a JSON string
+    if (!isJsonColumn) {
+      const isColumnHasStringAndNotEmpty = typeof columnValue === 'string' && columnValue.trim() !== ''
+
+      if (isColumnHasStringAndNotEmpty) {
+        const isJsonObjectString = columnValue.startsWith('{') && columnValue.endsWith('}')
```
rathboma, line 155: You'll need to trim whitespace here too -- spaces at the end or start of this string will break the detection.

This should have a unit test

apps/studio/src/lib/data/jsonViewer.ts
```diff
@@ -137,3 +138,37 @@ export function eachPaths(
...
+          isJsonColumn = true
+        }
+      }
+    }
+
+    if (isJsonColumn) {
+      try {
+        data[column.field] = JSON.parse(data[column.field])
+      } catch (e) {
+        console.warn(`Failed to parse JSON for column ${column.field}:`, e)
```
rathboma, line 168: This should use bksLogger

rathboma (changes requested):

fightbulc:
Hey @azmy60, is there still anything needed from my end?

azmy60:
Hey @fightbulc , thanks for the heads up, I think we are good. We were about to put this in 5.2 but something happened so this will be in the next release.

fightbulc:
alright, all the best!

azmy60 (approved):

azmy60:
@rathboma merging is blocked. Need your approval.

PR #2893, opened 2025-03-16
Feat/mysql postgres entra

ClarLabs:
This branch uses CLI authentication for MS SQL, MySQL and Postgres to authenticate using the clients AZ CLI.


apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,6 +171,31 @@ export class AzureAuthService {
     }
   }
 
+  private async getAzureCLIToken(): Promise<AuthConfig | null> {
+    return new Promise((resolve, reject) => {
+      const command = "az account get-access-token --resource https://ossrdbms-aad.database.windows.net --output json";
```
rathboma, line 176: Beekeeper will need to specifically resolve the az command? Like with backup/restore we should auto-detect this, but if we can't find it we'll need the user to provide the path to it.

We should be really explicit in the form about the fact it is using the `az` client to authenticate.

Also -- this doesn't seem to use the clientSecret or the msiendpoint?

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,6 +171,31 @@ export class AzureAuthService {
...
+      exec(command, { encoding: "utf8" }, (error, stdout) => {
+        if (error) {
+          console.error("Error getting token:", error);
+          reject(null);
```
rathboma, line 180: why reject(null)? Shouldn't we reject with the error?

Also -- you need to use @bksLogger, vs console for logging

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,6 +171,31 @@ export class AzureAuthService {
     }
   }
 
+  private async getAzureCLIToken(): Promise<AuthConfig | null> {
+    return new Promise((resolve, reject) => {
+      const command = "az account get-access-token --resource https://ossrdbms-aad.database.windows.net --output json";
+      exec(command, { encoding: "utf8" }, (error, stdout) => {
```
rathboma, line 177: What happens to the stderr here? We just discarding it? Might want to put that in the logs

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,6 +171,31 @@ export class AzureAuthService {
     }
   }
 
+  private async getAzureCLIToken(): Promise<AuthConfig | null> {
+    return new Promise((resolve, reject) => {
```
rathboma, line 175: Also might be easier to promisify the exec function so you can use async await:

```
// don't do it this way, but you get the idea
const exec = util.promisify(require('child_process').exec);
```

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,6 +171,31 @@ export class AzureAuthService {
...
+          const tokenData = JSON.parse(stdout);
+          resolve(  {
+            type: 'azure-active-directory-default',
+            options: {
+              token: tokenData.accessToken,
+            }
+          });
+        } catch (parseError) {
+          console.error("Error parsing token JSON:", parseError);
+          reject(null);
```
rathboma, line 193: same question about the reject. How does the upstream code know what error happened?

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,6 +171,31 @@ export class AzureAuthService {
     }
   }
 
+  private async getAzureCLIToken(): Promise<AuthConfig | null> {
```
rathboma, line 174: I think it would be better to move the CLI auth code to a new file, so it's not mixed in with the cloud SSO code

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -99,6 +115,8 @@ export class AzureAuthService {
         return this.msiVM(options);
       case AzureAuthType.ServicePrincipalSecret:
         return this.servicePrincipal(options);
+      case AzureAuthType.CLI:
+        return await this.getAzureCLIToken();
```
rathboma, line 119: This doesn't use the MsiEndpoint or the ClientSecret at all. Is that intended?

apps/studio/src/lib/db/authentication/azure.ts
rathboma, line 176: If this URL is static, it should be added to the `default.ini` file.

rathboma (changes requested):
Hey Chris, this looks good overall -- I added some questions about how this works, resolving the `az` client, and telling the user what this is doing. 

Also looking at other applications, we will likely need to support a version that doesn't require clientSecret. DBeaver calls this 'enterprise application' authentication.

tl;dr - it needs to be a lot more clear what auth method is being used, and what fields are needed to use it.

Seems like we'll need a couple dropdowns:

Authentication method: [username/password, Azure EntraId]
- If authenticating with EntraID, user should be able to select from 'cloud sso', 'use the az cli', etc. Then only provide the options necessary for that authentication type

apps/studio/src/lib/db/authentication/azure.ts
ClarLabs: That is correct, this does not require anything, the CLI will do the heavy lifting, as az is already authenticated it will get the access token based on the user, tenant and subscription the az authentication.

rathboma:
@ClarLabs let me know when this is re-ready for review

apps/studio/src/lib/db/authentication/azure.ts
ClarLabs: Discussed this Wednesday decided it's best to keep all azure code in one place to avoid future issues of where certain things are, ensuring everything azure related are localised.

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,163 @@
...
+        >help_outlined</i>
+        </label>
+        <input
+          name="tenantId"
+          type="text"
+          class="form-control"
+          v-model="config.azureAuthOptions.tenantId"
+        >
+      </div>
+      <div class="form-group" v-show="showClientSecret">
```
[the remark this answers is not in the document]
ClarLabs: This commonEntraId template is used for MSSQL as well which supports the other auth methods

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,163 @@
+<template>
+  <div v-show="azureAuthEnabled" class="host-port-user-password">
```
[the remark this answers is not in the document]
ClarLabs: Added info section, yes this is used for MSSQL now as well ClientId and other items are not required for CLI, but are for MSSQL

apps/studio/src-commercial/entrypoints/main.ts
```diff
@@ -160,6 +161,15 @@ async function initBasics() {
   return settings
 }
 
+ipcMain.handle('which-tool', async (_event, toolName: string) => {
```
rathboma, line 164: This should be done in the utility process, not the main process, as utility process is where any cli execution will happen

apps/studio/src-commercial/entrypoints/preload.ts
```diff
@@ -39,6 +39,13 @@ export const api = {
   isReady() {
     ipcRenderer.send('ready');
   },
+  rawInvoke(channel: string, ...args: any[]) {
```
rathboma, line 42: I don't want to expose raw invoke to the renderer, this will have to go I'm afraid.

apps/studio/src/common/globals.ts
```diff
@@ -25,6 +25,9 @@ export default {
   iamRefreshTime: 13 * 60 * 1000, // 13 minutes
   iamExpiryTime: 15 * 60 * 1000, // 15 minutes
   iamRefreshBeforeTime: 2 * 60 * 1000, // 2 minutes
+  azSQLLoginScope: 'https://ossrdbms-aad.database.windows.net',
```
rathboma, line 28: We should move these to the default ini file, so users can change them.

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,222 @@
...
+        this.errorSigningOut = e
+        this.$emit('error', e)
+      } finally {
+        this.signingOut = false
+      }
+    },
+    async tryFindAzCli() {
+      if (!this.config.azureAuthOptions.cliPath) {
+        try {
+          const result = await window.main.rawInvoke('which-tool', 'az')
```
rathboma, line 206: this will need to be replaced -- you should be able to easily add a handler in the utility process for this, it should be properly scoped to only look for the `az` binary

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,222 @@
...
+          if (result.success) {
+            this.config.azureAuthOptions.cliPath = result.path;
+          } else {
+            this.config.azureAuthOptions.cliPath = 'NO CLI FOUND, PLEASE REFER TO DOCS'
```
rathboma, line 210: This shouldn't be assigned to the `cliPath` variable, have a `cliError` variable on the component and show that to the user when you can't find the path.

apps/studio/src/components/connection/SqlServerForm.vue
```diff
@@ -51,105 +51,7 @@
...
-          <input
-            name="msiEndpoint"
-            type="text"
-            class="form-control"
-            v-model="config.azureAuthOptions.msiEndpoint"
-          >
-        </div>
-      </div>
-    </div>
+    <common-entra-id v-show="azureAuthEnabled" :auth-type="authType" :config="config" />
```
rathboma, line 54: love the addition of the shared component, much cleaner.

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,17 +172,43 @@ export class AzureAuthService {
     }
   }
 
+  private async getAzureCLIToken(options: AzureAuthOptions): Promise<AuthConfig> {
+    if(!options?.cliPath){
+      throw new Error('AZ command not specified')
+    }
+
+    const command = `${options?.cliPath} account get-access-token --resource ${globals.azSQLLoginScope} --output json`;
```
rathboma, line 180: `options.cliPath` needs to be properly escaped. Also I don't think you imported `execAsync`?

rathboma: To do proper escaping you should use `spawn`, that takes a command, and a list or arguments, that way the `cliPath` will be appropriately escaped.

apps/studio/src/lib/db/clients/utils.ts
```diff
@@ -39,6 +42,34 @@ export function joinQueries(queries) {
   return results.join("")
 }
 
+export async function whichTool({ toolName }: { toolName: string }) {
+  const command = `${platformInfo.isWindows ? 'where' : 'which'}`
+
+  return new Promise<string>((resolve, reject) => {
+    const proc = spawn(command, [toolName], { shell: true });
```
rathboma, line 49: See -- we're properly using spawn here to run a command, we should use spawn in the other method also. Maybe make a nice spawn wrapper so you don't have to rely on proc.stdout.on and such

apps/studio/src/components/connection/CommonEntraId.vue
rathboma: Tell them they need to set the path explicitly in their settings file, then link directly to that page on how to do it.

rathboma (changes requested):
Looks GREAT, Requests are: tighten things up a little, adhere to the bks way of doing things, and use the settings files!

apps/studio/src/components/connection/CommonEntraId.vue
ClarLabs: I need to investigate this one more, as I can do window.main.whichTool('az') as that same code is used for the dump tool, the issue is, I don't know if the dump tool is over engineered or has a reason to be convoluted and abstracted, will take a closer look and get back to you.

not-night-but: I can answer that for you: it's over-engineered into the freaking stratosphere! The idea was to make it super easy to add new forms/tools, and then the processes split happened and added a whole new layer of complexity to everything. I've been meaning to refactor it when I get the chance because it is now just a pain to maintain.

docs/user_guide/connecting/azure-entraid.md
```diff
@@ -0,0 +1,85 @@
+# Connecting to Azure Entra ID
```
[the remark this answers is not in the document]
ClarLabs: This does have an SSO section, CLI section and principal section, I have added troubleshooting section.

apps/studio/src-commercial/entrypoints/main.ts
ClarLabs: I can't seem to do this in the utility.ts file, however, I have changed it to be more useful, but left the IPC handler in place, as it seems the only place electron registers. Happy to be proven wrong ;), but I have spent ages trying to get it working in the utility file.

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,17 +171,64 @@ export class AzureAuthService {
...
+      const proc = spawn(options?.cliPath, [
+      'account',
+      'get-access-token',
+      '--resource',
+      BksConfig.azure.azSQLLoginScope,
+      '--output',
+      'json'
+    ]);
+
+      proc.stdout.on('data', (chunk) => {
```
rathboma, line 193: @ClarLabs Several issues here with spawn:

1. `data` events can trigger multiple times, it's a buffer so it triggers again when there is more output.
2. A healthy process can emit both STDERR and STDOUT data, you can't use this to figure out if it's good or bad
3. You're not listening for the `close` event -- this is how you figure out if the process has ended (and whether it succeeded).
  ```
    ls.on('close', (code) => {
      console.log(`child process exited with code ${code}`);
    });
  ```

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,17 +171,64 @@ export class AzureAuthService {
...
+            options: {
+              token: tokenData.accessToken,
+              clientId: globals.clientId
+            }
+          });
+        }
+      });
+
+      proc.stderr.on('data', (chunk) => {
+        reject(chunk.toString());
```
rathboma: You can't reject here, this stderr output could be totally valid (eg log data)

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,17 +171,64 @@ export class AzureAuthService {
...
+      })
+
+      proc.on('error', (err) => {
+        reject(err);
+      })
+
+      proc.on('close', (code) => {
+        if (code != 0) {
+          reject('ERROR: Command exited with errors');
```
rathboma: When you reject here you should include the STDERR and STDOUT logs

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -153,17 +171,64 @@ export class AzureAuthService {
...
+      '--resource',
+      BksConfig.azure.azSQLLoginScope,
+      '--output',
+      'json'
+    ]);
+
+      proc.stdout.on('data', (chunk) => {
+        if (chunk) {
+          const data: string = chunk.toString().trim();
+          const tokenData = JSON.parse(data);
```
rathboma: You have to collect the output, you can't parse it here, what if there are more chunks?

apps/studio/src/lib/db/clients/utils.ts
```diff
@@ -39,6 +42,34 @@ export function joinQueries(queries) {
...
+
+    proc.stdout.on('data', (chunk) => {
```
rathboma: Same issues as discussed above.

rathboma (changes requested):
You still need to revise how you use `spawn`.

apps/studio/src-commercial/entrypoints/preload.ts
```diff
@@ -9,7 +9,7 @@ import username from 'username';
 import { execSync } from 'child_process';
 import 'electron-log/preload';
 import pluralize from 'pluralize';
-
+import * as utility from './utility'
```
not-night-but: We definitely shouldn't do this. The entrypoints should remain separate (it also doesn't get used, but keep that in mind)

apps/studio/src-commercial/entrypoints/main.ts
```diff
@@ -172,6 +173,15 @@ app.on('window-all-closed', () => {
...
+  try {
+    const path = await whichTool({ toolName });
+    return { success: true, path };
+  } catch (error) {
+    return { success: false, error: error.toString() };
+  }
+});
+
```
not-night-but: We shouldn't do this, we can run this request through the utility connection.

apps/studio/src-commercial/entrypoints/preload.ts
```diff
@@ -39,6 +39,13 @@ export const api = {
   isReady() {
     ipcRenderer.send('ready');
   },
+  whichTool(toolName: string) {
+    return ipcRenderer.invoke('which-tool', toolName);
+  },
```
not-night-but: We already have a mechanism for communicating to the backend with the UtilityConnection, we should use that instead of trying to run things in the main process (The main process should literally only be used for application setup and interacting with electron directly)

apps/studio/src-commercial/entrypoints/preload.ts
```diff
@@ -39,6 +39,13 @@ export const api = {
...
+  invoke(event: AppEvent, name: string, arg?: any) {
+    if (!Object.values<string>(AppEvent).includes(event)) return;
+    ipcRenderer.invoke(event, name, arg)
+  },
```
not-night-but: This doesn't seem to be used anywhere?

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,262 @@
...
+        this.errorSigningOut = e;
+        this.$emit("error", e);
+      } finally {
+        this.signingOut = false;
+      }
+    },
+    async tryFindAzCli() {
+      if (!this.config.azureAuthOptions.cliPath) {
+        try {
+          const result = await window.main.whichTool("az");
```
not-night-but: This should just be done through the utility process. So you can 
```suggestion
          const result = await this.$util.send('backup/whichDumpTool');
```

You can move it to another handler file if you want, and just update the other calls of it, but we don't need to have two mechanisms to call the same code.

apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -130,6 +155,23 @@ export default {
     toggleServerless() {
       this.config.redshiftOptions.isServerless = !this.config.redshiftOptions.isServerless
     },
-  }
+    async tryFindAWSCli() {
+      if (!this.config.redshiftOptions.cliPath) {
+        try {
+          const result = await window.main.whichTool("aws");
```
not-night-but: Same as above

apps/studio/src/lib/db/authentication/amazon-redshift.ts
```diff
@@ -30,6 +32,58 @@ export interface TemporaryClusterCredentials {
...
+            resolve(stdout.trim());
+          } catch (err) {
+            console.error(`Failed to parse token JSON: ${err}\nRaw output: ${stdout}`);
+            reject(`Failed to parse token JSON: ${err}\nRaw output: ${stdout}`);
+          }
+        } else {
+          console.error(`Process exited with code ${code}\nSTDERR: ${stderr}\nSTDOUT: ${stdout}`);
+          reject(`Process exited with code ${code}\nSTDERR: ${stderr}\nSTDOUT: ${stdout}`);
+        }
+      });
```
not-night-but: Just a note about  logging: if you want to leave it in, use the bksLogger. :
```suggestion
        log.error(`Error executing AWS CLI command: ${err}`);
        reject(err);
      });

      proc.on('close', (code) => {
        if (code === 0) {
          try {
            log.info(stdout)
            resolve(stdout.trim());
          } catch (err) {
            log.error(`Failed to parse token JSON: ${err}\nRaw output: ${stdout}`);
            reject(`Failed to parse token JSON: ${err}\nRaw output: ${stdout}`);
          }
        } else {
          log.error(`Process exited with code ${code}\nSTDERR: ${stderr}\nSTDOUT: ${stdout}`);
          reject(`Process exited with code ${code}\nSTDERR: ${stderr}\nSTDOUT: ${stdout}`);
        }
      });
```
Just means that in the case of these errors hitting users, it gets saved to a file we can see which helps us debug for them, as well as just working better with electron in general.

apps/studio/src/lib/db/clients/utils.ts
```diff
@@ -40,6 +43,37 @@ export function joinQueries(queries) {
   return results.join("")
 }
 
+export async function whichTool({ toolName }: { toolName: string }): Promise<string> {
```
not-night-but: So honestly I would just move this back to the backupHandlers. This type of thing should *ONLY* be run from the utility process. Since this file is imported from types, which I believe is then used in the frontend, this probably won't build for production (We run in a secure mode that means that absolutely no native code can be in the frontend.)

apps/studio/src/lib/db/types.ts
```diff
@@ -1,4 +1,6 @@
 import { CancelableQuery, DatabaseFilterOptions, ExtendedTableColumn, FilterOptions, ImportFuncOptions, NgQueryResult, OrderBy, PrimaryKeyColumn, Routine, SchemaFilterOptions, StreamResults, SupportedFeatures, TableChanges, TableColumn, TableFilter, TableIndex, TableInsert, TableOrView, TablePartition, TableProperties, TableResult, TableTrigger, TableUpdateResult } from './models';
+import {AzureAuthService} from "@/lib/db/authentication/azure";
+import {getEntraOptions} from "@/lib/db/clients/utils";
```
not-night-but: These aren't used?

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,262 @@
...
+            }"
+            >help_outlined</i
+          >
+        </label>
+        <input
+          name="server"
+          type="text"
+          class="form-control"
+          v-model="config.host"
+        />
```
not-night-but: We may want to make some of these masked inputs. Anything that you wouldn't want to be seen in a video or something. Check out `CommonServerInputs.vue` for examples

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,262 @@
...
+          v-model="config.defaultDatabase"
+        />
+      </div>
+      <div
+        class="advanced-connection-settings signed-in-as"
+        v-if="hasAccessTokenCache"
+      >
+        <div class="advanced-body">
+          <span class="info"
+            >Signed in{{ accountName ? ` as ${accountName}` : "" }}</span
```
not-night-but: I'd also censor this if privacyMode is enabled

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,262 @@
...
+      </div>
+      <div class="form-group" v-show="showMsiEndpoint">
+        <label for="msiEndpoint">MSI Endpoint</label>
+        <input
+          name="msiEndpoint"
+          type="text"
+          class="form-control"
+          v-model="config.azureAuthOptions.msiEndpoint"
+        />
+      </div>
```
not-night-but, line 125: Yeah I'd say these should all be masked

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,262 @@
...
+      ].includes(this.authType);
+    },
+    showClientSecret() {
+      return [AzureAuthType.ServicePrincipalSecret].includes(this.authType);
+    },
+    showMsiEndpoint() {
+      return [AzureAuthType.MSIVM].includes(this.authType);
+    },
+    showCli() {
+      return [AzureAuthType.CLI].includes(this.authType);
```
not-night-but: If it just has one element in the array, I'd rather we just check if they are equal.
```suggestion
      return AzureAuthType.ServicePrincipalSecret === this.authType;
```

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -134,14 +134,19 @@ async function configDatabase(
   database: IDbConnectionDatabase
 ): Promise<mysql.PoolOptions> {
 
+  let awsCLIToken = undefined;
+  if( server.config.redshiftOptions.authType === 'iam_cli') {
+    awsCLIToken = await getAWSCLIToken(server.config, server.config.redshiftOptions);
+  }
```
not-night-but: I really don't like us using redshiftOptions everywhere now. That was meant to be a redshift only thing. I'd prefer we have `iamAuthOptions` or something so we can just use it db agnostically.

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -134,14 +134,19 @@ async function configDatabase(
...
+
   const config: mysql.PoolOptions = {
     authPlugins: {
       'client_ed25519': ed25519AuthPlugin(),
     },
     host: server.config.host,
     port: server.config.port,
     user: server.config.user,
-    password: await refreshTokenIfNeeded(server.config.redshiftOptions, server, server.config.port || 3306) || server.config.password || undefined,
+    password: awsCLIToken || await refreshTokenIfNeeded(server.config.redshiftOptions, server, server.config.port || 3306) || server.config.password || undefined,
```
not-night-but: This could be a problem. If we aren't in iam_cli mode, it's still going to try to refresh the token. This should be in the if statement up above, just check if the token is undefined after getting it, then refresh if needed. Then here you just do this
```suggestion
    password: awsCLIToken || server.config.password,
```

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1317,16 +1329,27 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult> {
...
+    if( server.config.redshiftOptions.authType === 'iam_cli') {
+      awsCLIToken = await getAWSCLIToken(server.config, server.config.redshiftOptions);
+    }
+
     const config: PoolConfig = {
       host: server.config.host,
       port: server.config.port || undefined,
-      password: await refreshTokenIfNeeded(server.config?.redshiftOptions, server, server.config.port || 5432) || server.config.password || undefined,
+      password: awsCLIToken || await refreshTokenIfNeeded(server.config?.redshiftOptions, server, server.config.port || 5432) || server.config.password || undefined,
       database: database.database,
```
not-night-but: Same notes as in mysql. (I know it wasn't added in this PR, but we shouldn't be running it every time regardless)

not-night-but (changes requested):
Looks good, but some issues with how our processes are separated.

- Any code that interacts with the OS should be in utility, here we are running some stuff in the main process which should not be done. That is mainly just for setting up the application and some general lifetime hooks for the app.
- To interact with that code, just use the UtilityConnection. Handles all the messaging bs for you.
- And then just some things about privacyMode (which I always forget that we have lol)

apps/studio/src-commercial/backend/handlers/awsHandlers.ts
```diff
@@ -0,0 +1,43 @@
...
+        reject(err);
+      });
+
+      proc.on("close", (code) => {
+        if (code === 0) {
+          const profiles = stdout.trim().split("\n");
+          console.log("AWS Profiles:", profiles);
+          resolve(profiles);
+        } else {
+          console.error(`AWS CLI failed with code ${code}`);
```
not-night-but: If we want to keep these logs, we should use bksLogger

apps/studio/src-commercial/entrypoints/main.ts
```diff
@@ -26,6 +26,7 @@ import { uuidv4 } from '@/lib/uuid';
 import installExtension, { VUEJS_DEVTOOLS } from 'electron-devtools-installer';
 import { UtilProcMessage } from '@/types'
 import { manageUpdates } from '@/background/update_manager'
+import {whichTool} from "@/lib/db/clients/utils";
```
not-night-but: This isn't used

apps/studio/src/components/connection/CommonEntraId.vue
not-night-but: @ClarLabs Still need to make these masked inputs. Anything remotely sensitive should use that instead of raw input tags

apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -14,122 +12,217 @@
...
-          <label for="Access Key ID">
-            Access Key ID
-          </label>
-          <input
-            type="text"
-            class="form-control"
-            v-model="config.redshiftOptions.accessKeyId"
-          >
+          <label for="Access Key ID">Access Key ID</label>
+          <input type="text" class="form-control" v-model="config.iamAuthOptions.accessKeyId">
```
not-night-but: A lot of these inputs should also be masked inputs

apps/studio/src/lib/db/authentication/amazon-redshift.ts
not-night-but: @ClarLabs This still applies. I would also note: be careful about logging anything that could be sensitive data. There are some places in the app that we currently do this and it's just gross tbh 😂 . I don't want users to send us logs and accidentally send us the creds to their database or something.

apps/studio/src/lib/db/types.ts
not-night-but: @ClarLabs I still don't think these are used

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+      <div class="form-group" v-show="showUser">
+        <label for="user">User</label>
+        <input
+          name="user"
+          type="text"
+          class="form-control"
+          v-model="username"
+        />
+      </div>
+      <div class="form-group" v-show="showPassword">
```
not-night-but: This is the one private piece of info I would say shouldn't be masked, just because it's hidden by default and I don't think we should mess with that usual flow. Annoying cause I just told you to make everything masked, oversight on my part

apps/studio/src/components/connection/MysqlForm.vue
```diff
@@ -45,52 +46,66 @@
...
       accountName: null,
       signingOut: false,
       errorSigningOut: null,
     }
   },
   computed: {
     ...mapGetters(['isCommunity']),
+    showServerInputs() {
+      return !this.azureAuthEnabled && !this.iamAuthenticationEnabled
+    }
```
not-night-but, line 53: I'm not sure if this is currently functioning? The server inputs disappear for azure, but not for iam.

apps/studio/src/components/connection/PostgresForm.vue
```diff
@@ -99,15 +108,21 @@ export default {
...
     }
   },
   computed: {
     ...mapGetters(['isCommunity']),
     isCockroach() {
       return this.config.connectionType === 'cockroachdb'
     },
+    showServerInputs() {
+      return !this.azureAuthEnabled && !this.iamAuthenticationEnabled
+    }
```
not-night-but, line 122: Also not working here

not-night-but (changes requested):
Just a UI bug with the postgres and mysql/maria forms, and clean up the masked inputs. 

Tbh idk what to do with the privacy mode stuff because it's something we forget about all the time, and idk if anyone uses it but it was a community contribution and I don't want it to die lol. I also don't want someone that was relying on privacy mode to accidentally leak there creds online lol. Anyways, just make sure that anything you think should be censored if you were to say livestream your connections screen is masked input, besides passwords because they're technically just handled by our default handling of passwords.

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+        <masked-input :value="config.azureAuthOptions.clientId" @input="val => config.azureAuthOptions.clientId = val" />
+      </div>
+      <div class="form-group" v-show="isServicePrincipal">
+        <label for="clientSecret">Client Secret</label>
+        <masked-input :value="config.azureAuthOptions.clientSecret" @input="val => config.azureAuthOptions.clientSecret = val" :type="'password'" />
+      </div>
+      <div class="form-group" v-show="showMsiEndpoint">
+        <label for="msiEndpoint">MSI Endpoint</label>
+        <masked-input :value="config.azureAuthOptions.msiEndpoint" @input="val => config.azureAuthOptions.msiEndpoint = val" />
+      </div>
```
not-night-but, line 125: These do nothing unless you pass in privacyMode. Look at `CommonServerInputs.vue` to see how it works

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+        </label>
+        <masked-input :value="config.azureAuthOptions.tenantId" @input="val => config.azureAuthOptions.tenantId = val" />
+      </div>
+      <div class="form-group" v-show="isServicePrincipal">
+        <label for="clientSecret">Client ID</label>
+        <masked-input :value="config.azureAuthOptions.clientId" @input="val => config.azureAuthOptions.clientId = val" />
+      </div>
+      <div class="form-group" v-show="isServicePrincipal">
+        <label for="clientSecret">Client Secret</label>
+        <masked-input :value="config.azureAuthOptions.clientSecret" @input="val => config.azureAuthOptions.clientSecret = val" :type="'password'" />
```
not-night-but: You shouldn't pass in the password type. If you want it to be hidden by default, it should be like how we typically make our password forms. Look at `CommonServerInputs.vue`

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+        <input
+          name="user"
+          type="text"
+          class="form-control"
+          v-model="username"
+        />
+      </div>
+      <div class="form-group" v-show="showPassword">
+        <label for="password">Password</label>
+        <masked-input :value="config.password" @input="val => config.password = val" :type="'password'" />
```
not-night-but: Idk why you marked the conversation as resolved, but this still shouldn't be a masked input. It should be hidden by default, look at how we do passwords in `CommonServerInputs.vue`

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+      </div>
+      <div class="form-group" v-show="showUser">
+        <label for="user">User</label>
+        <input
+          name="user"
+          type="text"
+          class="form-control"
+          v-model="username"
+        />
+      </div>
```
not-night-but: I believe we typically hide the username in privacyMode as well, so a masked input for this as well. Remember to pass in privacy mode

apps/studio/src/components/connection/CommonEntraId.vue
not-night-but: @ClarLabs Still need to do this

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+            style="padding-left: 0.25rem"
+            v-tooltip="{
+              content:
+                'This is the <code>\'Server name\'</code> field on your Sql Server in Azure, <br/> you might also think of this as the hostname. <br/> Eg. <code>example.database.windows.net</code>',
+              html: true,
+            }"
+            >help_outlined</i
+          >
+        </label>
+        <masked-input :value="config.host" @input="val => config.host = val" />
```
not-night-but: Pass in privacymode

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+            style="padding-left: 0.25rem"
+            v-tooltip="{
+              content:
+                'This can be found in the <code>\'Microsoft Entra ID\'</code> section of Azure, <br/> in the Overview labelled <code>\'Tenant ID\'</code>',
+              html: true,
+            }"
+            >help_outlined</i
+          >
+        </label>
+        <masked-input :value="config.azureAuthOptions.tenantId" @input="val => config.azureAuthOptions.tenantId = val" />
```
not-night-but: Pass in privacy mode

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+  },
+  data() {
+    return {
+      azureAuthEnabled: this.config.azureAuthOptions?.azureAuthEnabled,
+      accountName: null,
+      signingOut: false,
+      errorSigningOut: null,
+    };
+  },
+  computed: {
```
not-night-but, line 151: You need to map the privacyMode state here so you can pass it into the masked inputs

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+    username: {
+      get() {
+        return this.config.username || this.config.user;
+      },
+      set(value) {
+        this.config.username = value;
+        this.config.user = value;
+      },
+    },
```
not-night-but: I don't really understand why we are doing it like this. Also, user is not defined on the `IConnection` interface. Why are we setting both? Are they meant to be separate? If so why are we setting them both to the same thing? Doesn't seem like it would be necessary to have both if they are the same thing.

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+    isServicePrincipal() {
+      return AzureAuthType.ServicePrincipalSecret === this.authType;
+    },
+    showMsiEndpoint() {
+      return AzureAuthType.MSIVM === this.authType;
+    },
+    showCli() {
+      return AzureAuthType.CLI === this.authType;
+    },
+    hasAccessTokenCache() {
```
not-night-but: A bit of a nitpick, but I'd prefer the variables to be on the left of these comparisons. You're checking if the auth type is equal to something, Not if the enum is equal to auth type, if that makes sense?

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,245 @@
...
+        <masked-input :value="config.azureAuthOptions.msiEndpoint" @input="val => config.azureAuthOptions.msiEndpoint = val" />
+      </div>
+    </div>
+  </div>
+</template>
+<script>
+import { AzureAuthType } from "@/lib/db/types";
+import { AppEvent } from "@/common/AppEvent";
+import _ from "lodash";
+import { onMounted } from "vue";
```
not-night-but: This seems unnecessary? Have never seen this imported before, seems like it might be from the other vue api or something

apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -14,122 +12,221 @@
...
-          <label for="Access Key ID">
-            Access Key ID
-          </label>
-          <input
-            type="text"
-            class="form-control"
-            v-model="config.redshiftOptions.accessKeyId"
-          >
+          <label for="Access Key ID">Access Key ID</label>
+          <masked-input :value="config.iamAuthOptions.accessKeyId" @input="val => config.iamAuthOptions.accessKeyId = val" :type="'password'" />
```
not-night-but: Need to pass in privacyMode. Also, would make this not a masked input, but our typical password input with the toggleable visibility

apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -14,122 +12,221 @@
...
-          <label for="Secret Access Key">
-            Secret Access Key
-          </label>
-          <input
-            type="password"
-            class="form-control"
-            v-model="config.redshiftOptions.secretAccessKey"
-          >
+          <label for="Secret Access Key">Secret Access Key</label>
+          <masked-input :value="config.iamAuthOptions.secretAccessKey" @input="val => config.iamAuthOptions.secretAccessKey = val" :type="'password'" />
```
not-night-but: Pass in privacy mode

apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -14,122 +12,221 @@
...
+
       <div v-show="isRedshift">
         <div class="form-group">
           <label for="Cluster Identifier">Cluster Identifier or Workgroup Name</label>
-          <input
-            type="text"
-            class="form-control"
-            v-model="config.redshiftOptions.clusterIdentifier"
-          >
+          <input type="text" class="form-control" v-model="config.redshiftOptions.clusterIdentifier">
```
not-night-but: Should probably be masked

apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -14,122 +12,221 @@
...
-        <label for="AWS Region">
-          AWS Region
-        </label>
-        <input
-          type="text"
-          class="form-control"
-          v-model="config.redshiftOptions.awsRegion"
-        >
+        <label for="AWS Region">AWS Region</label>
+        <input type="text" class="form-control" v-model="config.iamAuthOptions.awsRegion">
```
not-night-but: Should probably be masked

apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -14,122 +12,221 @@
...
     return {
-      iamAuthenticationEnabled: this.config.redshiftOptions?.iamAuthenticationEnabled,
-      isServerless: this.config.redshiftOptions?.isServerless
+      iamAuthenticationEnabled: this.config.iamAuthOptions?.iamAuthenticationEnabled,
+      isServerless: this.config.redshiftOptions?.isServerless,
+      cliError: false,
+      profilesError: false,
     };
   },
   computed: {
```
not-night-but, line 143: Need to map the privacy mode state here

not-night-but (changes requested):
The forms still need a some work, currently it seems like privacy mode just doesn't function at all. Try to take a look at `CommonServerInputs.vue` for how we typically handle things.

apps/studio/src/migration/20250820_migrate_redshift_to_iam_options.js
```diff
@@ -0,0 +1,46 @@
...
+      'awsProfile',
+      'iamAuthenticationEnabled',
+      'accessKeyId',
+      'secretAccessKey',
+      'awsRegion',
+      'authType'
+    ];
+
+    for (const table of tables) {
+      const records = await runner.query(`
```
azmy60, line 17: What if this returns 1 or 0 result? Are we sure it's always `array`?

azmy60:
@not-night-but ~Could we have a short list of what's being fixed besides the broken IAM auth? I wanna make sure I don't lose context.~ Will continue the review tomorrow morning.

Edit: Nevermind. It's a feature PR.

apps/studio/src/components/connection/SqlServerForm.vue
```diff
@@ -187,61 +88,31 @@
         } else {
           if (this.$store.getters.isCommunity) {
             // we want to display a modal
+            console.log('MODAL')
```
azmy60: 'MODAL'

apps/studio/src/components/LostConnectionModal.vue
```diff
@@ -56,8 +56,8 @@ export default Vue.extend({
   },
   watch: {
     connError(value) {
```
wmontgomery, line 58: Is this needed any longer since it doesn't show anything?

apps/studio/src/migration/20250820_migrate_redshift_to_iam_options.js
azmy60: Just making sure it doesn't break the migration

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,233 @@
...
+    <div class="form-group col">
+      <div v-show="showCli" class="form-group">
+        <label for="cliPath"
+          >AZ Tool
+          <i
+            class="material-icons"
+            style="padding-left: 0.25rem"
+            v-tooltip="{
+              content:
+                'You are signing in using the <code>\'Azure CLI\' Beekeeper Studio will attempt to use the AZ tool in path specified.</code>',
```
azmy60: Just an opinion. I feel like this type of info is better to be shown right away (like using a `<div class="alert alert-info">`) instead of hidden under a tooltip.

apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -14,10 +12,42 @@
...
               href="https://docs.aws.amazon.com/redshift/latest/mgmt/generating-user-credentials.html"
-            >Amazon Docs</a> - <a href="https://docs.beekeeperstudio.io/user_guide/connecting/amazon-rdb">Beekeeper Docs</a>
+            >Amazon Docs</a> - <a href="https://docs.beekeeperstudio.io/user_guide/connecting/amazon-rds">Beekeeper Docs</a>
           </div>
         </div>
       </div>
+
+      <div class="form-group col" v-show="showCli">
+        <div class="form-group">
+          <label for="cliPath">AWS Tool
```
azmy60: Labeling this as `AWS Tool` feels a little vague to me, unless it's a common label. It's also not consistent with the docs. i personally would label this `AWS CLI` or `AWS CLI Path`.

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,233 @@
...
+          href="https://docs.beekeeperstudio.io/user_guide/connecting/azure-entraid"
+          >Beekeeper Docs</a
+        >
+        for more information
+      </div>
+    </div>
+    <div class="form-group col">
+      <div v-show="showCli" class="form-group">
+        <label for="cliPath"
+          >AZ Tool
```
azmy60: Same as `AWS Tool`. This should probably be `Azure CLI`, or better, `Azure CLI Path`.

apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -14,10 +12,42 @@
...
+              class="material-icons"
+              style="padding-left: 0.25rem"
+              v-tooltip="{
+                content:
+                  'You are signing in using the <code>AWS CLI</code>. Beekeeper Studio will attempt to use the AWS CLI tool at the specified path.',
+                html: true,
+              }">help_outlined</i>
+          </label>
+
+          <div class="alert alert-danger" v-show="!cliFound">
```
azmy60, line 33: It think this alert box should be placed underneath the input rather than underneath the label?

Image

apps/studio/src/components/connection/CommonIam.vue
azmy60: Also the error message says `No CLI found`, not `No AWS Tool found`. I think it would make a difference if we fix the label, no?

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -0,0 +1,233 @@
...
+            style="padding-left: 0.25rem"
+            v-tooltip="{
+              content:
+                'You are signing in using the <code>\'Azure CLI\' Beekeeper Studio will attempt to use the AZ tool in path specified.</code>',
+              html: true,
+            }"
+            >help_outlined</i
+          ></label
+        >
+        <div class="alert alert-danger" v-show="!config.azureAuthOptions.cliPath">
```
azmy60: Should this be using `cliFound`? See `CommonIAM.vue` lines 33 and 157.

apps/studio/src/components/connection/SqlServerForm.vue
azmy60: 'MODAL' 🗣️

apps/studio/src/components/connection/SqlServerForm.vue
```diff
@@ -187,61 +88,31 @@
...
+      },
     },
     computed: {
       ...mapState(['connection']),
-      showUser() {
-        return [AzureAuthType.Password].includes(this.authType)
-      },
       showPassword() {
         return [AzureAuthType.Password].includes(this.authType)
-      },
```
azmy60, line 216: clean!

docs/user_guide/connecting/amazon-rds.md
```diff
@@ -49,3 +49,24 @@ Below is an example config in Beekeeper Studio once IAM permissions are setup, y
...
+### 1. AWS CLI Authentication (PostgreSQL, MS SQL)
+
+Uses your local AWS CLI session to retrieve an access token.
+
+![AWS IAM CLI](../../assets/images/aws-iam-cli.png)
+
+[Install AWS CLI From Amazon](https://aws.amazon.com/cli/)
+
+**Steps:**
+1. Make sure AWS Credentials arne set
```
azmy60: `arne` typo

apps/studio/src-commercial/backend/handlers/awsHandlers.ts
```diff
@@ -0,0 +1,43 @@
...
+      proc.on("close", (code) => {
+        if (code === 0) {
+          const profiles = stdout.trim().split("\n");
+          resolve(profiles);
+        } else {
+          rawLog.error(`AWS CLI failed with code ${code}`);
+          reject(
+            `aws failed (code ${code})\nSTDERR: ${stderr}\nSTDOUT: ${stdout}`
+          );
+        }
```
azmy60, line 39: Not something we have to do now, but we might want to have a class helper that handles all of these. There are 4 codes in total that are similar to this.

apps/studio/src/components/LostConnectionModal.vue
azmy60: Yeah, it's probably a mistake?

apps/studio/src-commercial/backend/handlers/awsHandlers.ts
not-night-but: Yeah I agree. In general I don't love this

apps/studio/src/components/LostConnectionModal.vue
not-night-but: Oh I think this is from when I was debugging another issue, but afaik we don't really use this modal right now. I'll revert this change, I think the LostConnectionModal was created ages ago when I was experimenting with ways to determine when a connection died

PR #2895, opened 2025-03-17
feat: Add ctrl+pgup and ctrl+pgdown tab change

Jvillani-fend:
I added the two shortcut ctrl+pgup and ctrl+pgdown to navigate between tabs like browsers. 

I'll give you some info:

I needed to add keycodes on renderer.ts as you can see, because pg up and down was not supported in the library despite in mapping file in v-hotkey github were mapped, they didn't work without adding on renderer

**This part is not about this PR let me know if i have to open an issue**    
Other consideration, i needed to modify the start command on windows using nvm, because it had problems finding yarn path, here the modified start command:
    "electron:serve": "concurrently -c blue,green,yellow -n esbuild,vite,electron \"npx yarn dev:esbuild\" \"npx yarn dev:vite\" \"npx electron .\"",
 
 With npx worked fine, my colleague tried to start it on linux with defatult command and everything was fine.


not-night-but:
@Jvillani-fend I believe you just need to make sure you have installed yarn globally in the nvm environment. So go into the beekeeper directory, `nvm use`, and then `npm install -g yarn`.

apps/studio/src/components/CoreTabs.vue
```diff
@@ -401,6 +401,8 @@ import { TransportOpenTab, setFilters, matches, duplicate } from '@/common/trans
       const result = this.$vHotkeyKeymap({
         'tab.nextTab': this.nextTab,
         'tab.previousTab': this.previousTab,
+        "ctrl+pageup": this.previousTab,
+        "ctrl+pagedown": this.nextTab,
```
not-night-but, line 405: This isn't how we do things anymore since our ini config system was merged. Very easy to fix though! Just take a look at `apps/studio/default.config.ini` for the keybinds for tabs (under the `[keybindings.tab]` header). You can see how to add multiple keybindings for the same action at the refresh action in general, or the focusSearch action in quickSearch

not-night-but (changes requested):

apps/studio/src/components/CoreTabs.vue
not-night-but: (if it wasn't clear, you just need to remove these lines, the other bindings here will handle multiple bindings from the config file properly)

not-night-but (approved):

Issue #2898, opened 2025-03-17
BUG: Can't add rows to tables that have JSON columns


JeanPaiva42:
**Bug description:**
- Users can't manually add a row to a table when there is a JSON column on it.

**Steps to reproduce the behavior:**

1. On a table that contains a JSON column.
2. Click on "Add Row"
3. Notice that no row is added to the table.
4. See error on the console.

**Expected behavior**
- A row should always be added to the table when "Add Row" button is clicked.

**Screenshots / Gifs**

https://github.com/user-attachments/assets/04394fa6-2762-4247-b733-40d778d73f6b

**Version Information (please complete the following information):**
 - OS: Mac
 - App Version: 5.1.5.beta.1
 - Database type: Posgresql

PR #2891, opened 2025-03-15
Fix #2882: BUG: "Add row" button raises an error on a table with a JSON column #2882

tapiwakundi:
Added an undefined check before calling `friendlyJsonObject`

Issue #2882, opened 2025-03-14
BUG: "Add row" button raises an error on a table with a JSON column


franquis:
**Describe the bug**
When I click on the '**Add row**' button (on the bottom right side of the main window) to insert a new row into a table that contains a JSON column, the application raises an error.

```
TypeError: Object.defineProperties called on non-object
    at Function.defineProperties (<anonymous>)
    at friendlyJsonObject (index-D8UEDnXV.js:3267:225063)
    at p.jsonMutator (index-D8UEDnXV.js:3454:12183)
    at Object.<anonymous> (index-D8UEDnXV.js:3454:11260)
    at index-D8UEDnXV.js:3171:38456
    at index-D8UEDnXV.js:3182:29181
    at Array.forEach (<anonymous>)
    at ColumnManager.traverse (index-D8UEDnXV.js:3182:29165)
    at tc.transformRow (index-D8UEDnXV.js:3171:38194)
    at tc.rowDataChanged (index-D8UEDnXV.js:3171:37583)
```

**To Reproduce**

1. Create a new table with a JSON column

```sql
create table
  "public"."test" (
    "id" serial not null,
    "created_at" timestamp not null default NOW(),
    "data" JSON null,
    constraint "test_pkey" primary key ("id")
  )
```

2. Open the table using the left nav
3. Click on the '**Add row**' button 
4. See error in the Developer Tools

**Expected behavior**
I expect the application add a new row in the current table.

**Screenshots / Gifs**
<img width="1360" alt="Image" src="https://github.com/user-attachments/assets/db6acbb6-f278-4371-82b7-0627b4d0b34b" />

**Version Information (please complete the following information):**
 - OS: macOS 15.3.1
 - App Version : 5.1.5
 - Database type : Postgresql

EnergoKissel:
Same, new row stopped working on any table with json/jsonb columns

OS: macOS 15.3.1, windows 10
App Version : 5.1.5
Database type : Postgresql

gus-fabacus:
I am also experiencing the same issue, as reported.

OS: Ubuntu 22.04.5
App Version (snap): 5.1.5
Database type : Postgresql

tapiwakundi:
I'll investigate this

tapiwakundi:
@rathboma can you assign this to me?

rathboma:
Sure!

PR #2899, opened 2025-03-18
fix: add and clone row not working with tables with json column

therealrinku:
fixes: #2898 

Add row and Clone row both weren't working on tables with jsonb columns previously. 

before:
https://github.com/user-attachments/assets/67070416-9188-418a-b01f-4b459052b636

after:
https://github.com/user-attachments/assets/81e14b01-fc05-4413-abf8-03aeb801df51


EnergoKissel:
I believe this also relates to issue #2882 and pr #2891

not-night-but (approved):

PR #2911, opened 2025-03-21
Feature/2157 import can create tables

### LINKED ISSUE #2157: FEAT: Import should be able to create a new table

wmontgomery:
resolves #2157 
feature-2157


apps/studio/src/shared/lib/dialects/clickhouse.ts
```diff
@@ -74,6 +74,14 @@ export const ClickHouseData: DialectData = {
     const matched = value.match(UNWRAPPER);
     return matched ? matched[1] : value;
   },
+  importDataType: {
+    stringType: 'String',
```
rathboma, line 78: You'll probably need to check the length of the strings, if > 255 I'd make this a TEXT rather than a varchar. Also what varchar length do you use by default? That should probably be explicit

apps/studio/src/store/modules/LicenseModule.ts
```diff
@@ -53,14 +53,17 @@ export const LicenseModule: Module<State, RootState>  = {
       return state.licenses.length === 0
     },
     isUltimate(state) {
+      // return true // for some reason on macos, the switching license type doesn't work so this will do for now...
```
rathboma, line 56: remove comments? Sounds like the menu may be broken? Worth fixing?

apps/studio/tests/unit/lib/import/import.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  beforeEach(() => {
+    const mockOptions = {
+      trimWhitespaces: true,
+      nullableValues: ['NULL', 'NONE', 'EMPTY']
+    }
+    const mockConnection = {}
+    importInstance = new ImportClass('test.csv', mockOptions, mockConnection)
+  })
+
+  describe('checkDataType', () => {
```
rathboma, line 15: Thanks for adding some tests!

rathboma: I'd ask claude to help you write more if there are more tests you want to add but don't have time

rathboma (dismissed):
Will this looks awesome. I have a fairly minor change suggestion around the type detection. Oh, it's probably also worth trying to detect JSON objects or arrays, for most DBs that will map to TEXT, for postgres it should map to JSONB

rathboma:
If you wanna make a new ticket for JSON support that's fine, do it separately after merging this.

apps/studio/src/shared/lib/dialects/clickhouse.ts
wmontgomery: I can add a length check to determine which is which. I'll add another "string" type so I can explicitly say which one to use based on connection type. I use the varchar default per connection, but I can go back and get more specific. It's different per connection

apps/studio/src/store/modules/LicenseModule.ts
wmontgomery: I was leaving that there because I've accidentally committed "return true" more often than I care to admit :D 

I wasn't sure if it was worth fixing. I've looked into it a couple times and it didn't feel a "juice worth the squeeze" cliche especially since it was only on mac as far as I could tell.

rathboma:
Yeah it really shouldn't create the table to the end, and for sure we don't want to force primary key -- maybe there shouldn't even be a primary key on the table? Users should be able to deselect it

rathboma:
Let's do this:
1. Add a search box to the first page so users can find a table by searching
2. The table creation shouldn't happen until they click 'execute' at the end
3. Primary keys should not be null, should be optional, and should be any column

wmontgomery:
> Let's do this:
> 
>     1. Add a search box to the first page so users can find a table by searching
> 
>     2. The table creation shouldn't happen until they click 'execute' at the end
> 
>     3. Primary keys should not be null, should be optional, and should be any column

Sure sounds good to me. I'll move the table search to the first thing, then only create the table at the very end if they want a table. I'll keep primaries out, but if they want one, they can select it. I'll set it up like the table structure deal.

wmontgomery:
@rathboma and @JeanPaiva42 , changes have been made and this is ready for re-review. I updated the gif in the description to match the new layout

wmontgomery:
@JeanPaiva42 and @rathboma , this is ready for testing/code review now. The docs have been updated and merged.

wmontgomery:
Thanks, @JeanPaiva42, for looking through this. The first issue I can look into, but it's not much more than a UI count issue (might even just take it out, who knows what value it even adds outside of not being right sometimes).

For the second, I don't know if we have a radio button for tabulator but I can look into it if the need is necessary. Our current create table allows a user to select multiple columns as primary key (mind you, it does error), so I was following suit there.

3) I agree, but that also happens on the current create table as well. From a UI standpoint, I'm all for saving people from themselves, but it might be a larger discussion from a UX standpoint (nullable and primary become radio buttons on the rows vs checkbox, but then how do you make sure only one primary key, etc?)

4) Oh oops. That one I can for sure look into. I'm going to add a note here to remind me to add if those columns get imported or not (that would be a bad thing if they did). Edit: I can't replicate this one. I ignored a couple columns, went to the preview tab, and they were gone. I remember fixing it in another commit

rathboma:
@wmontgomery 4 should be fixed, 1 should be fixed, the multiple-primary and primary+nullable thing -- let's just fix this in a different PR, it's current behavior and unrelated to this change.

After those fixes, let's merge

JeanPaiva42:
Hey all, just gave it an extra pass:

(Since we agreed last meeting on release this even with some of the known issues, I am just limiting myself on retesting the ones that were previously reported)


So for issue 1, we still somewhat have it. Now the `number of ignored columns` is not a negative number, but it is still incorrect (see video 1)

https://github.com/user-attachments/assets/74058946-041b-4587-aa66-5ef16013bc29


For issue 4 we also still have it. Notice that we choose to ignore `Section` column and `total questions`. `Section` column was still in the preview while `total questions` was not. See video 2:

https://github.com/user-attachments/assets/5ae7abc1-088d-4f34-a2c1-0f109d9684f1

JeanPaiva42:
We are not displaying the ignored column numbers anymore (which of fixes the issue number 1) 

image

Besides that, we are not displaying the ignored columns in the preview, so that takes care of 4.


I believe that we are good to go, considering that we just wanted to fix these 2 problems before the release!

Keep in mind that we still have some bugs, but I am guessing that we should be able to deal with them later.

apps/studio/src/components/TabTableBuilder.vue
```diff
@@ -185,6 +194,10 @@ export default Vue.extend({
         this.running = false
       }
     },
+    async createAndImport() {
```
azmy60, line 197: @wmontgomery It looks like this isn't completed yet?

azmy60 (approved):

PR #2912, opened 2025-03-21
FEAT(i18n): Chinese Localization, Contents with Internationalization Texts

juratjan123:
I am a long-time user of this project. I would like to contribute by providing Chinese localization support and developing internationalization extension components for this project. Currently, I have completed the internationalization of most content on the initial interface, but there are still some parts that need further work. If you find these changes valuable, please review the modifications made.

The key updates in this pull request include:
- Introduction of i18n Plugin: Integrated an i18n (internationalization) plugin into the project to facilitate multi-language support.
- Internationalization of Initial Interface Content: Replacing front-end display texts with localized versions.
- Added Language Switching Button: Added a language switching button on the main interface, allowing users to quickly switch languages.
- Addition of Chinese Language Files: Ensuring all feature descriptions are accurately translated into Chinese.
- Quality Assurance: Through personal testing, no significant bugs or issues were found in the localized sections.

Given the broad scope of these changes, if you believe they add value to the project, I recommend a thorough review to ensure everything functions as expected. Although no issues were identified during my tests, additional reviews are always beneficial.

If these changes meet the project's standards and expectations, I will continue working on the remaining localization tasks, including but not limited to other interfaces, error messages, and documentation.

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -113,6 +113,7 @@
     "vue": "^2.7.16",
     "vue-class-component": "^7.2.3",
     "vue-clipboard2": "^0.3.1",
+    "vue-i18n": "8",
     "vue-js-modal": "^1.3.33",
     "vue-property-decorator": "^8.4.2",
     "vue-select": "^3.20.0",
--- a/apps/studio/src-commercial/entrypoints/renderer.ts
+++ b/apps/studio/src-commercial/entrypoints/renderer.ts
@@ -31,6 +31,7 @@ import { UtilityConnection } from '@/lib/utility/UtilityConnection'
 import { VueKeyboardTrapDirectivePlugin } from '@pdanpdan/vue-keyboard-trap';
 import App from '@/App.vue'
 import { ForeignCacheTabulatorModule } from '@/plugins/ForeignCacheTabulatorModule'
+import i18n, { setI18nLanguage } from '@/i18n'
 
 (async () => {
 
@@ -138,6 +139,7 @@ import { ForeignCacheTabulatorModule } from '@/plugins/ForeignCacheTabulatorModu
     const app = new Vue({
       render: h => h(App),
       store,
+      i18n,
     })
 
     Vue.prototype.$util = new UtilityConnection();
@@ -149,7 +151,17 @@ import { ForeignCacheTabulatorModule } from '@/plugins/ForeignCacheTabulatorModu
         log.log('Received port in renderer with sId: ', sId);
 
         Vue.prototype.$util.setPort(port, sId);
-        app.$store.dispatch('settings/initializeSettings');
+        app.$store.dispatch('settings/initializeSettings').then(() => {
+          // After initialization, check language settings and apply
+          // Only apply saved settings if the user has manually set the language
+          const savedLanguageSetting = app.$store.state.settings.settings.language
+          if (savedLanguageSetting && savedLanguageSetting.value) {
+            // Check if the user has manually set the language (contains user value)
+            if (savedLanguageSetting._userValue) {
+              setI18nLanguage(savedLanguageSetting.value)
+            }
+          }
+        });
       }
     }
 
--- a/apps/studio/src/components/ConnectionInterface.vue
+++ b/apps/studio/src/components/ConnectionInterface.vue
@@ -14,30 +14,30 @@
           <div class="card-flat padding" :class="determineLabelColor" v-else>
             <div class="flex flex-between">
               <h3 class="card-title" v-if="!pageTitle">
-                New Connection
+                {{ $t('connection.newConnection') }}
               </h3>
               <h3 class="card-title" v-if="pageTitle">
                 {{ pageTitle }}
               </h3>
               <ImportButton :config="config">
-                Import from URL
+                {{ $t('connection.importFromUrl') }}
               </ImportButton>
             </div>
-            <error-alert :error="errors" title="Please fix the following errors" />
+            <error-alert :error="errors" :title="$t('connection.pleaseFixErrors')" />
             <form @action="submit" v-if="config">
               <div class="form-group">
-                <label for="connection-select">Connection Type</label>
+                <label for="connection-select">{{ $t('connection.connectionType') }}</label>
                 <select name="connectionType" class="form-control custom-select" v-model="config.connectionType"
                         id="connection-select"
                 >
                   <option disabled hidden value="null">
-                    Select a connection type...
+                    {{ $t('connection.selectConnectionType') }}
                   </option>
                   <option :key="`${t.value}-${t.name}`" v-for="t in communityConnectionTypes" :value="t.value">
-                    {{ t.name }}
+                    {{ $t(`database.${t.value}`) }}
                   </option>
                   <option :key="`${t.value}-${t.name}`" :value="t.value" v-for="t in ultimateConnectionTypes">
-                    {{ t.name }}
+                    {{ $t(`database.${t.value}`) }}
                   </option>
                 </select>
               </div>
@@ -85,8 +85,8 @@
                     <input :disabled="!isUltimate" class="form-control" id="readOnlyMode" type="checkbox" name="readOnlyMode"
                            v-model="config.readOnlyMode"
                     >
-                    <span>Read Only Mode</span>
-                    <i v-if="!isUltimate" v-tooltip="'Upgrade to use Read Only Mode'" class="material-icons">stars</i>
+                    <span>{{ $t('connection.readOnlyMode') }}</span>
+                    <i v-if="!isUltimate" v-tooltip="$t('connection.upgradeForReadOnly')" class="material-icons">stars</i>
                     <!-- <i class="material-icons" v-tooltip="'Limited to '">help_outlined</i> -->
                   </label>
                 </div>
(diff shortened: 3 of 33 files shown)


juratjan123:
Hi @rathboma , Sorry to bother. Could you please take a look at this PR when you have a moment? It adds Chinese localization support and lays the groundwork for internationalization. Looking forward to your feedback! Thanks!

juratjan123:
Hi @not-night-but ,

Thank you very much for your suggestions. Based on your feedback, I have proposed an improved solution: in the frontend, we can directly use hardcoded content as keys, for example, {{ $t('Import from URL') }}. When the user's language setting is English (en), the key itself will be returned directly; if another language is selected, the corresponding translation value will be returned according to the relevant JSON file. This way, when writing Vue files, all that's needed is to wrap originally hardcoded English text with $t("...").

Regarding Git Hooks, as I understand, it can automatically identify $t("xxx") formatted content within the code and check whether the corresponding translation exists in the JSON file. If it does not exist, the hook can perform the necessary actions such as adding or prompting for translation.

While machine translations offer significant convenience and automation, I'm worried that the resulting texts might appear stiff and lack a natural flow. If there is a need for Chinese localization, I am more than happy to help. Besides, how do you feel about the feasibility and sustainability of this solution? If you think it’s acceptable, I can revert this PR, then apply the new approach to one or two Vue files for testing to see the results.

not-night-but:
I don't think we want to use any machine translations (I'm assuming you mean google translate or something like that). I am well acquainted with the fact that most of the time google translate and similar programs are typically quite terrible.

This in general will have to be something that we rely on the community for (specifically people like yourself). We do have a couple multi-lingual people on the team, but we obviously don't have every possible language. 

If you're up to work more on this, I'd love if you can get a minimal proof of concept working so the rest of the team can see how maintainable it will be. Let me know if you want any help with this as well! I can do some work getting the git hooks set up to see if that might be a feasible solution.

juratjan123:
Hi @not-night-but,

I've made these changes. The main improvements include:

- Adjusted the layout to place the Language Switch button above the Account button.
- Performed internationalization on some hard-coded content within Vue files such as `mysqlForum` and `commonServerInput`, wrapping them with `$t("xxx")`.
- The localization logic is as follows: The default language is set to English. Under this setting, hard-coded English content in the frontend serves as the translation keys. When the user switches languages, the system retrieves the corresponding key-value pairs from the respective JSON files for display. If no language switch occurs, the original hard-coded content will be displayed by default.

not-night-but:
Great! This definitely looks like it will be more manageable for day-to-day development. A couple questions:
- How does this work with interpolations like dates? Can we still use them like in your last iteration?
- How does this affect memory? Seems like we're loading all the strings at once, so if we have multiple languages would we end up with large memory usage? Is it possible for us to load the files dynamically?

not-night-but:
@juratjan123 I just pushed a precommit hook that will extract any new keys that are added in changes. You can manually run it by running bin/extract-translations.js. Let me know if this is helpful or not (for it to run before commits you'll need to run a yarn install).

juratjan123:
Hi @not-night-but !

Of course, you can directly use interpolation like this:
```html
<template>
  <div>{{ $t('Welcome, {name}!', {name: username}) }}</div>
</template>

<script>
export default {
  data() {
    return {
      username: 'Test'
    }
  }
}
</script>
```
And the json file might be:
```json
{
  "Welcome, {name}!": "欢迎，{name}！"
}
```

When the language is set to English, it will directly return the key itself, and the i18n plugin will automatically handle the interpolation. For other languages, it retrieves the corresponding value from the JSON file and automatically handles the interpolation.

Regarding the second issue, the current vue-i18n solution does indeed load all languages into memory at once (except for English, as it uses its own keys directly). The risk is that memory usage will increase linearly with the number of languages and translation strings. However, vue-i18n supports dynamic language loading, meaning only the languages needed by the user are loaded. This should significantly reduce memory usage.

Also, thanks for the new file! I think it will be very useful!

not-night-but:
Hey @juratjan123 !

Well then it looks like things are moving along! If you're up for it, I'd love to see more of the app translated. Let me know if you have any questions. I think we should probably move to dynamically loading the languages just to be safe, as ideally we will start getting more languages once we finish the bulk of the work here. You up for it?

juratjan123:
Hi @not-night-but. 
Sure, let me start translating the Vue files step by step. However, it might take a considerable amount of time since there are quite a lot of frontend files. lol. But I will do my best to adjust them gradually. I believe the outcome will be great.

juratjan123:
Hi @not-night-but ,

I have completed the localization of all frontend components, but I haven't had the time to refine the translations yet. With the upcoming holiday, we will be off for three days, so I won't be able to continue this work during that period. Will enjoy the break, lol! Upon my return, I will resume the localization efforts. In the meantime, I welcome any reviews or feedback on the code. I hope to contribute meaningfully to BKS with these updates!

Thank you!

apps/studio/src/App.vue
```diff
@@ -175,20 +178,32 @@ export default Vue.extend({
     notifyFreeTrial() {
       Noty.closeAll('trial')
       if (this.isTrial && this.isUltimate) {
-        const ta = new TimeAgo('en-US')
         const validUntil = this.status.license.validUntil
+        const dateStr = this.$bks.getTimeAgo().format(validUntil)
+        const localDateStr = validUntil.toLocaleDateString()
+        
+        let text = ''
+        if (this.$i18n.locale === 'en') {
```
Architrixs, line 186: hey @juratjan123
I'm not very familiar with Vue or this codebase yet, but I noticed that `fallbackLocale` is already being used in Vue I18n. Would it be possible to rely on it here and remove the `if-else` check, so `$t()` handles the default English text automatically? Just curious and eager to learn more!

July-Tea: Thank you for your suggestion! I understand that this piece of code could be expressed in a more concise way. Actually, my intention was to make the code easier to read and understand at a glance.(Just personal coding habit.)

jinzhongjia:
Hey, is there any further progress on this?

yunjintw:
Hi @July-Tea and @not-night-but,
I'm a long-time database tool user from Taiwan, and I'd love to contribute Traditional Chinese (zh-TW) localization to this project.
I noticed this PR has been quiet since April — is there anything blocking the merge, or any way I can help move it forward?
What I can offer:

Create and maintain zh-TW.json translation file
Review existing zh-CN.json for quality and consistency
Help test the i18n implementation on different platforms

Traditional Chinese is used by ~30 million people in Taiwan, Hong Kong, and overseas Chinese communities. Having both zh-CN and zh-TW would significantly expand Beekeeper Studio's reach in the Chinese-speaking market.
I'm happy to start working on the zh-TW translation once the i18n architecture is merged, or I can fork from @July-Tea's branch and add it there if that's more helpful.
Looking forward to your thoughts!

not-night-but:
@yunjintw I really think localisation would be huge for the app, but I really worry about maintainability going forward. Not only of the translations, but of the app as a whole. I guess saying that, does anyone in this thread have experience with localizing apps? If we were to move forward with this PR are there any best practices we should be following? I just don't want to merge these changes to have it abandoned and then make the experience worse for everyone if that makes sense, so I want to be cautious. That's my perspective at least

yunjintw:
> @yunjintw I really think localisation would be huge for the app, but I really worry about maintainability going forward. Not only of the translations, but of the app as a whole. I guess saying that, does anyone in this thread have experience with localizing apps? If we were to move forward with this PR are there any best practices we should be following? I just don't want to merge these changes to have it abandoned and then make the experience worse for everyone if that makes sense, so I want to be cautious. That's my perspective at least

Hi @not-night-but,
I understand your concern — I can't promise long-term maintenance either. But I can help create the initial zh-TW translation because I want to use it myself.
Here's a suggestion for a flexible approach:
Concept:

Allow users to load custom language files — Users download a translation file (e.g. zh-TW.json) and place it in a designated folder like ~/.config/beekeeper-studio/locales/. After restart, the new language option appears.
Fallback to English — If a string isn't translated, just show English. The app won't break.
Let the community grow organically — If there's demand, contributors will show up. No need for the team to maintain every language.

Technical approach (should work with current vue-i18n setup):

On app startup, scan the user locale folder for any .json files
For each file found, call i18n.setLocaleMessage('zh-TW', content) to register it
The language switcher automatically picks up new locales from $i18n.availableLocales
fallbackLocale: 'en' already handles missing translations

This way, the core team doesn't take on the burden of maintaining translations. The community handles it if there's real demand.
What do you think?

not-night-but:
@yunjintw It will have to be a community effort no matter what. I believe I wrote a script to extract new keys to make development a bit easier for us. I'll look into merging main back into this branch so we can at least get an idea of what would be required now.

not-night-but:
@yunjintw Okay so I'm just finishing merging main back into this branch, and I've also brought up some things with the team, so there is a lot that needs to be done. I don't love how the keys are set up now (I think I'm the one that originally suggested this method, so my bad lol). I think we need to move to a more conventional key setup (ie `{ "filterPlaceholder": "Filter keys by text or /regex/"}` instead of what we have right now which is the keys match the value). Along with that, a full scale review of the whole app will need to be done as almost 2000 commits have been merged since this PR was last updated. I almost think it would be better if we cherry picked the vue-i18n setup code into a brand new branch and just started from scratch. Luckily from there it's not super complex work, just a bunch of menial find and replace. If there's anyone willing to take on that work, great! Otherwise we'll have to find some time to do this ourselves

rathboma:
@July-Tea I'm sorry for the late response here. I think your work here is very thorough and I certainly want to be able to provide localization to users. You've done a great job to implement it.

The reason we haven't merged this, despite being enthusiastic about what it provides, is the difficulty in managing translations and how abstract it makes the text in the components, so instead of seeing english, we see a lookup string. It sounds small, but it makes it very difficult to develop a user interface.

I know this is probably not what you want to hear, but as-is I don't think we can merge this PR.

I do think there's a LOT of great work here, so I'd like to use this as a base to implement localization in a future PR.

Thanks again, and sorry :-/

PR #2914, opened 2025-03-21
Mongo Shell

not-night-but:
This PR adds a mongosh shell integration to the app. It uses mongo's open source libraries, specifically the electron-runtime and the node-driver-service-provider. This allows users to use mongosh in the app while still having access to the query result table, autocomplete, and some other features including json folding and error reporting. There is probably a lot more that we can do with this, but I think this is a good first step.
image
image
image


apps/studio/src-commercial/backend/lib/db/clients/mongodb.ts
```diff
@@ -595,29 +604,81 @@ export class MongoDBClient extends BasicDatabaseClient<QueryResult> {
...
+      }
+      results.push({
+        rows: ev.printable,
+        rowCount: ev.printable?.length,
+        fields,
+        tableName: ev.source?.namespace?.collection ?? 'mycollection',
+        command: queryText
+      })
+    }
+    log.info("RESULTS: ", results)
```
rathboma, line 679: probably we want this to be dubug.

Any additional tests that we need given the extra code added here?

apps/studio/src/common/globals.ts
```diff
@@ -10,6 +10,7 @@ export default {
   maxColumnWidth: 1000,
```
rathboma, line 10: Don't we have the config system now? Probably this should be moved there

apps/studio/src/components/TabShell.vue
```diff
@@ -0,0 +1,471 @@
...
+  <div
+    class="query-editor"
+    ref="container"
+    v-hotkey="keymap"
+  >
+    <div
+      class="top-panel"
+      ref="topPanel"
+    >
+      <mongo-shell
```
rathboma, line 11: Any way to limit the characters in the prompt? I can imagine users having super=duper-long hostnames and database names, then the prompt is the whole width of the screen

apps/studio/src/components/TabShell.vue
```diff
@@ -0,0 +1,471 @@
...
+  import rawlog from '@bksLogger'
+  import ErrorAlert from './common/ErrorAlert.vue'
+  import MergeManager from '@/components/editor/MergeManager.vue'
+  import { AppEvent } from '@/common/AppEvent'
+  import { PropType } from 'vue'
+  import { TransportOpenTab } from '@/common/transport/TransportOpenTab'
+
+  const log = rawlog.scope('query-editor')
+
+  export default {
```
rathboma, line 120: If using typescript you should use `Vue.extend`, in order to get more typings

apps/studio/src/components/TabShell.vue
```diff
@@ -0,0 +1,471 @@
...
+      },
+      unsavedChanges() {
+        this.tab.unsavedChanges = this.unsavedChanges
+      },
+      async active() {
+        if (!this.editor.initialized) {
+          return
+        }
+
+        // HACK: we couldn't focus the editor immediately each time the tab is
```
rathboma, line 260: `nextTick` doesn't work?

apps/studio/src/plugins/CMMongoHint.js
```diff
@@ -0,0 +1,29 @@
+import CodeMirror from "codemirror";
+(function(mod) {
+  mod(CodeMirror);
+})(function(CodeMirror) {
+  "use strict";
+  var Pos = CodeMirror.Pos;
+
+  CodeMirror.registerHelper("hint", "mongo", async function(editor, options) {
```
rathboma, line 8: Any tests for this file? Feels like it could be tested without codemirror

rathboma (changes requested):
Looks awesome. I think this is 99% totally fine -- had some minor suggestions, and slightly larger question about what should be tested here

apps/studio/src/components/TabShell.vue
not-night-but: No idea, that comment came from the sqltexeditor I believe

rathboma (approved):

PR #2925, opened 2025-03-23
Fix #2695: Unable to Export Queries from Saved Queries

### LINKED ISSUE #2695: BUG: Unable to Export Queries from Saved Queries
**Describe the bug**

Users are not able to use the "Export" button when accessing the Saved Query list.

**To Reproduce**
1. Go to Saved Queries
2. Right click on any of the saved queries
3. Select Rename
4. Notice that a modal will show up, and you can if needed Rename the query
5. Now after the right click, select "Export".
6. Notice that no feedback is provided to the user, and the query is not exported

**Expected behavior**
- Users should be able to export the query from user flow. 

- If the export is not allowed from this point, some sort of feedback should be provided to the users.

**Screenshots / Gifs**

- The video below provides a visual guide for the repro steps. Initially, a rename on the query is made simply to show that the right click actions should be working as expected. After that I proceed to try to export the query.

https://github.com/user-attachments/assets/36af3103-6581-44cf-81b8-3af300e43e01

 - OS: Mac
 - App Version: 5.0.9
 - Database type and version: sqlite

**Additional context**
- This issue was reproduced consistentlywith two different sqlite databases, and was happening even after re-opening the application.

tcctr:
Fix #2695: Changed the way defaultPath was given a name by removing a function that was not returning anything (getLastExportPath) and was breaking the code. Also added a test case. Now the saved queries can be exported correctly.

### DIFF
--- a/apps/studio/src-commercial/entrypoints/preload.ts
+++ b/apps/studio/src-commercial/entrypoints/preload.ts
@@ -86,7 +86,7 @@ export const api = {
 
     return [];
   },
-  async getLastExportPath(filename?: string) {
+  async defaultExportPath(filename?: string) {
     return path.join(homedir(), filename);
   },
   showOpenDialogSync(args: any) {
--- a/apps/studio/src-commercial/entrypoints/preload.ts
+++ b/apps/studio/src-commercial/entrypoints/preload.ts
@@ -87,10 +87,7 @@ export const api = {
     return [];
   },
   async getLastExportPath(filename?: string) {
-    return await SettingsPlugin.get(
-      "lastExportPath",
-      path.join(homedir(), filename)
-    );
+    return path.join(homedir(), filename);
   },
   showOpenDialogSync(args: any) {
     return electron.dialog.showOpenDialogSync(args);
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -277,7 +277,6 @@ import { readWebFile } from '@/common/utils'
 import Noty from 'noty'
 import ConfirmationModal from './common/modals/ConfirmationModal.vue'
 import SqlFilesImportModal from '@/components/common/modals/SqlFilesImportModal.vue'
-import path from 'path';
 
 import { safeSqlFormat as safeFormat } from '@/common/utils';
 import { TransportOpenTab, setFilters, matches, duplicate } from '@/common/transport/TransportOpenTab'
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -277,6 +277,7 @@ import { readWebFile } from '@/common/utils'
 import Noty from 'noty'
 import ConfirmationModal from './common/modals/ConfirmationModal.vue'
 import SqlFilesImportModal from '@/components/common/modals/SqlFilesImportModal.vue'
+import path from 'path';
 
 import { safeSqlFormat as safeFormat } from '@/common/utils';
 import { TransportOpenTab, setFilters, matches, duplicate } from '@/common/transport/TransportOpenTab'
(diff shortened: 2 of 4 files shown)


rathboma (approved):

rathboma:
@not-night-but can you look at this and see if the original function deleted just got moved somewhere? If you're good with it, merge it.

not-night-but:
@rathboma The function that's being called is actually there, I think the issue is that we never actually set the lastExportPath. Are we fine with removing the functionality of remembering where exports go? Cause that's what this would do.

rathboma:
Maybe just fix the broken method?

tcctr:
I will try to fix the method

apps/studio/src/plugins/SettingsPlugin.ts
```diff
@@ -5,13 +5,16 @@ const log = rawLog.scope('settings-plugin')
 
 export const SettingsPlugin = {
 
-  async get(key: string, defaultValue?: any) {
+  async get(key, defaultValue) {
+    if (!Vue.prototype.$util) {
+      return defaultValue || null;
+    }
```
not-night-but, line 11: I don't like doing this, as we should always have access to Vue here. The reason we don't is because in this instance we are calling it from the preload script. So we should probably not call the SettingsPlugin from the preload script, we can probably have a method there for returning the default export path, and then call the settingsplugin method directly from where we were calling `getLastExportPath` last time, if you know what I mean?

not-night-but (changes requested):

apps/studio/src/plugins/SettingsPlugin.ts
tcctr: I think i got it.

rathboma:
@not-night-but can you re-review?

apps/studio/src/plugins/SettingsPlugin.ts
```diff
@@ -5,13 +5,14 @@ const log = rawLog.scope('settings-plugin')
 
 export const SettingsPlugin = {
 
-  async get(key: string, defaultValue?: any) {
+  async get(key, defaultValue) {
     const result = await Vue.prototype.$util.send('appdb/setting/get', { key });
-    log.info("get", JSON.stringify(result))
+    result.value = defaultValue;
+    log.info("get", JSON.stringify(result));
```
not-night-but: We shouldn't set the value like this, as this would get rid of any user value/platformDefault/defaultValue that's configured for that setting. We should only return the defaultValue if there was no result returned from the backend.

apps/studio/src-commercial/entrypoints/preload.ts
```diff
@@ -87,10 +87,7 @@ export const api = {
     return [];
   },
   async getLastExportPath(filename?: string) {
-    return await SettingsPlugin.get(
-      "lastExportPath",
-      path.join(homedir(), filename)
-    );
+    return path.join(homedir(), filename);
```
not-night-but: Really stupid small thing here. We aren't getting the lastExportPath here anymore, we're just getting the defaultExportPath. I think we should rename this to reflect that

not-night-but (changes requested):
Sorry for the wait, just two small things

apps/studio/src-commercial/entrypoints/preload.ts
tcctr: No problem! I just pushed the changes you asked for.

apps/studio/src/components/CoreTabs.vue
```diff
@@ -277,6 +277,7 @@ import { readWebFile } from '@/common/utils'
 import Noty from 'noty'
 import ConfirmationModal from './common/modals/ConfirmationModal.vue'
 import SqlFilesImportModal from '@/components/common/modals/SqlFilesImportModal.vue'
+import path from 'path';
```
not-night-but: I'm so sorry, I missed this. We can't import any node libraries in the frontend unfortunately for security (that's what preload is for). Remove this import and we can get this merged. Sorry again lol

tcctr: Done.
No problem at all, i did not know about that. Thanks for your patience!

apps/studio/src/components/CoreTabs.vue
```diff
@@ -833,11 +833,14 @@ import { TransportOpenTab, setFilters, matches, duplicate } from '@/common/trans
       }
     },
     async handlePromptQueryExport(query) {
-      const safeFilename = query.title.replace(/[/\\?%*:|"<>]/g, '_')
+      const safeFilename = query.title.replace(/[/\\?%*:|"<>]/g, '_');
+      const fileName = `${safeFilename}.sql`;
+
+      const defaultExportPath = await Vue.prototype.$settings.get("defaultExportPath", await window.main.defaultExportPath(fileName));
```
not-night-but: Okay last thing I swear. When I said change it defaultExportPath I meant just the function in preload. i believe the setting in the db is still called lastExportPath

not-night-but (changes requested):

apps/studio/src/components/CoreTabs.vue
tcctr: Done

tcctr:
Just pinging this thread so we can close the issue. If there is any aditional changes needed, i am happy to help.

not-night-but (approved):
Sorry this took a while, thanks for this!

PR #2926, opened 2025-03-23
Fix #2697: UI Issues due to long connection names

### LINKED ISSUE #2697: BUG: UI Issues due to long connection names
**Describe the bug**
Due to the fact that we are not limiting the size of connection names, if a user decides to add a very long string on this field, certain UI aspects can break.

 
**To Reproduce**
1. Go to `Edit Connection`
2. On the modal that appeared, enter a new connection name, make it very long.
3. Play around on the UI (opening and closing tabs), up until the new name is reflected on the screen.
4. Noticed how it glitches the header of the screen.
6. Disconnect from this DB, and look at your `Connection List`
7. Select the connection with the very long name.
8. Notice how the modal is glitched to support it's name.

**Expected behavior**

**Screenshots / Gifs**
- We can see the issues happening on the two videos below. On the first one the glich appears after editing the name, and on the second one, we can already see it from the `Connection Selection` Modal.

https://github.com/user-attachments/assets/11e4520a-eea2-44aa-994e-b947af9518db

https://github.com/user-attachments/assets/72be2479-fe82-4da8-bf0f-4eb4fee97b65

 - OS: Mac (air, 13'')
 - App Version 5.0.9
 - Database type and version SQLite
 
**Additional context**

- I understand that this issue might sound very edge case (and it is), but the fact that the screen can become non-interactable because of a connection name might be something to be concerned.

T-MSD:
When the connected database name is excessively large, the window title overflows into the rest of the application. The fix
truncates the window title when the database name is longer than 30 characters.


rathboma:
@azmy60 can you review this? Could this have just been done with CSS `overflow`?

azmy60 (changes requested):
I agree with @rathboma. It looks like this could be done simply by using css. I would play around with `max-width`, `overflow` and/or `text-overflow`.

It'd be nicer if you we could add a little bit of `margin-left` and `margin-right` on the text too so it wouldn't touch the other elements on each side, giving some space to breathe.

T-MSD:
That was my first approach. It works, but with overflow the end of the window title will always be clipped, meaning that ' - Beekeeper Studio' won't be displayed (with excessively large database names). Should I keep it that way?
Thank you for your feedback.

azmy60:
Yeah, I think that having the name "Beekeeper Studio" truncated in result to that is fine.

azmy60 (approved):
LGTM!

PR #2931, opened 2025-03-25
Use UI Kit SQL Text Editor in the main app

azmy60:


apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -479,43 +493,6 @@
       result() {
         return this.results[this.selectedResult]
       },
-      individualQueries() {
```
azmy60, line 482: @not-night-but I removed a bunch of queries related computations because it's being done inside the SQL Text Editor now. I feel like this shouldn't affect the new JavaScript Text Editor right?

rathboma (approved):

rathboma:
Merge when ready

PR #2934, opened 2025-03-25
Feat/streamer mode

### LINKED ISSUE #2856: FEAT: Streamer Mode
I'm always frustrated when streamers can't use great tools because they lack a streamer mode.
(and when I can't screenshare)

View -> Streamer Mode: Hides IP in a "saved" block and connection settings.
Also, it would be great to have a toggle icon here:
Image

It can be a paid feature.

RuiSoares333:
Closes #2856 

The motivation for contributing to an open source project came from a faculty course unit, and in the scope of that I have detailed all the findings and the fix solution [here](https://github.com/thePeras/beekeeper-studio/wiki/Issue-%232856).
On the details section I layed out our Wiki documentation on this specific Issue.

Fell free to suggest improvements.
I think I implemented all that the issuer wanted, but something could have slipped.

<details>

# Issue and Requirements

Issue [#2856](https://github.com/beekeeper-studio/beekeeper-studio/issues/2856) describes a feature known as "Streamer Mode", but we'll name it **"Privacy Mode"**, because it isn't just made for streamers.

This is a common feature in multiple applications, more in particular, applications that contain sensitive data. This feature aims to hide information for when the user is streaming/screensharing.

The reporter suggested that a toggle icon for enabling/disabling the Privacy Mode would also be an ergonomic way of using this feature.

| Privacy Mode Toggle Button |
| - |
| 420608304-4ba7d49d-f49a-4863-b267-9217a5f98898|
| Fig.1 User Print 1 |

The reporter also shared the specific fields that they wanted hidden, namely:

| Host / Port / DB in Saved Connections | The pop-up with the full URL on Hover |
| - | - |
| 421115860-eb9ab99a-c50e-41d1-8a8b-a24295171219 | 421115885-5b5ae907-6f51-4193-a668-c14be7cf7a9c |
| Fig.2 User Print 2 | Fig.3 User Print 3 |

| Host / Port / DB in Connection Settings |  URL when hovering over the DB name after connecting |
| - | - |
| 421115905-1838c709-ff2c-4a99-bb3a-15c7c1cde6d4 | 421115915-1875bcee-971c-4245-a4b1-6f0a0c23b6af |
| Fig.4 User Print 4 | Fig.5 User Print 5 |

Expected behavior: When activating Privacy Mode, the above specified fields should become hidden.

This issue probably won't depend on the operating system, but it could be important to specify that I am using Windows 10.

# Source Code Files

The files related to the issue are:

| File | Description |
| ---- | ----------- |
| ConnectionSidebar.vue | A Vue.js component that renders a sidebar interface for managing database connections. |
| ConnectionListItem.vue | A Vue.js component that renders a clickable list item for a database connection, displaying connection details. |
| CommonServerInputs.vue | A Vue.js component that provides a form for configuring database connection details. |
| ConnectionButton.vue | A Vue.js component that displays a connection button in the UI that shows the current database connection details. |
| MaskedInput.vue | A Vue.js component that was created in order to simplify the censored fields in the CommonServerInputs. |
| SettingsStoreModule.ts | A Vuex store module that manages user settings. |

# Design of the Fix

While trying to find a solution for this feature, I just needed to use the selecion tool provided by the 'Inspect'-like feature on the right side of the BeekeeperStudio app. This allowed me to find where were the components that needed to be hidden by the Privacy Mode somewhat quickly.

Note: The Inspect is really useful from a maintenance POV, well done.

The Privacy Mode feature is expected to work like described in the following Sequence Diagram from Figure 6: 

The user clicks the Privacy Mode Icon that is located on the Sidebar, which will update the privacyMode variable to the new value (true/false). Then, the SettingsStore notifies all components and the UI re-renders the affected fields.

| Expected Sequence Diagram |
|-|
| Privacy Mode Sequence Diagram |
| Fig.6 Privacy Mode Sequence Diagram |

The following State Diagram show sthe binary toggle between modes. The notes in the diagram show an example of what the final result of the Privacy Mode should look like.

| Expected State Diagram |
|-|
| Privacy Mode State Diagram |
| Fig.7 Privacy Mode State Diagram |

As I am not 100% familiar with Vue, in the beggining I was trying to propagate the privacyMode variable value from the ConnectionSidebar.vue file to the other components. However, I ended up realizing that using the Store feature from Vue would make it way simpler.

# Update Source Code

## ConnectionSidebar.vue

```jsx
<div class="actions">
    <a 
        @click="togglePrivacyMode" 
        :title="privacyMode ? 'Disable Privacy Mode' : 'Enable Privacy Mode'"
    >
        <i class="material-icons">{{ privacyMode ? 'visibility_off' : 'visibility' }}</i>
    </a>
    <a @click.prevent="refresh"><i class="material-icons">refresh</i></a>
</div>
```

This adds the new button for toggling the Privacy Mode.

```jsx
<connection-list-item
  // ...
  :pinned="true"
  :privacy-mode="privacyMode"
  @edit="edit"
  // ...
/>
```

This sends the value of privacy-mode to the ConnectionListItem.

```jsx
  computed: {
    ...mapState('settings', ['privacyMode']),
  }
```
This maps the privacyMode from the Vuex 'settings' module so that the component always reflects it

```jsx
  methods: {
    ...mapActions({
      togglePrivacyMode: 'settings/togglePrivacyMode',
    }),
```
This makes sure that when we're calling the `togglePrivacyMode` function, it's the one from the settings, which changes the original privacyMode variable.

## ConnectionListItem.vue

```jsx
<div class="subtitle">
  <span
    class="bastion"
    v-if="this.config.sshBastionHost && !privacyMode"
```
This will show the subtitle if sshBastionHost exists and if the Privacy Mode is deactivated.

```jsx
    v-if="this.config.sshHost && !privacyMode"
```
The same as the sshBastionHost but with the sshHost.

```jsx
<span class="connection">
  <span>
    {{ privacyMode ? '******' : subtitleSimple }}
```
If the Privacy Mode is On, the subtitle will only show '******', otherwise it will show the subtitle normally.

```jsx
  title() {
    return this.privacyMode ? 
      'Connection details hidden by Privacy Mode' : 
      this.$bks.buildConnectionString(this.config)
  },
```
When Privacy Mode is On, hovering a Connection will show 'Connection details hidden by Privacy Mode, otherwise it will show the Connection String normally.

## CommonServerInputs.vue

```jsx
  <label for="port">Port</label>
  <masked-input
    :value="config.port"
    :privacyMode="privacyMode"
    :type="'number'"
    @input="val => config.port = val"
  />
```
This hides the Port value as '******' when Privacy Mode is On, otherwise it will show normally.

```jsx
  <label for="Host">Host</label>
  <masked-input
    :value="config.host"
    :privacyMode="privacyMode"
    @input="val => config.host = val"
  />
```
This hides the Host value as '******' when Privacy Mode is On, otherwise it will show normally.

```jsx
  <label for="user">User</label>
  <masked-input
    :value="config.username"
    :privacyMode="privacyMode"
    @input="val => config.username = val"
  />
```
This hides the Username value as '******' when Privacy Mode is On, otherwise it will show normally.

## ConnectionButton.vue

```jsx
  <div
    class="connection-button flex flex-middle"
    v-if="config"
    :title="privacyMode ? 
      'Connection details hidden by Privacy Mode' : 
      $bks.buildConnectionString(config)"
  >
```
This hides the connection details while hovering when Privacy Mode is on.

## MaskedInput.vue

This is a new file that handles the repetitive logic of the censored inputs in the CommonServerInputs.

```jsx
<template>
  <input
    :type="computedType"
    :value="displayedValue"
    @input="handleInput"
    :readonly="privacyMode"
    class="form-control"
  >
</template>
```
Defines an input element that binds its type, value, and readonly state to dynamic component properties and handles input events.

```jsx
props: {
    value: {
      type: [String, Number],
      default: ''
    },
    privacyMode: {
      type: Boolean,
      default: false
    },
    type: {
      type: String,
      default: 'text'
    },
    mask: {
      type: String,
      default: '******'
    }
  },
```
Props for value, privacy mode, input type, and mask string.

```jsx
  computed: {
    computedType() {
      return this.privacyMode ? 'text' : this.type;
    },
    displayedValue() {
      return this.privacyMode ? this.mask : this.value;
    }
  },
```
Calculates the actual input type and displayed value based on the privacy mode, switching between provided values and defaults.

```jsx
  methods: {
    handleInput(event) {
      if (!this.privacyMode) {
        this.$emit('input', event.target.value);
      }
    }
  }
```
Implements a handler that emits user input only when privacy mode is disabled, ensuring proper input management.

## SettingStoreModule.ts

```ts
interface State {
  privacyMode: boolean
}
```
Adds the the privacyMode variable (which is a boolean) to the State interface.

```ts
const M = {
  state: () => ({
    privacyMode: false
  }),
```
Adds the variable privacyMode to the context.

```ts
    SET_PRIVACY_MODE(state, value: boolean) {
      state.privacyMode = value;
    }
```
Function that changes the value of privacyMode.

```ts
    if (key === 'privacyMode') {
        const setting = context.state.settings[key] || await Vue.prototype.$util.send('appdb/setting/new');
        setting.key = key;

        if (_.isBoolean(value)) setting.valueType = UserSettingValueType.boolean;
        setValue(setting, value);
        const newSetting = await Vue.prototype.$util.send('appdb/setting/save', { obj: setting });
        _.merge(setting, newSetting);
        context.commit(M.ADD, setting);
    }
```
Checks if the setting key is 'privacyMode', prepares or creates a new setting object, assigns a boolean type if applicable, updates its value, saves it via an API call, merges the result, and commits it to the store.

```ts
    togglePrivacyMode({ commit, state }) {
      const newPrivacyMode = !state.privacyMode;
      commit('SET_PRIVACY_MODE', newPrivacyMode);
    },
```
Toggles the privacyMode state by inverting its current value and committing the updated value to the store.

# Final Result

| Privacy Mode Off | Privacy Mode On |
| - | - |
|image|image |
| Fig.8 Button Off | Fig.9 Button On |

| Hidden Host / Port / DB in Saved Connections | The pop-up with the full URL on Hover |
|-|-|
|image | image|
| Fig.10 - Hidden Feature 1 | Fig.11 - Hidden Feature 2 |

| Host / Port / DB in Connection Settings |  URL when hovering over the DB name after connecting |
| - | - |
|image | image |
| Fig.12 - Hidden Feature 3 | Fig.13 - Hidden Feature 4 |

# Pull Request

Proposed fix submitted in the [#2934](https://github.com/beekeeper-studio/beekeeper-studio/pull/2934) PR
</details>

### DIFF
--- a/apps/studio/src/components/connection/CommonServerInputs.vue
+++ b/apps/studio/src/components/connection/CommonServerInputs.vue
@@ -38,10 +38,12 @@
       <div class="col s3 form-group" v-if="supportsSocketPathWithCustomPort">
         <label for="port">Port</label>
         <input
-          type="number"
+          :type="streamerMode ? 'text' : 'number'"
           class="form-control"
           name="port"
-          v-model.number="config.port"
+          :value="streamerMode ? '******' : config.port"
+          @input="!streamerMode && (config.port = $event.target.value)"
+          :readonly="streamerMode"
         >
       </div>
     </div>
@@ -56,16 +58,20 @@
           class="form-control"
           @paste="onPaste"
           name="host"
-          v-model="config.host"
+          :value="streamerMode ? '******' : config.host"
+          @input="!streamerMode && (config.host = $event.target.value)"
+          :readonly="streamerMode"
         >
       </div>
       <div class="col s3 form-group">
         <label for="port">Port</label>
         <input
-          type="number"
+          :type="streamerMode ? 'text' : 'number'"
           class="form-control"
           name="port"
-          v-model.number="config.port"
+          :value="streamerMode ? '******' : config.port"
+          @input="!streamerMode && (config.port = $event.target.value)"
+          :readonly="streamerMode"
         >
       </div>
     </div>
@@ -173,8 +179,10 @@
         <input
           type="text"
           name="user"
-          v-model="config.username"
+          :value="streamerMode ? '******' : config.username"
+          @input="!streamerMode && (config.username = $event.target.value)"
           class="form-control"
+          :readonly="streamerMode"
         >
       </div>
       <div class="col s6 form-group">
@@ -214,6 +222,7 @@ import FilePicker from '@/components/common/form/FilePicker.vue'
 import ExternalLink from '@/components/common/ExternalLink.vue'
 import { findClient } from '@/lib/db/clients'
 import ToggleFormArea from '../common/ToggleFormArea.vue'
+import { mapGetters } from 'vuex'
 
   export default {
     props: {
@@ -236,6 +245,12 @@ import ToggleFormArea from '../common/ToggleFormArea.vue'
       }
     },
     computed: {
+      ...mapGetters({
+        settings: 'settings/settings'
+      }),
+      streamerMode() {
+        return this.settings?.streamerMode || false
+      },
       hasAdvancedSsl() {
         return this.config.sslCaFile || this.config.sslCertFile || this.config.sslKeyFile
       },
--- a/apps/studio/src/components/connection/CommonServerInputs.vue
+++ b/apps/studio/src/components/connection/CommonServerInputs.vue
@@ -37,14 +37,12 @@
       </div>
       <div class="col s3 form-group" v-if="supportsSocketPathWithCustomPort">
         <label for="port">Port</label>
-        <input
-          :type="privacyMode ? 'text' : 'number'"
-          class="form-control"
-          name="port"
-          :value="privacyMode ? '******' : config.port"
-          @input="!privacyMode && (config.port = $event.target.value)"
-          :readonly="privacyMode"
-        >
+        <masked-input
+          :value="config.port"
+          :privacyMode="privacyMode"
+          :type="'number'"
+          @input="val => config.port = val"
+        />
       </div>
     </div>
     <div
@@ -53,26 +51,20 @@
     >
       <div class="col s9 form-group">
         <label for="Host">Host</label>
-        <input
-          type="text"
-          class="form-control"
-          @paste="onPaste"
-          name="host"
-          :value="privacyMode ? '******' : config.host"
-          @input="!privacyMode && (config.host = $event.target.value)"
-          :readonly="privacyMode"
-        >
+        <masked-input
+          :value="config.host"
+          :privacyMode="privacyMode"
+          @input="val => config.host = val"
+        />
       </div>
       <div class="col s3 form-group">
         <label for="port">Port</label>
-        <input
-          :type="privacyMode ? 'text' : 'number'"
-          class="form-control"
-          name="port"
-          :value="privacyMode ? '******' : config.port"
-          @input="!privacyMode && (config.port = $event.target.value)"
-          :readonly="privacyMode"
-        >
+        <masked-input
+          :value="config.port"
+          :privacyMode="privacyMode"
+          :type="'number'"
+          @input="val => config.port = val"
+        />
       </div>
     </div>
 
@@ -176,14 +168,11 @@
     <div class="row gutter">
       <div class="col s6 form-group">
         <label for="user">User</label>
-        <input
-          type="text"
-          name="user"
-          :value="privacyMode ? '******' : config.username"
-          @input="!privacyMode && (config.username = $event.target.value)"
-          class="form-control"
-          :readonly="privacyMode"
-        >
+        <masked-input
+          :value="config.username"
+          :privacyMode="privacyMode"
+          @input="val => config.username = val"
+        />
       </div>
       <div class="col s6 form-group">
         <label for="password">Password</label>
@@ -222,6 +211,7 @@ import FilePicker from '@/components/common/form/FilePicker.vue'
 import ExternalLink from '@/components/common/ExternalLink.vue'
 import { findClient } from '@/lib/db/clients'
 import ToggleFormArea from '../common/ToggleFormArea.vue'
+import MaskedInput from '@/components/MaskedInput.vue'
 import { mapState } from 'vuex'
 
 export default {
@@ -236,7 +226,8 @@ export default {
   components: {
     FilePicker,
     ExternalLink,
-    ToggleFormArea
+    ToggleFormArea,
+    MaskedInput
   },
   data() {
     return {
--- a/apps/studio/src/components/connection/CommonServerInputs.vue
+++ b/apps/studio/src/components/connection/CommonServerInputs.vue
@@ -38,12 +38,12 @@
       <div class="col s3 form-group" v-if="supportsSocketPathWithCustomPort">
         <label for="port">Port</label>
         <input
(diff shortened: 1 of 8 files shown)


apps/studio/src/components/connection/CommonServerInputs.vue
```diff
@@ -56,16 +58,20 @@
...
+          :value="streamerMode ? '******' : config.host"
+          @input="!streamerMode && (config.host = $event.target.value)"
+          :readonly="streamerMode"
         >
       </div>
       <div class="col s3 form-group">
         <label for="port">Port</label>
         <input
-          type="number"
+          :type="streamerMode ? 'text' : 'number'"
```
rathboma, line 69: Looks like we're repeating this pattern a lot in the component - 3 times it looks like. Probably makes sense to consolidate this logic into a component or other re-usable thing

apps/studio/src/components/connection/CommonServerInputs.vue
```diff
@@ -236,6 +245,12 @@ import ToggleFormArea from '../common/ToggleFormArea.vue'
       }
     },
     computed: {
+      ...mapGetters({
+        settings: 'settings/settings'
+      }),
+      streamerMode() {
```
rathboma, line 251: Probably we should call this 'Privacy Mode' -- it's not just for streamers, also if I share my screen on a zoom call this would be useful

rathboma (changes requested):
Some minor changes to the code requested. Also before this can be merged, you'll need to add it to the docs, ideally with some screenshots to show exactly what it does. See the docs/ folder in the repo. You can run the docs in dev by doing:

```
pip install -r requirements.txt 
mkdocs serve
# go to localhost:8080 to view
```

apps/studio/src/components/connection/CommonServerInputs.vue
rathboma: or maybe 'Presentation mode'? Something a little more general than Streamer mode

apps/studio/src/components/connection/CommonServerInputs.vue
RuiSoares333: Moved repetitive logic into [MaskedInput.vue](https://github.com/beekeeper-studio/beekeeper-studio/blob/009b0efd2d8eb29d704c4997a2f801b1161cbce2/apps/studio/src/components/MaskedInput.vue). Please let me know if it wasn't what you had in mind.

apps/studio/src/components/connection/CommonServerInputs.vue
RuiSoares333: Changed all Streamer's to Privacy's!

rathboma (approved):
This is awesome! I suspect there are more inputs that need the privacy controls, maybe that would be better as a plugin so we could add it to any existing input?

rathboma:
If you enjoyed making this PR and would love to contribute more code -- checking all the database-specific forms for password / host fields would be very helpful!

Issue #2863, opened 2025-03-11
BUG: Excel export showing null string on empty table cells


JeanPaiva42:
From a costumer:

"I noticed with a newer version that when exporting (in my case MySQL) to excel that nulls are exported as literal nulls instead of “blank” which causes issues with numbers in the spreadsheet."

---

The issue was reported under MySQL usage but I also confirmed it sqlite (so I am guessing this is probably happening regardless of the database type.

**To Reproduce**
1. While connected to a database;
2. Perform a query that should return columns that contain NULL values
3. Download the query results as `Excel`
4. Open the generated `Excel` file and notice that we have cells with `null` as their values

**Expected behavior**

- NULL cells should be exported as `blank` instead of being exported with a `null` string.

**Screenshots / Gifs**

![Image](https://github.com/user-attachments/assets/710d0fc5-9891-44a5-9e0b-5e51b4ba7855)

**Version Information (please complete the following information):**
 - OS: Mac
 - App Version 5.1.4
 - Database type and version: MySQL, SQLite

PR #2938, opened 2025-03-26
fix: excel export null values

therealrinku:
fixes: #2863 

- also refactored the code for download function

before:
Screenshot 2025-03-26 at 12 10 39

after:
Screenshot 2025-03-26 at 12 11 25

### DIFF
--- a/apps/studio/src/components/editor/ResultTable.vue
+++ b/apps/studio/src/components/editor/ResultTable.vue
@@ -223,9 +223,9 @@
       download(format) {
         let formatter = format;
         const dateString = dateFormat(new Date(), 'yyyy-mm-dd_hMMss');
-        const title = this.query.title ? _.snakeCase(this.query.title) : "query_results";
+        const title = this.query.title ? _.snakeCase(this.query.title) : 'query_results';
 
-        if(format === "md"){
+        if(format === 'md'){
           formatter = (rows, options, setFileContents) => {
             const values = rows.map(row => row.columns.map(col => typeof col.value === 'object' ? JSON.stringify(col.value) : col.value));
             setFileContents(markdownTable(values), 'text/markdown')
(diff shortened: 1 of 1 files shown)


apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -220,19 +221,39 @@
...
+          formatter = (rows, options, setFileContents) => {
+             const newValue = JSON.stringify(this.dataToJson(this.tabulator.getData(), false), null, "  ");
+             setFileContents(newValue, 'text/json');
+          };
+        }
+
+        // Fix Issue #2863 replacing null values with empty string
+        if(format === "xlsx"){
+          formatter = (rows, options, setFileContents) => {
+             const values = rows.map(row => row.columns.map(col => col.value === null ? "" : typeof col.value === "object" ? JSON.stringify(col.value) : col.value));
```
not-night-but, line 247: I don't love this long tertiary statement. Could you just make these if statements?

apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -220,19 +221,39 @@
...
+             const wb = XLSX.utils.book_new();
+
+             XLSX.utils.book_append_sheet(wb, ws, title);
+             const excel = XLSX.write(wb, { type: "buffer" });
+             setFileContents(excel);
+          }
+        }
 
         // xlsx seems to be the only one that doesn't know what 'all' is it would seem https://tabulator.info/docs/5.4/download#xlsx
         const options = typeof formatter !== 'function' && formatter.toLowerCase() === 'xlsx' ? {} : 'all'
```
not-night-but, line 259: This no longer makes sense if we're going to handle the xlsx download ourselves, as `formatter.toLowerCase() === 'xlsx'` will always be false now., I think you can probably just pass `'all'` as the options directly.

not-night-but (changes requested):
Looks good! Just a couple issues.

therealrinku:
hey @not-night-but!
thanks for checking, I've made the changes!

not-night-but (approved):

PR #2940, opened 2025-03-26
fixed: BUG: Query Cancellation Unavailable After Result Pane Refresh …

### LINKED ISSUE #2870: BUG: Query Cancellation Unavailable After Result Pane Refresh
**Description:**
From a customer:

"Maybe I am still missing something, but I have encountered a problem when running a faulty query (after all Beekeeper is for developing and also a bit of debugging I think) that runs into an infinite loop you can stop/cancel it as long as the result pane hasn't changed. But this can sometimes disappear. I had an erroneous query running the other day, which infinitely kept inserting records into a table. Normally you can just press "cancel" in the result pane below the running query. But at some time (trying to find the error causing the infinite loop)  I ran a query in the same window, something like "select max(Date) from table" to see if the query was just not ready yet, or if it had passed the maximum date and was running infinitely.  At that point the result pane is refreshed and the "cancel" option also is gone. So the only way to stop the infinite loop (luckily that worked) was to close Beekeeper all together. It would be nice if there was a possibility somehow to always quickly cancel a running query."

**TL;DR**: If a new query refreshes the results pane, the "Cancel" button disappears, making it impossible to stop an infinite loop without closing Beekeeper.  
 
**To Reproduce**
 
- While not sure yet if this is exactly the situation where the user faced, we can see that it's not possible to cancel the original query (see video below) and that could cause the situation described.

 
1. Execute a set of queries that would take some seconds to be executed (I will provide the queries that I used at the bottom of the page)
3. Execute another one on top of them, (I also used `SELECT MAX(last_update) FROM actor;`)
4. Notice that a small refresh happened on the result pane, and you can click to `Cancel` the latest query.
5. Click to `Cancel` it.
6. Notice that you are not taken to a position where you could cancel the previous set of queries even though they are most likely running still. So if this query is running in an infinite loop, you are not allowed to cancel it.

**Screenshots / Gifs**
- See video below:

https://github.com/user-attachments/assets/e4389ae4-9d67-4981-9bc9-aca53af75dc9

 - OS: Mac/Linux
 - App Version: happening on 5.1.4
 - Database type and version: MySQL, SQLite

- Queries:
```
WITH RECURSIVE series(n) AS (
    SELECT 1
    UNION ALL
    SELECT n + 1 FROM series WHERE n < 6000000
)
INSERT INTO actor (first_name, last_name, last_update)
SELECT 'data1', 'data2', datetime('now') FROM series;
```
- Reproduced using the [sakila.db](https://www.kaggle.com/datasets/atanaskanev/sqlite-sakila-sample-database).

The0nis:
Closes #2870

This PR addresses query cancellation reliability issues (#2870) by:  

1. **Implementing Single-Query Execution**  
   - Added checks to prevent concurrent queries  
   - Auto-cancel previous queries on new execution  

2. **UI/UX Improvements**  
   - Persistent cancel button during query execution  
   - Disabled run buttons while queries are active  

Developed as part of Software Comprehension and Evolution requirements, this solution demonstrates:  

<details>
# **Issue & Requirements**

Issue [#2870](https://github.com/beekeeper-studio/beekeeper-studio/issues/2870) describes a bug related to Query Cancellation Unavailable After Result Pane Refresh.

As a user, I want to cancel any running query at any time, even if I accidentally run a new query, so that I can avoid infinite loops without restarting the application.

Description:
From a customer:

"Maybe I am still missing something, but I have encountered a problem when running a faulty query (after all Beekeeper is for developing and also a bit of debugging I think) that runs into an infinite loop you can stop/cancel it as long as the result pane hasn't changed. But this can sometimes disappear. I had an erroneous query running the other day, which infinitely kept inserting records into a table. Normally you can just press "cancel" in the result pane below the running query. But at some time (trying to find the error causing the infinite loop) I ran a query in the same window, something like "select max(Date) from table" to see if the query was just not ready yet, or if it had passed the maximum date and was running infinitely. At that point the result pane is refreshed and the "cancel" option also is gone. So the only way to stop the infinite loop (luckily that worked) was to close Beekeeper all together. It would be nice if there was a possibility somehow to always quickly cancel a running query."

TL;DR: If a new query refreshes the results pane, the "Cancel" button disappears, making it impossible to stop an infinite loop without closing Beekeeper.

**To Reproduce**

* While not sure yet if this is exactly the situation where the user faced, we can see that it's not possible to cancel the original query (see video below) and that could cause the situation described.

**Steps to reproduce the behavior:**
1. Execute a set of queries that would take some seconds to be executed (I will provide the queries that I used at the bottom of the page)
2. Execute another one on top of them, (I also used SELECT MAX(last_update) FROM actor;)
3. Notice that a small refresh happened on the result pane, and you can click to Cancel the latest query.
4. Click to Cancel it.
5. Notice that you are not taken to a position where you could cancel the previous set of queries even though they are most likely running still. So if this query is running in an infinite loop, you are not allowed to cancel it.

**Screenshots / Gifs**
### See video below:

https://github.com/user-attachments/assets/0c51342f-4b41-43bf-9fd1-c428e58cc13f

***

**Version Information**
* OS: Mac/Linux
* App Version: happening on 5.1.4
* Database type and version: MySQL, SQLite
* Queries:

***
```SQL
WITH RECURSIVE series(n) AS (
    SELECT 1
    UNION ALL
    SELECT n + 1 FROM series WHERE n < 6000000
)
INSERT INTO actor (first_name, last_name, last_update)
SELECT 'data1', 'data2', datetime('now') FROM series;
```
***
* Reproduced using the [sakila.db](https://www.kaggle.com/datasets/atanaskanev/sqlite-sakila-sample-database).

**Expected Behavior**
_As a user, I want to cancel any running query at any time, even if I accidentally run a new query, so that I can avoid infinite loops without restarting the application._

**Functional Requirements**
* Only one query should execute at a time.
* The "Cancel" button must remain visible and functional until all queries are stopped.
* Disable "Run" buttons/keyboard shortcuts while a query is running.

**Non-Functional Requirements**
* Performance: Cancellation must terminate the query immediately.
* UX: Clear visual feedback (e.g., disabled buttons, progress bar) during execution.

# Source Code Files

The files related to the issue are:
- **TabQueryEditor.vue** – A Vue.js component for the Query Tab, where the SQL query is written and executed.

# Design of the fix
Below are some diagrams to explain the process before the fix and where it fails.

# **State Diagram**

***

* The state diagram map out all possible outcomes of a query. Starting from an **idle state**, the system transitions to **"running"** when a query begins. From there, it should branch into three paths: **success (if the query completes)**, **cancellation (if the user stops it)**, or **error (if it fails)**. All paths must eventually return to the idle state, ready for new actions. _A note should explain that during the "running" state, the cancel button remains visible and functional._

# **Sequence Diagram**

***

* The sequence diagram show how the system handles overlapping query requests. When a user runs a query **(Query 1)** and then starts a second query **(Query 2)** before the first finishes, the system cancels **Query 1** automatically before executing **Query 2**. This ensures only one query runs at a time, and users never lose the ability to cancel an ongoing operation. The final results shown should always reflect the most recent query.

# **Class Diagram**

***

* The class diagram should illustrate the relationships between the core components. The QueryEditor must connect to both the Connection (which talks to the database) and the Query (which represents the running operation). Key actions like "execute" and "cancel" should appear as methods in their respective classes, showing how the editor controls query execution. Arrows between boxes should clarify which components interact, ensuring the design matches the actual code structure.

Together, these diagrams verify that the system prevents overlapping queries, maintains clear user controls, and follows logical workflows. A simple visual inspection should confirm that all critical interactions and states are represented accurately.

To fix the issue where the cancel button disappears when a new query is executed while another is running, we need to ensure only one query runs at a time and handle cancellation properly. Here's how to modify the code:

**Step-by-Step Changes:**

1. **Prevent Multiple Query Execution:** Add checks in methods that trigger queries to prevent running new queries while one is already executing.

2. **Cancel Existing Queries:** Before starting a new query, cancel any existing one to ensure only one query runs at a time.

3. **Disable UI Elements During Execution:** Disable run buttons and keyboard shortcuts while a query is running to prevent user from triggering new queries.

# Code Modification
1. **In submitTabQuery and submitCurrentQuery methods, add a check for this.running:**
`TabQueryEditor.vue`
```jsx
async submitTabQuery() {
  if (this.running) return; // Add this line
  const text = this.hasSelectedText ? this.editor.selection : this.unsavedText
  this.runningType = this.hasSelectedText ? 'selection' : 'everything'
  if (text.trim()) {
    await this.submitQuery(text)
  } else {
    this.error = 'No query to run'
  }
},
async submitCurrentQuery() {
  if (this.running) return; // Add this line
  if (this.currentlySelectedQuery) {
    this.runningType = 'current'
    await this.submitQuery(this.currentlySelectedQuery.text)
  } else {
    this.results = []
    this.error = 'No query to run'
  }
},
```
2. **Modify the submitQuery method to cancel any existing query before starting a new one:**
`TabQueryEditor.vue`
```jsx
async submitQuery(rawQuery, fromModal = false) {
  if (this.remoteDeleted) return;
  
  // Cancel existing query before starting a new one
  if (this.running && this.runningQuery) {
    await this.cancelQuery();
  }

  this.tab.isRunning = true;
  this.running = true;
  this.error = null;
  this.queryForExecution = rawQuery;
  this.results = [];
  this.selectedResult = 0;
  // ... rest of the existing code
}
```
3. **Update the template to disable run buttons when running is true:**
`TabQueryEditor.vue`
```html
<x-button
  class="btn btn-primary btn-small"
  v-tooltip="'Ctrl+Enter'"
  @click.prevent="submitTabQuery"
  :disabled="running"
>
  <x-label>{{ hasSelectedText ? 'Run Selection' : 'Run' }}</x-label>
</x-button>

<x-button
  class="btn btn-primary btn-small"
  menu
  :disabled="running"
>
  <!-- ... other elements ... -->
</x-button>
```

# Pull Request

Proposed fix submitted in the [#2940](https://github.com/beekeeper-studio/beekeeper-studio/pull/2940)
</details>


not-night-but (approved):
Looks good! Thanks for doing this!

PR #2941, opened 2025-03-27
fast ui kit hmr

azmy60:


rathboma (changes requested):
This feels like a hack. Should we totally separate the UI kit into it's own repo and just use the npm published version?

PR #2946, opened 2025-03-29
Feat/mongo sql editor

not-night-but:
Marking this as ready for review, but I'm sure I'm missing something (I know for sure imports don't work yet).

But this allows you to use a postgres-like sql dialect to query your mongodb instance. It does modify the UI to be able to open either a shell or a query tab, see our friend the lil grey triangle below:
image

I have managed to get exports to work as well, for both tables and queries, and also copy to sql for most changes (this is where I'm suspecting I forgot a feature)


apps/studio/src-commercial/backend/lib/db/clients/mongodb.ts
```diff
@@ -239,12 +249,63 @@ export class MongoDBClient extends BasicDatabaseClient<QueryResult> {
     }
   }
 
-  async selectTopSql(_table: string, _offset: number, _limit: number, _orderBy: OrderBy[], _filters: string | TableFilter[], _schema?: string, _selects?: string[]): Promise<string> {
-    log.error("MongoDB does not support generating SQL scripts");
-    return '';
+  async selectTopSql(table: string, offset: number, limit: number, orderBy: OrderBy[], filters: string | TableFilter[], _schema?: string, selects?: string[] = ['*']): Promise<string> {
```
rathboma, line 252: Can't you use the implementation from the postgresql driver?

rathboma (approved):
I'm going to approve and merge it -- but it really needs tests before we consider the mongo driver, and the mongo sql implementation out of beta.

PR #2948, opened 2025-03-30
Fix #2907: Prevent "ß" from printing when toggling the sidebar

### LINKED ISSUE #2907: BUG: ß character (or equiv.) prints when toggling the side bar
**Describe the bug**
Your keyboard shortcut to toggle the sidebar also registers as a key press in the query editor.

**To Reproduce**
1. Open a query editor tab
2. Option+s
3. See the side bar opens/closes and an ß is printed in the editor.

**Expected behavior**
See the side bar opens/closes but no character is printed in the editor.

 - OS: Mac amd64 15.3.2
 - App Version 5.1.5

First
Image

Second
Image

Third
Image

SMV-21:
The issue occurs when the keybind is activated, causing the cursor in the text editor to trigger and print "ß". To resolve this, I modified the keymap() function to prevent the default behavior (printing) on Mac when the keybind is triggered, ensuring only the sidebar is toggled.

Close #2907

### DIFF
--- a/apps/studio/src/components/CoreInterface.vue
+++ b/apps/studio/src/components/CoreInterface.vue
@@ -82,9 +82,16 @@
         "secondarySidebarSize",
       ]),
       keymap() {
-        return this.$vHotkeyKeymap({
+        const result = this.$vHotkeyKeymap({
           'general.openQuickSearch': this.showQuickSearch
-        })
+        });
+        if (this.$config.isMac) {
+          result['alt+s'] = (event) => {
+            event.preventDefault();
+            this.toggleSidebar();
+          };
+        }
+        return result;
       },
       splitElements() {
         return [
--- a/apps/studio/src/components/CoreInterface.vue
+++ b/apps/studio/src/components/CoreInterface.vue
@@ -85,12 +85,6 @@
         const result = this.$vHotkeyKeymap({
           'general.openQuickSearch': this.showQuickSearch
         });
-        if (this.$config.isMac) {
-          result['alt+s'] = (event) => {
-            event.preventDefault();
-            this.toggleSidebar();
-          };
-        }
         return result;
       },
       splitElements() {
--- a/apps/studio/src/common/menus/MenuItems.ts
+++ b/apps/studio/src/common/menus/MenuItems.ts
@@ -165,7 +165,7 @@ export function menuItems(actionHandler: IMenuActionHandler, settings: IGroupedU
     primarySidebarToggle: {
       id: 'menu-toggle-sidebar',
       label: 'Toggle Primary Sidebar',
-      accelerator: "Alt+S",
+      accelerator: platformInfo.isMac? "CommandOrControl+B" : "Alt+S",
       click: actionHandler.togglePrimarySidebar,
     },
     secondarySidebarToggle: {


apps/studio/src/components/CoreInterface.vue
```diff
@@ -80,9 +80,16 @@
...
           'general.openQuickSearch': this.showQuickSearch
-        })
+        });
+        if (this.$config.isMac) {
+          result['alt+s'] = (event) => {
+            event.preventDefault();
+            this.toggleSidebar();
+          };
+        }
+        return result;
```
not-night-but, line 92: Are you on a Mac? Can you confirm that this works? I believe that the keybinds that are set in the menubar are actually run by electron, unless we tell electron not to (take a look at the quick-search item in MenuItems.ts, we have `registerAccelerator: false` there).

I think this should also not be added to the object this way. Maybe add a key to the `default.config.ini` file under general for toggle sidebar, and then register it here how we register quick search.

not-night-but (changes requested):
Thanks for this! I think we should make sure that this keybind isn't being handled by both electron and vue at the same time, and maybe add a key to the configuration for toggling the sidebar.

apps/studio/src/components/CoreInterface.vue
SMV-21: Hi! I’m on macOS and can confirm that this works. I looked into your suggestions and noticed that even with registerAccelerator set to false, the shortcut still seems to be handled in both places. When I added the key to default.config.ini, it caused the window to flash briefly (open and then close), which suggests it’s still being triggered somehow.

For debugging, I temporarily changed the quick-search keybind to Alt+P in both files, and when I did that, it printed a “π” character inside the quick-search modal.

Would it make sense to consider changing the default keybinding to something like ControlOrCmd+B to help avoid this conflict?

Thanks again for your help and for taking the time to review this!

https://github.com/user-attachments/assets/20330357-645c-461e-b76a-1d70ab522dd5

not-night-but: Hmmm register accelerator *should* work, we use it in a couple other places (cut, quickSearch, closeTab). Are you seeing similar behaviour with those shortcuts? Ideally I don't want us to have to change the shortcut, but we might have to. I want to be able to do this with the same paradigm as some of the other shortcuts in the native menu.

SMV-21: While testing the default keybindings in `default.config.ini` and `MenuItems.ts`, when switching Cut and Quick Search from Ctrl to Alt shortcuts (Alt+X and Alt+P) and using these combinations, I noticed that the corresponding characters (« for Alt+X and π for Alt+P) are inserted wherever the cursor is. This suggests that the key events are being interpreted as text too. This doesn’t happen with the Close Tab shortcut, possibly because the cursor gets unfocused when that action is triggered.

From what I could gather, it seems like `registerAccelerator` isn’t being used for keybindings in `MenuItems.ts`. Instead, keybinds are sent through `NativeMenuActionHandler.js`, and since `NewAppMenu.vue` isn’t involved, the `registerAccelerator === false` condition is never checked, so Electron continues to send the shortcut to the app regardless.

To test this, I added a conditional check for `registerAccelerator` in `NativeMenuActionHandler.js`. With this in place, I confirmed with the terminal that the keybind wasn’t sent when `registerAccelerator` was false. However, the character was still being inserted. I suspect this has to do with the nature of Alt key combinations, they often map to special characters while Ctrl combinations don't behave the same way.

That being said, I was able to prevent the character from being printed using `event.preventDefault()` in `CoreInterface.vue`, though I’m not sure if that’s the cleanest or most appropriate approach here.

I've attached two videos that demonstrate the behaviour with the new keybindings for Cut and Quick Search, in case it's helpful!
Note: with cut I needed to spam the keybind for the behaviour to be seem due to its nature.

https://github.com/user-attachments/assets/65aa5ac2-08a5-48bc-8319-59d31eec74ab

https://github.com/user-attachments/assets/83cf1ebe-889e-49d6-8881-21fee40687bf

wmontgomery: I'm running this on Sonoma 14.6.1 and not getting the sidebar to toggle. I'm no longer seeing the ß though, so that's nice

SMV-21: That happened when I rebased about two weeks ago to include the latest changes from the secondary sidebar. At that point, the method for toggling the sidebar was changed from `toggleSidebar()` to `handleToggleOpenPrimarySidebar()`, so I adjusted my code accordingly to continue development on top of those updates. I didn’t commit those changes at the time because I was still unsure what direction to proceed in, and I wanted to avoid committing anything that might be incorrect.

not-night-but: @wmontgomery You're a mac user, do you have any thoughts on which direction we should proceed with this? Should we just avoid all `alt` keybinds as this seems to often conflict with some native mac keybinds? Or should we override those keybinds to keep the experience the same across all platforms?

wmontgomery: @not-night-but , I would say avoid all *alt* keybinds. Sure it's nice to have the same stuff across platforms, but macos in general has a different keyboard layout what with cmd vs ctrl.

SMV-21: Hi @not-night-but , I was wondering if I should change the keybind only on Mac or if I should change it for all platforms? Also, which keybind should I use? Should I use `CommandOrControl+B` ?

SMV-21: Hi @not-night-but, just wanted to kindly follow up on my previous question. I understand you might be busy, but I’d really appreciate any input whenever you have a moment. Thanks in advance!

not-night-but: @SMV-21 So sorry I let the ball drop on this!! My bad. I think we should only change it for Mac if possible. It's kinda just different philosophies between the OS's, so I don't think we should force other OS users to conform to the Mac defaults if that makes sense.

SMV-21: Hi @not-night-but, no problem! I made the changes and committed. Take your time, don’t worry! Thanks.

not-night-but:
@SMV-21 Sorry this took a bit, merging now

not-night-but (approved):

Issue #2485, opened 2024-10-09
QUESTION: How to disable fuzzy search ?


NaWer:
**Details:**
When you toggling filter type to fuzzy search, every keystroke triggers a query : this causes errors until the request is valid. (with the message ‘Error loading data, see below’).

Is it possible to disable this behaviour and just type enter when the input is complete to search?

**Version Information [might be relevant to your issue]**
Operating System: Ubuntu 22.04
App Version [help -> about]: 4.6.2
Database type and version: all

**Screenshot / GIF**:
![image](https://github.com/user-attachments/assets/42905283-2a85-4cea-953e-d23b7620a7c7)

not-night-but:
Yeah we should probably debounce this

PR #2949, opened 2025-03-31
fix: submit raw query filter only on form submit

therealrinku:
fixes: #2485 

removing watch method for raw filter and only search when user clicks enter or clicks search icon, 
same behavior as query filter builders.


rathboma (approved):

rathboma:
Thanks again :-)

PR #2951, opened 2025-03-31
Feature/1816 export connection

### LINKED ISSUE #1816: FEAT: Export a connection in multiple formats
If I have a bunch of connections saved in Beekeeper Studio, it's helpful to be able to copy them for use in other applications.

Some examples:
- I want to use the database in my rails APP
- I want to connect to the database from Java
- I want to connect to the database from the Sequelize ORM
- I want to connect to the database using ODBC

For version 1 support the following exports (at least for our core databases - psql, sqlite, mysql, mariadb, mssql, oracle):

- Export as JDBC connection string
- Export in Ruby on Rails database.yml format
- Export in SQLAlchemy format for Python
- Export as a generic connection string (url)
- Export as JSON - Just use Beekeeper's fields

Just needs to be for a single connection, can be a dropdown button at the bottom of the connection form

wmontgomery:
resolve #1816 

Note: I can add a codemirror look for the code if desired. Wasn't sure if it would make sense to have or not.

1816_export-connection


rathboma (changes requested):
I’m not sure the exported configs are all valid. Like in rails you use ‘sqlite3’, not SQLite, and many drivers aren’t supported, like duckdb.

each connection or dialect will need ways to convert into the right language, or not if it doesn’t work. I think we should also have a  3 dot menu or gear menu button next to save, vs a big copy button.

PR #2957, opened 2025-03-31
Fix #2627: Foreign key button not visible if column content is truncated

### LINKED ISSUE #2627: BUG: Foreign key button not visible if column content is truncated
When I want to click into a foreign key, I have to expand the column until the link shows up.

I want the link to show up even if the content isn't fully expanded - i.e., it should show up at the end of the visible column, not only when the column is fully expanded.

**Additional context**

Only shows up when content is expanded:

image

Would like it to show up even when the content isn't

image

joaovicente3:
Added `position: absolute` and `right: 5px` to position the button correctly.
Set `background-color: $theme-bg` to improve visibility when overlapping text.

Close #2627


rathboma (approved):

rathboma:
Thanks!

PR #2958, opened 2025-04-01
CHORE: Add load test container to Docker Compose

JeanPaiva42:
# Add Load Test Container to Docker Compose

## Overview

This PR introduces a new container service to the existing Docker Compose setup to facilitate stressing beekeeper. 

## Changes

- **New Container Service**: `psql15-load-test` added to the `docker-compose.yml`.
  - Configured with a separated volume.
  - Includes initialization scripts to create 15000 tables and populate them with 100 rows of data per table (we could go harder but it's good enough to make the app stressed.
  - Exposes the container on port `5435`.

## How to Use

1. `docker compose up psql15-load-test`

### DIFF
--- a/apps/studio/src/migration/dev-8.js
+++ b/apps/studio/src/migration/dev-8.js
@@ -1,0 +1,33 @@
+import _ from 'lodash'
+import { SavedConnection } from '../common/appdb/models/saved_connection'
+import platformInfo from '../common/platform_info'
+
+export default {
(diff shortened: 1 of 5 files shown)


rathboma (changes requested):
You should also add it to the develop-mode only database migrations, so it's available by default in the dev version of the app. See for example dev-7.js, and migrations/index

rathboma (approved):

PR #2960, opened 2025-04-02
fix some deprecation warnings and dark theme

not-night-but:
This fixes deprecation warnings in our sass relating to global builtins that sass is deprecating (lighten, darken, etc). There are more deprecations that should be dealt with at some point, specifically `@import`, but that will require a more involved refactor of our styles. `@import` makes everything global, which we've used to our advantage with variables, but the new way of doing things, `@use` doesn't make those declarations global, which means we need to possibly rethink how our styles are organized.

This also fixes the issues with dark theme by adding a dark theme variables file. It seems that vite or something is not properly reloading the sass when we switch themes and since we don't have a variables file in master for dark theme, it ends up using the light theme variables.


azmy60:
@rathboma I feel like the part where we didn't import dark theme variables was because our css defines colors using dark theme by default?

not-night-but:
@azmy60 yeah dark theme vars are set as the default, and since we only import other variables for the other themes, but not for dark them, we should be getting the dark theme values. Which is why I don't like this solution and why I'm confused as to why this stopped working. Like I said, I think sass is improperly reloading, and somehow the media queries are forcing them to reload, which is why system theme still works.

azmy60 (approved):

rathboma:
@not-night-but it's approved, so merge

PR #2962, opened 2025-04-02
Fix: Restore SQL auto-complete suggestions after ESC key

### LINKED ISSUE #2884: BUG: Issue with auto-complete suggestion
**Bug Description:**
 
- When writing queries on the application, if a user ignores the auto-complete suggestion, by pressing ESC for example, the auto-complete will not come back to suggest the same thing again. This will also happen if we pick a suggestion, but change our minds and decide to go with a different thing (a different table name for example). Suggestions will only come back if we delete the query up until the last SQL reserved word. 

**To Reproduce**
1. Go to a query editor space;
2. Start to write a simple query that would trigger auto-complete (such as a SELECT that has a valid table)
3. Continue writing the table name, but don't pick from the auto-complete. 
4. Now delete the name (completely or partially) and notice that no suggestion is provided.

**Expected behavior**

- The application should keep providing suggestions, even if the user ignores them.

**Screenshots / Gifs**

https://github.com/user-attachments/assets/32b7e2ca-f288-460e-9da7-6b092cae988c

 - OS: Mac
 - App Version: 5.1.5.beta.1

**Additional context**
This issue was initially reported by an user through email, it was not informed the version in which they faced it, but I was able to repro it on the latest beta (5.1.5.beta.1)

rathboma:
## Summary
- Modified auto-complete handler to show suggestions again after ESC key is pressed
- Updated show-hint.js to track dismissed suggestions and allow them to reappear
- Added tests to verify auto-complete suggestions still appear after using ESC
- Applied fix to both main app and UI kit package

## Problem
When writing queries on the application, if a user ignores the auto-complete suggestion (by pressing ESC) or deletes text after selecting an item, the auto-complete will not show the same suggestion again. The user must delete the query up to the last SQL reserved word to get suggestions back.

## Test plan
1. Open query editor
2. Start to write a simple query that triggers auto-complete (like SELECT FROM users)
3. Press ESC to dismiss the suggestion popup
4. Verify that suggestions appear again when continuing to type
5. Run tests: 

Fixes #2884

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/apps/studio/src/lib/editor/CodeMirrorPlugins.ts
+++ b/apps/studio/src/lib/editor/CodeMirrorPlugins.ts
@@ -132,10 +132,11 @@ export function autocompleteHandler(
     "Period": "period",
   };
 
-  if (instance.state.completionActive) return;
+  // Fix for issue #2884: Don't prevent completions when pressing Escape or deleting text
+  if (instance.state.completionActive && e.key !== "Escape") return;
   
-  // Handle delete and backspace keys
-  if (e.key === "Backspace" || e.key === "Delete") {
+  // Handle delete and backspace keys or Escape key (when completion was active)
+  if (e.key === "Backspace" || e.key === "Delete" || (e.key === "Escape" && instance.state.completionActive)) {
     // Get context at current cursor position
     const cursor = instance.getCursor();
     const line = instance.getLine(cursor.line);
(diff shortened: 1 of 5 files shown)


rathboma:
@azmy60 claude did this, I don't know if it's sane at all

apps/studio/src/lib/editor/CodeMirrorPlugins.ts
```diff
@@ -132,10 +132,11 @@ export function autocompleteHandler(
     "Period": "period",
   };
 
-  if (instance.state.completionActive) return;
+  // Fix for issue #2884: Don't prevent completions when pressing Escape or deleting text
```
azmy60, line 135: No need to change this.

I think the whole autocomplete plugin from apps/studio can be removed completely since we are using the SQL Text Editor from UI Kit.

And we can remove other plugins too. Though some of the functions might be related to the query magics which is the only plugin we use here so be cautious.

azmy60: @not-night-but Does the mongo shell use some of the plugins in this file?

apps/studio/src/vendor/show-hint/index.js
```diff
@@ -78,6 +78,20 @@ import CodeMirror from "codemirror";
   Completion.prototype = {
     close: function() {
       if (!this.active()) return;
+      
+      // Fix for issue #2884: Store the completion state before clearing it
+      // This helps with re-triggering the same suggestions after pressing Escape
+      if (!this.cm.lastCompletionState && this.data) {
```
azmy60, line 84: This should also be applied in UI Kit too?

azmy60 (changes requested):

rathboma:
closing this, it was claude and I don't have enough confidence in it.

PR #2969, opened 2025-04-04
Add global status bar

azmy60:
Screenshot 2025-04-04 130521

### DIFF
--- a/apps/studio/src/components/GlobalStatusBar.vue
+++ b/apps/studio/src/components/GlobalStatusBar.vue
@@ -1,0 +1,15 @@
+<template>
+  <div class="global-status-bar">
+    <connection-button />
+    <portal-target name="global-status-bar"></portal-target>
+  </div>
+</template>
+
+<script>
+import { PortalTarget } from 'portal-vue'
+import ConnectionButton from './sidebar/core/ConnectionButton.vue'
+
+export default {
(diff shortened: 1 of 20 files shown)


rathboma:
@azmy60 so the connection is no longer contained in the sidebar visually? Not sure I like that, like in your screenshot - the connection button is bleeding into the context of the tab.

I get that this is a practical thing because of the right sidebar, but maybe the statusbar can be split in the same proportions as the sidebar?

azmy60:
@rathboma It'll be too hard to make it proportional to the elements outside of it. The closest we can get is putting the connection button back in the sidebar which should be practical too I feel like.

azmy60:
primary-secondary-sidebar-new

@rathboma I just changed it the way you described it. It turns out not that hard.

apps/studio/src/components/GlobalStatusBar.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <div class="global-status-bar">
+    <div
+      class="connection-button-wrapper"
+      :style="{
+        ['--connection-button-icon-width']: connectionButtonIconWidth + 'px',
+        width: connectionButtonWidth + 'px',
```
azmy60, line 7: This might look hacky. But it's like how splitjs works under the hood.

rathboma (approved):

Issue #2892, opened 2025-03-16
BUG: JSON in booleans not showing


ChrisToxz:
**Describe the bug**
In JSON the plain values `true` or `false` are valid. (Note: No quotes or brackets needed)
However this value breaks the data viewer. No rows are visible.
When you run the query manually, the json is properly parsed

**To Reproduce**
1. At least MySQL databae (Not tested others)
2. Have a JSON column
3. Add `true` or `false` as value
4. Try to see rows in data viewer.

**Expected behavior**
Parse booleans properly in JSON columns

**Screenshots / Gifs**
No rows visible while rows exists + error:
![Image](https://github.com/user-attachments/assets/476c0c4f-bfc5-4bc6-ad66-b27a03fbd96d)
Structure:
![Image](https://github.com/user-attachments/assets/cd89cd64-26c4-4a0b-a801-873d1ba50345)
Manually query shows there is data:
![Image](https://github.com/user-attachments/assets/b80af093-9158-4fd6-a9bb-df29ca51a3c6)

**Version Information (please complete the following information):**
 - OS: Linux, ZorinOS 17
 - App Version: 5.1.5
 - MySQL 8

**Additional context**
https://datatracker.ietf.org/doc/html/rfc8259#section-3
```
A JSON value MUST be an object, array, number, or string, or one of
   the following three literal names:
      false
      null
      true
```

PR #2978, opened 2025-04-05
fix: boolean values in jsonb column error

therealrinku:
fixes: boolean values in jsonb causing error that causes table to be blank

closes: #2892 

before:

https://github.com/user-attachments/assets/cee37203-ad17-401e-975b-72787cf05c1c

after:

https://github.com/user-attachments/assets/e9eb8bfd-6e08-4d51-beb4-80cd04e6cd83

### DIFF
--- a/apps/studio/src/lib/data/tools.ts
+++ b/apps/studio/src/lib/data/tools.ts
@@ -101,7 +101,7 @@ export const Mutators = {
 
   /** Stringify json data for MySQL column */
   jsonMutator(value: any): JsonFriendly {
-    if (_.isString(value) || _.isNull(value) || _.isUndefined(value)) return value
+    if (_.isString(value) || _.isNull(value) || _.isUndefined(value) || _.isBoolean(value)) return value
     return friendlyJsonObject(value)
   },
 }


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1302,7 +1302,11 @@ export default Vue.extend({
         const data = { ...row.getData() }
         const dataParsed = Object.keys(data).reduce((acc, d) => {
           if (!this.primaryKeys?.includes(d)) {
-            acc[d] = data[d]
+            if(data[d] !== null && (typeof data[d] === "object" || typeof data[d] === "boolean")) {
+              acc[d] = JSON.stringify(data[d])
```
azmy60, line 1306: Was this tested on all database types and column types? If not, would it be better to specifically apply this to mysql and a column that is `json` only?

azmy60: How does it affect the JSON Viewer if the cloned row selected?

azmy60 (changes requested):
We probably need to narrow down the fix to mysql and `json` column only to avoid new issues. Maybe a little comment that links to the issue would be great also! :)

apps/studio/src/components/tableview/TableTable.vue
therealrinku: hey @azmy60
yeah you're right, clone issue might need some more testing.

so I guess we can just merge fix for #2892 on this PR
and 
check out the clone issue on other PR.
what do you think?

there's not any issue open for that yet, maybe it's happening just for me, not exactly sure yet.

azmy60: @therealrinku Sounds great! But we'll need to test this PR first with other database types other than MySQL. For example, SQLite is not strict with data types depending on the column type name, so def need to check that.

therealrinku:
hey @azmy60 
I've updated the code that just fixes the issue #2892, you can check it out. I think this should be good.
I can checkout the clone issue later when I get the chance.

azmy60 (approved):

PR #2979, opened 2025-04-05
Fix #2699: Prevent exposing internal errors in UI

### LINKED ISSUE #2699: BUG: Exposing internal code issues in the UI
**Describe the bug**

When trying to run a query to a file, sometimes we are displaying JS errors or informations on the UI, which can cause confusion and affect the user experience on the app.

**To Reproduce**
1. Open a Query
2. When running it, select `Run Selection to File`
   image

3. Go through the steps needed to Run the query
4. Notice that on the toast that appear on the bottom right side, we can read "Undefined rows"
   image
5. Now add some error to the query, and do the same steps (1-3)
6. Notice that we are displaying a "object Promise" error to our users.
   image

**Expected behavior**
- We shouldn't be exposing implementation related errors to our end users. This error should have appropriate handling.

**Screenshots / Gifs**

- On top of the screenshots provided on the repro steps, see the video below for details on the issue:

https://github.com/user-attachments/assets/71045556-23ea-45ad-91f3-af75d27e1b9e

 - OS: Mac
 - App Version: 5.0.9
 - Database type and version: SQLite

**Additional context**
- On the step 3 of the repro steps I mention `Undefined rows`,  it's because I am working under the impression that the `undefined` is coming from JS `undefined`.. if that's not the case, we should just consider the `object Promise` part of the issue.

TomasSilv:
This PR addresses issue #2699 by improving error handling during export operations.
Previously, when running a query to a file, internal erros were being exposed in the UI,
exposing implementation related errors to end users. This fix ensures that errors are properly handled and displayed as user-friendly messages instead of raw internal values.

Changes in this PR:

Errors from the export process are now caught and replaced with a generic, user-friendly message.
Added fallback for undefined row count during export notifications.
Introduced a test to verify error message display when the export fails.

Screenshot of error


azmy60 (approved):
LGTM!

PR #2983, opened 2025-04-07
Fix #2809: Improve JSON Row Viewer Truncated Text

### LINKED ISSUE #2809: Improve JSON Row Viewer Truncated Text
Hi! First, thank you for making this app and putting it out there with a free version!

I'm a bit surprised that you can't properly preview the data when you select a row. Or am I doing this wrong? I want to be able to select a row and easily read the text that's in the fields.

Is the truncated JSON view the way you inspect your data in Beekeeper Studios?

It's a bummer because the rest of the app seems awesome, and I really want to use it over TablePlus.  But not being able to properly view the row data at a glance is kind of a dealbreaker for my usage.

DanielFF2:
Fix #2809:
Captura de ecrã 2025-04-09 124320

Problem: Users missed clickable truncation
Solution: Blue+underlined ellipsis
Behavior: Double-click still expands
Impact: Better feature discoverability


rathboma:
@azmy60 Can you take a look for me?

azmy60 (changes requested):
Thanks for making a PR! But this doesn't seem to solve the issue. If you are looking for the JSON Row Viewer, that should be in `apps/studio/src/components/sidebar/JSONViewer.vue` and `apps/studio/src/lib/data/jsonViewer.ts`

Particularly, the ellipsis marker creation is located ~[here](https://github.com/beekeeper-studio/beekeeper-studio/blob/58a83c9189bb2d819905c1e4e0b57f0eec6d597d/apps/studio/src/lib/data/jsonViewer.ts#L46-L58)~ [here](https://github.com/beekeeper-studio/beekeeper-studio/blob/58a83c9189bb2d819905c1e4e0b57f0eec6d597d/apps/studio/src/lib/data/jsonViewer.ts#L61-L74). So I think that would be the best place to fix the issue.

Also, I think this will be great if we can change the ellipsis to a "button" that says "Show more". Many apps do this especially chrome devtools. 
Screenshot 2025-04-15 123657

DanielFF2:
@azmy60 So sorry for not responding earlier, didn't have a hold of my pc! Going right to work on it!

Thank you!

azmy60:
@DanielFF2 no worries! :)

DanielFF2:
Hi @azmy60 ! Wanted to first send a picture of what I made it look like! 
Captura de ecrã 2025-04-26 003915

rathboma:
@azmy60 looks like this needs a re-review?

azmy60:
@DanielFF2 Oh sorry I got caught up with other work. Looking at your screenshot, it looks like a link which could work but mostly links are used for opening a new page or window, so it doesn't really fit the situation. For example, people might expect that clicking this will open a new tab or a browser.

So is it possible to style it like a button? like the screenshot I posted? Thanks!

apps/studio/src/assets/styles/components/json-viewer.scss
```diff
@@ -174,3 +174,13 @@
     }
   }
 }
+
+.truncatable-value .more {
```
azmy60, line 178: Can we make it look like a button? You might need to make the font smaller too.

azmy60 (changes requested):

mrsimb:
What do you think about using ellipsis `...` instead of `Show more`? A bit shorter, saves space.

DanielFF2:
@azmy60 I will get that done in no time! So, so sorry for the delay!

DanielFF2:
@azmy60 how do you like this one? Again, very sorry for the time it took me to see this, indeed an easy change to apply, tried to make it as simiral as I could to the picture! Could also have it underlined when an user hovers over it. Again sorry and thank you!
Captura de ecrã 2025-06-01 182308

azmy60:
@DanielFF2  No worries :) That looks like if "Show more" is actually a part of the text. I would add a background color so it really looks like a button.

DanielFF2:
@azmy60 I was looking at backround color and ended up with a very suttle grey, first screenshot is how it looks like and second is how it looks like when you hover over it! 
Captura de ecrã 2025-06-02 110805

image

azmy60:
@DanielFF2 Looks good to me!

PR #2997, opened 2025-04-11
fix ansi output from mongosh

not-night-but:
Updated the mongosh output to convert ansi to html.

So we go from this:
image

To this:
image

### DIFF
--- a/apps/studio/src/components/common/texteditor/MongoShell.vue
+++ b/apps/studio/src/components/common/texteditor/MongoShell.vue
@@ -106,8 +106,6 @@ export default {
       const doc = this.shell.getDoc();
       const lastLineNum = this.shell.lastLine();
       let output = value.output;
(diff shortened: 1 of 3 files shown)


apps/studio/src/components/common/texteditor/MongoShell.vue
```diff
@@ -103,8 +106,18 @@ export default {
       const doc = this.shell.getDoc();
       const lastLineNum = this.shell.lastLine();
       let output = value.output;
+      console.log('OUTPUT: ', output)
```
azmy60, line 109: Should we use `@bksLogger` for this?

not-night-but: i thought I removed that :upside_down_face:

azmy60 (approved):

PR #3016, opened 2025-04-15
disable shell for non-mongo dialects

not-night-but:


rathboma (approved):

rathboma:
@not-night-but this needs to be merged against rc-52

PR #3023, opened 2025-04-17
Update electron to v35

### LINKED ISSUE #2968: CHORE: Update Electron to v35
Beekeeper studio won't build on NixOS because Electron 31 is EOL (https://endoflife.date/electron)

Upgrade to the latest Electron 35 release.

bradynwalsh:
Fixes #2968 (I am also running into the same NixOS build error, so decided to contribute)

As part of this update, also required to update Node to v22 and several other dependencies that are incompatible with these Node/Electron versions. It appears to pass all tests I can run locally, but I don't have a Windows box to test the changes required in the Electron build config

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5804,6 +5804,16 @@ bcrypt-pbkdf@^1.0.2:
   dependencies:
     tweetnacl "^0.14.3"
 
+bcryptjs@^3.0.2:
+  version "3.0.2"
+  resolved "https://registry.yarnpkg.com/bcryptjs/-/bcryptjs-3.0.2.tgz#caadcca1afefe372ed6e20f86db8e8546361c1ca"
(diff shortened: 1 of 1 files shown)


azmy60:
@bradynwalsh Thanks for the PR! Sorry this took a while. I have a windows machine so I can test this ASAP!

not-night-but:
@bradynwalsh Unfortunately I think we are going to have to close this without merging. Electron updates are a nightmare, and this build seems to be broken in release mode (from our testing on both Arch Linux and Windows). This may be something we can look into properly doing in the future, but unfortunately a combination of issues makes it unlikely for now (some of our dependencies do NOT like the upgraded node, so we may be blocked in that regard until they are updated).

Thanks for attempting this though! We love our community, just unfortunately it's look like this upgrade will have to be put on hold.

Issue #3024, opened 2025-04-17
BUG: Mongo ObjectID Columns In JSON View Show As Bits


mslosarek:
**Describe the bug**
When looking at the JSON view of a MongoDB records, the ObjectID renders as individual bits. 
```
"_id": {
    "0": 103,
    "1": 250,
    "2": 250,
    "3": 60,
    "4": 45,
    "5": 253,
    "6": 220,
    "7": 172,
    "8": 64,
    "9": 157,
    "10": 38,
    "11": 74
  }
```

**To Reproduce**
1. Select a MongoDB Record
2. Right click on the record and choose "See Details", or have the JSON viewer open
3. The _id shows as an object of bits
**Expected behavior**
Expect a string

**Screenshots / Gifs**
<img width="151" alt="Image" src="https://github.com/user-attachments/assets/ec8bb255-5c45-4fd9-8da3-3d528ce0a6d5" />

**Version Information (please complete the following information):**
 - OS: Mac
 - App Version: Version 5.2.0-beta.1 (5.2.0-beta.1)
 - Database type and version MongoDB 6.0.21

**Additional context**
Anything else you think might be helpful

not-night-but:
Thank you so much for your report! Glad to see some people are messing around with our mongo integration, I'll take a look at this

mslosarek:
I think you should be able to something like
```
      if (value && typeof value === 'object' && 'id' in value && ArrayBuffer.isView(value.id)) {
        return typedArrayToString(value.id as ArrayBufferView, this.binaryEncoding)
      }
```
in the [JSON viewer replacer](https://github.com/beekeeper-studio/beekeeper-studio/blob/552aa727779c60a0d45cc94f330916b8d7020fe9/apps/studio/src/components/sidebar/JsonViewer.vue#L345)

not-night-but:
@mslosarek Oh probably! I'm working on another fix right now, so I'll get to it eventually, but if you have a fix also feel free to put up a PR!

itsamrit:
this is still coming when you search Mongo ObjectId Columns in JSON viewer

<img width="793" height="706" alt="Image" src="https://github.com/user-attachments/assets/d71cb14f-7eca-4c66-ad31-1502ab5b6ea9" />

itsamrit:
attaching pr :- https://github.com/beekeeper-studio/beekeeper-studio/pull/3530

PR #3025, opened 2025-04-17
fix-3024: Process JSON View records using JSON.stringify and JsonSourceMap since JsonSourceMap does not support replacer

mslosarek:
Convert JSON objects to a JSON save object using JSON.stringify and the existing replacer method


mslosarek:
FIx for https://github.com/beekeeper-studio/beekeeper-studio/issues/3024

not-night-but (approved):

PR #3032, opened 2025-04-18
Upgrade SQLTextEditor to use CM6 with LSP support

azmy60:
# CM5 -> CM6 Upgrade in UI Kit

## What's in this PR?

- Upgraded SQLTextEditor and TextEditor components to use CodeMirror 6 (previously CodeMirror 5) in UI Kit.
- Added support for the Language Server Protocol (LSP).
- Kept backward compatibility for common props (not all yet) while adding new ones.
- Kept the old text editor components using CodeMirror 5 in v1 folder (I'm planning to remove this once all props are implemented)
- Included docs on the upgraded text editor and LSP integration.
- Old text editor in `v1` folder. New text editor in `v2` folder.
- Export `<bks-text-editor>` as a component . Previously we have SQL Text Editor but no Text Editor.

## What happens after merging?

- Tab query editor will be affected.
- Other components are not affected because only query editor uses UI Kit
- Enabling vim mode will work but configuring it won't work yet (sorry Day!).

This table compares the props between TextEditor v1 and v2, showing which props are implemented, deprecated, or new.

## Base TextEditor Props

| Prop | v1 | v2 | Notes |
|------|----|----|-------|
| value | ✅ | ✅ | String, default: "" |
| readOnly | ✅ | ✅ | Boolean, default: false |
| height | ✅ | ⚠️ | Deprecated in v2, use CSS instead |
| focus | ✅ | ✅ | Boolean |
| forceInitialize | ✅ | ✅ | Triggers initialization |
| keymap | ✅ | ✅ | v1: "default", "vim", "emacs", "sublime"<br>v2: "default", "vim", "emacs" (removed "sublime") |
| lineWrapping | ✅ | ✅ | Boolean |
| lineNumbers | ✅ | ✅ | Boolean, default: true |
| keybindings | ✅ | ✅ | Object |
| contextMenuItems | ✅ | ✅ | Array or Function |
| mode | ✅ | ❌ | Replaced by languageId in v2 |
| hint | ✅ | ❌ | Replaced by languageId in v2 |
| vimConfig | ✅ | ❌ | Not implemented yet in v2 |
| markers | ✅ | ❌ | Not implemented yet in v2 |
| autoFocus | ✅ | ❌ | Not implemented yet in v2 |
| foldGutter | ✅ | ❌ | Not implemented yet in v2 |
| vimKeymaps | ✅ | ❌ | Not implemented yet in v2 |
| clipboard | ✅ | ❌ | Not implemented yet in v2 |
| languageId | ❌ | ✅ | New in v2, replaces mode/hint, type: "plaintext" or "sql" |
| lsConfig | ❌ | ✅ | New in v2, Language Server Configuration. |
| replaceExtensions | ❌ | ✅ | New in v2, for extending functionality |

## SqlTextEditor Specific Props

| Prop | v1 | v2 | Notes |
|------|----|----|-------|
| entities | ✅ | ✅ | Array of entities for autocompletion |
| defaultSchema | ✅ | ✅ | String, default: "public" |
| columnsGetter | ✅ | ✅ | Function for getting columns by entity name |
| formatterDialect | ✅ | ✅ | String, default: "sql" |
| identifierDialect | ✅ | ✅ | String, default: "generic" |

## Supported LSP Features in Beekeeper Studio

- ✅ Syntax highlight based on semantic tokens
- ✅ Code completion
- ✅ Diagnostics - a.k.a. syntax validation
- ✅ Formatting

The text editor in **Beekeeper Studio** supports the following Language Server Protocol (LSP) features:

### Document Formatting
- `textDocument/formatting` — Full document formatting  
- `textDocument/rangeFormatting` — Formatting of selected text ranges

### Semantic Tokens
- `textDocument/semanticTokens/full` — Full semantic token information  
- `textDocument/semanticTokens/full/delta` — Delta updates for semantic tokens  
- `textDocument/semanticTokens/range` — Semantic tokens for a specific range

#### Other Text Document Methods
- `textDocument/didOpen`  
- `textDocument/didChange`  
- `textDocument/hover`  
- `textDocument/completion`  
- `completionItem/resolve`  
- `textDocument/definition`  
- `textDocument/codeAction`  
- `textDocument/rename`  
- `textDocument/prepareRename`  
- `textDocument/signatureHelp`  
- `textDocument/publishDiagnostics`

### ⚙️ Initialization Methods
- `initialize`  
- `initialized`


apps/ui-kit/lib/components/sql-text-editor/v2/SqlContextAnalyzer.ts
```diff
@@ -0,0 +1,172 @@
+import { EditorState, Text } from "@codemirror/state";
+import { syntaxTree } from "@codemirror/language";
+import { Entity } from "../../types";
+import getAliases from "./getAliases";
+
+export type ColumnsGetter = (entity: Entity) => string[] | Promise<string[]>
+
+export class SqlContextAnalyzer {
```
rathboma, line 8: Is this implementing a basic LSP? Or is this what we need customers to implement? Be good to add some comments to this file to outline what it does and why

rathboma: Does this file have any tests? Doesn't look like it?

apps/ui-kit/lib/components/sql-text-editor/v2/SqlTextEditor.vue
```diff
@@ -0,0 +1,82 @@
+<template>
+  <div class="BksUiKit BksTextEditor BksSqlTextEditor" ref="editor"></div>
```
rathboma, line 2: [nitpick] really with css you should use `bks-ui-kit` vs `BksUiKit`

apps/ui-kit/lib/components/sql-text-editor/v2/extensions/triggerAutocomplete.ts
```diff
@@ -0,0 +1,35 @@
...
+        }
+      );
+
+      if (spaceInserted) {
+        const cursor = update.state.selection.main.head;
+        const line = update.state.doc.lineAt(cursor);
+        const textBefore = line.text.slice(0, cursor - line.from);
+
+        // Check if we just typed a space after FROM or JOIN
+        if (/\b(FROM|JOIN)\s$/i.test(textBefore)) {
```
rathboma, line 28: Is this the same logic as in the Analyzer but for tables? Why is this in a different file? Also seems like while we're here we should add more statements to this list.

apps/ui-kit/lib/components/sql-text-editor/v2/utils.ts
```diff
@@ -0,0 +1,54 @@
+import { Entity } from "../../types";
+import { Completion } from "@codemirror/autocomplete";
+import getAliases from "./getAliases";
+
+export { getAliases };
+
+/**
+ * Convert column names to auto completion options
+ */
+export function columnsToCompletions(columns: string[]): Completion[] {
```
rathboma, line 10: tests?

rathboma (approved):
Looks good overall.

Some requests:
- Testing?
- copilot has some ideas
- Should all similar logic be in the same place?

Bigger question:
- We provide the SQL Analyzer, but a user could also add a LSP. Would the LSP replace the analyzer?
    - Like for Beekeeper Studio, if we provide a Postgres LSP, but we don't have one yet for MySQL - what happens?

apps/ui-kit/lib/components/sql-text-editor/v2/SqlContextAnalyzer.ts
azmy60: This implements the autocomplete for the `entities` prop. Mainly for analyzing the query so it can autocomplete columns after the SELECT keyword, and trigger autocomplete after FROM or JOIN keyword. Forgot to give this file a comment.

apps/ui-kit/lib/components/sql-text-editor/v2/SqlTextEditor.vue
azmy60: This was inspired by Material UI honestly lol. They name their components like `MuiBase-root`.

apps/ui-kit/lib/components/sql-text-editor/v2/SqlContextAnalyzer.ts
azmy60: wait no, the first part is correct. The second part about FROM or JOIN is not handled here. This exists mainly because: CodeMirror's sql extension can't show column completion after SELECT keyword.

So if you have: `SELECT | FROM users;` or `SELECT | FROM users u;` with your cursor in between the SELECT and FROM, it doesn't show you the columns of users;

apps/ui-kit/lib/components/sql-text-editor/v2/extensions/triggerAutocomplete.ts
azmy60: It's different. This is only for triggering autocomplete like after typing FROM or JOIN like how our text editor was. I should've not named the other file Sql Analyzer really. Claude was hallucinating.

azmy60:
@rathboma I'll merge this so we can see the updated docs. I'll address your feedbacks in a follow up PR.

apps/ui-kit/lib/components/sql-text-editor/v2/extensions/triggerAutocomplete.ts
azmy60: Forget my comment. You are right! :D

azmy60:
@rathboma just want to answer your questions here. 

> We provide the SQL Analyzer, but a user could also add a LSP. Would the LSP replace the analyzer?
Like for Beekeeper Studio, if we provide a Postgres LSP, but we don't have one yet for MySQL - what happens?

First, with a follow up PR, it shouldn't really be called SQL Analyzer (my mistake), it's really just about a "context aware" autocomplete feature, we had this in our original query editor but kind of placed everywhere, so anything related to that is now in a file called `sqlContextAutocomplete.ts`.

> Would the LSP replace the analyzer?

Simple answer, yes. I tried one postgres LSP and it turns out only showing tables and columns and no keywords. So it replaces the analyzer, but other completions like keywords were not provided by the LSP, so we still need the SQL Text Editor in that sense just to have the keywords.

> If we provide a Postgres LSP, but we don't have one yet for MySQL - what happens?

If we provide a Postgres LSP, then we should not pass `entities` to the text editor to avoid table and column duplicates. If we don't have one yet for MySQL, then just use the SQL Text Editor like normally. Don't activate the LSP.

PR #3047, opened 2025-04-23
Fix composite keys and colliding table name keys

### LINKED ISSUE #3018: BUG: Foreign keys not separated by schema
**Describe the bug**
When 2 schemas have identical sets of tables and they have a foreign key with the same name, both foreign keys are shown in the Relations tab for either table.

**To Reproduce**
Define two schemas with an identical foreign key relation.

```
-- Create schemas
CREATE SCHEMA shop1;
CREATE SCHEMA shop2;

-- Create tables in shop1
CREATE TABLE shop1.orders (
    id SERIAL PRIMARY KEY
);

CREATE TABLE shop1.items (
    id SERIAL PRIMARY KEY
    order_id INTEGER NOT NULL REFERENCES shop1.orders(id)
);

-- Create tables in shop2
CREATE TABLE shop2.orders (
    id SERIAL PRIMARY KEY
);

CREATE TABLE shop2.items (
    id SERIAL PRIMARY KEY
    order_id INTEGER NOT NULL REFERENCES shop2.orders(id)
);
```

**Expected behavior**

Relations tab should only show relations actually relevant to that Table - a foreign key defined on a table in another schema should not be included.

**Screenshots / Gifs**

This is the Relations tab for shop1.items:
Image

 - OS: Arch Linux
 - App Version: 5.1.5
 - Database type and version: Postgres 17

### LINKED ISSUE #3021: Foreign key constraint leads to the "wrong" table
**Describe the bug**
The situation is a bit more nuanced than the title might imply, I understand why this happens. Nonetheless, it had met panicked for a bit because I thought I had a faulty foreign key relation in my production database.

What happens? When I click on a column with a foreign relation, beekeeper leads me to the foreign record. In some situations where both a simple and a compound foreign key exists, it leads to the wrong record. Or to be more precise, to no record. Because it uses only the column value in the foreign table, while a compound key uses by definition multiple values.

**To Reproduce**
I had to fiddle around a bit to make a demo situation, because not every compound foreign key triggers this bug. For example when I made two foreign key constraints, one with just an int to an INT and one compound key using two INTs, everything worked as expected. 

```
/* Clean slate --------------------------------------------------- */
DROP TABLE IF EXISTS source;
DROP TABLE IF EXISTS not_a_target;
DROP TABLE IF EXISTS target;

/* 1) Primary table the single‑column FK *should* point to -------- */
CREATE TABLE target (
  id   int(10) unsigned NOT NULL AUTO_INCREMENT,
  name varchar(255)     NOT NULL,
  PRIMARY KEY (id)
) ENGINE=InnoDB
  DEFAULT CHARSET=utf8mb4
  COLLATE=utf8mb4_unicode_520_ci;

/* 2) Table with a composite PK that re‑uses target_id ------------ */
CREATE TABLE not_a_target (
  target_id int(10) unsigned NOT NULL,
  ref       varchar(255)     NOT NULL,
  description varchar(255),
  PRIMARY KEY (target_id, ref),

  CONSTRAINT fk_nat_target
    FOREIGN KEY (target_id)
      REFERENCES target(id) ON DELETE CASCADE
) ENGINE=InnoDB
  DEFAULT CHARSET=utf8mb4
  COLLATE=utf8mb4_unicode_520_ci;

/* 3) Source table that contains BOTH FKs ------------------------- */
CREATE TABLE source (
  id        int(10) unsigned NOT NULL AUTO_INCREMENT,
  target_id int(10) unsigned NOT NULL,
  ref       varchar(255) DEFAULT NULL,
  payload   varchar(100),
  PRIMARY KEY (id),

  /* ---------- composite FK first (target_id, ref) --------------- */
  CONSTRAINT a_fk_source_targetid_ref
    FOREIGN KEY (target_id, ref)
      REFERENCES not_a_target(target_id, ref),

  /* ---------- single‑column FK second (target_id) --------------- */
  CONSTRAINT b_fk_source_targetid
    FOREIGN KEY (target_id)
      REFERENCES target(id) ON DELETE CASCADE
) ENGINE=InnoDB
  DEFAULT CHARSET=utf8mb4
  COLLATE=utf8mb4_unicode_520_ci;

/* Seed rows so there’s something to click on -------------------- */
INSERT INTO target (id, name) VALUES (1, 'row in target');

INSERT INTO not_a_target (target_id, ref, description)
VALUES (1, 'BTC', 'row in not_a_target');

INSERT INTO source (id, target_id, ref, payload)
VALUES (100, 1, 'BTC', 'row in source');
```

1. Go to the source table
2. Click on the icon in the target_id column 

Image

3. You will be in the not_a_target table, with a filter on just target_id.

**Expected behavior**
One of two things I would have expected:
* I would have expected to be in the target table, with a filter on target_id
* Or, we would be in the not_a_target table with a filter on both target_id AND ref.

Arguably, because you are clicking on target_id, I would expect that one has priority over the compound key.

 - OS: [Ubuntu]
 - App Version: [5.1.5]
 - Database type and version [MariaDB]

**Additional context**

not-night-but:
fix #3021 
fix #3018 

Some of our queries weren't taking schema into account, so we ended up with more keys being pulled in if there was a collision of table names.

For composite keys, looks like we just never adjusted the query to actually handle them at all.

### DIFF
--- a/apps/studio/src-commercial/backend/lib/db/clients/anywhere.ts
+++ b/apps/studio/src-commercial/backend/lib/db/clients/anywhere.ts
@@ -465,7 +465,8 @@ export class SQLAnywhereClient extends BasicDatabaseClient<SQLAnywhereResult> {
         c_for.column_name AS from_column,
         CAST(fk.foreign_table_id AS VARCHAR) + '_' + CAST(fk.primary_table_id AS VARCHAR) AS constraint_name,
         'NO ACTION' AS on_update,
-        'NO ACTION' AS on_delete
+        'NO ACTION' AS on_delete,
+        ic_for.sequence AS sequence
       FROM SYS.SYSFKEY fk
       JOIN SYS.SYSIDXCOL ic_for ON fk.foreign_index_id = ic_for.index_id AND ic_for.table_id = fk.foreign_table_id
       JOIN SYS.SYSCOLUMN c_for ON ic_for.table_id = c_for.table_id AND ic_for.column_id = c_for.column_id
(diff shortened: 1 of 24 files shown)


rathboma (approved):
Wow, this ended up being a massive amount of code, I thought it would just be a few lines. Good stuff! Great tests too.

Issue #2804, opened 2025-02-25
Improve Query History to Avoid Duplicate Entries


rathboma:
Feature Request: Enhance the query history functionality by preventing duplicate entries for unchanged SQL queries. This improvement is necessary as the current system becomes cluttered and difficult to use when repeatedly executing the same query without modifications. A mechanism should be implemented to check if the executed query is identical to the previous one, and in such cases, avoid creating a new entry.
***

*This issue was extracted from an email thread by [IssueWizard](https://issuewizard.com)*

PR #3070, opened 2025-04-28
fix: avoid history duplication if current and last query are same

therealrinku:
fixes: #2804 

before:

https://github.com/user-attachments/assets/835ac422-8008-4bbd-9a05-f4e9bf22aaae

after:

https://github.com/user-attachments/assets/973cb260-6abf-495a-8a74-cb9eaf6e8898

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -955,7 +955,13 @@
           console.log("non empty result", nonEmptyResult)
           this.selectedResult = nonEmptyResult === -1 ? results.length - 1 : nonEmptyResult
 
-          this.$store.dispatch('data/usedQueries/save', { text: query, numberOfRecords: totalRows, queryId: this.query?.id, connectionId: this.usedConfig.id })
+          const lastQueryText = this.$store.state['data/usedQueries']?.items[0]?.text
+          const isDuplicate = lastQueryText === query
+          
+          if(!isDuplicate){
+            this.$store.dispatch('data/usedQueries/save', { text: query, numberOfRecords: totalRows, queryId: this.query?.id, connectionId: this.usedConfig.id })
+          }
+          
           log.debug('identification', identification)
           const found = identification.find(i => {
             return i.type === 'CREATE_TABLE' || i.type === 'DROP_TABLE' || i.type === 'ALTER_TABLE'


apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -955,7 +955,13 @@
           console.log("non empty result", nonEmptyResult)
           this.selectedResult = nonEmptyResult === -1 ? results.length - 1 : nonEmptyResult
 
-          this.$store.dispatch('data/usedQueries/save', { text: query, numberOfRecords: totalRows, queryId: this.query?.id, connectionId: this.usedConfig.id })
+          const lastQueryText = this.$store.state['data/usedQueries']?.items?.[0]?.text
+          const isDuplicate = lastQueryText === query
+          
+          if(!isDuplicate){
```
rathboma, line 961: should we change the date of the query do you think? So we know WHEN it was last run?

therealrinku: that make sense. updated!

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -969,7 +969,23 @@
           console.log("non empty result", nonEmptyResult)
           this.selectedResult = nonEmptyResult === -1 ? results.length - 1 : nonEmptyResult
 
-          this.$store.dispatch('data/usedQueries/save', { text: query, numberOfRecords: totalRows, queryId: this.query?.id, connectionId: this.usedConfig.id })
+          const lastQuery = this.$store.state['data/usedQueries']?.items?.[0]
+          const isDuplicate = lastQuery?.text === query
```
not-night-but: I do wonder about this check, as adding more whitespace would make this return false, no?

Should we maybe trim and format both sides before comparing? Or do we think a naive check like this is good enough? @rathboma thoughts?

not-night-but (commented):
I just had one question, we'll see if Matt can weigh in. 

Also, so sorry this seems to have flown under the radar for a bit.

apps/studio/src/components/TabQueryEditor.vue
therealrinku: hey @not-night-but! thanks for checking it out!
that make sense, I've updated the code to add the trim before comparing the queries!

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -967,7 +967,23 @@
...
+
+          const queryObj = { 
+            text: query, 
+            numberOfRecords: totalRows, 
+            queryId: this.query?.id, 
+            connectionId: this.usedConfig.id 
+          }
+
+          if(lastQuery && isDuplicate){
+            queryObj.createdAt = new Date();
```
rathboma: Why are we setting `createdAt`? We should be setting `updatedAt`. CreatedAt - First time the query was run, updatedAt - last time it was run

rathboma: Let's make sure we keep with conventions. Might mean the used query list needs ordering by updatedAt instead of createdAt

therealrinku: updated!

not-night-but (approved):
Cool, thanks!

PR #3076, opened 2025-04-28
Fix add database

### LINKED ISSUE #3067: BUG: The feature help -> add beekeeper's database is not working (Version 5.1.5)
The feature help -> add beekeeper's database is not working (Version 5.1.5)
Image

original comment: https://github.com/beekeeper-studio/beekeeper-studio/issues/1645#issuecomment-2831704365

not-night-but:
fix #3067 

The default transform function wasn't checking if the received object was null/undefined, so it was constructing an empty object, which then caused things in the frontend to think results were being returned when they weren't.


rathboma (approved):

rathboma:
@not-night-but should this be against 52?

Issue #3079, opened 2025-04-28
BUG: MongoDB (and others) don't use transcoders on query results


not-night-but:
![Image](https://github.com/user-attachments/assets/6ea55967-7a66-4a11-af24-34eef4139025)
We are currently displaying ObjectIds in query results incorrectly. Instead of `ObjectId('<ID>')` we are displaying a JSON object.

This is likely also an issue with other dbs, as we don't run transcoders on query results.

azmy60:
Yeah, we don't use transcoders when querying arbitrary sql or in query editor yet. The reasons were:

1. When querying, it's hard to get the right info to say what `bksType` this is. This isn't an issue with `selectTop` because we are able to use "schema_information" very easily.
2. It's hasn't been proven but I'm afraid there'll be performance degradation to this too, because imagine serializing 1000 rows or 10K rows times 1 or 3 columns for a single query, though I'm not very sure that's a common thing to do. Again, not an issue with table view because we use pagination. Table view has better control on the query creation so it can determine the data limit.

@rathboma I need your thoughts on this cause I feel like one quick solution to this is by fixing the formatter instead of the transcoder, which we have in this [PR](https://github.com/beekeeper-studio/beekeeper-studio/pull/3082).

Other databases currently have no issues with query result that has binary type values because of the formatter.

![Image](https://github.com/user-attachments/assets/34b4e8e6-7dd2-4ba0-9b3b-c4a8af9754f0)

The ultimate solutions to this are possibly one of these:
1. As suggested by Day, we can make a new class for transcoding query results
2. We can skip converting in transcoder but still carry the `bksType` to frontend. We do this if we want to address the concern about performance, and electron uses `structuredClone` by default, which makes it possible to pass buffer objects or even ObjectID. https://www.electronjs.org/docs/latest/api/ipc-renderer/#ipcrenderersendchannel-args

PR #3082, opened 2025-04-29
Fix objectid formatter

azmy60:
This does not fix the serializer but fixes the format of ObjectID in table result and json viewer.

Before:
image

After:
image

One commit for json viewer was cherry picked from a PR that was merge in master. https://github.com/beekeeper-studio/beekeeper-studio/pull/3025


azmy60:
This is supposed to be a very quick fix https://github.com/beekeeper-studio/beekeeper-studio/issues/3079, not a final solution.

rathboma:
hacks are fine for now, def need fixing long term though

Issue #3101, opened 2025-05-02
FEAT: Copy Column Name from Data View


JeanPaiva42:
**Is your feature request related to a problem? Please describe.**
From a customer:
"When working in the data view on Ubuntu, it’s currently difficult or nearly impossible to copy just the column name. Right now, the only available right-click action copies the entire column’s data, which makes it inconvenient when I just need the column header."

![Image](https://github.com/user-attachments/assets/ca7e99d2-5805-4e4b-9303-87469d875221)

**Describe the solution you'd like**

Add a right-click context menu on the column header in the data view with an option to "Copy Column Name", or when copying the data, the option to "Copy with column header". This would allow users to quickly copy just the name of the column without having to manually retype or extract it in other ways.

**Describe alternatives you've considered**

Currently, users might try to manually retype column names or use a SQL query to get metadata, which is unnecessarily time-consuming for such a simple task.

**Database type, if appropriate**
Applies to all databases

PR #3103, opened 2025-05-02
feat: add copy column name right click menu

therealrinku:
closes: #3101 

added "Copy Column Name" menu on query result view and table view.

- if user selects single column , it copies that column name
- if user selects multiple columns, it copies all those selected column names

https://github.com/user-attachments/assets/1bc12916-d987-4309-b6cb-ab5b701d848e


not-night-but (approved):

not-night-but:
Thanks again, nice little convenience feature!

PR #3113, opened 2025-05-05
Fix Mongo Filtering

### LINKED ISSUE #3080: BUG: MongoDB TableTable filtering doesn't work with all datatypes.
FIltering is not working for every type of data. Currently we are able to do it for string types, but for booleans and for ObjectIds, for example the filtering will not return results. 
Repro Steps:
1. Connect to mongoDB 
2. On a table with booleans, objectIds and string, try to filter for the boolean or the ObjectId
3. Notice that no items are returned even though the filter is valid in theory.

not-night-but:
Fix #3080 

Fix some type issues with mongodb filtering in table table.


rathboma (approved):

rathboma:
merge when good

PR #3118, opened 2025-05-06
fix title bar position

azmy60:
- Windows/Linux
titlebar-shrink
- Mac
titlebar-shrink-mac


azmy60:
titlebar-shrink-vscode

I was thinking of making it work like vscode, which to shrink the menu instead of the title, but it requires larger changes with how the app menu works right now.

rathboma (approved):

PR #3124, opened 2025-05-07
add json & html support for text editor v2

azmy60:


MiriamCall:
I think this change is great.

rathboma (approved):

Issue #3126, opened 2025-05-07
BUG: No way to specify profile from AWS credentials file


Ryanjso:
I just purchased the professional plan for $168, and many of my teammates were going to follow suit. Unfortunately the issue describes below is blocking us and I will need to cancel my subscription if this can't be resolved. 

**Describe the bug**
When signing in with an AWS credentials file `~/.aws/config` (Postgres RDS), there is no way to distinguish which profile from the credentials file to use. 

**To Reproduce**
1. Select Postgres from connection type
2. select IAM using Credentials File from Auth method
3. Note an `IAM Authentication` box will appear below with a profile field, but this field does not seem to read from the credentials file and is instead used for something else. I can confirm these because I am able to connect to my dev database even when I type random stuff in here.

**Expected behavior**
A way to clarify which profiles credentials to use

**Screenshots / Gifs**
If applicable, add screenshots or GIFS to help explain your problem.

**Version Information (please complete the following information):**
 - OS: [e.g. Windows/Mac/Ubuntu]
 - App Version [e.g. help -> About Beekeeper Studio]
 - Database type and version [eg Posgresql 9.3]

**Additional context**
Anything else you think might be helpful

rathboma:
Hey Ryan,

Thanks for reporting this, we'll get right on it and get a patch release out this week with the fix.

rathboma:
@Ryanjso I'm pushing 5.2.4 right now, should be available in ~1 hour. Can you let me know if it solves the problem, and if so, close this issue?

rathboma:
Huh, that's interesting, so prod and dev use a different profile is that right?

So it's "remembering" the first profile somehow and not switching even when you ask it to? Is that right?

Can you confirm this is the broken workflow:

1. Connect to prod, all good
2. Disconnect prod
3. Connect to dev, doesn't work
4. Restart beekeeper
5. Connect to dev, all good

Ryanjso:
> Huh, that's interesting, so prod and dev use a different profile is that right?
> 
> So it's "remembering" the first profile somehow and not switching even when you ask it to? Is that right?
> 
> Can you confirm this is the broken workflow:
> 
> 1. Connect to prod, all good
> 2. Disconnect prod
> 3. Connect to dev, doesn't work
> 4. Restart beekeeper
> 5. Connect to dev, all good

Yes that's correct

Ryanjso:
[...]

PR #3128, opened 2025-05-07
fix: Updated signer to check for region and ensure profile is used

ClarLabs:
Related: https://github.com/beekeeper-studio/beekeeper-studio/issues/3126


not-night-but:
@ClarLabs this needs to be against rc-52, not master

PR #3131, opened 2025-05-07
fix: Allows region retrieval from shared config and ensures profile set

ClarLabs:
Related to https://github.com/beekeeper-studio/beekeeper-studio/issues/3126


not-night-but:
@ClarLabs weird thing, the awsProfile doesn't show up for me until I make changes in the iam form:
image
(Also looking at the implementation, I don't think the keys are supposed to be there?)

rathboma:
@ClarLabs why does is ask for key/secret when you select 'credentials file'?

It shouldnt. Those should only be required if not using a credentials file, right?

Issue #2836, opened 2025-03-04
FEAT: Support a common theme spec - VSCode or Textmate/Sublime


rathboma:
As we're implementing theme plugins we should consider existing theme specifications. Providing support would give us access to thousands of themes for free, right away.

Sublime Text did this with TextMate themes back in the day, it was awesome. We should do the same!

I don't know which spec is better to support - vscode or ST/TM, but we should support something standard for sure.

Here's an example repo with 300 themes for both ST and TM - https://github.com/rainglow/sublime

dxps:
Yes, please! 🙏 

I just bought the Pro license, thus I intend to seriously use it on a daily basis.

But I'd really prefer another flavor of a dark theme, something Gruvbox inspired one.
And I was wondering if I can contribute to this.

PR #3137, opened 2025-05-08
feat: new theme manager system (#2836)

zanuka:
# notes 

* **dedicated theme service architecture:**
  * created specialized theme service (`initializeThemeService`) and content handlers (`setupThemeContentHandlers`) in the `src/background` directory
  * integrated with Electron's main process for proper initialization during application startup
  * uses a modular approach separating service initialization from theme application

* **theme asset organization:**
  * organized theme assets under structured directories (`src/assets/styles/themes/*`)
  * implemented multiple theme variations (dark, light, solarized-dark, solarized-light)
  * created dedicated variable files to separate theme configuration from application styling

* **app-wide theme integration:**
  * modified the main process (`main.ts`) to initialize themes during app bootstrap
  * added proper sequencing of theme initialization after menu setup and before window creation
  * ensured themes apply consistently across all UI components

* **component-Level theme compatibility:**
  * enhanced various components to properly respond to theme changes
  * modified UI elements to use theme-aware styling variables
  * updated CSS to ensure consistent theming across the application

* **development & testing considerations:**
  * maintained backward compatibility with existing components
  * ensured theme changes don't interfere with core application functionality
  * preserved the existing project structure while adding new theming capabilities

* **performance optimizations:**
  * deferred theme initialization until after core application services are ready
  * used asynchronous loading pattern to minimize impact on startup time
  * designed theme system to only update necessary UI elements when themes change

I tried my best to follow best practices by separating concerns (service vs. handlers), maintaining your existing code patterns, and ensuring backward compatibility with the codebase.

feedback welcome! ...and we could also look to adding more themes or pulling from a themes api :-)


zanuka:
https://github.com/user-attachments/assets/ed409381-1163-4e30-af19-a2a3ecdd3414

(demo)

apps/studio/src/App.vue
```diff
@@ -42,194 +33,538 @@
...
 
-    if (this.url) {
-      try {
-        await this.$store.dispatch('openUrl', this.url)
-      } catch (error) {
-        console.error(error)
-        this.$noty.error(`Error opening ${this.url}: ${error}`)
-        throw error
+      // Also listen for window events
+      window.addEventListener("theme-preview", (event) => {
```
rathboma, line 220: You should use our Vue event bus for this - $this.$root.emit, and `registerHandlers`, along with `AppEvent`

apps/studio/src/App.vue
```diff
@@ -42,194 +33,538 @@
...
+      this.licenseInterval = setInterval(
+        () => this.$store.dispatch("licenses/updateAll"),
+        globals.licenseCheckInterval
+      );
 
+      // Initialize settings
+      await this.$store.dispatch("settings/initializeSettings");
 
-    this.$nextTick(() => {
-      window.main.isReady();
```
rathboma, line 157: Can we move all of this theme logic into it's own component? Maybe <theme-manager>?

zanuka:
@azmy60 I left in console logs for testing and can remove after feedback is addressed

apps/studio/src/App.vue
```diff
@@ -42,194 +33,538 @@
...
+        (t) => t.id === themeId
+      );
+
+      if (!theme) {
+        console.error(`App.vue: Theme not found in store: ${themeId}`);
+        return;
+      }
+
+      // Special handling for built-in themes that have their own SCSS files
+      const baseThemes = [
```
rathboma, line 296: Should these themes be migrated to your new system?

apps/studio/src/background/theme-handlers.ts
```diff
@@ -0,0 +1,267 @@
...
+
+    // Inject the CSS
+    const key = await webContents.insertCSS(css, { cssOrigin: 'user' });
+    console.log(`CSS injected for theme ${themeName} with key ${key.substring(0, 10)}...`);
+
+    // Store the key for later removal
+    themeKeys[`${webContents.id}_${themeName}`] = key;
+
+    // Also set the theme class directly via JavaScript
+    await webContents.executeJavaScript(`
```
rathboma, line 191: This doesn't seem like something we should be doing in a handler. Can the CSS and theme name not be passed back to the front end, and we have the front-end apply it?

rathboma (changes requested):
I haven't done a full review yet, but I do have questions about how we're mutating webContents from the electron side. I think it's better to let the front-end deal with applying styles.

not-night-but:
@zanuka This is so cool! I also had a bit of a poke around and I think maybe something went wrong with your rebase? Non beekeeper themes don't seem to be working properly (they all end up with something halfway between dracula and our dark theme).

zanuka:
@not-night-but thanks... glad you're testing it out and I'll await more feedback and then make updates. yes the the theme colors need to be more accurate for sure

apps/studio/src/background/theme-handlers.ts
zanuka: totally

apps/studio/src/App.vue
zanuka: yes indeed, good catch

apps/studio/src/App.vue
zanuka: ok cool, will get that sorted out

apps/studio/src/App.vue
zanuka: 👍🏼

rathboma:
Also, funny bug: when changing to a light theme it breaks the theme manager, because it stays half dark, half light.

I suspect there are other components that aren't changing either, like the json sidebar only partially restyles, and the table table doesnt seem to at all

apps/studio/src/background/theme-parsers.ts
```diff
@@ -0,0 +1,528 @@
...
+      foreground?: string;
+      background?: string;
+      fontStyle?: string;
+    }
+  }>;
+}
+
+export function setupThemeContentHandlers() {
+  // Apply a theme to a window
+  ipcMain.handle('themes/apply', async (event, { name }) => {
```
azmy60, line 68: We do this stuff via utility process. For example, please look at the [handlers/fileHandlers.ts](https://github.com/beekeeper-studio/beekeeper-studio/blob/master/apps/studio/src/handlers/fileHandlers.ts).

So instead of this:
```js
export function setupThemeContentHandlers() {
    ipcMain.handle('themes/apply', () => {})
    ipcMain.handle('themes/removeActive', () => {})
    ipcMain.handle('themes/apply', () =>{})
}
```

We should really do this:

```js
// app/studio/src/handlers/themeHandlers.ts

export const ThemeHandlers: IThemeHandlers = {
  "themes/apply": async function (...): Promise<void> {
    // apply theme here
  },
  "themes/removeActive": ...,
  "themes/register": ...,
}
```

And then register the `ThemeHandlers` in [utility.ts](https://github.com/beekeeper-studio/beekeeper-studio/blob/64f885acfd12745a8dbe06d39b19d4eb4f9c3eca/apps/studio/src-commercial/entrypoints/utility.ts#L42):
```js
export const handlers: Handlers = {
  // other handlers ...
  ...LicenseHandlers,
  ...TabHistoryHandlers,
  ...ThemeHandlers, // <--- Theme handlers here
  ...(platformInfo.isDevelopment && DevHandlers),
};

```

To call the handler from the frontend, you can use `this.$util.send("themes/register")` or `Vue.prototype.$util.send("themes/register")`. There is a doc about utility process here: https://github.com/beekeeper-studio/beekeeper-studio/wiki/Utility-Process.

Also, most of the codes in `handleFileUpload()`  -> `ThemesManagerModal.vue` can be placed inside a new utility handler if you want. After that, you might just need to call `this.$util.send("themes/import", filePath)` inside the modal.

azmy60 (changes requested):
@zanuka This is a massive PR! Thank you! I have one concern about the theme palette overall but I'm not going to talk about it now cause we'll need to think about it more carefully.

And sorry I had to nitpicks here and there lol. There is a lot of changes caused by the linter? I don't usually nitpick this :D cause many times they are small. It'll just clutter the git history, but I'm fine if you want to leave it too, as long as it works :)

zanuka:
@azmy60 thanks for the thorough review! I should be able to get things updated over the next day or two and agree with your feedback. Can fix the linter errors as well.

zanuka:
been a busy week @azmy60 ... trying to carve out some time to wrap up these changes today

apps/studio/src/App.vue
zanuka: updated

apps/studio/src/App.vue
zanuka: the original ones now work with the new ones

apps/studio/src/types/global.d.ts
```diff
@@ -0,0 +1,12 @@
+interface Window {
+  showThemeManagerModal?: () => void;
```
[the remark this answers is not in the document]
zanuka: updated

apps/studio/src/store/modules/ui.ts
```diff
@@ -0,0 +1,27 @@
+import { Module } from 'vuex';
+
+interface UIState {
+  showThemeManager: boolean;
+}
+
+const uiModule: Module<UIState, any> = {
+  namespaced: true,
+  state: () => ({
+    showThemeManager: false,
```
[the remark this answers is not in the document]
zanuka: removed

apps/studio/src/components/settings/ThemeManagerModal.vue
```diff
@@ -0,0 +1,1474 @@
...
+  padding: 15px 20px;
+  border-bottom: 1px solid rgba(255, 255, 255, 0.1);
+}
+
+.theme-manager-header h3 {
+  margin: 0;
+  font-size: 18px;
+}
+
+.close-button {
```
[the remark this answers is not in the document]
zanuka: fixed

apps/studio/src/components/settings/ThemeManagerModal.vue
```diff
@@ -0,0 +1,1474 @@
...
+}
+
+.theme-manager-footer {
+  padding: 15px 20px;
+  border-top: 1px solid rgba(255, 255, 255, 0.1);
+  display: flex;
+  justify-content: flex-end;
+}
+
+.tabs {
```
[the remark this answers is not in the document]
zanuka: fixed

apps/studio/src/background/theme-parsers.ts
zanuka: updated

zanuka:
note to reviewers: I have a few more updates to make so that this is all working properly

zanuka:
@azmy60 I am pushing some new updates soon with a simplified approach to toggling the new theme manager modal. Due to the complexities of handling it from the MenuItems component through IPC, I think having a simple button on the bottom right by the Avatar icon will be a cleaner approach. In addition, I'm just going to add the 5 standard themes to the list of themes that appears in the modal.

zanuka:
@rathboma @azmy60 @not-night-but here's a preview of the new version, updated code to be pushed soon so disregard the current files since there's a lot of cleanup I'm doing

https://github.com/user-attachments/assets/5e60547b-8951-415b-94ad-0c135d9a03a2

zanuka:
please pull branch and test thoroughly, I just added more cleanup and enabled theme persistence so user can close/open app and their selected theme will load.

rathboma:
Hey @zanuka! This is a lot of awesome work and love the direction you've taken this visually.

Given the extent of changes, and the mess our current theme system is in, I just don't think I can merge this. I know that's not want you want to hear, but honestly I think unless we make some significant changes to how the app handles theming (moving to a cleaner set of css variables), this will just become unmaintainable.

I'm going to close this PR, but have plans to re-open the work in a month or two once we have our themes organized a bit better

zanuka:
all good, makes sense to me... was fun getting it to work anyway and I'll stay tuned

lundjrl:
hey @rathboma any update here? How can we help?

not-night-but:
@lundjrl this is still something we want to do for sure! It is just going to require a ton of sweeping changes to all of the styles in the app. I think over the years a lot of the styles have become fragmented and we really just need to standardize a lot of things so that this system would become maintainable

lundjrl:
> @lundjrl this is still something we want to do for sure! It is just going to require a ton of sweeping changes to all of the styles in the app. I think over the years a lot of the styles have become fragmented and we really just need to standardize a lot of things so that this system would become maintainable

PR's welcome for this?? 😄

not-night-but:
@lundjrl Unfortunately I think this might be something we have to do ourselves. It's a pretty massive foundational change to the app with a lot of careful planning required from the team and probably some standardized guidelines to be written out and enforced across the project. Thank you for the offer though!

lundjrl:
@not-night-but  godspeed 🏃🏻

Issue #3138, opened 2025-05-08
BUG: 5.2.4 will not start on NixOS Linux


mkellyxp:
**Describe the bug**
For over a year, I've maintained my own Nix derivation based on the Appimage.  And it has always worked perfectly.  You can see the code here https://github.com/mkellyxp/nix/blob/main/common/beekeeper.nix

When I updated to 5.2.4, I now get an error when starting

```
A JavaScript error occurred in the main process
Uncaught Exception:
Error: EACCES: permission denied, copyfile '/nix/store/i8n3b01vksvvahax7yvg19rh6rja0w42-beekeeper-studio-5.2.4-extracted/resources/default.config.ini' -> '/home/mkelly/.config/beekeeper-studio/default.config.ini'
    at copyFileSync (node:fs:3034:11)
    at func (node:electron/js2c/node_init:2:2559)
    at copyBundledConfig (/nix/store/i8n3b01vksvvahax7yvg19rh6rja0w42-beekeeper-studio-5.2.4-extracted/resources/app.asar/dist/main.js:278367:31)
    at loadConfig (/nix/store/i8n3b01vksvvahax7yvg19rh6rja0w42-beekeeper-studio-5.2.4-extracted/resources/app.asar/dist/main.js:278384:5)
    at mainBksConfig (/nix/store/i8n3b01vksvvahax7yvg19rh6rja0w42-beekeeper-studio-5.2.4-extracted/resources/app.asar/dist/main.js:278431:25)
    at Object.<anonymous> (/nix/store/i8n3b01vksvvahax7yvg19rh6rja0w42-beekeeper-studio-5.2.4-extracted/resources/app.asar/dist/main.js:278458:15)
    at Module._compile (node:internal/modules/cjs/loader:1484:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1564:10)
    at Module.load (node:internal/modules/cjs/loader:1295:32)
    at Module._load (node:internal/modules/cjs/loader:1111:12)
```

When I delete the ~/.config/beekeeper-studio folder, the first time I try and run it, I get this, and then it hangs and never loads

```
[mkelly@garage:~/.config]$ beekeeper-studio
14:15:59.349 [MAIN] (BksConfig)           › Loading configs.
14:15:59.352 [MAIN] (BksConfig)           › Copying bundled config default.config.ini to /home/mkelly/.config/beekeeper-studio/default.config.ini.
[...]

PR #3141, opened 2025-05-09
fix NixOS extraResources

azmy60:
Other extra resources seem to work without issues. Don't know why config.ini fails the copy permission on NixOS. One obvious difference between config files and the other extra resources is the way we define them in the electron builder config, so I guess matching it with the other configs would work?

### DIFF
--- a/apps/studio/electron-builder-config.js
+++ b/apps/studio/electron-builder-config.js
@@ -55,10 +55,17 @@ module.exports = {
       to: 'public'
     },
     {
-      from: ".",
-      to: ".",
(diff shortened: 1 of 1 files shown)


azmy60:
@emaryn So this is from an issue #3138 . The error says `EACCES: permission denied, copyfile`. But that could also mean that the file does not exist, rather than lacks of copy permission?

If we look at #3159, it says that the file does not exist, which means that the app was trying to access the wrong file path, or the electron-builder did not bundle the config files correctly. So I made a new PR for this here #3160.

Issue #3147, opened 2025-05-09
BUG: cmd+D for row duplication not working


drochag:
**Describe the bug**
After updating to v5.2.5 cmd+D on a row no longer works

**Version Information (please complete the following information):**
 - OS: MacOS Ventura 13.7.5
 - App Version 5.2.5

rathboma:
The fix for this will be in the next release!

PR #3148, opened 2025-05-10
fix: clone row shortcut not working

therealrinku:
fixes: #3147 

Screenshot 2025-05-10 at 08 22 05


rathboma:
Hey! I actually have a fix for this in the rc-52 branch (where I'll be making a patch release from)

PR #3161, opened 2025-05-13
fix: close all not removing tabs from appdb

therealrinku:
fixes: #3185

issue was with the `close all` option only.  other close options seems to be working fine!

before:

https://github.com/user-attachments/assets/f595cd7c-6875-4bc2-94a0-7ba32fc2e482

after:

https://github.com/user-attachments/assets/30418c46-0fa8-4baa-ae75-abaabfde819c


rathboma (approved):
Thanks so much!

PR #3181, opened 2025-05-16
feat: support regex filter in json viewer

mrsimb:
I wanted to filter some keys by a pattern in a json viewer, and it seemed logical to use regex for that, so I added support for it.

Here's an example of only showing keys with "smart" or "gam" in them:


mrsimb:
Seems like a similar feature in entities list filter would be useful for [some people](https://github.com/beekeeper-studio/beekeeper-studio/discussions/1798)
I can add it there too.

rathboma (approved):
This is awesome, thanks so much! Yes please add patches for the other search boxes if you are willing :-)

Issue #3180, opened 2025-05-16
BUG: Utility Process crash for large result sets


bmercernccer:
**Describe the bug**
Normally when running a query that would return a large result set, BeeKeeper will instead return only the first 50K rows and report the total rows available. 

However when the result set is very large, I get a "Utility Process Crashed" error instead. It seems t

**To Reproduce**
1. Run a query which returns a large number of rows (in my case about 3 million rows of mixed text and number values
2. Wait for the query to finish
3. Utility Process Crashed error appears and session is broken

**Expected behavior**
1. Run a query which returns a large number of rows
2. Either
    Receive the first 50K rows with the total row count as happens for smaller results sets
    OR
    Receive a meaningful error explaining why the results can't be returned
3. Retain connectivity with the database

**Version Information (please complete the following information):**
 - OS: Windows 11
 - App Version 5.2.7
 - Database type and version Postgres 15 running on Debian 12

**Additional context**
I have not seen this error before, even though I have previously worked with similar numbers of rows on other databases. I believe the issue is not the number of rows but the size of the result set. I can run the same query successfully by removing some columns and making the results smaller.

At first I thought it was a timeout, but I have encountered the error after only 30-40 seconds in one case and after 60 seconds in another case so not sure. 

When running my specific query, I was able to use a limit to determine that the error started happening between 2 million rows and 3 million rows. 
By reducing the number of columns I could increase the successful row count. 

Some logs showing different runs attached. 

[logs.7z.zip](https://github.com/user-attachments/files/20258305/logs.7z.zip)

not-night-but:
@bmercernccer Thank you for the detailed report! I suspect that the utility node process is running out of memory.

Just some things I'm noticing right now: we don't truncate the result in the utility process. THIS IS BAD. This means we grab the entire resultset, send it to the frontend, and then truncate it. That's a lot of wasted memory. Second, we need a way to truncate the resultset in progress. Ideally we should be able to only get the first 50000 rows, and then just discard the rest of the query, because 3 million rows in the UI is just not a good idea, I suspect that tabulator would freak out at that point.

not-night-but:
![Image](https://github.com/user-attachments/assets/3f6f9a63-e00e-4de5-9d0a-4820c9772b4a)

Replicated. Yeah the util process is just running out of memory.

PR #3182, opened 2025-05-16
Util out of mem

not-night-but:
This just has the script that replicates #3180 right now. 

### TODO
- [ ] We need to make sure that we're truncating all result sets to the maxResults from the config BEFORE we send them to the frontend
- [ ] We probably need to figure out how to stream query results for selects.


bmercernccer:
Maybe you could do something with FETCH? It's pretty cross-platform.

PR #3199, opened 2025-05-22
add workflow to check for missing lodash imports

not-night-but:
We were missing a couple lodash imports, and since this seems to keep happening, I've added a workflow that checks for any files that use `_.something` but don't import lodash.

### DIFF
--- a/.github/workflows/studio-test.yml
+++ b/.github/workflows/studio-test.yml
@@ -59,6 +59,9 @@ jobs:
       - name: Check for bad log imports
         run: bin/check-for-electron-log-imports.sh
 
+      - name: Check for missing lodash imports
+        run: bin/check-lodash.sh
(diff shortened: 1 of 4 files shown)


rathboma:
shouldn't running `tsc --no-imit` do this?

not-night-but:
Maybe? I've noticed tsserver doesn't always report it as an error, hence why we end up with it happening as a runtime error. I'll check

not-night-but:
@rathboma Yeah I just looked into it, and a couple things: It doesn't actually seem to catch the missing lodash imports all the time, it seems to just assume it's a global or something? Also, it catches a lot more errors that aren't actually errors, so we'd have to do some cleanup type-system wise to actually get that to function how we'd want to put it in a workflow

Issue #2826, opened 2025-03-03
FEAT: Improve pagination/navigation


not-night-but:
Add a 'jump to last page' button and hide the next-page button when on the last page for improved navigation.

joaovicente3:
Hi, @not-night-but,
@SMV-21 and I are working on this feature and had a question. Since the next and previous page buttons have keybinds, which ones should we add for these buttons? We were thinking maybe ``ctrl+L`` for "last page" and either ``ctrl+K`` or ``ctrl+J`` for the first, since ``ctrl+F`` is already taken and those keys are closest to L? 
What do you think?

not-night-but:
@joaovicente3 I'm not totally sure. Are there comparable keybinds in other programs? Like excel/google sheets?

joaovicente3:
@not-night-but thanks for your reply! 
I'm not sure I follow though, don't excel and google sheets display everything in a single page, unlike Beekeeper Studio? Both of them use ```Ctrl+↓``` to jump to the last row of the table and so does BKS but within the current page.
We also looked into other database tools and generally they don't have a keybind for this so maybe it is not necessary after all?

While working on the feature, we came across another question:
Since the user can manually set the current page, what should happen if they enter a number larger than the total number of pages?
Should we automatically jump to the last page, or follow the same behavior as when a negative number is entered—i.e., do nothing?

not-night-but:
@joaovicente3 doh my bad lol. I'm just thinking of you can think of any applications that have similar behaviours that we can use those keybinds. If not, we can probably use something like ctrl+H and ctrl+L (from vim).

I'm not sure what to do in the case of a page being greater than the number of existing pages, as I'm not sure we can definitively know that. I believe we only really get an estimate of table size so we can't really determine if it'll be within the limit. What's the behaviour now?

joaovicente3:
Nothing comes to mind for us, those from vim seem like the best options in our opinion.

In our solution to create the button to jump to the last page we use the ```getTableLength``` method, that retrieves the number of rows, and then dividing by the page limit to calculate the number of pages. We could use the same approach here. 
Currently, if a user sets a page number that's higher than the total pages, it just navigates there and shows “No Data,” which is the same behaviour as navigating manually past the last page with data.

not-night-but:
[...]

PR #3216, opened 2025-05-26
Feat #2826: Improve pagination/navigation

joaovicente3:
Improves pagination by adding navigation buttons that allow users to jump directly to the first and last pages. This enhancement provides quicker access to the beginning and end of the content, improving overall user experience.

To support this functionality, a new function, jumpToLastPage(), was implemented. Alongside this, keyboard shortcuts were introduced to enable faster navigation through the table pages.

Finally, the commit includes conditional logic that hides the jump, next, and previous buttons whenever navigation is not possible. 

We tried to do everything as per discussed in the issue, but please feel free to request any changes. 

https://github.com/user-attachments/assets/d08645ef-fd25-490b-a81a-aceecbaaa8b8

### DIFF
--- a/apps/studio/default.config.ini
+++ b/apps/studio/default.config.ini
@@ -139,3 +139,5 @@ nextPage = ctrlOrCmd+right
 previousPage = ctrlOrCmd+left
 focusOnFilterInput = ctrlOrCmd+f
 openEditorModal = shift+enter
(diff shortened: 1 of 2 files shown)


not-night-but (approved):

not-night-but:
Looks good! Thanks so much for this!

PR #3219, opened 2025-05-27
New tab type for AI plugin

azmy60:
- A new tab type called `plugin-shell` which looks like Mongo shell or Query Editor but the top part is a plugin iframe.
- Add 2 more APIs:
  - `expandTableResult` - For expanding the result table in the `plugin-shell`
  - `updateTabTitle` - As the name suggests, it updates the title and save it
- Plugin system will notify plugins with `themeChanged` with a css string so now `getTheme` API is not needed.

cool-ai-2

### DIFF
--- a/apps/studio/src/common/transport/TransportOpenTab.ts
+++ b/apps/studio/src/common/transport/TransportOpenTab.ts
@@ -7,7 +7,7 @@ export type TabType = 'query' | 'table' | 'table-properties' | 'settings' | 'tab
 
 const pickable = ['title', 'tabType', 'unsavedChanges', 'unsavedQueryText', 'tableName', 'schemaName']
 
-interface BaseTransportOpenTab extends Transport {
+export interface TransportOpenTab<Context = {}> extends Transport {
   tabType: TabType,
   unsavedChanges: boolean,
   title: string,
@@ -24,27 +24,18 @@ interface BaseTransportOpenTab extends Transport {
   workspaceId?: number,
   filters?: string,
   lastActive?: Date|null,
-  deletedAt?: Date|null 
+  deletedAt?: Date|null
   isRunning: boolean, // not on the actual model, but used in frontend
+  context: Context
 }
 
-interface TransportDefaultOpenTab extends BaseTransportOpenTab {
-  tabType: Exclude<TabType, 'plugin-shell'>;
-  context: unknown;
-}
-
-export interface TransportPluginShellTab extends BaseTransportOpenTab {
-  tabType: "plugin-shell";
-  context: {
-    pluginId: string;
-    pluginTabTypeId: string;
-    /** A plugin can save the state of the tab here. For example, an AI plugin
+export type TransportPluginShellTab = TransportOpenTab<{
+  pluginId: string;
+  pluginTabTypeId: string;
+  /** A plugin can save the state of the tab here. For example, an AI plugin
      * can save the chat conversation here */
-    data: any;
-  }
-}
-
-export type TransportOpenTab = TransportDefaultOpenTab | TransportPluginShellTab;
+  data: any;
+}>
 
 export namespace TabTypeConfig {
   interface BaseTabTypeConfig {
@@ -132,7 +123,7 @@ export function matches(obj: TransportOpenTab, other: TransportOpenTab): boolean
   }
 
   switch (other.tabType) {
-    case 'table-properties': 
+    case 'table-properties':
       return obj.tableName === other.tableName &&
         (obj.schemaName || null) === (other.schemaName || null) &&
         (obj.entityType || null) === (other.entityType || null) &&


azmy60:
~I haven't cleaned this PR yet, but I'll open it anyway for review.~ (DONE)

apps/studio/src/components/TabPluginShell.vue
```diff
@@ -0,0 +1,220 @@
+<template>
+  <div
+    v-if="isCommunity && tab.context.pluginId === 'bks-ai-shell'"
```
rathboma, line 3: Do we mean to have a hard-coded check here for the plugin-id?

Can't the plugin config have a `communityEnabled: false` field? That way we could add this to any plugin.

apps/studio/src/common/transport/TransportOpenTab.ts
```diff
@@ -28,6 +28,55 @@ export interface TransportOpenTab extends Transport {
   isRunning: boolean, // not on the actual model, but used in frontend
 }
 
+interface TransportDefaultOpenTab extends BaseTransportOpenTab {
```
rathboma: This is ok, but would it be simpler to have a single interface for TransportOpenTab, but just add an optional field for the context?

rathboma: Happy as-is for now, but just something to think about

apps/studio/src/migration/20250527_add_context_to_tabs.js
```diff
@@ -0,0 +1,13 @@
+export default {
+  name: 'add-context-to-tabs',
+  async run(runner) {
+    const queries = [
+      `
+        ALTER TABLE tabs ADD COLUMN "context" text NULL DEFAULT '{}'
```
rathboma, line 6: see we're just making it available in the DB for all tabs, so might be good to just reflect that in the front end as a single model, otherwise the model and the DB are drifting...

apps/studio/src/migration/20250529_add_plugin_settings.js
```diff
@@ -0,0 +1,9 @@
+export default {
+  name: "20250529_add_plugin_settings",
+  async run(runner) {
+    await runner.query(`
+      INSERT INTO user_setting(key, defaultValue, valueType)
+        VALUES ('pluginSettings', '{}', 3);
```
rathboma, line 6: Not using config files for this?

apps/studio/src/services/plugin/PluginManager.ts
```diff
@@ -1,7 +1,12 @@
 import _ from "lodash";
 import PluginRegistry from "./PluginRegistry";
 import PluginFileManager from "./PluginFileManager";
-import { Manifest, PluginRegistryEntry, PluginRepository } from "./types";
+import {
```
rathboma, line 4: should this go in src/backend?

apps/studio/src/services/plugin/comm/notificationTypes.ts
```diff
@@ -1,13 +1,23 @@
+import { ThemeType, WindowEventInits, WindowEventClass } from "./commonTypes";
```
rathboma, line 1: Are these exclusively backend, or shared with the front end? If both, put in src/common, if backend, put in src/backend

apps/studio/src/services/plugin/web/WebPluginLoader.ts
```diff
@@ -114,19 +144,66 @@
...
+          KeyboardEvent,
+          Event,
+        };
+
+        if (!windowEventMap[notification.args.eventClass]) {
+          log.warn(`Unknown event class: ${notification.args.eventClass}`);
+        }
+
+        document.dispatchEvent(
+          new windowEventMap[notification.args.eventClass](
```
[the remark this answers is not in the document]
rathboma: looks like using eventClass directly might call it if it's a method?

apps/studio/src/common/transport/TransportOpenTab.ts
azmy60: Yes. That's better actually. I'll change it.

apps/studio/src/services/plugin/web/WebPluginLoader.ts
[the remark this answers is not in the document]
azmy60: Yes. The security check just wants to prevent calling undefined.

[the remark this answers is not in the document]
azmy60: FIxed.

apps/studio/src/services/plugin/comm/notificationTypes.ts
azmy60: I'm making an npm package for this.

azmy60: These are front end types

azmy60:
@rathboma I'll think about your comments in the next iteration. I think it makes sense to move some of the definitions somewhere to match our codebase pattern. I'll merge this now so everyone can test our bumblebee 🐝

Issue #3218, opened 2025-05-27
FEAT: Live Query Elapsed Time Indicator


JeanPaiva42:
*Is your feature request related to a problem? Please describe.*

Adapted from a customer message:
"When I run long queries in Beekeeper Studio and can't see how long they’ve been running until they finish. Most other database GUIs I’ve used display an active timer during query execution, which is really helpful for monitoring and decision-making."

*Describe the solution you'd like*
I’d like Beekeeper Studio to display a live "elapsed time" indicator while a query is still running — ideally near the results panel or in the status bar. It would update in real-time and reset with each new query execution.

*Describe alternatives you've considered*
Other tools like DBeaver or TablePlus show the ongoing query duration as it runs. I currently rely on those for long-running queries purely because of this feature, even though I prefer Beekeeper’s interface overall.

*Additional context*
"This would be a small but powerful quality-of-life improvement for users who often deal with long-running operations and want to know if something is stuck, slow, or just taking expected time."

Database type, if appropriate
All database types

PR #3225, opened 2025-05-29
Issue #3218 - added timer to display during execution

Landro3:
#3218 

First time contributing to the project. Added a timer that displays after one second during execution time, and hides when it's done. Seems like this may want to be tied to a setting but open to ideas/improvements.


apps/studio/src/components/editor/QueryEditorStatusBar.vue
```diff
@@ -259,6 +269,9 @@ export default {
       }
       return `Execution time: ${humanizeDuration(this.executeTime)}`
     },
+    elapsedTimeText() {
+      return formatSeconds(this.elapsedTime + 7200);
```
not-night-but: This looks like debug code to me? Otherwise everything reads as over 2 hours running.

not-night-but (changes requested):
Looks good! Looks like you left some debug code in there though

apps/studio/src/components/editor/QueryEditorStatusBar.vue
Landro3: removed test code

not-night-but (approved):
Thanks so much for your contribution! Feel free to submit more PRs if you feel like it

PR #3232, opened 2025-06-02
Feat #2963: Implemented Query Variables Support

### LINKED ISSUE #2963: FEAT: Query Variables Support

When developing and optimizing complex queries, especially in PostgreSQL, it's useful to define parameters or variables to easily tweak query conditions without manually editing the SQL each time.
Some tools, like DBeaver, allow for inline variable definition (e.g., %magic% = 'poof'), which makes iterative query development more efficient. Beekeeper Studio currently doesn’t support this, making query testing and optimization less flexible.

It would be great to have a way to define and use variables directly within the query editor. Ideally, users could set a variable once and reference it multiple times within a query without needing to rewrite values manually. This would make it easier to explore data, test different filtering conditions, and optimize queries without modifying the core SQL structure.

* Manually modifying queries each time a value needs to change, which is less efficient.

It seems like this was initially thought to PostgreSQL (based on the message our user said) but I am guessing that we could potentially expand that for other databases also (?).

TomasSilv:
Resolves #2963

This PR adds support for inline variables in SQL queries, allowing users to define 
and use variables directly within the query text using special comment syntax. Variables can be 
declared using single-line comments like -- %var% = value or block comments such as /* VARS: ... */.

During parsing, these variables are extracted and removed from the original query,
 then substituted with the appropriate value based on their type. The substitution logic handles 
 quoting for strings, preserves numeric and null values, supports JSON objects, and formats lists 
 in a SQL-compatible way — all without introducing syntax errors.

The implementation includes robust error handling for malformed declarations and a
 comprehensive test suite that covers a wide range of use cases: valid and invalid inputs,
  repeated variables, optional substitutions, strings with spaces, floats, nulls, JSON, and lists.
  (We also added examples and information in editor.md)

https://github.com/user-attachments/assets/c14ac31b-c84d-4786-88b7-7fcdf88ce61e

### DIFF
--- a/apps/studio/src/common/utils.ts
+++ b/apps/studio/src/common/utils.ts
@@ -322,3 +322,82 @@ export function toRegexSafe(input: string) {
     return null;
   }
 }
+
+export function extractVariablesAndCleanQuery(query: string) {
+  const variables: Record<string, string> = {};
+
+  // Expression for valid lines: -- %var% = valor
+  const linePattern = /^\s*--\s*%(.*?)%\s*=\s*(.+)$/gm;
+
+  // Expression for invalid lines: -- %var% = (no value)
+  const invalidLinePattern = /^\s*--\s*%(.*?)%\s*=\s*$/gm;
+  const invalidMatch = query.match(invalidLinePattern);
+  if (invalidMatch) {
+    throw new Error(`Malformed variable: "${invalidMatch[0]}"`);
+  }
+
+  let match;
+  while ((match = linePattern.exec(query)) !== null) {
+    const [_, name, value] = match;
+    if (!name || !value || value.trim() === '') {
+      throw new Error(`Malformed variable: "${match[0]}"`);
+    }
+    variables[name] = value.trim();
+  }
+
+  // Variables inside blocks: /* VARS: ... */
+  const blockPattern = /\/\*\s*VARS:(.*?)\*\//gs;
+  let blockMatch;
+  while ((blockMatch = blockPattern.exec(query)) !== null) {
+    const blockContent = blockMatch[1];
+
+    const varPattern = /%(.*?)%\s*=\s*(.*)/g;
+    let varMatch;
+    while ((varMatch = varPattern.exec(blockContent)) !== null) {
+      const [_, name, value] = varMatch;
+      if (!name || !value || value.trim() === '') {
+        throw new Error(`Malformed variable: "${varMatch[0]}"`);
+      }
+      variables[name] = value.trim();
+    }
+  }
+
+  // Remove variable definitions from the query
+  const cleanedQuery = query
+    .replace(linePattern, '')
+    .replace(blockPattern, '')
+    .trim();
+
+  return { variables, cleanedQuery };
+}
+
+export function substituteVariables(query: string, variables: Record<string, string>): string {
+  let substitutedQuery = query;
+
+  for (const [key, rawValue] of Object.entries(variables)) {
+    let value = rawValue.trim();
+
+    let isJSON = false;
+
+    try {
+      const parsed = JSON.parse(value);
+      isJSON = typeof parsed === 'object';
+    } catch {
+      // not JSON
+    }
+
+    const isList = value.startsWith('(') && value.endsWith(')');
+    const isQuoted = /^'.*'$/.test(value);
+    const isNull = value.toLowerCase() === 'null';
+    const isNumber = /^-?\d+(\.\d+)?$/.test(value); 
+
+    if (!isJSON && !isList && !isQuoted && !isNull && !isNumber) {
+      value = `'${value}'`;
+    }
+
+    const regex = new RegExp(`%${key}%`, 'g');
+    substitutedQuery = substitutedQuery.replace(regex, value);
+  }
+
+  return substitutedQuery;
+}
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -358,6 +358,8 @@
   import { registerQueryMagic } from "@/lib/editor/CodeMirrorPlugins";
   import { getVimKeymapsFromVimrc } from "@/lib/editor/vim";
   import { monokai } from '@uiw/codemirror-theme-monokai';
+  import { extractVariablesAndCleanQuery, substituteVariables } from '@/common/utils'
+
 
   const log = rawlog.scope('query-editor')
   const isEmpty = (s) => _.isEmpty(_.trim(s))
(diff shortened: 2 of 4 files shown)


not-night-but:
Hey @TomasSilv ! Thanks for the contribution! I haven't looked at the code just yet, but I wanted to ask: how does this handle not clashing with existing sql syntax? For example: if I define a variable called `%id%` and later have a like statement:

```sql
-- %id% = 7

SELECT * FROM table WHERE id = %id% AND name LIKE '%id%';
```
1. Does it replace both instances of `%id%`?
2. What if the user really wanted `%id%` to be the actual text in the like statement?

I understand this is a corner case, just want to know how we plan on handling situations like that.

TomasSilv:
Hi @not-night-but.Thanks for pointing this out!

We've now updated the implementation to avoid replacing variables that appear inside SQL string literals. This ensures that a placeholder like %id% inside a string (e.g., LIKE '%id%') will remain untouched, while still being correctly substituted when used outside of strings.

For example
```
-- %id% = 7

SELECT * FROM table WHERE id = %id% AND name LIKE '%id%';
```
The resulting query becomes:
```
SELECT * FROM table WHERE id = 7 AND name LIKE '%id%';
```
This prevents accidental substitutions in places where the user meant to keep the original %id% as a string. We also added a dedicated test case to cover this scenario and help ensure consistent behavior going forward.

apps/studio/src/common/utils.ts
```diff
@@ -322,3 +322,128 @@ export function toRegexSafe(input: string) {
     return null;
   }
 }
+export function extractVariablesAndCleanQuery(query: string) {
+  const variables: Record<string, string> = {};
+
+  // Expression for valid lines: -- %var% = valor
+  const linePattern = /^\s*--\s*%(.*?)%\s*=\s*(.+)$/gm;
```
not-night-but: This regex means that the following will be a match:
```
--


%var%


=


valor
```
Which I'm not sure we want? For the most part I think the `\s` should be replaced with just `[\ \t]`, right? You're welcome to argue with me here lol.

NOTE:
Technically this case would be caught by the malformed vars check below, but only because of the newline after the `=` (because in `m` mode, `$` matches a newline as the end of a line). Realistically I think we just need to modify both of these checks to ensure that the var declaration is all on one line

apps/studio/src/common/utils.ts
```diff
@@ -322,3 +322,128 @@ export function toRegexSafe(input: string) {
...
+    let varMatch;
+    while ((varMatch = varPattern.exec(blockContent)) !== null) {
+      const [_, name, value] = varMatch;
+      if (!name || !value || value.trim() === '') {
+        throw new Error(`Malformed variable: "${varMatch[0]}"`);
+      }
+      variables[name] = value.trim();
+    }
+  }
+
```
not-night-but: Same issue as above. Not sure how we want it to work but:

```
/*
	VARS:

	%name%



	=



	'Beekeeper''s Database'
*/
```

This matches fine in the system, I'm just not sure if we want to allow it? Do you have any thoughts on this?

apps/studio/src/common/utils.ts
```diff
@@ -322,3 +322,128 @@ export function toRegexSafe(input: string) {
...
+      i++;
+      continue;
+    }
+
+    if (!insideSingleQuote && char === '"') {
+      insideDoubleQuote = !insideDoubleQuote;
+      result += char;
+      i++;
+      continue;
+    }
```
not-night-but: Okay so a couple things:

Not all dialects necessarily use single/double quotes for strings, if you look at our dialect datas  in `apps/studio/src/shared/lib/dialects/` you can see that each dialect may have its own way of escaping strings/quoting identifiers/etc. So I'm not sure we should be hardcoding this.

But regardless of that, how do other applications handle this? I believe the original issue mentions dbeaver. Does it allow for replacing within a string? Does it allow for a variable to be an identifier? eg:
```sql
-- %var% = name

SELECT %var% FROM connection;
```
Does that work in dbeaver? Is it something we should think about?

I want us to avoid accidentally malforming the query obviously, but we should look at what is supported in those other systems so we can attempt feature parity. Interested to hear your thoughts on this.

apps/studio/src/common/utils.ts
```diff
@@ -322,3 +322,128 @@ export function toRegexSafe(input: string) {
...
+
+      try {
+        const parsed = JSON.parse(value);
+        isJSON = typeof parsed === 'object';
+      } catch {
+        // Not JSON
+      }
+
+      const isList = value.startsWith('(') && value.endsWith(')');
+      const isQuoted = /^'.*'$/.test(value);
```
not-night-but: Is any of this dialect specific? Can this be reliably used with postgres, mysql, bigquery, etc? If you need to make dialect specific changes for this, I would add the information you need to the dialect data for that dialect.

not-night-but (changes requested):
Okay this looks good ! Thanks so much for the contribution. I had a few questions, let me know your thoughts on them. 

Also, thank you so much for testing/documenting your feature! That helps out a bunch. Yall are legends!

TomasSilv:
Hi @not-night-but , thanks a lot for the detailed review and thoughtful suggestions!

Apologies for the delay — I’ve been a bit tied up recently, but I’m currently working on addressing your comments.

You're absolutely right about the regex pattern — that's something I’ll definitely need to improve to avoid multiline matches.
Regarding dialect-specific behavior, I’m not entirely sure yet how to approach that — right now it’s not fully adapted for all dialects, so that’s something I’ll need to explore further, especially considering how DBeaver handles similar cases.

I’ll push the updates as soon as I have everything tested and stable. Thanks again for your patience!

TomasSilv:
Hi @not-night-but,

Thanks so much for the thoughtful and detailed feedback — really appreciate how carefully you went through this!
I’ve now pushed the updates:

Refined the regex to ensure variable declarations stay on a single line, as you suggested — this avoids accidental multiline matches.

Verified and tested string literal handling for several dialects. The quote handling is done simply in the code and covers the main cases we need.

Confirmed that variables are not substituted inside string literals — this matches how DBeaver behaves too, and prevents unexpected replacements.

Unlike DBeaver, which does allow variable substitution for identifiers, this implementation deliberately skips that for now to keep the logic simple and safe.

Thanks again for all the great input — really appreciate it!

not-night-but:
@TomasSilv So sorry, this kinda fell under the radar, and I've been thinking hard about this feature. I want the functionality it provides, but I don't think the regex approach is the way to go. I think we probably need to add an integration to [sql-query-identifier](https://github.com/coresql/sql-query-identifier) to allow the parsing to be a lot more robust. I also think we could probably add some more constraints to this to make it a bit simpler: declarations should only happen at the top of the file for example. This would require a bit more work to get functional, so if you don't have the time/motivation to do that, I understand. But I think that's where we stand with this at the moment. Let me know if you're down to do the changes and then we can discuss what actually needs to be done in the identifier repo

rathboma:
@TomasSilv Awesome work implementing this.

Unfortunately I'm going to close it in favor of moving the work to https://github.com/coresql/sql-query-identifier (which we also control).

If you want to add the identification of variables to sql-query-identifier it would definitely be appreciated.

PR #3240, opened 2025-06-03
Plugin System Documentation

azmy60:


docs/user_guide/plugin_development/index.md
```diff
@@ -0,0 +1,65 @@
...
+Plugins are web applications (HTML, CSS, JavaScript) that run inside Beekeeper Studio to provide additional functionality. They can:
+
+- Add custom interfaces for database interaction
+- Create specialized tools for data analysis and manipulation
+- Integrate with external services and APIs
+- Provide domain-specific functionality for your workflow
+
+## Plugin Types
+
+### Tab Plugins (Shell Type)
```
rathboma: Can you add some other plugin types that you've planned, but are 'coming soon'?

docs/user_guide/plugin_development/index.md
```diff
@@ -0,0 +1,65 @@
...
+summary: "Learn how to develop, distribute, and manage plugins for Beekeeper Studio to extend its functionality."
+icon: material/puzzle
+---
+
+# Plugin Development
+
+!!! warning "Beta Feature"
+    The plugin system is currently in beta. APIs and functionality may change in future releases. Please provide feedback and report issues to help us improve the system.
+
+The Beekeeper Studio Plugin System allows developers to extend the application's functionality through web-based plugins. Plugins run in secure, sandboxed environments and communicate with the main application through a structured API.
```
rathboma: Add a screenshot

mkdocs.yml
```diff
@@ -109,6 +109,11 @@ nav:
     - user_guide/backup-restore.md
     - Cloud Sync: user_guide/cloud-storage-team-workspaces.md
     - user_guide/privacy-mode.md
+  - Plugin Development:
```
rathboma: You need a 'Plugins' page in the user guide for **users** of plugins, detailing what they are, who they're available to, and how to install them

docs/user_guide/plugin_development/index.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Tab Plugins (Shell Type)
+The main plugin type that creates new tab interfaces with:
+- Custom HTML/JavaScript interface at the top
+- Built-in result table at the bottom for displaying query results
+- Full access to the Plugin API for database operations
+
+## Getting Started
+
+### For Users
```
rathboma: Make this a separate page in the user guide

docs/user_guide/plugin_development/index.md
rathboma: Who are plugins available to? Make this clear both here and in the user docs

docs/user_guide/plugin_development/index.md
```diff
@@ -0,0 +1,65 @@
...
+npm install github:beekeeperstudio/plugin
+```
+
+Then import API functions in your plugin code:
+
+```javascript
+import { getTables, runQuery, setTabTitle } from '@beekeeperstudio/plugin';
+```
+
+### Key Components
```
rathboma: Are these key components for the developer? Not really.

I'd have this be `plugin lifecycle` and say stuff like:

1. Install the beekeeper plugin API
2. Create your plugin
3. Push to your own repository
4. Publish a build on GitHub
5. Submit a PR to the Beekeeper Studio community plugins repository

docs/user_guide/plugin_development/index.md
rathboma: A basic 'how plugins work' here is helpful, say stuff like:

Plugin installation is handled natively by the Beekeeper Studio application, community plugins are published on the Beekeeper Studio community plugin repository (link)

rathboma: You'll need a version of this for the user docs too, with screenshots

docs/user_guide/plugin_development/api-reference.md
```diff
@@ -0,0 +1,286 @@
...
+    The plugin system is currently in beta. APIs and functionality may change in future releases. Please provide feedback and report issues to help us improve the system.
+
+The Beekeeper Studio Plugin API is accessible through the `@beekeeperstudio/plugin` package. The API provides three main functions for communication between your plugin and the main application.
+
+## Installation
+
+Install the plugin package from GitHub:
+
+```bash
+npm install github:beekeeper-studio/plugin
```
rathboma: We don't have an NPM package for this?

docs/user_guide/plugin_development/api-reference.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+## Import API Functions
+
+```javascript
+import { request, notify, addNotificationListener } from '@beekeeperstudio/plugin';
+```
+
+## How It Works
```
rathboma: Can we add the example plugin (the one you have screenshots of) link here, so folks can look at the real code?

docs/user_guide/plugin_development/api-reference.md
```diff
@@ -0,0 +1,286 @@
...
+
+The `@beekeeperstudio/plugin` package is a thin wrapper around `window.postMessage` for communication between your plugin (iframe) and Beekeeper Studio. It provides type safety, promises, and automatic request/response matching.
+
+## Core API Functions
+
+The plugin API consists of three main functions:
+
+### `request(name, args?)`
```
rathboma: Why implement these like `bks.request('getTables')` rather than `bks.getTables()`? The second way means that for each method we can add details JSDocs and then publish those as the reference using TypeDoc or JSDoc binaries.

No? Is there some advantage to this system that I'm missing?

To be clear: I'm fine with this for now, but seems like a lot more work to maintain

docs/user_guide/plugin_development/api-reference.md
```diff
@@ -0,0 +1,286 @@
...
+Send requests to Beekeeper Studio to retrieve data or execute actions. Internally uses `window.parent.postMessage()` with a unique ID and waits for a response.
```
rathboma: How are errors handled with request functions? What if malformed request is sent? How should the user debug?

Are there error classes or codes they should pay attention to?

docs/user_guide/plugin_development/api-reference.md
```diff
@@ -0,0 +1,286 @@
...
+
+### `notify(name, args)`
```
rathboma: What does this return when the notification is successful / unsuccessful? What errors can be thrown here?

rathboma (changes requested):
I think the docs need a bit of a reorganize, and also need expanding on a little.

Long term I want you to think about using actual methods, vs a generic `request` method.

docs/user_guide/plugin_development/api-reference.md
azmy60: I'll add those to the docs.

If you call `request` and there is an error, it will just throw an error. 

When error happens, it usually happens in the plugin system, so the plugin system will respond with an object that goes `{ error: "Error message here" }`. 

We don't have custom classes for errors yet. I just used simple `Error` class with a message text.

Also, you can debug the messages by calling `setDebugComms(true)` which will log the three main functions.

docs/user_guide/plugin_development/api-reference.md
azmy60: So the communication module was inspired by LSP which basically has two ways of communications:
- Request/Response: If a plugin requests something, it expects something in return
- Notification: If the plugin system notifies the plugin, it does not expect a return

An example to notification is `switchTheme`: when user chooses a theme, the plugin system notifies all plugins (without expecting an answer from all plugin) with `switchTheme`.

But good questions though, will make the docs clearer.

docs/user_guide/plugin_development/api-reference.md
azmy60: `request` is a very thin wrapper of `window.parent.postMessage()`. Basically it is this:

```js
export async function request(name, args) {
  return new Promise((resolve, reject) => {
      const id = generateUUID();
      const data = { id, name, args };
      pendingRequests.set(id, { name, resolve, reject });
      window.parent.postMessage(data, "*");
  });
}
```

It's the very basic function for communicating that can't be replaced. In Beekeeper Studio, it is like `this.$util.send`.

I think we could add JSDocs to request function too right? the request function is [typed](https://github.com/beekeeper-studio/plugin/blob/203cb3ec170126dca20f78bf5e5e586299d6b1ee/src/comms.ts#L84). We could then add JSDocs this way:

```ts
/** jsdocs here */
export async function request(name: "getTables", args?: GetTablesRequest["args"]): Promise<GetTablesResponse>;

/** jsdocs here */
export async function request(name: "getColumns", args: GetColumnsRequest["args"]): Promise<GetColumnsResponse>;

/** jsdocs here */
export async function request(name: "getConnectionInfo"): Promise<GetConnectionInfoResponse>;
```

I'd be down to do `bks.getTables` too if it's better. It's gonna be a wrapper of the `request` function.

docs/user_guide/plugin_development/api-reference.md
azmy60: I'll add them today or next week.

rathboma:
@azmy60 some changes requested here before we merge

docs/user_guide/plugin_development/index.md
azmy60: > Can you add some other plugin types that you've planned, but are 'coming soon'?

Only sidebar plugins currently (beside json sidebar). But before that, we probably want to have sidebar tabs that can be dragged around. For example, JSON Viewer can be moved to the global sidebar and vice versa with mouse drag. Not sure, would love to discuss about that.

azmy60:
@rathboma changes were made. Anything else?

rathboma (approved):
I think this is great! Let's merge this, but don't add it to the mkdocs navbar yet until we're ready to announce it.

azmy60:
Awesome! I also just published the vite plugin for the hot reload stuff. I'll add it here, and then merge.

azmy60:
https://www.npmjs.com/package/@beekeeperstudio/vite-plugin

PR #3242, opened 2025-06-03
Pin security

### LINKED ISSUE #3243: FEAT: React to power state changes (idle, locked, suspend)
On suspend/idle/lock users might need to do the following things:

- disconnect from the db
- prompt for a pin/password

Implementation notes:

- Power Monitor in Electron has the hooks we need: https://www.electronjs.org/docs/latest/api/power-monitor

rathboma:
Fix #3243


azmy60:
### Update PIN (Tools > Update pin)
Screenshot 2025-06-04 232619

### After pressing connect/while connecting
Screenshot 2025-06-04 232638

### Creating a new PIN first time installing the app
Screenshot 2025-06-04 232650

azmy60:
The last thing from me is to rebase the PR to rc-52 😢

azmy60:
@rathboma I think the `disconnectOnIdle` and the other flags should be `false` by default, right? it can be enabled without having to set the `lockMode`.

rathboma:
@azmy60 When users are adding a new pin - have them enter it twice to confirm it

apps/studio/src/backend/lib/security.ts
```diff
@@ -0,0 +1,51 @@
...
+import rawLog from "@bksLogger";
+import { powerMonitor } from "electron";
+
+const log = rawLog.scope("security");
+
+let idleCheckInterval: NodeJS.Timer;
+
+let initialized = false;
+
+export function initializeSecurity() {
```
rathboma, line 13: Future: This smells like a class, maybe it could be refactored into a class, then the initialize method can just be the class initializer (not a blocker for release)

azmy60:
Ready to merge this. Please also look at the docs PR for the `forgot-pin` link. I made it against `master`. https://github.com/beekeeper-studio/beekeeper-studio/pull/3245/files

Issue #2900, opened 2025-03-18
FEAT: MySQL User & Role Management


JeanPaiva42:
**Feature Request from a user:**

**Describe the solution you'd like**

A feature to manage database users and roles, similar to MySQL Workbench’s "Users & Privileges". It should allow defining what each connection can do, with an option to open multiple windows for easier data comparison.

**Describe alternatives you've considered**
- User mentioned using MySQL Workbench to achieve such goals;

**Additional context**

- This feature is essential for security-focused companies and non-monolithic architectures. A screenshot from MySQL’s "Users & Privileges" was provided for reference.

![Image](https://github.com/user-attachments/assets/1f7d8fc0-71ab-4fd7-98a4-8d249c165a0d)

**Database type, if appropriate**
MySQL

rathboma:
This would be a great feature, but I think it should be implemented as a plugin. Whether we bundle it or not is a later decision, but seems like a great self-contained plugin

PR #3253, opened 2025-06-06
Feat #2900: Implement MySQL User & Role Management

joao-sntos:
A feature to manage database users and roles, similar to MySQL
Workbench’s Users & Privileges:
 - User listing: View all users and their respective hosts.
 - Creation and editing: Create new users, modify name, host,
   authentication type, password, and resource limits.
 - Schema-level privilege management: View and edit privileges
   per schema, with bulk select/deselect.
 - Global privilege management: Assign and revoke privileges,
   including “Revoke All Privileges.”
 - Password management: Set, change, and expire passwords.
 - User removal: Delete users with confirmation.
 - UI/UX: Tab separation (login, limits, privileges), contextual
   actions, and confirmation modals.
 - Backend Tests: Integration tests were implemented for all
   backend queries related to User Management.

 Co-authored-by: Rodrigo Perestrelo <rodrigo.perestrelo@tecnico.ulisboa.pt>
       
Preview of the feature: https://www.youtube.com/watch?v=OMdnmc-MuA0


apps/studio/src-commercial/backend/handlers/connHandlers.ts
```diff
@@ -521,4 +536,53 @@ export const ConnHandlers: IConnectionHandlers = {
     checkConnection(sId);
     return await state(sId).connection.getQueryForFilter(filter);
   },
+
+  'conn/hasAdminPermission': async function({ sId }: { sId: string}) {
+    return await state(sId).connection.hasAdminPermission();
+  },
```
not-night-but: With all of these handlers that you've added here, we should make sure we check that the connection exists: `checkConnection(sId);`

apps/studio/src/components/CoreTabs.vue
```diff
@@ -434,6 +443,7 @@ import { TransportOpenTab, TransportPluginShellTab, setFilters, matches, duplica
         { event: AppEvent.beginImport, handler: this.beginImport },
         { event: AppEvent.restoreDatabase, handler: this.restoreDatabase },
         { event: AppEvent.switchUserKeymap, handler: this.switchUserKeymap },
+        { event: 'userManagementOpen', handler: this.userManagementOpen },
```
not-night-but: This should be added to the AppEvent enum in `apps/studio/src/common/AppEvent.ts`, rather than just using a string for the event name.

apps/studio/src/components/TabUserManagement.vue
```diff
@@ -0,0 +1,152 @@
...
+    selectUser(user) {
+      this.updateSelectedUser(user);
+    },
+    async addUser(newUser) {
+      try {
+        newUser.isNew = true;
+        this.updateSelectedUser(newUser);
+        await this.$store.dispatch('updateUsersList');
+      } catch (error) {
+        console.error('Failed to add user:', error);
```
not-night-but: Any logging that is done should be done through electron log.

```ts
import rawLog from "@bksLogger"

const log = rawLog.scope("TabUserManagement")

log.error('Failed to add user:', error);
```

This has all the same methods as console, just is better for electron, and it saves those logs to files as well. I know there are a bunch of them, but I'm not going to tag them all.

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,1107 @@
+<template>
```
not-night-but, line 1: I think the tabs here should be split into their own components, as this one is already pretty big. I'd take a look at how we do things with the tabs in `apps/studio/src/components/TabTableProperties.vue`

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,1107 @@
...
+      backupDetails: {
+      authType: '',
+      maxQueries: 0,
+      maxUpdates: 0,
+      maxConnections: 0,
+      maxUserConnections: 0,
+    },
+      objectPrivileges: ['SELECT', 'INSERT', 'UPDATE', 'DELETE', 'EXECUTE', 'SHOW VIEW'],
+      ddlPrivileges: ['CREATE', 'ALTER', 'REFERENCES', 'INDEX', 'CREATE VIEW', 'CREATE ROUTINE', 'ALTER ROUTINE', 'EVENT', 'DROP', 'TRIGGER'],
+      otherPrivileges: ['GRANT OPTION', 'CREATE TEMPORARY TABLES', 'LOCK TABLES'],
```
not-night-but: These look like they could possibly change if we were to implement this feature in a different dialect, no? May be worth moving these to the dialectData for mysql

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,1107 @@
...
+          }
+        }
+
+        return Object.entries(schemaGrantsMap).map(([name, privileges]) =>
+              this.normalizeSchemaPrivileges({ name, host: this.localUser.host || '%', privileges }));
+
+      } catch (error) {
+        console.error('Error obtaining User Grants:', error);
+      }
+    },
```
not-night-but: It seems to me like a lot of this could be done in the backend, no? I saw the other place we call this handler does very similar processing after we get the request. We should probably just come up with an object and format the data in the backend (like we do with tables, columns, views, etc). That way when we try to do this for other dialects, we don't have to litter this with ifs and switches.

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,1107 @@
...
+          for (const row of grants) {
+            const grantStr = Object.values(row)[0];
+            if (typeof grantStr !== 'string') continue;
+            const match = grantStr.match(/GRANT (.+) ON [`]?(.+?)[`]?.\* TO/i);
+            if (!match) continue;
+            const privsStr = match[1];
+            const schema = match[2];
+            const privileges = privsStr.split(',').map(p => p.trim().toUpperCase());
+            if (!schemaGrantsMap[schema]) schemaGrantsMap[schema] = {};
+            for (const priv of privileges) schemaGrantsMap[schema][priv] = true;
```
not-night-but: Again, like above, I think this processing should be done in the backend when we call `showGrantsForUser`

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,1107 @@
...
+    },
+    async applyChanges() {
+      if (this.localUser.password && !this.passwordsMatch) {
+        this.$noty.error('Password and confirmation do not match.');
+        return;
+      }
+      const changes = [];
+
+      if (this.localUser.isNew) {
+        // New User Creation - [0, user, host, password, authType]
```
not-night-but, line 359: Make an object for this! Storing this as an array of arrays can make this super unclear when another dev comes along. Take a look at how we build changes in `TableTable.vue`. Since these all seem to have the same members it could just be 

```ts
export interface UserChange {
	type: UserChangeType,
	user: string,
	host: string,
	password: string,
	authType: AuthType
}
```
Also, the first element should not just be a magic number. We should have an enum (I called it UserChangeType, but feel free to make your own decision on naming). That makes it easy to see what's going on at a glance.

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,1107 @@
...
+          insert: false,
+          update: false,
+          delete: false,
+          create: false,
+          alter: false,
+          drop: false,
+          index: false,
+          references: false
+        }
+      })
```
not-night-but: I think we should probably type this object somehow. Make it an interface/class. That way we at least have some way of finding out what structure is to be expected.

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,1107 @@
...
+      if (schemaIndex !== -1) {
+        this.$set(this.localUser.schemas[schemaIndex].privileges, privilege, value);
+      }
+    },
+    normalizeSchemaPrivileges(schema) {
+      const allPrivileges = [
+        ...this.objectPrivileges,
+        ...this.ddlPrivileges,
+        ...this.otherPrivileges
+      ];
```
not-night-but: I've seen this done a couple times, maybe just make it a computed property?

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,1107 @@
...
+    if (!this.localUser.isNew) { 
+      this.getAccountDetails();
+      this.getUserResourceLimits();
+    }
+  },
+}
+</script>
+
+<style scoped>
+.user-detail-view {
```
not-night-but: So the styling here, while looking nice, is not necessarily in line with the rest of the app. It isn't reactive to the themes, and diverges from the standard form styles that we have for the app. I would recommend looking into the `TabTableProperties.vue` component, as well as the child components for which classes you should use. I also mentioned above that we should split this out into different components for each tab, so also reference this component to see how to do that. Again, it's not that yall did a bad job or anything, just that it looks foreign in the app, and the extra styling is unnecessary when we have classes set up for this. If you do need to add custom styles for certain things, I'd recommend adding a specific stylesheet for it, rather than putting it in the style tag here.

apps/studio/src/components/TabUserManagement_UserList.vue
```diff
@@ -0,0 +1,179 @@
...
+  methods: {
+    addAccount() {
+      const newUser = {
+        user: 'newuser',
+        host: '%',
+        authType: 'caching_sha2_password',
+        password: '',
+        confirmPassword: '',
+        isNew: true
+      }
```
not-night-but: Again, I think we need a model for this. I think we probably need to move all the tabuser components into their own directory within the components directory, and have a models file there for all this stuff. (Like the stepper or entity list).

apps/studio/src/components/TabUserManagement_UserList.vue
```diff
@@ -0,0 +1,179 @@
...
+      this.$emit('user-added', newUser)
+    }
+  }
+}
+</script>
+
+<style scoped>
+.user-list-view {
```
not-night-but: I'd review the styles for this to see if there's something similar in the app.

apps/studio/src/components/TabUserManagement_UserList.vue
```diff
@@ -0,0 +1,179 @@
...
+              <span class="user-host">{{ user.host || 'unknown' }}</span>
+            </td>
+            <td>
+              <button 
+                class="btn-icon primary" 
+                @click="$emit('user-selected', user)" 
+                title="View Details"
+              >
+                <i class="material-icons">visibility</i>
+              </button>
```
not-night-but: We shouldn't have this button, you should just be able to double click the row and have it open the user.

apps/studio/src/components/TabUserManagement_UserList.vue
```diff
@@ -0,0 +1,179 @@
...
+      <table class="user-table">
+        <thead>
+          <tr>
+            <th>User</th>
+            <th>From Host</th>
+            <th>Actions</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr v-for="user in users" :key="`${user.user}@${user.host}`">
```
not-night-but: I think we should probably use tabulator here just to help gel with the rest of the app. Check out the TableSchema component for an example of what I mean.

apps/studio/src/components/sidebar/core/UserManagement.vue
```diff
@@ -0,0 +1,285 @@
...
+    },
+    openAllSettingsTab() {
+      this.updateSelectedUser(null);
+      this.$root.$emit('userManagementOpen');    }
+  }
+};
+</script>
+
+<style scoped>
+.dialog-buttons {
```
not-night-but, line 238: Review these styles to see if there's already available classes in the app, check out some of the other sidebar components and modals just to make sure. (Stuff like `.btn` probably doesn't need to be modified)

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,1107 @@
...
+        <i class="material-icons">block</i> Revoke All Privileges
+      </button>
+      <button class="btn" @click="revertChanges">
+        <i class="material-icons">undo</i> Revert Changes
+      </button>
+      <button class="btn success" @click="showApplyModal()">
+        <i class="material-icons">check</i> Apply Changes
+      </button>
+    </div>
+  </div>
```
not-night-but: I think we can move these to be in the statusbar, take a look at `TableTable.vue` or any of our other tabs to see how we do that.

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+  async hasAdminPermission(): Promise<boolean> {
+    const sql = "SELECT 1 FROM mysql.user LIMIT 1";
+    const result = await this.driverExecuteSingle(sql);
+    if (result.rows.length === 0 || result.rows[0] === undefined) {
+      return false;
+    }
+    return true;
+  }
+
+  async getListOfUsers(): Promise<any[]> {
```
not-night-but: We need to create a model for this, not just return an `any[]`. That way it can be standardized across other dialects.

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+    const sql = "SELECT User, Host FROM mysql.user";
+    const { rows } = await this.driverExecuteSingle(sql);
+    return rows;
+  }
+
+  async getUserAuthenticationDetails(
+    user: string,
+    host: string,
+    connection?: Connection 
+  ): Promise<any[]> {
```
not-night-but: Same thing here, create a model. Just needs to be an interface (Like we do with table columns or tables or any of our other stuff. We don't want to just return the raw rows from the client.)

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+    });
+  
+    return rows;
+  }
+  
+  async getUserPrivileges(
+    user: string,
+    host: string,
+    connection?: Connection
+  ): Promise<any[]> {
```
not-night-but: Model

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+    });
+
+    return rows;
+  }
+
+  async getUserResourceLimits(
+    user: string,
+    host: string,
+    connection?: Connection
+  ): Promise<any[]> {
```
not-night-but: Model

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+    });
+
+    return rows;
+  }
+
+  async showGrantsForUser(
+    user: string,
+    host: string,
+    connection?: Connection
+  ): Promise<any[]> {
```
not-night-but: Model

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+
+    const { rows } = await this.driverExecuteSingle(sql, {
+      params,
+      connection,
+    });
+
+    return rows;
+  }
+
+  async applyUserChanges(changes: any[][]): Promise<{ success: boolean; error?: string }> {
```
not-night-but: We should be accepting some sort of structured data here, not an array of arrays of any. I think I made a note in the frontend, but this should be taking something like `UserChange[]`.

Also the return value shouldn't need to be this, you can probably just return void, if it fails throw an error and that can be caught in the frontend.

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+      const [type, ...args] = change;
+  
+      switch (type) {
+
+        case 0: {
+          const [user, host, password, authType] = args;
+          const escapedUser = escapeString(user);
+          const escapedHost = escapeString(host);
+          const escapedPassword = escapeString(password);
+          const escapedAuth = escapeString(authType);
```
not-night-but: You do this in every case, you should just do it before the switch

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+  async applyUserChanges(changes: any[][]): Promise<{ success: boolean; error?: string }> {
+    const statements: string[] = [];
+
+    for (const change of changes) {
+      const [type, ...args] = change;
+  
+      switch (type) {
+
+        case 0: {
+          const [user, host, password, authType] = args;
```
not-night-but: This will change when this is an object, but also we don't want to use magic numbers. What is case 0? Type should be an enum

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+      FROM information_schema.schemata
+      WHERE schema_name NOT IN (
+          'information_schema',
+          'performance_schema',
+          'sys')`;
+
+    const { rows } = await this.driverExecuteSingle(sql);
+    return rows;
+  }
+
```
not-night-but: We already have `listSchemas` for this, no?

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+    const { rows } = await this.driverExecuteSingle(sql);
+    return rows;
+  }
+  
+  async deleteUser(user: string, host: string): Promise<NgQueryResult> {
+    const sql = `DROP USER '${escapeString(user)}'@'${escapeString(host)}'`;
+
+    const result = await this.driverExecuteSingle(sql); 
+    return result;
+  }
```
not-night-but: We don't do anything with this return value, might as well just not return anything.

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+
+  async renameUser(user: string, host: string, newName: string): Promise<NgQueryResult> {
+    const sql = `RENAME USER '${escapeString(user)}'@'${escapeString(host)}' TO '${escapeString(newName)}'@'${escapeString(host)}'`;
+
+    const result = await this.driverExecuteSingle(sql);
+    return result;
+  }
```
not-night-but: Same as above

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+
+  async expireUserPassword(user: string, host: string): Promise<NgQueryResult> {
+    const sql = `ALTER USER '${escapeString(user)}'@'${escapeString(host)}' PASSWORD EXPIRE`;
+
+    const result = await this.driverExecuteSingle(sql);
+    return result;
+  }
```
not-night-but: Same as above

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1445,239 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+
+  async revokeAllPrivileges(user: string, host: string): Promise<NgQueryResult> {
+    const sql = `REVOKE ALL PRIVILEGES, GRANT OPTION FROM '${escapeString(user)}'@'${escapeString(host)}'`;
+    return await this.driverExecuteSingle(sql);
+  }
```
not-night-but: Same as above

apps/studio/src/lib/utility/ElectronUtilityConnectionClient.ts
```diff
@@ -110,6 +110,60 @@ export class ElectronUtilityConnectionClient implements IBasicDatabaseClient {
...
+    const privileges = await Vue.prototype.$util.send('conn/getUserPrivileges', {user, host});
+    return privileges;
+  }
+
+  async getUserResourceLimits(user: string, host: string): Promise<any[]> {
+    const resourceLimits = await Vue.prototype.$util.send('conn/getUserResourceLimits', {user, host});
+    return resourceLimits;
+  }
+
+  // Each row in the result represents a Schema and its corresponding privileges for the user
```
not-night-but, line 138: Make a model for this!

apps/studio/src/lib/utility/ElectronUtilityConnectionClient.ts
```diff
@@ -110,6 +110,60 @@ export class ElectronUtilityConnectionClient implements IBasicDatabaseClient {
...
+  async showGrantsForUser(user: string, host: string): Promise<any[]> {
+    const grants = await Vue.prototype.$util.send('conn/showGrants', {user, host});
+    return grants;
+  }
+
+  async applyUserChanges(changes: any[][]): Promise<{ success: boolean; error?: string }> {
```
not-night-but: Model!

apps/studio/src/store/index.ts
```diff
@@ -72,6 +72,9 @@ export interface State {
   versionString: string,
   connError: string
   expandFKDetailsByDefault: boolean
+  hasAdminPrivileges: boolean,
```
not-night-but: We should probably just have a separate module for UserManagement, rather than having it at the root and in the sidebar store.

apps/studio/src/store/modules/SidebarModule.ts
```diff
@@ -2,6 +2,7 @@ import _ from "lodash";
 import { Module } from "vuex";
 import { State as RootState } from "../index";
 import { SmartLocalStorage } from "@/common/LocalStorage";
+import Vue from "vue";
```
not-night-but: Yeah we shouldn't have this stuff here, we should just have it in a separate store module.

not-night-but (changes requested):
Okay so the functionality is all there! My main issue with this is some of the code style, as well as the styling of the feature itself. We want to make sure this looks like it's part of the app, and while it looks cool right now, it definitely looks foreign when compared to other components we have. I made some recommendations on what this should look like, as well as some code organization tips. 

Thanks so much for taking this on! Let me know if you have any questions/concerns regarding any of my comments here.

RodrigoPerestrelo:
Thank you very much for your feedback. We appreciate your insights and will work on implementing the suggested changes.

apps/studio/src/components/CoreTabs.vue
```diff
@@ -961,6 +971,23 @@ export default Vue.extend({
...
+        tabType: 'user-management',
+        title: 'User Management',
+        userId: 'user-management',
+        unsavedChanges: false,
+      };
+
+      const existing = this.tabItems.find(
+        (t) => t.tabType === 'user-management'
+      );
+      if (existing) {
```
not-night-but: You should switch this to use the `matches` function, just so we are using the same patterns as other tab types. Look below at the `openTableProperties` function for an example (you'll have to add an entry to the switch in the `matches` function.

apps/studio/src/components/TabUserManagement.vue
```diff
@@ -0,0 +1,155 @@
...
+.header h1 {
+  font-size: 1.5em;
+  margin-bottom: 5px;
+}
+
+.header h2 {
+  font-size: 1.2em;
+  margin-top: 0;
+  font-weight: 400;
+}
```
not-night-but: These are all still custom styles. Look at our `TabTableProperties.vue` component for how I think we should style this. It looks nice, it just doesn't fit with the rest of our styles (and it doesn't change when the theme does)

apps/studio/src/components/TabUserManagementUser.vue
not-night-but: @joao-sntos @RodrigoPerestrelo I still think we need to do this. These custom styles don't fit with the rest of the app, and they aren't reactive to the theme state

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,741 @@
...
+          <i class="material-icons">block</i> Revoke All Privileges
+        </button>
+        <button class="btn" @click="revertChanges">
+          <i class="material-icons">undo</i> Revert Changes
+        </button>
+        <button class="btn success" @click="showApplyModal()">
+          <i class="material-icons">check</i> Apply Changes
+        </button>
+      </div>
+    </StatusBar>
```
not-night-but: The style seems to be bugged here, it should be the same height as the connection button on the left.
image

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,741 @@
...
+
+export interface UserSchemaPrivileges {
+  [privilege: string]: boolean;
+}
+
+export interface UserSchema {
+  name: string;
+  host?: string;
+  privileges: UserSchemaPrivileges;
+}
```
not-night-but: I'd put these in their own `models.ts` file so they can be used elsewhere (for example, in the `applyUserChanges` function, you're still accepting an `any[]`. Would be more useful if that was typed.)

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,741 @@
...
+            };
+          }
+          
+          Object.keys(grant).forEach(key => {
+            if (!key.match(/schema|host|grantOptionPrivilege/i)) {
+              const privName = key.replace(/Privilege$/i, '')
+                                .replace(/([A-Z])/g, ' $1')
+                                .replace(/_/g, ' ')
+                                .trim()
+                                .toUpperCase();
```
not-night-but: You do this same thing up above, make it a function instead and just reuse it.

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,741 @@
...
+      }
+
+      if (changes.length === 0) {
+        this.$noty.info('No changes to apply.');
+        return;
+      }
+
+      const result = await this.connection.applyUserChanges(changes);
+
+      if (result.success) {
```
not-night-but: I still don't think this needs to return an object. Just return void, and throw an error if something bad happens and then catch it here.

apps/studio/src/components/TabUserManagementUser.vue
```diff
@@ -0,0 +1,741 @@
...
+    TabUserManagementUserLogin,
+    TabUserManagementUserLimits,
+    TabUserManagementUserPrivileges,
+    StatusBar,
+  },
+}
+</script>
+
+<style>
+
```
not-night-but, line 595: A lot of these styles are unnecessary (statusbar-actions is already styled, you shouldn't have to modify it, right?)

apps/studio/src/components/sidebar/core/UserManagement.vue
```diff
@@ -0,0 +1,266 @@
...
+              @delete="showDeleteModal"
+              @contextmenu.prevent="openContextMenu(user, $event)"
+            />
+          </template>
+          <template v-else>
+            <div style="padding: 24px; color: #888; text-align: center; font-size: 1em;">
+              <i class="material-icons" style="font-size: 2em;">info</i>
+              <div>No users found</div>
+            </div>
+          </template>
```
not-night-but: I feel like we could reues the styles from the history list for this. Or at least emulate the styles for it

apps/studio/src/components/sidebar/core/UserManagement.vue
```diff
@@ -0,0 +1,266 @@
...
+              @select="selectUser"
+              @open="openUserTab"
+              @all_settings="openAllSettingsTab"
+              @rename="showRenameModal"
+              @delete="showDeleteModal"
+              @contextmenu.prevent="openContextMenu(user, $event)"
+            />
+          </template>
+          <template v-else>
+            <div style="padding: 24px; color: #888; text-align: center; font-size: 1em;">
```
not-night-but: Don't hardcode colours here. Switching themes can mean these will be hard to read. This is an issue with a lot of the styles for this feature. eg:
image

apps/studio/src/components/sidebar/core/UserManagement.vue
```diff
@@ -0,0 +1,266 @@
...
+  margin-bottom: 16px;
+}
+
+.form-control {
+  width: 100%;
+  padding: 8px;
+  border: 1px solid #ddd;
+  border-radius: 4px;
+}
+
```
not-night-but: I'm still not sure these styles are necessary? We have form-control and form-group styles already set for the whole app. Like I said, I'd just emulate what the history list does.

apps/studio/src/components/sidebar/core/user_management_list/UserManagementListItem.vue
```diff
@@ -0,0 +1,96 @@
...
+      style="height: 100%; width: 100%; display: flex; align-items: center; gap: 8px;"
+    >
+      <i class="item-icon material-icons">group</i>
+      <div style="display: flex; align-items: center; gap: 12px; flex: 1; min-width: 0;">
+        <span style="flex: 1; white-space: nowrap; overflow: hidden; text-overflow: ellipsis;">
+          {{ item.user }}
+        </span>
+        <span style="font-size: 0.85em; display: flex; gap: 4px; color: #666; padding-right: 12px;">
+          <span>{{ item.host }}</span>
+        </span>
```
not-night-but: Again, I don't like having the styles in style tags here. If it's 100% necessary, make a class, but I think you should be able to reuse some classes that we already have for `HistoryListItem`

apps/studio/src/components/sidebar/core/user_management_list/UserManagementListItem.vue
```diff
@@ -0,0 +1,96 @@
...
+    active: {
+      type: Boolean,
+      default: false
+    }
+  },
+  data: () => ({
+    timeAgo: new TimeAgo('en-US')
+  }),
+  computed: {
+    ...mapState('data/queryFolders', {'folders': 'items'}),
```
not-night-but: This doesn't seem to need to be here? Possibly copied from another component?

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1445,6 +1451,321 @@ export class MysqlClient extends BasicDatabaseClient<ResultType> {
...
+      trigger: !!privileges["TRIGGER"],
+      grant_Option: !!privileges["GRANT OPTION"],
+      lock_Tables: !!privileges["LOCK TABLES"],
+      create_Temporary_Tables: !!privileges["CREATE TEMPORARY TABLES"],
+    }));
+  
+    return userPrivileges;
+  }
+  
+  async applyUserChanges(changes: any[]): Promise<{ success: boolean; error?: string }> {
```
not-night-but: The changes should be typed here with the model that you made earlier. I also still don't think that this needs to return anything, you can just throw any errors that come up and catch them in the frontend.

apps/studio/src/store/index.ts
not-night-but: @joao-sntos @RodrigoPerestrelo  I still think we should create a separate vuex module for UserManagement.

apps/studio/src/store/modules/SidebarModule.ts
not-night-but: ^^^

apps/studio/tests/integration/lib/db/clients/mysql.privileges.spec.ts
```diff
@@ -0,0 +1,164 @@
+/*
```
not-night-but: These tests should be able to just be added to the mysql client test file. Ideally we'd have this functionality for every dialect and we could just add it to the common tests, but adding it to the mysql file should work for now.

not-night-but (changes requested):
This is better! There are still some issues with styles. 
- `TabUserManagementUser` should behave like `TabTableProperties.vue`, and it should be styled the same way (you should be able to just use the same classes that are used there). That way the feature will react to the theme so we don't end up with stuff like this happening:
image
- Reusing those styles also means that this doesn't look out of place in the app, right now it looks like it was inserted from somewhere else if that makes sense.
- The tests should be able to run within the existing mysql test suite in `apps/studio/tests/integration/lib/db/clients/mysql.spec.js`. 

As you're going along, it's also helpful if you can mark certain conversations as resolved so I know it has been seen and addressed

Thanks again for all your work on this, I know this is a big feature, so we really appreciate it!

apps/studio/src/components/CoreTabs.vue
RodrigoPerestrelo: This issue has been seen and addressed.
It is now using the matches function.

apps/studio/src/components/TabUserManagement.vue
RodrigoPerestrelo: This issue has been seen and addressed.
We reduced the custom styles to only the necessary: .table-properties-header and .table-properties-divider.
It now changes when the theme does.

apps/studio/src/components/TabUserManagementUser.vue
RodrigoPerestrelo: This issue has been seen and addressed.
We created new tabs for each component: TabUserManagementUserLimits; TabUserManagementUserLogin and TabUserManagementUserPrivileges.
They are now reactive to the theme state.

apps/studio/src/components/sidebar/core/UserManagement.vue
RodrigoPerestrelo: This issue has been seen and addressed.
We reviewed and most of the styles were already in available classes.
We decided to keep the styles only for the .dialog-buttons,  so that it would not appear like this:
image
but like this:
image

apps/studio/src/lib/db/clients/mysql.ts
RodrigoPerestrelo: This issue has been seen and addressed.

apps/studio/src/lib/utility/ElectronUtilityConnectionClient.ts
RodrigoPerestrelo: This issue has been seen and addressed.

apps/studio/src/store/index.ts
RodrigoPerestrelo: This issue has been seen and addressed.
We created a separate module for UserManagement called UserManagementModule.

apps/studio/src/store/modules/SidebarModule.ts
RodrigoPerestrelo: This issue has been seen and addressed.

apps/studio/src/components/TabUserManagementUser.vue
RodrigoPerestrelo: This issue has been seen and addressed.

apps/studio/src/components/TabUserManagementUser.vue
RodrigoPerestrelo: This issue has been seen and addressed.
Interfaces and enum moved to models.ts file.
The applyUserChanges function was also updated.

apps/studio/src/components/TabUserManagementUser.vue
RodrigoPerestrelo: This issue has been seen and addressed.
The function formatPrivilegeName was created, in order to reuse it.

apps/studio/src/components/TabUserManagementUser.vue
RodrigoPerestrelo: This issue has been seen and addressed.

apps/studio/src/components/TabUserManagementUser.vue
RodrigoPerestrelo: This issue has been seen and addressed.
A lot of these styles were removed but we kept some that we found necessary.

apps/studio/src/components/sidebar/core/UserManagement.vue
RodrigoPerestrelo: This issue has been seen and addressed.
We tried to reuse the styles from other components.

apps/studio/src/components/sidebar/core/UserManagement.vue
RodrigoPerestrelo: This issue has been seen and addressed.

apps/studio/src/components/sidebar/core/UserManagement.vue
RodrigoPerestrelo: This issue has been seen and addressed.
The styles were not necessary.

apps/studio/src/components/sidebar/core/user_management_list/UserManagementListItem.vue
RodrigoPerestrelo: This issue has been seen and addressed.
We tried to replicate what the other Lists did: we created a user-management-list.scss that is similar (but with a few changes) to other scss for lists already created; updating the base.scss to import it.

apps/studio/src/components/sidebar/core/user_management_list/UserManagementListItem.vue
RodrigoPerestrelo: This issue has been seen and addressed.
We copied it from other component but it was not necessary.

apps/studio/src/lib/db/clients/mysql.ts
RodrigoPerestrelo: This issue has been seen and addressed.

apps/studio/tests/integration/lib/db/clients/mysql.privileges.spec.ts
RodrigoPerestrelo: This issue has been seen and addressed.
We moved the tests to the mysql.spec.js client test file.

rathboma:
Been almost a year outstanding. I think there's a lot to like here, especially in the backend, but I can't merge this with the UX in the state it is in unfortunately.

UX and UI design is hard! We try really hard to adhere to the vibe of the application when we build new features, while powerful, the new UI this introduces does not line up at all with our current UI.

I'd like to implement user management. @joao-sntos if you have time and you're still interested, a great place to start would be to follow up with a PR that just adds the backend methods, then we can do a separate UI for the front-end components.

joao-sntos:
Sounds good, I’ll follow up by the end of the week with a new PR focused on the back-end

Issue #2939, opened 2025-03-26
FEAT: Name suggestion on creating a Foreign Key


jmperro:
Hi there!

It would be a great idea if, when creating a Foreign Key, Beekeeper suggested a name for it using some kind of nomenclature based on the names of the tables involved in the relationship.

For example:
If the table I'm creating/editing is called **clients** and the table I'm referencing is called **status**, the suggested name could be something like **FK_clients_status**

HideOnArbusto:
Hi! @not-night-but 
I would like to work on the implementation of this feature with @tomasmbrito . Could you please assign this issue to us?

Thanks!

PR #3255, opened 2025-06-06
feat: suggest foreign key names based on table names (#2939)

tomasmbrito:
This adds foreign key name suggestions using the format FK_fromTable_toTable. The feature is implemented in two parts:

SQL Text Editor: When the user writes or pastes a CREATE TABLE statement that includes a foreign key with no name or a non-standard name, a suggestion popup appears with the recommended name. If the user dismisses the suggestion, it won’t appear again for that specific foreign key.

Table Relations View: After a table is created, suggestions are shown for existing foreign keys that don’t follow the expected naming format. The user can apply or ignore the suggestion.

The goal is to improve consistency in foreign key naming. Suggestions are based on the names of the two tables involved in the relationship and follow the naming convention described in issue #2939.

Example:
Screenshot 2025-06-06 at 16 38 25
Screenshot 2025-06-06 at 16 38 52

If the table is "orders" and it references "users", the suggestion will be "FK_orders_users".

Co-authored-by: Henrique José Cavaco Ramalho [henrique.j.cavaco@tecnico.ulisboa.pt]

### DIFF
--- a/apps/studio/src/components/common/texteditor/SQLTextEditor.vue
+++ b/apps/studio/src/components/common/texteditor/SQLTextEditor.vue
@@ -1,22 +1,68 @@
 <template>
-  <text-editor
-    v-bind="$attrs"
-    :value="value"
-    @input="$emit('input', $event)"
-    :hint="hint"
-    :mode="dialectData.textEditorMode"
-    :extra-keybindings="keybindings"
-    :hint-options="hintOptions"
-    :columns-getter="columnsGetter"
-    :context-menu-options="handleContextMenuOptions"
-    :plugins="plugins"
-    :auto-focus="true"
-    @update:focus="$emit('update:focus', $event)"
-    @update:selection="$emit('update:selection', $event)"
-    @update:cursorIndex="$emit('update:cursorIndex', $event)"
-    @update:cursorIndexAnchor="$emit('update:cursorIndexAnchor', $event)"
-    @update:initialized="$emit('update:initialized', $event)"
-  />
+  <div>
+    <text-editor
+      v-bind="$attrs"
+      :value="value"
+      @input="handleInput"
+      :hint="hint"
+      :mode="dialectData.textEditorMode"
+      :extra-keybindings="keybindings"
+      :hint-options="hintOptions"
+      :columns-getter="columnsGetter"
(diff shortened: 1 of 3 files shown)


tomasmbrito:
Hi @not-night-but  please let us know if there’s anything you’d like us to adjust to better match what you had in mind.

not-night-but:
@tomasmbrito So sorry I missed this! I'll look through it right away

not-night-but (changes requested):
@tomasmbrito Okay so I think this looks great! I'm not sure how I feel about the modal just popping up while typing though, as I could see that becoming somewhat annoying to users. I like the suggestions popup you have in the table schema better. It's less intrusive, and doesn't interrupt the flow. I wonder if we could add that same flow to the query editor? Or maybe write an autocomplete plugin with this functionality instead? @azmy60 thoughts?

Also, since I think this needs some changes in that area, I didn't go too intensely through the code, but one question: I notice we have some regex there to identify foreign keys. How locked down to one dialect is that? Does it work with bigquery, postgres, mysql, cassandra, etc?

azmy60:
Yeah I'm on the same page with @not-night-but. I think adding those as autocomplete suggestions should be much better. And honestly, I might be wrong, but this seems to be very opinionated? Or is this a feature that other apps have but we don't? What do y'all think @not-night-but @rathboma ? Personally, I like the idea. I could imagine using this many times cause naming things is the most difficult task in engineering lol. I'm just not sure if this is a core feature or a plugin feature.

rathboma:
I like the idea of suggesting foreign key names, but the UX of a modal isn't appropriate for the feature.

Modals are great when you want to `stop` the user from doing something or draw them into a sub workflow of some sort. In this case we just want to give them a hint on how they should do something.

Way I think this should be handled:

1. SQL Editor: Some sort of autocomplete or 💡hint, which when clicked shows a tooltip with the suggestion and ability to accept/ignore
2. Table relations tab: It could auto-complete the name of the foreign key based on what options are selected, or even provide the same 💡system of accept/ignore

@tomasmbrito are you still willing to work on this? If so, please let us know!

rathboma:
Going to close this. Happy to reopen if the OP comes back and moves it forwards

Issue #2504, opened 2024-10-15
Enhancement Request: Auto-refresh on Double-clicking an Already Open Table


rathboma:
Currently, in Beekeeper Studio, if a user double clicks a table that is already open, there is no auto-refresh for that table. This can lead to confusion when new records are added via an API or other methods, as users expect to see updated data immediately upon interaction. 

Suggested enhancement: Implement an auto-refresh feature that triggers when a user double-clicks on a tab of an already open table. This would improve user experience by ensuring that the latest data is always visible without the need for a manual refresh.

Steps to reproduce:
1. Open a table in Beekeeper Studio.
2. Add new data to the table externally (e.g., via API).
3. Double-click the tab of the already open table. 
4. Observe that the table does not auto-refresh to show the updated data.

This feature improvement could enhance workflow efficiency and reduce manual actions for users who frequently interact with dynamic datasets.
***

*This issue was extracted from an email thread by [IssueWizard](https://issuewizard.com)*

rathboma:
I think the AI was meant to say something like:

If a tabletable is open, and the user double-clicks the table again from the sidebar, it doesn't do anything other than focus the tab. Rather it should potentially refresh the tab somehow? What if there are pending changes though?

rathboma:
Maybe we should just have a `The table structure has changed` warning, with a 'refresh' button.

PR #3256, opened 2025-06-07
Feature #2504: Implement auto-refresh for an already open table

T-MSD:
Implemented the table refresh feature with corresponding unit test. Added visual indicators to suggest that the table tab is double-clickable.
Includes hover animations, a loading spinner while refreshing
and display notifications regarding the success of the table refresh.

### DIFF
--- a/apps/studio/src/components/CoreTabHeader.vue
+++ b/apps/studio/src/components/CoreTabHeader.vue
@@ -7,8 +7,6 @@
       class="nav-item"
       :title="title + scope"
       @contextmenu="$emit('contextmenu', $event)"
-      @mouseenter="showTooltip('Double click to refresh')"
-      @mouseleave="hideTooltip"
     >
       <a
         class="nav-link"
@@ -26,12 +24,7 @@
           v-if="scope && !$store.getters.minimalMode"
           class="tab-title-scope"
         >{{ scope }}</span></span>
-        <div class="tab-action">
-          <span v-if="isLoading" class="loading-spinner">
-            <svg class="spinner-svg" viewBox="0 0 50 50">
-              <circle class="spinner-path" cx="25" cy="25" r="20" fill="none" stroke-width="4"/>
-            </svg>
-          </span>
+        <div class="tab-action"> 
           <span
             class="tab-close"
             @mouseenter="hover=true"
@@ -51,12 +44,6 @@
         </div>
       </a>
     </li>
-    <div 
-      v-if="tooltip.visible"
-      class="tooltip"
-    >
-      {{ tooltip.message }}
-    </div>
     <portal to="modals">
       <modal
         :name="modalName"
@@ -108,36 +95,9 @@ import { mapState } from 'vuex'
         hover: false,
         sureOpen: false,
         lastFocused: null,
-        isLoading: false,
-        tooltip: {
-          visible: false,
-          timeout: null,
-          message: '',
-        },
       }
     },
     methods: {
-      showTooltip(message) {
-        this.tooltipTimeout = setTimeout(() => {
-          if (this.tab.tabType === 'table') {
-            this.tooltip.visible = true;
-            this.tooltip.message = message;
-
-            this.$nextTick(() => {
-              const tooltipElement = document.querySelector('.tooltip');
-              const rect = this.$el.getBoundingClientRect(); // Get the position of the tab header
-
-              // Dynamically set the tooltip's position to align with the center of the tab header
-              tooltipElement.style.top = `${rect.top - tooltipElement.offsetHeight}px`;
-              tooltipElement.style.left = `${rect.left + rect.width / 2 - tooltipElement.offsetWidth / 2}px`;
-            });
-          }
-        }, 1000);
-      },
-      hideTooltip() {
-        clearTimeout(this.tooltipTimeout);
-        this.tooltip.visible = false;
-      },
       beforeOpened() {
         this.lastFocused = document.activeElement
       },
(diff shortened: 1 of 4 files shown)


tcctr:
Hey, just following up on this feature. Happy to clarify anything.

rathboma:
hey! Code looks alright -- can you post a screenshot of what this does? Is the goal of this to refresh tab content on double click?

rathboma:
Oh I see, my original issue was an AI generated one. 

I think the issue was meant to be about double-clicking the table in the sidebar actually. If you already have the table open now and you double click the table in the sidebar, it doesn't do anything. The issue is to fix that.

rathboma:
We do have a way to refresh a table from within the table tab (click the refresh button in the statusbar). Interested to see the feature as a gif or something though! Hopefully we can merge. Is it useful when you're using it?

T-MSD:
Hey! Yes the goal was to refresh the opened table with a double click. Since the feature said to refresh an already open table we thought it was the table on the current tab.

Here is a video showing how our feature works.
https://www.youtube.com/watch?v=SmpcIt4YHHQ

apps/studio/src/components/CoreTabHeader.vue
```diff
@@ -7,10 +7,13 @@
       class="nav-item"
       :title="title + scope"
       @contextmenu="$emit('contextmenu', $event)"
+      @mouseenter="showTooltip('Double click to refresh')"
+      @mouseleave="hideTooltip"
     >
       <a
         class="nav-link"
         @mousedown="mousedown"
+        @dblclick="handleDoubleClick"
```
not-night-but, line 14: So I don't think this is where we want this functionality. I believe the issue was that, if you double click a table in the entity list on the left, and the table is already open, it should trigger a refresh. So I'd move this functionality to the entity list item most likely.

apps/studio/src/components/CoreTabHeader.vue
```diff
@@ -7,10 +7,13 @@
       class="nav-item"
       :title="title + scope"
       @contextmenu="$emit('contextmenu', $event)"
+      @mouseenter="showTooltip('Double click to refresh')"
+      @mouseleave="hideTooltip"
```
not-night-but: I personally don't think we need a tooltip, I think this is more of a silent feature.

apps/studio/src/components/CoreTabHeader.vue
```diff
@@ -24,6 +27,11 @@
           class="tab-title-scope"
         >{{ scope }}</span></span>
         <div class="tab-action">
+          <span v-if="isLoading" class="loading-spinner">
+            <svg class="spinner-svg" viewBox="0 0 50 50">
+              <circle class="spinner-path" cx="25" cy="25" r="20" fill="none" stroke-width="4"/>
+            </svg>
+          </span>
```
not-night-but: We should already have this functionality in the app. It's in the TabIcon component, you should just have to mark the tab as running: `tab.isRunning = true` (at least that's how it *should* function, but if that doesn't work then it may need to be updated to support this)

apps/studio/src/components/CoreTabHeader.vue
```diff
@@ -93,10 +107,37 @@ import { mapState } from 'vuex'
...
+              tooltipElement.style.top = `${rect.top - tooltipElement.offsetHeight}px`;
+              tooltipElement.style.left = `${rect.left + rect.width / 2 - tooltipElement.offsetWidth / 2}px`;
+            });
+          }
+        }, 1000);
+      },
+      hideTooltip() {
+        clearTimeout(this.tooltipTimeout);
+        this.tooltip.visible = false;
+      },
```
not-night-but: Yeah I don't think we need to show a tooltip for this at all, it should just silently refresh the table if the user tries to open a table that is already open.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -379,6 +379,10 @@ export default Vue.extend({
       expandablePaths: [],
     };
   },
+  created() {
+    // Listen for the 'refresh-active-table' event
+    this.$root.$on('refresh-active-table', this.refreshActiveTable);
```
not-night-but: These should probably just be in our AppEvent system, and then bind them using rootBindings. See the rootBindings function in this component to see how we handle the switch tab event.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -705,6 +710,40 @@ export default Vue.extend({
...
+
+        if (equalTable && equalRows) {
+          this.$noty.warning('Table already up to date!');
+        } else if (!updatedTable) {
+          this.$noty.error('Failed to refresh the table!');
+        } else {
+          this.$noty.success('Table refreshed successfully!');
+        }
+      }, 1000);
+    },
```
not-night-but, line 742: I don't think you should have to do all of this, you should just be able to call the existing `refreshTable()` function and that should handle it for you.

not-night-but (changes requested):
The functionality is nice, but I'm not sure this is what the issue was asking for. I think this functionality probably needs to be move to the EntityListItem component, and we should use the same method of emitting/handling root events that the rest of the app does. Great work though! Let me know if you're up to refactoring this

T-MSD:
Hey @not-night-but, I'm up to refactor and already working on it!
I've already removed the tooltip and used the AppEvent system as requested. We already use the `refreshTable()` but we have extra logic to choose which message to display considering if the table had any new data to display or not.
Regarding the loading-spinner, using the `this.isRunning = true`  did not work, I also tried to using set and commit.

tcctr:
Hey, we have tried again using this.isRunning = true but it does not seem to work... maybe we are missing something. Is there anything more that we should be doing besides changing the value to true?

T-MSD:
Hi, just committed the changes, it might help understanding our problem.

not-night-but:
@T-MSD sorry I missed this. I'll take a look and see if I can't fix the isRunning issue. You should just be able to set that flag to true, so if that's not the case then something else is broken. In that case I'll fix and file a pr

PR #3268, opened 2025-06-11
Upgrade JSON Viewer to use UI Kit

### LINKED ISSUE #3050: Add `marker` property to UI Kit Text Editor

azmy60:
This also adds marker so this closes #3050


apps/studio/src/lib/editor/extensions/partialReadOnly.ts
```diff
@@ -0,0 +1,424 @@
+import {
```
azmy60, line 13: A lot of codes because many of these were handled in cm5, but now we have to track almost everything (checking marker positions, states, etc.) by ourselves lol.

rathboma: Damn this is a lot of code to handle this! Is this general enough to make it a full CM6 plugin that we can commit back to them?

apps/ui-kit/lib/components/text-editor/v2/extensions/lineGutters.ts
```diff
@@ -0,0 +1,91 @@
...
+ * Line Gutters Extension for CodeMirror 6
+ * 
+ * Adds visual indicators (gutters) to specific lines in the editor.
+ * Supports different gutter types like "changed" for marking modified lines.
+ * 
+ * Usage:
+ * 1. Add to extensions: lineGutters({ lineGutters: gutterArray })
+ * 2. Update gutters: applyLineGutters(view, newGutters)
+ */
+import { StateField, StateEffect, Extension } from "@codemirror/state";
```
rathboma, line 11: Damn so many extensions!

rathboma (approved):
Damn, that's a lot of custom CM6 code we need. Going to need a lot of testing

apps/studio/src/lib/editor/extensions/partialReadOnly.ts
azmy60: I think our best chance is to publish our own CodeMirror extension.

azmy60: Need a little tweak before we can do that though.

PR #3272, opened 2025-06-12
Remake query magics autocompletion for cm6

azmy60:


rathboma (approved):
Oh man, so many CM6 extensions

azmy60:
It looks a lot mostly because cm6 extension is verbose 😢

Issue #2431, opened 2024-09-15
FEAT: Connection Quick Switcher ⚡


rathboma:
As a stop gap between where we are and full multi-connection mode, let's first just make it much easier to quickly switch connections.

This is meant to happen before #1667 

This is in two parts

- [ ] A new menu item to the bottom-left connection menu
- [ ] New entries in quicksearch (cmd/ctrl+p) for connections

How does it work?
- This follows our existing disconnect/reconnect logic. It doesn't try to maintain multiple connections at the same time. It's the same as clicking 'disconnect', going back to the connection screen, then clicking 'connect' on a different database.

Why?
Building full multi-connect behavior is a much larger piece of work. This would give folks ~70% of what they need with ~20% of the work. Seems like a win

![Multiple Connections quick fix](https://github.com/user-attachments/assets/d9b91962-4b36-47d0-ad13-fd2dfeffc7f9)

tjrchampion:
is this being developed? I like your product, but i find having many windows per db connection open very frustrating.

not-night-but:
Reopening as I believe we still want to have these options in quicksearch, can be done in a new PR.

PR #3275, opened 2025-06-12
feat #2431: Connection Quick Switcher

SMV-21:
Closes: #2431
Introduced a Connection Quick Switcher accessible via the connection menu, allowing users to quickly switch between recent or saved connections.

- Added "Switch Connection" button to the status bar menu
- Added quick-switcher-panel with recent and saved connections toggle
- Populates recent connections (max 5) and recent connections excluding the active one

### DIFF
--- a/apps/studio/src/assets/styles/app/sidebar/sidebar.scss
+++ b/apps/studio/src/assets/styles/app/sidebar/sidebar.scss
@@ -618,3 +618,77 @@
   }
 
 }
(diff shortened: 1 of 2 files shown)


rathboma:
This is great! Probably need to do a little style tidy up, but we can do that in Master :-)

rathboma (approved):
LGTM

rathboma:
Thanks so much!

PR #3277, opened 2025-06-12
feat(redis): basic redis support

### LINKED ISSUE #80: Redis connection option
Would love to see a Redis connection option similar to Table Plus. Great work, will be watching this project for updates!

mrsimb:
As per title, added basic redis support. Resolves #80.

Stuff that works:
- listing keys, their types, ttl, and occupied memory
- filtering keys (use `raw filter`)
- querying (similar to redis-cli, thanks to the [redis-splitargs](https://github.com/luin/splitargs) author for this)
- listing redis server info
- pagination (though it needs cursor pagination, which is only supported when exporting data in beekeeper)
- selecting database
- redis connection form
- showing/editing key content in table view
- showing/editing key content in JSON viewer
- syntax highlighting and completions in query editor
- changing key name
- changing key TTL in seconds
- streams

Stuff that doesn't (yet):
- showing pub/subs (not sure if i want to add it)


mrsimb:
@rathboma I'm going to add `TableTable.updateDetailsViewer()` in addition to existing `updateJsonViewer()` and `BasicDatabaseClient.getRowDetails()` which will return an array of detailed views, including:
- text data view (will use codemirror as is)
- list view? (may use json view for that)
- binary data view (need to find a hex viewer)
- json data view (will use JsonViewer)

Default implementation will return just one detailed view - json, which is existing behavior.
Since there could be multiple data views, we would be able to see the data in different slices, or interpretations.

What do you think?

mrsimb:
Oh and `getRowDetails()` is `async`, meaning we can retrieve additional info to display it (key's content in case of Redis).
Since Redis key can contain wast amount of data, i do not display nor load it in TableTable when `selectTop()` is called.

mrsimb:
This might be expanded to showing graphs/plots (if applicable) when selecting multiple rows, but i'm not sure if there will be demand on such feature, since you can do the same by copying the data to excel sheet.

rathboma:
This is awesome! Thanks for starting on this. Damn I didn't think we'd have Redis support for ages :-D.

Do you have time to work on the remaining items?

Haven't done a full review yet, but your code looks nice, clean, and well organized so far, so great job!

rathboma:
Most boring bit will be testing, but I bet we can get the AI to help out there

mrsimb:
I only have time on weekends, but yeah, I'll get back onto this soon

mrsimb:
However you can take it from here if you'd like, i definitely wouldn't mind :)

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -0,0 +1,634 @@
...
+      "ZSCORE",
+      "KEYS",
+      "SCAN",
+      "FLUSHDB",
+      "FLUSHALL",
+      "INFO",
+      "PING",
+    ];
+  }
+
```
not-night-but: This is actually a mongo only thing, so you shouldn't need to add this. Unless you think it would be useful to add frontend support for returning specific keywords for specific dialects?

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -0,0 +1,634 @@
...
+const NEWLINE_RG = /[\r\n]+/;
+
+export class RedisClient extends BasicDatabaseClient<RedisQueryResult> {
+  redis: Redis;
+
+  constructor(server: IDbConnectionServer, database: IDbConnectionDatabase) {
+    super(null, redisContext, server, database);
+  }
+
+  async connect(): Promise<void> {
```
not-night-but, line 92: You should call `await super.connect()` at the beginning of this function. Just sets up a couple things for the class (and it handles ssh tunneling if you want to get that working)

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -0,0 +1,634 @@
...
+
+  async queryStream(): Promise<StreamResults> {
+    throw new Error("Not implemented");
+  }
+
+  async rawExecuteQuery(
+    _query: string
+  ): Promise<never> {
+    throw new Error("Where is this called from?");
+  }
```
not-night-but: Raw execute query is called from the BasicDatabaseClient. Typically what we like to do is just implement this, and then when we want to query from within the driver, we call `this.driverExecuteSingle` or `this.driverExecuteMultiple` (execute a single query or execute multiple). It provides guards for readonly mode and does some other things. I'm not sure how relevent this will be for redis, as I'm honestly not too familiar with the db. We did have to completely cut this out of our mongo driver for example, because it just didn't fit that paradigm.

apps/studio/src/shared/lib/dialects/redis.ts
```diff
@@ -0,0 +1,74 @@
+import {  DialectData } from "./models";
+
+export const RedisData: DialectData = {
+  columnTypes: [],
+  usesOffsetPagination: true,
+  queryDialectOverride: 'generic', // TODO: idk what to put there
+  textEditorMode: 'text/x-pgsql', // TODO: there's no "redis" mode for codemirror
```
not-night-but: The dialect override I think was specifically used for mongo, but I don't think it's relevant here. In the case of mongo, we have the ability to query with a postgres-like syntax, which is why we override the query dialect for it (because otherwise it just uses the connection type). We may need to think about looking into developing a redis mode for codemirror

apps/studio/src/shared/lib/dialects/redis.ts
```diff
@@ -0,0 +1,74 @@
...
+  disabledFeatures: {
+    // rawFilters: true, // TODO: how to disable non-raw filters?
+    shell: true,
```
not-night-but, line 12: Does redis need a shell like experience? Currently our shell is pretty heavily tied to mongo, but we could definitely make it more generic if this is the case.

apps/studio/src/lib/db/clients/redis.ts
mrsimb: I thought this is required for auto completion to work. If custom redis codemirror plugin would do this, then this is not needed. I'll remove this. In any case, making a plugin for redis is a bit out of scope for me now, maybe after the basic stuff is finished.

apps/studio/src/lib/db/clients/redis.ts
mrsimb: Is this related to ability to select some lines in query editor to run them, isolated from the whole code?
Read only mode would be a nice feature to have for redis.

apps/studio/src/shared/lib/dialects/redis.ts
mrsimb: Agree about codemirror mode

apps/studio/src/shared/lib/dialects/redis.ts
mrsimb: I'm not sure i understand what shell is. Is this something similar to what is psql for postgres?

mrsimb:
Hey and thanks for the review, your help is appreciated

apps/studio/src/lib/db/clients/redis.ts
not-night-but: For sure! I think basic functionality is most important before we start getting in the weeds of stuff.

This function was only added for the mongo shell integration, but depending on whether or not this is a shell, we may want to keep this implementation.

apps/studio/src/lib/db/clients/redis.ts
not-night-but: No that's a frontend thing. When we select some lines and run them, it just takes the selection and sends that to the backend. This was just a way to have the same API for all drivers, while also having the ability to add checks to every call if that makes sense

apps/studio/src/shared/lib/dialects/redis.ts
not-night-but: image

This is shell. If you're wanting redis to have a REPL shell like this, then we'll need to update the shell to be less tied to mongo (I can help with this).

not-night-but: (ie, I type a command, press enter, it returns the results, and I have another prompt to interact with)

mrsimb: oh now i get it
no, redis has no shell feature

mrsimb:
@not-night-but  I added value loading and displaying for these types in the JSON sidebar: "string" | "list" | "hash" | "set" | "zset" | "ReJSON-RL" (JSON addon).

You can set some values in query editor to view them in "keys" table. See how to use redis-cli for help or generate some queries in chatgpt.

I originally planned to create a separate, generalized key-value storage view alongside the JSON viewer, but I ditched that idea since it’s too time consuming.

I’m going to have to leave this as is for now, as I have a pretty busy period at work. I’ll keep monitoring this thread though, so feel free to ask me anything!

mrsimb:
im back, gonna have a look at what's changed since my last pull from master

apps/studio/src/lib/db/clients/redis.ts
mrsimb: Still not quite understanding what it is for, will mark it with more explicit error (e.g. "not supported") for now.

mrsimb:
image

redis autocomplete ✨

mrsimb:
image

value editor sidebar (i ditched idea with using json viewer for editing values, it is not convenient)

mrsimb:
@not-night-but can you review this please so i know if our opinions still align?

not-night-but:
@mrsimb I'm going to take a look at this today, but would you be able to add a docker container to the `docker-compose.yml` and add a dev migration to add a test connection for us? (Look at the `dev-x.js` migrations for an example). Ideally we like to have a container for every database set up there with some test data so we can quickly reproduce issues specific to that db.

apps/studio/src/components/connection/RedisForm.vue
```diff
@@ -0,0 +1,122 @@
...
+
+    <div class="form-group">
+      <label for="port">Port</label>
+      <input
+        id="port"
+        v-model.number="config.port"
+        type="number"
+        class="form-control"
+        placeholder="6379"
+      >
```
not-night-but: These should be masked inputs so that the values are not shown when privacy mode is enabled.

apps/studio/src/components/connection/RedisForm.vue
```diff
@@ -0,0 +1,122 @@
...
+    <!-- Authentication -->
+    <div class="form-group">
+      <label for="password">Password (optional)</label>
+      <input
+        id="password"
+        v-model="config.password"
+        type="password"
+        class="form-control"
+        placeholder="password"
+      >
```
not-night-but: Look in `CommonServerInputs.vue` for how we style our password (mainly just the toggle button is missing here). In general we'd like this form to look as similar to the other forms as possible, ie how the host and port are normally on the same line: 

Image

apps/studio/src/components/connection/RedisForm.vue
```diff
@@ -0,0 +1,122 @@
...
+.form-control::placeholder {
+  color: var(--text-lighter);
+}
+.form-text {
+  margin-top: 0.25rem;
+  font-size: 0.875rem;
+}
+.text-muted {
+  color: var(--text-lighter);
+}
```
not-night-but: I don't think you should have to use this much manual styling for the form as long as you follow the same general structure as the rest of the forms we have

apps/studio/src/components/connection/RedisForm.vue
```diff
@@ -0,0 +1,122 @@
...
+    if (!this.config.port) {
+      this.$set(this.config, "port", 6379);
+    }
+    if (!this.config.defaultDatabase) {
+      this.$set(this.config, "defaultDatabase", "0");
+    }
+    if (!this.config.options) {
+      this.$set(this.config, "options", {});
+    }
+  },
```
not-night-but, line 101: We shouldn't be setting the defaults here. Host and port should both at the very least be set in the saved_connection model (look at the defaultPort getter, make sure you add an entry for redis).

apps/studio/src/components/sidebar/SecondarySidebar.vue
```diff
@@ -26,6 +26,14 @@
...
           />
         </template>
+        <template v-else-if="tab.id === 'value-editor'">
+          <value-editor
+            v-show="secondaryActiveTabId === 'value-editor'"
+            :key="tab.id"
+            :value-context="valueContext"
+            @value-editor-change="handleValueEditorChange"
+          />
+        </template>
```
not-night-but: I think this makes sense to have, may have to ask @azmy60 how he feels about it though.

One note: it is kind of hard to tell that you can click the value editor heading to switch tabs. May need a hover effect and change the pointer

apps/studio/src/components/sidebar/ValueEditor.vue
```diff
@@ -0,0 +1,254 @@
...
+  },
+  created() {
+    // Create debounced function with 500ms delay
+    this.debouncedEmitChange = debounce(this.emitChange, 500);
+  },
+};
+</script>
+
+<style scoped>
+.value-editor {
```
not-night-but: I would suggest not having the styling be in the vue component. We tend to like to have the bulk of our styles in their own .scss files, that way we can easily use the variables and such (of course they can be used here too, just a bit more effort imo)

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1837,6 +1851,24 @@ export default Vue.extend({
       this.selectedRowPosition = position
       this.selectedRowIndex = this.primaryKeys?.map((key: string) => data[key]).join(',');
       this.selectedRowData = parseRowDataForJsonViewer(cleanedData, this.tableColumns)
+
+      // TODO: move value editor update logic to a separate method?
```
not-night-but: I agree with this. And maybe make it so the even also doesn't get triggered if this isn't redis, just ignore the updates cause I don't think the value editor will be shown if it is not redis, unless we think that would be useful?

apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -83,6 +83,11 @@
                   :config="config"
                   :testing="testing"
                 />
+                <redis-form
+                  v-else-if="config.connectionType === 'redis' && isUltimate"
```
not-night-but: Is this going to be an ultimate dialect?

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
```diff
@@ -65,7 +85,7 @@ export interface BaseQueryResult {
 export abstract class BasicDatabaseClient<RawResultType extends BaseQueryResult> implements IBasicDatabaseClient {
   knex: Knex | null;
   contextProvider: AppContextProvider;
-  dialect: "mssql" | "sqlite" | "mysql" | "oracle" | "psql" | "bigquery" | "generic";
+  dialect: "mssql" | "sqlite" | "mysql" | "oracle" | "psql" | "bigquery" | "generic" | "redis";
```
not-night-but: I believe these are `sql-query-identifier` dialects, so we don't need to add this here.

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -0,0 +1,853 @@
...
+    };
+  }
+
+  async executeQuery(queryText: string): Promise<NgQueryResult[]> {
+    return this.executeCommand(queryText);
+  }
+
+  async listDatabases() {
+    // TODO: check actual database count from server info?
+    return new Array(16).fill(null).map((_, i) => String(i));
```
not-night-but: I think this would definitely be a much better option than what is here now

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -0,0 +1,853 @@
...
+    }
+
+    return {
+      type,
+      key,
+      ttl: await this.redis.ttl(key),
+      memory,
+      encoding,
+    };
+  };
```
not-night-but: Just a style thing, I'd rather these be declared in the same way the other functions are in this class. Just for consistency's sake

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -0,0 +1,853 @@
...
+      .split(NEWLINE_RG)
+      .map((c) => c.trim())
+      .filter(Boolean);
+
+    const results: NgQueryResult[] = [];
+
+    for (const command of commands) {
+      const args = splitargs(command);
+      try {
+        const result = await this.redis.call.call(this.redis, ...args);
```
not-night-but, line 283: Any dangers of injection attacks here?

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -0,0 +1,853 @@
...
+    return `SCAN 0 MATCH * COUNT ${limit}`;
+  }
+
+  async selectTopStream(): Promise<StreamResults> {
+    throw new Error("Not implemented");
+  }
+
+  async queryStream(): Promise<StreamResults> {
+    throw new Error("Not implemented");
+  }
```
not-night-but, line 607: Not sure if this is something redis users would want, but looking at the scan code you have, we could probably implement these relatively easy which would allow for 'table' export and 'query' export functionality

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -0,0 +1,853 @@
...
+
+  private getOpKey(op: TableUpdate | TableDelete): string {
+    return op.primaryKeys.find((pk) => pk.column === "key")!.value;
+  }
+
+  async executeApplyChanges(
+    changes: TableChanges
+  ): Promise<TableUpdateResult[]> {
+    log.debug("Redis executeApplyChanges called with:", changes);
```
not-night-but, line 750: Ignorant non-redis user question: are transactions possible, or anything approaching a transaction? Generally we try to wrap everything in this function in a transaction so we can roll back if one of the changes fails

apps/studio/vite.config.mjs
```diff
@@ -16,8 +16,16 @@ export default defineConfig({
   plugins: [vue(), commonjs()],
   base: '/', // Set the base URL for the app
   optimizeDeps: {
-    exclude: []
-  },
+    exclude: [
+      // Exclude native modules from optimization
+      // Without this, the build fails :(
```
not-night-but, line 21: Huh. That's a new one

not-night-but (commented):
These are just some general thoughts, but this is an insane contribution! Thank you so much for your work on this (especially the codemirror stuff, I know that can be a bloody nightmare). I'm not sure where you consider your work on this to be, but I'm available for questions/discussions if you need.

mrsimb:
> would you be able to add a docker container to the docker-compose.yml and add a dev migration to add a test connection for us? 

sure
as for now, you can just start redis-server and click connect in connection options, it should work with defaults

edit: added migration and seeds

apps/studio/src/components/connection/RedisForm.vue
mrsimb: yeah, most of vue code was done with claude code (im a react person myself) so i wasn't paying too much attention to styling. will fix it

apps/studio/src/components/sidebar/SecondarySidebar.vue
mrsimb: i'm still not sure if i really want to add this. feels like too much new stuff, that should be maintained too. but it is definitely more convenient to read long text content with it.

apps/studio/src/lib/db/clients/redis.ts
mrsimb: not sure redis can have those. [this is](https://github.com/luin/medis/blob/12c87a5a2cc3fd7fa616beb2eaed79413538769a/src/renderer/windows/MainWindow/InstanceContent/DatabaseContainer/Content/Terminal/index.jsx#L5) how medis does query stuff, i don't see any sanitization logic.

apps/studio/src/lib/db/clients/redis.ts
mrsimb: Redis has MULTI + EXEC command, which allow user to prepare multiple commands and then execute them atomically. But it doesn't have concept of rolling back if one of the commands failed (like doing INCR on a string). I can wrap these into MULTI just to have atomicity, if needed.

apps/studio/vite.config.mjs
mrsimb: Yeah, i've read somewhere in the code that someone wanted to get rid of those in client runtime, so that's how it is done.

apps/studio/src/components/tableview/TableTable.vue
mrsimb: I think it works for non-redis clients too! Haven't tested much, but at least it shows selected column's content in SQL databases.

apps/studio/src/components/ConnectionInterface.vue
mrsimb: I don't know, i thought so when started implementing this because json-viewer was a paid feature, and i used it for showing values.

apps/studio/src/lib/db/clients/redis.ts
mrsimb: The failed commands will just error out, and the other ones will be executed despite the errors.

mrsimb:
Added "value" column. I hesitated to do this at first, because there can be massive amounts of data in keys content, so i opted to lazy loading and adding the whole `ValueContext` thing. But beekeeper might just handle it. Also, maybe using json viewer isn't that bad...

image

mrsimb:
Yeah, i think i overcomplicated it. I'll remove value editor sidebar.

apps/studio/src/components/tableview/TableTable.vue
mrsimb: Removed these changes entirely.

apps/studio/src/components/sidebar/SecondarySidebar.vue
mrsimb: Removed these changes entirely.

apps/studio/src/components/sidebar/ValueEditor.vue
mrsimb: Removed entirely.

apps/studio/src/components/ConnectionInterface.vue
mrsimb: Made it available for non ultimate users, you guys decide which should it be later.

mrsimb:
@not-night-but I addressed most of the stuff you wrote, and reverted to using json-viewer for editing values. The implementation is almost complete, only value setting for complex types (it's broken right now) and some minor issues are left i think.

Also cleaned up the diffs.

mrsimb:
F-d up the rebase. Hold on...

mrsimb:
@not-night-but I think it can be considered readyish.
The only thing is the condition filter needs to be disabled for redis, only raw filter is supported. Let me know if you'll find something else.

mrsimb:
Added streams support, and disabled builder filter for redis (only raw filter is available, for filtering keys). I think it is feature complete now. Need to add tests though.

not-night-but:
@mrsimb Okay cool! A couple things:
- I demoed this for the team this morning, and everyone was super impressed, so once again thanks for your work on this.
- We did actually test this out with our AI shell, and it works out of the box (after telling the ai to use redis commands, not sql so we'll have to preload that into our messages but it's cool nonetheless).
- We've decided we want this to be a community feature, as it kind of feels wrong to take your amazing community contribution and then lock it behind a paywall. I get that the json-viewer may help with the use of redis, but with the editor modal I think this should be fine? If you want to add some redis specific behaviour to the editor modal let me know (right click on a cell and `Edit in modal` or just press `Shift+Enter` on a cell)
- I think the state of it is quite good, so we'll hopefully get it out in either 5.4 or 5.5, but as you seem to be quite familiar with redis, I'd ask that if you figure out any features you think would be a good add, file an issue so we keep track of it (of course you're more than welcome to keep adding things, but if not it's good for us to have a todo list to improve things)

apps/studio/src/components/connection/RedisForm.vue
```diff
@@ -0,0 +1,101 @@
...
+        id="database"
+        v-model="config.defaultDatabase"
+        type="text"
+        class="form-control"
+        placeholder="0"
+      >
+      <small class="form-text text-muted">
+        Not applicable to Redis Cluster
+      </small>
+    </div>
```
not-night-but: Perhaps we should make this a number input? Since it has to be a number anyways

mrsimb:
@not-night-but Good to hear (read?) that :)

I didn't know there's modal editor, it should probably be invoked by default when user tries to edit a redis value. Other than that, it suits perfectly. Just found a bug though, both when editing using the modal and when using the in-place cell input. I'll fix it tomorrow.

And yeah, i'll let you guys know if i have any new ideas for features.

apps/studio/src/lib/db/clients/redis.ts
mrsimb: Oh, it actually supports evaluating lua code https://redis.io/docs/latest/develop/programmability/eval-intro/

mrsimb:
@not-night-but I fixed issue when editing values in TableTable's cell or in modal. Also added proper validation for input, so its harder to f*** up your data. Oh and it reads real database count now.

The branch is rebased to the current master (had to create new file for migration, since dev-9 was occupied by Trino).

apps/studio/src/shared/lib/dialects/models.ts
```diff
@@ -161,8 +162,7 @@ export interface DialectData {
     headerSort?: boolean,
     duplicateTable?: boolean,
     export?: {
-      sql?: boolean,
-      stream?: boolean
```
not-night-but: We also still need this

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -0,0 +1,875 @@
...
+  }
+
+  addColumn(_item: SchemaItem): string {
+    throw new Error("Not supported");
+  }
+
+  dropColumn(_column: string): string {
+    throw new Error("Not supported");
+  }
+}
```
not-night-but: I get this is kinda a throwaway class, but I'd still put it in it's own file. (`apps/studio/src/shared/lib/sql/change_builder/`)

not-night-but (dismissed):
Okay just a couple little things. And it looks like possibly some merge issues? Looks like some trino stuff has been removed.

mrsimb:
@not-night-but yeah it was after rebase I guess. Do you mind if i just reset the branch to a state before the rebase, and you'll handle merge conflicts?

not-night-but:
@mrsimb yeah *rebasing* can be super annoying lol. Go for it

apps/studio/src/lib/db/clients/redis.ts
mrsimb: Moved

apps/studio/src/components/connection/RedisForm.vue
mrsimb: Changed the input type

mrsimb:
@not-night-but Done, you can check it out now

PR #3280, opened 2025-06-13
add show table in plugin shell

azmy60:
image
image


rathboma:
Simple solution! Don't like the icon though, feel free to merge and we can pick a better icon later.

rathboma (approved):

azmy60:
@rathboma I don't like it too lol. I mean that's probably one of the best we can get from material icons.

PR #3282, opened 2025-06-13
fix: handle read-only permission errors gracefully in PostgreSQL table properties

### LINKED ISSUE #3262: BUG: Relations not showing for read-only Postgres connection
**Describe the bug**
Table relations are not visible when the DB connection is read-only.

**To Reproduce**
1. Create a Postgres DB with tables `t1` and `t2`.
2. Create a column in `t1` referencing entries of `t2` via a foreign key constraint.
3.Connect to the Postgres DB with a user with read-only credentials.
4. Right click the `t1` entity in the left sidebar.
5. Click "View Structure".
6. Click the "Relations" tab.
7. Observe no relations are listed despite the column in `t1` referencing entries of `t2` via a foreign key constraint.

**Expected behavior**
The relation of `t1` referencing entries of `t2` via a foreign key constraint is listed in the "Relations" tab after opening the "View Structure" view.

 - OS: MacOS Sequoia 15.5
 - App Version: Beekeeper Studio Version 5.2.12
 - Database type and version: PostgreSQL 16.8 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 7.3.1 20180712 (Red Hat 7.3.1-17), 64-bit

rathboma:
Fixes issue where Relations tab shows empty for read-only PostgreSQL connections.

## Changes
- Modified getTableProperties method to handle permission errors gracefully
- Added individual error handlers for each query to prevent total failure
- Ensures relations are still displayed even when permission-sensitive queries fail
- Added comprehensive unit test to prevent regression

Resolves #3262

### DIFF
--- a/apps/studio/src/components/TabTableProperties.vue
+++ b/apps/studio/src/components/TabTableProperties.vue
@@ -86,6 +86,14 @@
                 >
                   <i class="material-icons-outlined">report_problem</i> Read Only
                 </span>
+                <span
+                  class="statusbar-item permission-warning"
+                  v-if="properties && properties.permissionWarnings && properties.permissionWarnings.length > 0"
+                  :title="permissionWarningsTooltip"
+                  @click="showPermissionWarnings"
+                >
+                  <i class="material-icons-outlined">warning</i> Limited Info
+                </span>
               </template>
             </div>
           </template>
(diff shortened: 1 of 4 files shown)


rathboma:
@claude could we be using different SQL that doesn't require elevated perms to do the same things here?

rathboma:
@claude looks like your tests don't work

rathboma:
@claude does this affect derivative drivers? Eg Cockroach / Redshift / others?

rathboma:
@claude we should have some way of telling the user which information we were unable to show due to permissions errors, currently that's hidden from them. Can you add some sort of error message to the response handling and figure out how to deal with that on the front-end? Maybe with something in the statusbar or as a noty notification?

PR #3283, opened 2025-06-13
fix: handle schema drops correctly in PostgreSQL client

### LINKED ISSUE #3260: Dropping a schema results in an unexpected (app) error
Dropping whole schemas currently does not work, at least for postgres databases (did not test others yet)

**To Reproduce**

1. Open a postgres database
2. Right click on any schema (eg. `public`)
3. Press `Drop`

Results in the following error:

```
Error performing drop: Cannot read properties of undefined (reading 'match')
```

**Expected behavior**
It should drop the whole schema

 - OS: macOS 15.5
 - App Version: 5.2.12
 - Database type and version: pgvector 16

rathboma:
This PR fixes the bug where dropping a PostgreSQL schema results in "Cannot read properties of undefined (reading 'match')" error.

## Changes
- Fixed `dropElement` method to treat schemas as top-level objects
- Added null/undefined checks to `wrapIdentifier` functions
- Now generates correct SQL: `DROP SCHEMA schema_name`

## Testing
- Verified fix handles both schema drops and table/view drops
- Added proper null safety to prevent crashes

Fixes #3260

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -997,7 +997,7 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult> {
   }
 
   wrapIdentifier(value: string): string {
-    if (value === '*') return value;
+    if (!value || value === '*') return value;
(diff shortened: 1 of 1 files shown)


rathboma:
@claude please add an integration test for this

PR #3284, opened 2025-06-13
fix: use pg_constraint system catalog for retrieving foreign keys in PostgreSQL

### LINKED ISSUE #1557: QUESTION: Should Viewing Relations Show All Foreign Keys on a Given Table?
**Details:**
Asking this as a question rather than logging a bug as I'm not sure if I just have the wrong expectations about what the `relations` tab does when viewing a table's structure. 

My expectation is that I would see any columns that referenced another table, and some details about that. 

When I run the following query: 

```
SELECT conrelid::regclass AS table_name, 
       conname AS foreign_key, 
       pg_get_constraintdef(oid) 
FROM   pg_constraint 
WHERE  contype = 'f' 
AND    connamespace = 'public'::regnamespace   
ORDER  BY conrelid::regclass::text, contype DESC;
```

I see the foreign keys I would expect for a given table, but they do not show up in the relations tab. 

So I guess my real question is: what should the relations tab do? 

Operating System: Mac OS 13.2.1 (Ventura)
App Version [help -> about]: 3.8.9
Database type and version [eg Posgresql 9.3]: PostgreSQL 12.12 on x86_64-pc-linux-gnu, compiled by x86_64-pc-linux-gnu-gcc (GCC) 7.4.0, 64-bit

Screenshot 2023-03-29 at 1 19 50 PM

Screenshot 2023-03-29 at 1 19 56 PM

rathboma:
Fixes #1557

The Relations tab was not showing foreign keys that should be visible because the getTableKeys() method was using information_schema views instead of the more direct pg_constraint system catalog approach.

## Changes
- Replace information_schema query with pg_constraint system catalog query
- Properly map PostgreSQL action codes to readable strings
- Maintain same return structure for compatibility
- Support both simple and composite foreign keys

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -557,35 +557,41 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult> {
   async getTableKeys(table: string, schema: string = this._defaultSchema): Promise<TableKey[]> {
     const sql = `
       SELECT 
-        tc.constraint_name,
-        kcu.column_name,
-        kcu.table_schema AS from_schema,
-        kcu.table_name AS from_table,
-        kcu2.column_name AS to_column,
-        kcu2.table_name AS to_table,
-        rc.unique_constraint_schema AS to_schema,
-        rc.update_rule,
-        rc.delete_rule,
-        kcu.ordinal_position
-      FROM 
-        information_schema.table_constraints AS tc 
-        JOIN information_schema.key_column_usage AS kcu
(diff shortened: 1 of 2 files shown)


rathboma:
@claude Questions to answer: 
1. How does this affect derivative drivers? (eg cockroach, redshift)
2. Can you reproduce the original error and add a test that verifies this fix?
3. Does this query require any different permissions vs the old query?

rathboma:
@claude, please fix the cockroack and postgres integration tests, you can run a specific integration test with `yarn test:integration <path-to-test-file>`

PR #3290, opened 2025-06-17
Mongo Shell CM6

not-night-but:


apps/ui-kit/lib/components/mongo-shell/MongoShell.vue
```diff
@@ -0,0 +1,321 @@
...
+          scrollIntoView: true
+        })
+      }
+
+      // this will need to update the shell prompt first
+      this.resetPrompt();
+    }
+  },
+  computed: {
+    replaceExtensions() {
```
azmy60, line 97: I'm fine with this, but another way to add extensions is by making a new class for Mongo Shell, and add `getExtensions()` method. See [SQLTextEditor](https://github.com/beekeeper-studio/beekeeper-studio/blob/master/apps/ui-kit/lib/components/sql-text-editor/v2/SqlTextEditor.ts#L50) for example. We do that to allow users to use `replaceExtensions` as a prop, without internal codes interfering.

apps/ui-kit/lib/components/mongo-shell/props.ts
```diff
@@ -0,0 +1,36 @@
...
+  promptSymbol: {
+    type: String,
+    default: "mongo> "
+  },
+  maxPromptSymbolLength: {
+    type: Number,
+    default: 30
+  },
+  output: Object,
+  extensions: [Array, Function] as PropType<Extension>,
```
azmy60, line 34: I like that we should really name it `extensions` rather than `replaceExtensions1

apps/ui-kit/lib/components/mongo-shell/MongoShell.vue
```diff
@@ -0,0 +1,321 @@
+<template>
+  <div class="BksUiKit BksTextEditor BksSqlTextEditor" ref="shell"></div>
```
azmy60: Might need to change the class name, not `BksSqlTextEditor`.

apps/ui-kit/lib/components/mongo-shell/MongoShell.vue
```diff
@@ -0,0 +1,321 @@
...
+    //await this.initialize();
+  }
+}
+</script>
+
+<style lang="scss">
+@use 'sass:color';
+@import '../../styles/_variables.scss';
+ 
+.cm-mongo-prompt > span {
```
azmy60: For UI Kit, I would make the styles to be really scoped with the component name, for example:

```scss
.BksSqlTextEditor {
  // all classes for mongo shell inside
}
```

apps/ui-kit/lib/components/mongo-shell/MongoShell.vue
```diff
@@ -0,0 +1,321 @@
...
+      this.textEditor.dispatchChange({
+        changes: { from: insertPos, to: length, insert: cmd },
+      });
+      const newLength = this.textEditor.getLength();
+      this.textEditor.dispatchChange({
+        selection: { anchor: newLength },
+        scrollIntoView: true
+      })
+    },
+    constructTextEditor() {
```
azmy60, line 278: If you use the mixin, this can be replaced with a MongoShell class.

apps/ui-kit/lib/components/mongo-shell/MongoShell.vue
```diff
@@ -0,0 +1,321 @@
+<template>
+  <div class="BksUiKit BksTextEditor BksSqlTextEditor" ref="shell"></div>
+</template>
+
+<script lang="ts">
+import { TextEditor } from '../text-editor/v2/TextEditor';
+import { monokai } from '@uiw/codemirror-theme-monokai';
```
azmy60: I don't know if we should use monokai to be a built-in thing for UI Kit? Should we actually?

apps/ui-kit/lib/components/mongo-shell/MongoShell.vue
azmy60: I'm not super strict with you with all these styling "rules", and I know it's not perfect. Might need to get some feedbacks too, but the idea is this; you can make your own `TextEditor` class with `constructTextEditor()` method. Let's say that MongoShell needs to add some new extensions. In that case, like SQLTextEditor, the pattern is:

```ts
// MongoShell.vue
{
  methods: {
    constructTextEditor() {
      return new MongoShell();
    },
  },
}
```
```ts
// MongoShell.ts
class MongoShell extends TextEditor {
  // NOTE: this is where we customize the extensions internally
  protected getExtensions(config: ExtensionConfiguration): Extension[] {
    const baseExtensions = super.getExtensions(config);
    return [
      baseExtensions, 
      mongoShellExtensions,
    ];
  }
}
```

When MongoShell has new properties that needs to be passed to the class, you can use `contructTextEditor()` again:

```js
constructTextEditor() {
  const shell = new MongoShell(this.shellConfig) // maybe pass a config through constructor
  shell.setConfig(this.shellConfig) // or maybe through a method. do as you like!
  return shell; // and then ofc return it
}
```

azmy60: I like to think that all codemirror codes are outside of the vue components. I didn't directly use CodeMirror codes or like core stuff inside the vue files. I mean at some point we want to upgrade to Vue 3? It just feels right to give some room to separate both? Just want to give you my thoughts.

azmy60:
@not-night-but Fixed the theme!

apps/ui-kit/lib/components/mongo-shell/MongoShell.vue
azmy60: no need to do this now honestly

azmy60 (approved):

PR #3297, opened 2025-06-18
FEAT: Persist privacy mode setting

pvlvld:
- Add privacyMode boolean setting to the database with migration
- Load privacy mode from the database on startup
- Save privacy mode changes to the database on toggle

Bonus 🤩:
- Fix: saving settings with a bool value when it's false


apps/studio/src/migration/20250618_add_privacy_mode_setting.js
```diff
@@ -0,0 +1,7 @@
+export default {
+  name: "20250618_add_privacy_mode_setting",
+  async run(runner) {
+    const query = `INSERT INTO user_setting(key, defaultValue, valueType) VALUES ('privacyMode', 'false', 5)`;
+    await runner.query(query);
+  }
+}
```
not-night-but, line 7: You have to import this migration into `apps/studio/src/migration/index.js` and add it to the `realMigrations` array for it to run.

pvlvld: Oopsie. Done!

apps/studio/src/store/modules/settings/SettingStoreModule.ts
```diff
@@ -74,9 +80,10 @@ const SettingStoreModule: Module<State, any> = {
         context.commit(M.ADD, setting);
```
not-night-but: Sorry, I just noticed this, but there is code specifically in the save function for privacyMode? And it doesn't appear that was you? And that code also seems to be exactly the same as what's in the else statement? And it doesn't seem to be used anywhere? Any way you could clean that up?

not-night-but (changes requested):
Okay two things: 
1. Not at all related to your code, but related to the feature. Tell me to screw off if you want lol, I've just now noticed the if/else statement on line 63 in SettingStoreModule and now it's bugging me lol
2. It seems there's a bug when setting the value to false? if I toggle privacy mode off and restart the app, suddenly I am in privacy mode again

pvlvld:
All done.

@not-night-but I'm curious.
Why destructuring
`async save(context, { key, value }) {...}`
And not just
`async save(context, payload: Pick<TransportUserSetting, 'key' | 'value'>) {...}`
this one doesn't create additional memory and always reflects key & value types from the TransportUserSetting type.

not-night-but (approved):
Looks good! Thanks again!

PR #3298, opened 2025-06-19
FIX: plugins manager styles

pvlvld:
Small fixes:
- Plugin description image no longer [overflows horizontally](https://ibb.co/album/n7FDvz).
- Plugin description image is no longer draggable, [preventing an error](https://youtu.be/8kjSbsXG0KA).
- Wider modal overall.

@azmy60


apps/studio/src/assets/styles/app/modals/plugin-manager-modal.scss
```diff
@@ -3,7 +3,7 @@
 
   .v--modal {
     height: 100% !important;
-    max-width: 800px;
+    max-width: 80%;
```
azmy60: Can we use fixed width instead?

It's a small thing but it will stretch the modal to 80% of the screen, and we don't want folks with BIG monitors to tilt their head left to right to read a sentence :) Moving your cursor from the center of the screen to the close button can be too far as well.

azmy60 (changes requested):
Thanks for the PR! I like the draggable image fix. Although there is a UX issue with the `max-width`

Also not sure if this is intentional or not?
image

Could you keep it stretched like before? The white space on the right just doesn't feel right.

I have an idea. Maybe the max-width can be 800px or lower, and then if you double click the plugin, the max-width can be larger?

apps/studio/src/assets/styles/app/modals/plugin-manager-modal.scss
pvlvld: Added new class for modal with a 1200px width that enabled only when plugin description is open

azmy60 (approved):
LGTM

PR #3301, opened 2025-06-20
Ai shell docs - WIP

rathboma:


azmy60 (changes requested):
The ini plugin looks broken in security page?

image

PR #3302, opened 2025-06-21
disable plugins from config files

azmy60:
This allows users to disable plugins from config files (not from UI. we're not able to write/replace config files in a way that preserve comments yet!)

image
image
image

### DIFF
--- a/apps/studio/src/background/lib/electron/ProtocolBuilder.ts
+++ b/apps/studio/src/background/lib/electron/ProtocolBuilder.ts
@@ -5,6 +5,7 @@ import * as fs from 'fs'
 import { URL } from 'url'
 import rawLog from '@bksLogger'
 import platformInfo from '@/common/platform_info'
+import bksConfig from "@/common/bksConfig";
 
 const log = rawLog.scope('ProtocolBuilder')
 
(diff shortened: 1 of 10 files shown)


apps/studio/src/background/lib/electron/ProtocolBuilder.ts
```diff
@@ -78,6 +79,11 @@ export const ProtocolBuilder = {
       const normalized = path.normalize(pathName)
       const fullPath = path.join(platformInfo.userDirectory, "plugins", normalized)
       log.debug("resolving", pathName, 'to', fullPath)
+      console.log("plugin", pathName, fullPath, {host:url.host})
```
rathboma: console log

apps/studio/src/background/lib/electron/ProtocolBuilder.ts
```diff
@@ -78,6 +79,11 @@ export const ProtocolBuilder = {
...
+      if (bksConfig.get(`plugins.${url.host}.enabled`) === false) {
```
rathboma: How do I identify the name of the plugin to set the disabled flag?

rathboma: Can the ID be put in the plugin manager somewhere?

rathboma (approved):
2 minor changes / clarifications -- then feel free to merge

apps/studio/src/background/lib/electron/ProtocolBuilder.ts
azmy60: You can look at the `id` in the `manifest.json` file. The id is also used for the folder name. So let's say we have ai plugin, and this is the manifest.json:

```json5
{
  "id": "bks-ai-shell",
  "author": "Beekeeper Studio",
   // other keys ...
}
```

That id is what we want to use inside the config files:

```ini
[plugins.bks-ai-shell]
enabled = false
```

The plugin is also located at the plugin id folder => `~/.config/beekeeper-studio/plugins/bks-ai-shell/`. Which is why `url.host` can also be used for identifying the plugin.

PR #3303, opened 2025-06-21
fix query selection for cm6

azmy60:


not-night-but:
@azmy60 Looks like there's a build issue

not-night-but:
@azmy60 Just ran this on my machine, it looks like selection is still broken? If you select a query in the editor, the button should change to `Run Selection` and it should only run that command. That functionality seems to have broken at some point.

not-night-but (approved):

PR #3306, opened 2025-06-23
broken e2e tests

rathboma:


rathboma:
@claude please fix the broken e2e tests. Run with yarn, see the studio-test.yml for stuff you need to start before running the tests.

rathboma:
@claude can you fix the workflow file then?

PR #3308, opened 2025-06-23
fix disabled plugins

azmy60:


apps/studio/src/background/lib/electron/ProtocolBuilder.ts
```diff
@@ -80,7 +80,7 @@ export const ProtocolBuilder = {
       const normalized = path.normalize(pathName)
       const fullPath = path.join(platformInfo.userDirectory, "plugins", normalized)
       log.debug("resolving", pathName, 'to', fullPath)
-      if (bksConfig.get(`plugins.${pluginId}.enabled`) === false) {
+      if (bksConfig.get(`plugins.${pluginId}.disabled`)) {
```
rathboma, line 83: lgtm. Would probably be good to have an api like `bksConfig.get('plugins', pluginId, 'disabled')` -- that takes any number of strings and joins them itself

apps/studio/src/components/plugins/PluginList.vue
```diff
@@ -10,7 +10,7 @@
       <div class="info">
         <div class="title">
           {{ plugin.name }}
-          <span class="badge" v-if="$bksConfig.plugins?.[plugin.id]?.enabled === false">disabled</span>
+          <span class="badge" v-if="$bksConfig.plugins?.[plugin.id]?.disabled">disabled</span>
```
rathboma, line 13: could one also use `bksConfig.get`? Or is that API not exposed?

rathboma (approved):
ship it

apps/studio/src/background/lib/electron/ProtocolBuilder.ts
azmy60: good idea!

apps/studio/src/components/plugins/PluginList.vue
azmy60: yes we can do that.

azmy60: Actually that's better, idk why I was not very convinced to use `bksConfig.get` on this one.

PR #3316, opened 2025-06-25
Organize Plugin System Errors

azmy60:
This PR aims to make error messages a bit clearer when developing plugins.

The errors are transferrable through the utility-process <-> renderer IPC.


rathboma:
@azmy60 finish it or kill it

apps/studio/src-commercial/entrypoints/utility.ts
```diff
@@ -70,6 +70,8 @@ interface Reply {
   type: 'reply' | 'error',
   data?: any,
   error?: string
+  errorName?: string
+  errorCode?: number
```
not-night-but: Were you intending for code to be an enum? Because in the error we define in `errors.ts` it's a string

apps/studio/src/components/plugins/PluginManagerModal.vue
```diff
@@ -220,6 +221,20 @@ export default Vue.extend({
       }
     },
     async openPluginPage({ id }) {
+      try {
+        const info = await this.$util.send("plugin/repository", { id });
+        this.selectedPluginReadme = info.readme;
+      } catch (e) {
+        log.error(e);
+
+        if (e instanceof PluginSystemError && e.code === "PLUGIN_NOT_FOUND") {
```
not-night-but, line 230: Yeah the type we set on the interface is a number, but here we're testing for a string name. Probably just makes sense to use an enum?

not-night-but (changes requested):
Some issues with types for the errors codes. Otherwise looks good. Might be nice to expand this pattern for something like UtilityError, DriverError, etc. Just to separate errors we mean to throw and want to treat differently

apps/studio/src-commercial/entrypoints/utility.ts
azmy60: Hmm, you're right.

not-night-but (approved):

Issue #3189, opened 2025-05-20
FEAT: Add "Reconnect" Option


JeanPaiva42:
**Description:**

**From a user:**

"Reconnect option is missing from the connection indicator/label in the bottom of the left sidebar. Only options are Disconnect & Edit Connection. I tend to work over ssh tunnels and they tend to error out when my laptop resumes from sleep. I have to disconnect and connect manually each time.""
 
**Describe the solution you'd like**

I'd like to see a "Reconnect" option in the connection status indicator at the bottom of the left sidebar. Ideally, this would allow one-click reconnection without needing to fully disconnect and reconnect.

**Describe alternatives you've considered**

"Currently, I manually disconnect and then reconnect every time a connection fails. It's time-consuming and not ideal for frequent interruptions."

Achrou:
+1

PR #3322, opened 2025-06-26
feat: add connection reconnect button

therealrinku:
fixes: #3189 

Added **Reconnect** button on connection menu.
_I was also thinking of reloading tables, databases etc. on reconnect but not exactly sure of that._

https://github.com/user-attachments/assets/41a37c94-f2bd-45d8-aed0-a3ed50c3cbdf

### DIFF
--- a/apps/studio/src/components/sidebar/core/ConnectionButton.vue
+++ b/apps/studio/src/components/sidebar/core/ConnectionButton.vue
@@ -26,6 +26,12 @@
         >
           <x-label><i class="material-icons">power_settings_new</i>Disconnect</x-label>
         </x-menuitem>
+        <x-menuitem
+          @click.prevent="reconnect()"
+          class="red"
+        >
+          <x-label><i class="material-icons">sync</i>Reconnect</x-label>
+        </x-menuitem>
         <x-menuitem @click.prevent="$modal.show('config-save-modal')">
           <x-label v-if="config.id">
             <i class="material-icons">edit</i>Edit Connection
@@ -248,6 +254,14 @@ export default {
         this.$modal.show('running-exports-modal')
       } else {
         this.$store.dispatch('disconnect')
+      }
+    },
+    async reconnect(){
+      try{
+        await this.$store.dispatch('reconnect')
+        this.$noty.success("Reconnected successfully.")
+      } catch(error){
+        this.$noty.error("Unable to reconnect.")
       }
     },
     async syncDatabase() {


rathboma:
@therealrinku Is there a way to detect a connection dropping and prompt the user to reconnect do you think?

not-night-but:
We did attempt a way to detect whether or not  we were still connected and to prompt the user to reconnect (that's why `LostConnectionModal.vue` and the `reconnect` store action still exist. I think we still have some commented out code in `BasicDatabaseClient.ts` that was meant to handle this, but we weren't totally sold on the method so we disabled it.

Previously the thought was to literally just run `SELECT 1` before every query, if that fails, then we aren't connected. It felt kinda hacky, and I think we were also just worried about constantly querying like that.

I'm not exactly sure how we want to approach that, as from what I can tell, most of our drivers don't have anything simple like a `checkConnection` function or anything, and I'm not sure how reliable parsing errors would be. It may have to be done on a client by client basis.

azmy60:
Just want to mention - you guys might have know this - also not saying this is a feedback to this PR, but there is a [ping](https://sidorares.github.io/node-mysql2/docs/documentation/mysql-server#events-1:~:text=ping()%20%2D%20Default%3A%20return%20OK) function in mysql2. Detected in their [changelog](https://github.com/sidorares/node-mysql2/blob/a0bb6bfe8c9a3d62f2e7df2af29c7266f44d7e9a/Changelog.md?plain=1#L1119-L1120) as well. I'm not sure about other databases though.

therealrinku:
hi all, thanks for the comments! 

one idea I had is to somehow catch connection lost error while performing db operations and prompt the user to reconnect or reconnect automatically but not exactly sure yet.

 I will check this out further.

rathboma:
Yeah I think it should work this way:

1. The app should routinely `ping` the server to check connection status (like every 10s)
2. When running any query, we should detect disconnects, and throw an error.
3. In both cases we should use a special `Error` class that we define ourselves, like `DisconnectedError`, then we can use that to prompt the user.
4. The app can detect when it wakes from sleep/lock/hibernate, and ping the server then

Thoughts?

therealrinku:
> Yeah I think it should work this way:
> 
> 1. The app should routinely `ping` the server to check connection status (like every 10s)
> 2. When running any query, we should detect disconnects, and throw an error.
> 3. In both cases we should use a special `Error` class that we define ourselves, like `DisconnectedError`, then we can use that to prompt the user.
> 4. The app can detect when it wakes from sleep/lock/hibernate, and ping the server then
> 
> Thoughts?

yeah that make sense! I will see If I can make it work 😁

azmy60:
Hey @therealrinku , just checking in - is this still being worked on?

therealrinku:
> Hey @therealrinku , just checking in - is this still being worked on?

hey @azmy60 👋🏻 
sorry, I'm bit busy with other things, not sure If I can continue on this 😅

azmy60:
@therealrinku No worries! I'm closing this for now since it's stale. Feel free to continue or make a new PR if you'd like.

PR #3328, opened 2025-06-27
fix: do not convert empty arrays to (NULL) for display

### LINKED ISSUE #2665: BUG: Empty JSONB Arrays display as null
**Describe the bug**
When querying a JSONB field and the value is an empty array, the displayed data is `(NULL)` in the results.

**To Reproduce**
1. Connect to a PostgreSQL database
2. Open a new query window
3. Run the query `select '[]'::jsonb`
4. See the result displayed as `(NULL)`

**Expected behavior**
An empty array should be displayed as `[]` in the results.

**Screenshots**
Screenshot 2024-12-18 at 1 22 12 PM

**Version Information:**
 - OS: Mac
 - App Version 5.0.8
 - PostgreSQL 15.8

wrheinheimer:
Fix #2665 
Before
beekeeper-not-showing-empty-array

After
beekeeper-show-empty-array

Empty arrays were previously converted to `(NULL)` for display, which was misleading.

### DIFF
--- a/apps/studio/src/mixins/data_mutators.ts
+++ b/apps/studio/src/mixins/data_mutators.ts
@@ -17,9 +17,6 @@ export function emptyResult(value: any) {
   if (_.isString(value) && _.isEmpty(value)) {
     return buildNullValue('EMPTY')
   }
-  if (_.isArray(value) && value.length === 0) {
-    return buildNullValue('NULL')
-  }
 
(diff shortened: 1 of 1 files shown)


not-night-but (approved):
Thanks for this! Sorry the issue kinda fell by the wayside.

PR #3338, opened 2025-07-01
add clipboard APIs for plugins, fix plugin cache, right clicking tab

azmy60:
Clicking `copy` in AI shell does not work due to browser restriction (see https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts). The easiest workaround is by adding clipboard APIs from electron.

Next, the plugin will run `clipboard.writeText` similar to the [Clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard) interface from browser:

```js
import { clipboard } from "@beekeeperstudio/plugin";

await clipboard.writeText("this is copied from plugin");
```

The `clipboard` object is available in `v1.3.0-beta` of `@beekeeperstudio/plugin`. (It's beta because the clipboard APIs don't exist in bks yet until this PR is merged.)

AI Shell uses the new clipboard API in the draft release => https://github.com/beekeeper-studio/bks-ai-shell/releases/tag/untagged-51aa39a229e527754bed . (we just need to click "mark as latest" to actually release it)

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -36,7 +36,7 @@
     "@aws-sdk/shared-ini-file-loader": "^3.374.0",
     "@azure/msal-node": "^2.12.0",
     "@babel/plugin-transform-class-static-block": "^7.26.0",
-    "@beekeeperstudio/plugin": "^1.2.2",
+    "@beekeeperstudio/plugin": "^1.3.0-beta",
     "@cleverbrush/async": "^1.1.10",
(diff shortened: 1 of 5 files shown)


not-night-but:
@azmy60 we should probably merge this into `rc-53`, no?

azmy60:
@not-night-but right, I forgot lol

not-night-but (approved):

PR #3344, opened 2025-07-02
CHORE: Add E2E tests for export and copy results

JeanPaiva42:
## Description:

* This PR introduces E2E tests to the export and copy results features

* Add E2E code infrastructure and actions to deal with this part of the application, which will allow for more footer related features to be tested.

## Observations:
* The tests cover the export and copy features, with some caveats. 
    - For exports, we are not able to have a deeper verification to confirm if the files were actually downloaded (Playwright/Electron limitations), but we are able to confirm that things on our end were called (Buttons/Menus were accessed successfully).

### DIFF
--- a/apps/studio/e2e/pageActions/index.ts
+++ b/apps/studio/e2e/pageActions/index.ts
@@ -3,12 +3,14 @@ import { userActions as queryTabActions } from './queryTabActions';
 import { userActions as resultPaneActions } from './resultPanelActions';
 import { userActions as tableSideBarActions } from './tableSideBarActions';
 import { userActions as footerActions } from './footerActions';
+import { userActions as toggleSideBarActions } from './toggleSideBarActions';
 export const userActions = (page) => {
     return {
         ...newDatabaseConnectionActions(page),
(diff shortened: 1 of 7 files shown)


apps/studio/e2e/pageActions/footerActions/clickOnCopyToClipboardOption.ts
```diff
@@ -0,0 +1,5 @@
+import { Footer } from "../../pageComponents/Footer";
+
+export const clickOnCopyToClipboardOption = async (footer: Footer, fileType: string): Promise<void> => {
```
azmy60, line 3: Sorry if I'm being too nitpicky lol. Are we planning to add more stuff on each footer action file? I'd rather put all footer actions code in the `index.ts` so we could immediately see the code without jumping around files :)

JeanPaiva42: Yes, I am planning to add checks that would use the `Structure` button, number of results, pagination of results and such... and in this case I would expand this action set

PR #3375, opened 2025-07-19
Trino db

wmontgomery:
trino


apps/studio/src-commercial/backend/handlers/connHandlers.ts
```diff
@@ -41,7 +41,7 @@ export interface IConnectionHandlers {
   'conn/listViews': ({ filter, sId }: { filter?: FilterOptions, sId: string }) => Promise<TableOrView[]>,
   'conn/listRoutines': ({ filter, sId }: { filter?: FilterOptions, sId: string }) => Promise<Routine[]>,
   'conn/listMaterializedViewColumns': ({ table, schema, sId }: { table: string, schema?: string, sId: string }) => Promise<TableColumn[]>,
-  'conn/listTableColumns': ({ table, schema, sId }: { table: string, schema?: string, sId: string }) => Promise<ExtendedTableColumn[]>,
+  'conn/listTableColumns': ({ table, schema, database, sId }: { table: string, schema?: string, database?: string, sId: string }) => Promise<ExtendedTableColumn[]>,
```
not-night-but: I'm not gonna highlight the whole file, but the database in this instance is the catalog, no? Which means that it's always available in the client as `this.db` so we don't need to pass it through to these functions this way. We can just change anywhere that you reference `catalog` or `filter.database` to `this.db`, as that should always be correct (we create a new connection when you switch databases (catalogs) so there shouldn't be any races or anything like that.

apps/studio/src/components/common/TableLength.vue
```diff
@@ -37,7 +37,7 @@ export default Vue.extend({
       this.fetchingTotalRecords = true
       try {
         this.error = null
-        this.totalRecords = await this.connection.getTableLength(this.table.name, this.table.schema);
+        this.totalRecords = await this.connection.getTableLength(this.table.name, this.table.schema, this.database);
```
not-night-but: If we move to using `this.db` in the client, changes like these are unnecessary.

apps/studio/src/shared/lib/dialects/trino.ts
```diff
@@ -0,0 +1,93 @@
+import _ from "lodash";
+import { ColumnType, defaultEscapeString, DialectData, friendlyNormalizedIdentifier } from "./models";
+
+// TODO: find what these should be!
```
not-night-but: Are these correct?

apps/studio/src/store/index.ts
```diff
@@ -543,14 +545,14 @@ const store = new Vuex.Store<State>({
 
         // FIXME (matthew): We're doing another loop here for no reason
         // so we're looping n*2 times
-        // Also this is a little duplicated from `updateTableColumns`, but it doesn't make sense
+        // Also this is a little duplicated from ``, but it doesn't make sense
```
not-night-but: ?

docs/user_guide/connecting/trino.md
```diff
@@ -0,0 +1,25 @@
...
+---
+
+# Trino Support
+
+Check back frequently as we are making continual updates to our Trino support.
+
+Super excited to have Trino support here for you all. Please bear in mind that Trino is only in the early stages of implementation, but we wanted to release it to y'all for testing.
+
+Support has been limited to basically a "read only" state since any form of writing should be done in the confines of the database itself.
+
```
not-night-but, line 15: Add some basic instructions on how to connect

package.json
```diff
@@ -18,6 +18,7 @@
     "sqltools:dev": "yarn workspace sqltools serve",
     "electron:build": "yarn bks:build",
     "electron:serve": "yarn bks:dev",
+    "trino": "docker compose up -d mysql psql trino",
```
not-night-but: Is this script totally necessary? All you really need to do is `docker compose up -d trino`

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+      return {
+        columns,
+        rows: this.rowsToObject(columns, rows),
+        arrayMode: false
+      }
+    } catch (err) {
+      log.error(err)
+      throw err
+    }
+  }
```
not-night-but: We should NOT be implementing this directly. Put this in `rawExecuteQuery` and it will be called when you call `this.driverExecuteSingle`

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+    database: string
+  ): Promise<TableResult> {
+    const columns = await this.listTableColumns(table, schema, database)
+    let selectFields = [...selects]
+    if (!selects || selects?.length === 0 || (selects?.length === 1 && selects[0] === '*')) {
+      // select all columns with the column names instead of *
+      selectFields = columns.map((v) => v.columnName)
+    }
+
+    console.log('here first')
```
not-night-but: nuke

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+import {
+  createCancelablePromise,
+  joinFilters
+} from "@/common/utils"
+import {
+  AlterTableSpec,
+  TableKey
+} from "@shared/lib/dialects/models"
+import { IdentifyResult } from "sql-query-identifier/lib/defines"
+import { uuidv4 } from "@/lib/uuid"
```
not-night-but: Unused import

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+interface ResultColumn {
+  name: string
+  type: string
+}
+
+interface TrinoResult extends BaseQueryResult {
+  queryId?: string
+}
+
+type Result = TrinoResult
```
not-night-but: Is there a reason for us to alias this type here?

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+    table: string,
+    offset: number,
+    limit: number,
+    orderBy: OrderBy[],
+    filters: string | TableFilter[],
+    schema: string,
+    selects: string[],
+    database: string
+  ): Promise<TableResult> {
+    const columns = await this.listTableColumns(table, schema, database)
```
not-night-but:
```suggestion
    const columns = await this.listTableColumns(table, schema)
```

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+  }
+
+  async listSchemas(filter: SchemaFilterOptions): Promise<string[]> {
+    log.info('filters in listSchemas', filter)
+    if (filter.database == null) return []
+    const sql = `show schemas from ${filter.database}`
+    const result = await this.driverExecuteSingle(sql)
+
+    return result?.rows ? result.rows.map((row) => row.Schema) : []
+  }
```
not-night-but:
```suggestion
  async listSchemas(filter: SchemaFilterOptions): Promise<string[]> {
    log.info('filters in listSchemas', filter)
    const sql = `show schemas from ${this.db}`
    const result = await this.driverExecuteSingle(sql)

    return result?.rows ? result.rows.map((row) => row.Schema) : []
  }
```

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+    if (!filter || filter.database == null) return []
+    const sql = `select * from ${filter.database}.information_schema.tables`
+    const result = await this.driverExecuteSingle(sql)
+
+    return result.rows.map((row) => ({
+      schema: row['table_schema'],
+      name: row['table_name'],
+      entityType: 'table' as const
+    }))
+  }
```
not-night-but:
```suggestion
   async listTables(filter?: FilterOptions): Promise<TableOrView[]> {
    log.info('filters in listTables', filter)
    const sql = `select * from ${this.db}.information_schema.tables`
    const result = await this.driverExecuteSingle(sql)

    return result.rows.map((row) => ({
      schema: row['table_schema'],
      name: row['table_name'],
      entityType: 'table' as const
    }))
  }
```

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+
+  async listTableColumns(table: string, schema: string, catalog: string): Promise<ExtendedTableColumn[]> {
+    const sql = `
+      SELECT
+        *
+      FROM ${catalog}.information_schema.columns
+      WHERE table_schema = '${schema}'
+        AND table_name = '${table}'
+      ORDER BY ordinal_position
+    `
```
not-night-but:
```suggestion
  async listTableColumns(table: string, schema: string): Promise<ExtendedTableColumn[]> {
    const sql = `
      SELECT
        *
      FROM ${this.db}.information_schema.columns
      WHERE table_schema = '${schema}'
        AND table_name = '${table}'
      ORDER BY ordinal_position
    `
```

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+
+  async duplicateTableSql(): Promise<string> {
+    log.info("Trino doesn't support changing data")
+    return null
+  }
+
+  async setElementNameSql(): Promise<string> {
+    log.info("Trino doesn't support changing data")
+    return null
+  }
```
not-night-but: I'd say these logs probably aren't necessary for info mode? Maybe `debug` or `verbose` or something

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+          const queryResult = await this.driverExecuteSingle(queryText) 
+          
+          return queryResult
+        } catch (err) {
+          throw new Error(err)
+        }
+      })
+    )
+    return results
+  }
```
not-night-but: Again, should definitely not be implementing this. This should be contained within rawExecuteQuery.

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+        rows: data,
+        rowCount: data.length,
+        queryId: ''
+      })
+    }
+    return ret
+  }
+
+  async rawExecuteQuery(_query: string): Promise<Result[]> {
+    // TODO: Still need this?
```
not-night-but: Yes. This is how all database clients should be running queries. Implement all of the logic here, and then use the 'wrapper' versions (`driverExecuteSingle` and `driverExecuteMultiple`) to call this.

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+    const result = await this.driverExecuteSingle(
+      `SELECT count(*) as count FROM ${this.wrapIdentifier(database)}.${this.wrapIdentifier(schema)}.${this.wrapIdentifier(table)}`
+    )
+
+    const [row] = result.rows as { count: number }[]
+    return row.count
+  }
+
+  // TODO: ALL THE STREAMING!
+  // No exports right now so.... is this needed?
```
not-night-but: This doesn't seem to be fully implemented yet? I don't see a cursor being defined, `TrinoResult` doesn't have a data field, etc. If we aren't planning on supporting exports, I'd rather we just disable this method (return an empty object or something) than having non working code here.

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+    const result = await this.driverExecuteSingle(qs.countQuery)
+    const json = result.data as ResponseJSON<{ total: number }>
+    const totalRows = Number(json.data[0].total) || 0
+    const columns = await this.listTableColumns(table, schema, database)
+    return { totalRows, columns, cursor }
+  }
+
+  queryStream(_query: string, _chunkSize: number): Promise<StreamResults> {
+    throw new Error("Method not implemented.")
+  }
```
not-night-but, line 547: If we aren't implementing query streaming, then we should probably disable the feature in the frontend as well. Currently I think it's just disabled for mongo, but we should probably just add a feature in dialectData that can be disabled.

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+
+      fullFilterString = "WHERE " + joinFilters(filtersWithoutParams, filters)
+    }
+
+    return {
+      fullFilterString,
+    }
+  }
+
+  static buildSelectTopQuery(
```
not-night-but: Is there really a reason for this to be static? It's only used within this class, so really it could just be a private function, no?

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+      return value.toString()
+    }
+    if (typeof value === 'boolean') {
+      return value ? 'TRUE' : 'FALSE'
+    }
+    return `'${value.toString().replace(/'/g, "''")}'`
+  }
+
+
+  static buildFilterString(filters: TableFilter[], columns = []) {
```
not-night-but: I don't see the need for this to be static either.

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,770 @@
...
+        SELECT 
+          ${wrappedSelects},
+          ROW_NUMBER() OVER (${rowNumberOrderClause}) AS rownum
+        FROM ${databaseRef}.${tableRef}
+        ${filter}
+      )
+      SELECT *
+      FROM ranked
+      ${usePagination ? `WHERE rownum > ${safeOffset} AND rownum <= ${safeOffset + safeLimit}` : ""}
+    `
```
not-night-but: I'd rather this be a private function of the class, rather than an inlined one

not-night-but (changes requested):
Okay so this is looking good! I did have a couple notes:
- The trino docs page needs to be added to the nav in the `mkdocs.yml` file, otherwise it won't actually be accessible.
- Could you add a dev migration to add a default connection for the trino docker container? 
- With these changes, we're now passing the database along to a bunch of the handlers, but that information should already be available on the client itself? It is passed on connect, and we create a new connection when you change databases. So `this.db` (which is just a string for the name of whatever db, or catalog in this case, we are connected to) should be able to be used in all those cases, no?
- We also should NOT be overriding the `driverExecuteSingle` and `driverExecuteMultiple` methods. These are meant to be defined in the base class and nowhere else. Instead you should implement `rawExecuteQuery` which is called by both of those functions. You should be provided with a `multiple` boolean on the options object so you can then return the appropriate format (either an array or just a single query result object).

apps/studio/src/store/index.ts
wmontgomery: Oh I think I did a "removal all " for "updateTableColumns" or something and it hit the comment too. This is probably something from the "clickhouse" one I used as my base so might not mean anything

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
wmontgomery: I don't think exports are going to be a thing for this. I don't know if people use Trino for that to export tables, moreso run queries. I was leaving it there as a "do we want to"?

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
wmontgomery: Ah I must have missed that in the remove bits. Thought I had.

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
wmontgomery: You're right. I will go with this being a typescript thing and we all know my feelings on typescript :D

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
wmontgomery: No problemo. Normally I would have because it's cached in memory when already its own instead of being setup when the parent function is called, but guess I just went half brain on it

package.json
wmontgomery: Trino needs underlying databases to connect to I believe. I can try it without and see what happens, but my understanding is Trino just acts as a middleman to talk to already running DBs since you have to add connection details into trino connect files.

apps/studio/src-commercial/backend/handlers/connHandlers.ts
wmontgomery: Ah ok. Hadn't thought of that. THanks for the heads up! I'll make the change

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
wmontgomery: TIL those 2 functions shouldn't be overwritten. Groovy, I'll change rawExecute instead!

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
wmontgomery: loloops. Thought I axed all the console.logs I left

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
wmontgomery: It was a pattern I saw in the clickhouse file and emulated. Probably only reason was to not go through and have to change all Types of Result to TrinoResult.

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
wmontgomery: TIL

apps/studio/src/shared/lib/dialects/trino.ts
wmontgomery: OK, Grabbed the list out of the docs now. Not sure if this is totally what all can be covered because it supports so many different engines, but a good list to start

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
```diff
@@ -0,0 +1,705 @@
...
+          } satisfies NgQueryResult
+        })
+    )
+    return results
+  }
+
+  async rawExecuteQuery(sql: string): Promise<TrinoResult> {
+    try {
+      // The trino query parser doesn't particularly like semicolons. Who can blame it?
+      const result: AsyncIterableIterator<QueryResult> = await this.client.query(sql.trim().replace(/;$/, ''))
```
not-night-but, line 408: Isn't this dangerous? What if a value we're filtering by includes a semicolon?

apps/studio/src/lib/db/clients/index.ts
```diff
@@ -208,6 +209,17 @@ export const CLIENTS: ClientConfig[] = [
...
   },
+  {
+    key: 'trino',
+    name: 'Trino',
+    topLevelEntity: 'Catalog',
+    defaultPort: 8080,
+    disabledFeatures: [
+      'server:ssl',
+      'server:socketPath',
+      'cancelQuery', // TODO how to do this?
```
not-night-but, line 220: This is definitely something we need to figure out at some point. I know we have a few clients that don't totally support cancelling queries.

apps/studio/src/lib/db/types.ts
```diff
@@ -189,7 +190,7 @@ export interface IBasicDatabaseClient {
   listViews(filter?: FilterOptions): Promise<TableOrView[]>,
   listRoutines(filter?: FilterOptions): Promise<Routine[]>,
   listMaterializedViewColumns(table: string, schema?: string): Promise<TableColumn[]>,
-  listTableColumns(table: string, schema?: string): Promise<ExtendedTableColumn[]>,
+  listTableColumns(table: string, schema?: string, database?: string): Promise<ExtendedTableColumn[]>,
```
not-night-but: I don't think we're taking these in anymore, since we moved to `this.db`

not-night-but (approved):
Just some comments. At the very least, remove those parameters that we no longer need from the handlers, and if you don't think removing semicolons is an issue then merge away

apps/studio/src/lib/db/types.ts
wmontgomery: drat! Knew I was going to miss one

apps/studio/src-commercial/backend/lib/db/clients/trino.ts
wmontgomery: This will only remove the last semicolon from a string. Any other ones in there will remain

apps/studio/src/lib/db/clients/index.ts
wmontgomery: agreed. I might be able to get it when coming back at some point to get the cursors running. Trino returns a queryId as part of the response, so can potentially use that. I'll have to do a lot more reading on that.

PR #3425, opened 2025-08-05
Fix clickhouse query export

not-night-but:
Implemented missing function in the clickhouse client to allow for exporting full result sets.


azmy60 (commented):
Do we need to test this?

rathboma:
@claude add a few tests to make sure streaming is working for ClickHouse

PR #3426, opened 2025-08-05
Fix outdated documentation links in UI

zimm3rmann:
Updated 6 links that pointed to a old documentation URLs.


rathboma (approved):
Thank you!

PR #3427, opened 2025-08-06
feat: add numeric and date datatype check to right align non key columns

### LINKED ISSUE #424: FEAT: Right-justify columns containing number data types
Query results in Beekeeper show number columns as left-justified. This makes comparing values harder; it's also different from psql behavior.

Beekeeper:

Screen Shot 2020-10-30 at 9 36 17 AM

PSQL:

Screen Shot 2020-10-30 at 9 40 25 AM

Request changing the default display for those types.

wmontgomery:
resolves #424 
424_right-align

### DIFF
--- a/apps/studio/src/common/utils.ts
+++ b/apps/studio/src/common/utils.ts
@@ -322,3 +322,48 @@ export function toRegexSafe(input: string) {
     return null;
   }
 }
+
+export function normalizeDataType (dataType) {
+  return dataType.toLowerCase()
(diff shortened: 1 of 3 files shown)


wmontgomery:
> This looks great!
> 
> At some point we need to start providing this type information from the drivers I think

I was thinking about that to have each driver start sending back "is this column a right aligning one" but kind of testing the waters here before going wild with it. Hopefully one function can handle it for most cases, so complicate it when the world tells us it has to be :D

Issue #3418, opened 2025-08-04
BUG: Unable To paste multiple line in the query editoe


Rajesh9998:
**Describe the bug**
I was unable to paste from my clipboard into the query editor, after i copied  some sql query using ctrl+c  into my clipboard and tried to paste in the query editor in beekeeper studio, It is pasting nothing, this was the case if the text/query was more than 1 line.

**To Reproduce**
1. Go to 'any database'
2. Click on '+ to open a query editor tab'
3. and paste a multi lined  sql query, precisely  More than 1 line 
4. See nothing is pasted, if it is a single line it works.
5. I see this in the Latest version 5.3.3

**Expected behavior**
Able paste multi line sql quey.

**Version Information (please complete the following information):**
 - OS:  Windows 11
 - App Version : 5.3.3
 - Database type and version : postgres 17

not-night-but:
@Rajesh9998 Hey! I'm not able to reproduce this. Are you seeing any errors in the console?

SamBrishes:
Hello,

I am experiencing the same issue. When pasting queries into the console... Single-line queries paste correctly using Ctrl+V but Multi-line queries do not paste using Ctrl+V, but do work when using Right-click > Paste.

The following error occurred on the console

```
3index-Cs1poal4.js:3461 RangeError: Selection points outside of document
    at checkSelection (index-Cs1poal4.js:3458:14482)
    at new Transaction (index-Cs1poal4.js:3458:22735)
    at Transaction.create (index-Cs1poal4.js:3458:22908)
    at resolveTransaction (index-Cs1poal4.js:3458:25334)
    at EditorState.update (index-Cs1poal4.js:3458:27971)
    at EditorView.dispatch (index-Cs1poal4.js:3464:6739)
    at Object.paste (index-Cs1poal4.js:3911:440)
    at index-Cs1poal4.js:3463:7118
    at InputState.runHandlers (index-Cs1poal4.js:3463:5032)
    at InputState.handleEvent (index-Cs1poal4.js:3463:4857)
```

I'm using Windows 11 and Beekeeper Studio v5.3.4.

~Sam.

azmy60:
I'm able to replicate this on windows. Thanks for the report! It's being worked on now.

basonjui:
Yes, I was going to open a ticket for this. I am also having this issue on Windows 11, Beekeeper Studio v5.3.4.

[...]

PR #3429, opened 2025-08-06
use clipboard filter instead of replacing the dom event

azmy60:
Before this, we replaced the built-in behavior of "pasting"  by adding our own paste handler so we could use the `removeQueryQuotes` which isn't ideal. Luckily, there is a facet that allows us to modify the clipboard text before it gets applied in the editor using [clipboardInputFilter](https://codemirror.net/docs/ref/#view.EditorView^clipboardInputFilter) (see the actual implementation [here](https://github.com/codemirror/view/blob/3cea8dba19845fe75bea4eae756c6103694f49f3/src/input.ts#L676)).

What do we fix in this PR?

1. https://github.com/beekeeper-studio/beekeeper-studio/issues/3418
2. Pasting on multiple cursors. To replicate:
  i. Create new cursors by pressing `Ctrl+Click` (or `Ctrl+Click+Drag` to make selection)
  ii. Press `Ctrl+V`
  iii. The pasted value is applied on one cursor. It should be applied to all cursors.
3. Possibly other bugs that we haven't caught!

### DIFF
--- a/.github/workflows/studio-test.yml
+++ b/.github/workflows/studio-test.yml
@@ -66,7 +66,7 @@ jobs:
         run: yarn workspace beekeeper-studio run test:ci --ci --silent
 
       - name: Unit Tests (UI Kit)
-        run: yarn workspace @beekeeperstudio/ui-kit run test --ci --silent
+        run: yarn workspace @beekeeperstudio/ui-kit run test --silent
 
 
   integration:


azmy60:
@rathboma I use vitest cause that's what we have in the beginning of UI Kit. The setup is super simple too. But let me know if we want to change that.

PR #3434, opened 2025-08-09
feat: create in statement from selected data

### LINKED ISSUE #3390: FEAT: Add 'Copy as...' option to improve workflow efficiency
I write queries which can give results up to 500 (or more, or less). For those 500 results I would need all the id's or tokens to write an update query. With the current way, I copy the whole column and ask chat to seperate them with a comma or add strings to the tokens. It's not a big task to do this but if you have to do this multiple times, it begins to annoy me..

I think it would be amazing to have a copy option where it is comma seperated and considering the column type, with added parentheses so you can easily copy the list in your update .. IN statement 

I don't have an alternative right now..

**Additional context**

Image

Postgresq

wmontgomery:
resolve #3390


Stef-Verniers:
Thank you for this!! 🖤💛

rathboma (approved):

PR #3441, opened 2025-08-12
rewrite query for getTableKeys

not-night-but:
We were using a lot of information schema views for this query, at large scales it seems it was terrible for performance (reported to take around ~12 seconds to actually run).


rathboma (changes requested):
Couple of questions, one minor issue:

Question: Will this break postgres-subclients? Redshift, Cockroach, etc?

Observation from AI:


    The mapping of key columns (from and to columns) in the new query uses ANY (c.confkey), which matches all referenced columns. In multi-column foreign keys, you need to ensure the order and pairing matches the old query’s behavior.

    The old query uses ordinal_position to map columns, which is less error-prone for multi-column FKs.

not-night-but:
@rathboma Updated the query. In terms of postgres subclients, redshift overrides this function so that isn't a concern, and cockroach seems to still function fine? The tests are passing at the very least.

Issue #1675, opened 2023-08-21
FEAT: Scoping of run-history by connection


rathboma:
By default the run history should only show queries from the current connection, but have the option to change it to all connections.

![image](https://github.com/beekeeper-studio/beekeeper-studio/assets/279769/398c8433-ca67-4d3e-bcd0-8596d0e2c6d2)

PR #3442, opened 2025-08-13
feat: scoping of history by connection

therealrinku:
resolves: #1675 

https://github.com/user-attachments/assets/b80ca968-09e8-4acd-b63b-12384c11a4ec


azmy60:
@therealrinku Code looks good to me. The UI needs a little tweak though. The icon size and the font size seem to be disproportionate with the rest of the elements below it. I would reduce the size so it visually looks the same. Also, maybe the font can be a little transparent so it's not too distracting?

not-night-but (dismissed):
I agree with Azmy that the UI could use some tweaking. I think this is probably a good first pass at improving the used query system, but I think I'ma write up another issue to possibly improve it even more. Thanks as always for the contribution

therealrinku:
@azmy60 @not-night-but  I have updated the styling a bit. let me know if this looks good?
Screenshot 2025-08-17 at 17 55 11

azmy60:
That looks better to me, thanks! 

Actually, I feel like there is another small one :). Instead of "Show All History", it might as well can be:
- "Filter by this connection"
- "Filter by connection"
- "Scope by connection"
- "Include other connections"

It just makes it a little clearer to me that by not showing all, we only show a partial and based on **what**? Based on the current connection! What do you all think (and @rathboma, since you made the issue)?

Honestly, Show All History is still good to me. You can ignore this.

azmy60 (approved):

PR #3448, opened 2025-08-15
Feat/plugin minappversion

azmy60:
## What is `minAppVersion`

`minAppVersion` is a property in manifest.json that defines which version of Beekeeper Studio that a plugin support. 

## What are the conditions

1. If app version < minAppVersion, the plugin is NOT compatible.
2. If minAppVersion === undefined, it's always compatible.
3. If compatible, the plugin can be installed, all good.
4. If not compatible, user can't install it
    - After you click install, it should show an error.
    - Error message = "Plugin requires version ... Please update bks or click here for how to install plugin manually"
5. If not compatible and already installed, the plugin should not load
    - Happens when you downgrade bks
    - Error message = same

### Other changes (not all but maybe worth mentioning)

- Added tests and many test utils for plugins (including edge cases for `minAppVersion`)
- Refactored plugin manager and other plugin classes so they are testable, and applied fixes after tests
- Added `pluginManagerStatus` state in vuex for easy tracking in the UI
- Custom error classes for plugin - Not the best way. There's a better way to handle errors (see #3468)

NOTE: We still fetch the latest release of plugins, so no changes on that.

<details>
<summary>TODOs [done]</summary>

- [x] Allow installing specific plugin versions (not necessarily from UI)
- [x] Check minAppVersion in manifest.json
- [x] If minAppVersion is missing, the plugin is compatible to all version
- [x] Block install/load if app version < minAppVersion
- [x] Show error in Plugin Manager UI when blocked
- [x] Use semantic version comparison
- [x] Add tests
- [x] Docs: how to install a plugin version manually and disable auto-update
</details>

### DIFF
--- a/apps/studio/src/migration/20250529_add_plugin_settings.js
+++ b/apps/studio/src/migration/20250529_add_plugin_settings.js
@@ -2,8 +2,8 @@ export default {
   name: "20250529_add_plugin_settings",
   async run(runner) {
     await runner.query(`
-      INSERT INTO user_setting(key, defaultValue, valueType, version, userValue, linuxDefault, macDefault, windowsDefault)
-        VALUES ('pluginSettings', '{}', 3, 0, '{}', '{}', '{}', '{}');
+      INSERT INTO user_setting(key, defaultValue, valueType)
+        VALUES ('pluginSettings', '{}', 3);
     `);
   },
 };


apps/studio/src/migration/20250529_add_plugin_settings.js
```diff
@@ -2,8 +2,8 @@ export default {
   name: "20250529_add_plugin_settings",
   async run(runner) {
     await runner.query(`
-      INSERT INTO user_setting(key, defaultValue, valueType)
-        VALUES ('pluginSettings', '{}', 3);
+      INSERT INTO user_setting(key, defaultValue, valueType, version, userValue, linuxDefault, macDefault, windowsDefault)
+        VALUES ('pluginSettings', '{}', 3, 0, '{}', '{}', '{}', '{}');
```
not-night-but: Do we need to add another migration for users that may already have this user setting in their local db?

azmy60: I'll remove it. This isn't needed.

PR #3451, opened 2025-08-16
fix: autocomplete case sensitive columns

### LINKED ISSUE #3432: BUG: Say something here
**Describe the bug**
When building a query against a Postgres database the editor autocomplete no longer double quotes case sensitive column names like it does with table names.

**To Reproduce**
1. Open up new query start and start building a query like `select * from "Users" u where u."firstName" = 'Bob'` using autocomplete to complete the the table and column names.

**Expected behavior**
That the column name is double quoted (eg: `"firstName"`).

**Screenshots / Gifs**
N/A

 - OS: Mac
 - App Version 5.3.4
 - Database Postgres 16

**Additional context**
I'm pretty sure back a few releases this worked before!

Redfox:
resolves #3432 

[beekeeper.webm](https://github.com/user-attachments/assets/92b94854-3c37-4e10-aa2c-64fbe887728c)

### DIFF
--- a/apps/ui-kit/lib/components/sql-text-editor/utils.ts
+++ b/apps/ui-kit/lib/components/sql-text-editor/utils.ts
@@ -5,15 +5,39 @@ import getAliases from "./getAliases";
 export { getAliases };
 
 /**
+ * Check if a column name needs quoting
+ * @param column - The column name to check
+ * @returns True if the column name needs quoting, false otherwise
+ */
+function needsQuoting(column: string): boolean {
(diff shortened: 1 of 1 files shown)


Redfox:
Hi @azmy60, that makes sense, I am still trying to understand the codebase and dependencies
I'll try to fix it that way

recck:
@Redfox Were you able to update it based on @azmy60 's suggestion? It looks like this would resolve https://github.com/beekeeper-studio/beekeeper-studio/issues/3525

PR #3464, opened 2025-08-25
limit query result rows for plugins

### LINKED ISSUE #3454: BUG: query results should be limited to 10K if requested by plugins

azmy60:
closes #3454

### DIFF
--- a/apps/studio/default.config.ini
+++ b/apps/studio/default.config.ini
@@ -183,5 +183,9 @@ openEditorModal = shift+enter
 firstPage = ctrlOrCmd+h
 lastPage = ctrlOrCmd+l
 
+[plugins.config]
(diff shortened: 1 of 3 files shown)


azmy60:
@rathboma Not sure if this is correct, or maybe we want to limit the query result from the driver? like putting `LIMIT 10000` at the end of query?

not-night-but:
@azmy60 What is the reason for limiting query results? Do we want to do it to save memory? If that's the case it should be done within the query.

rathboma:
Did someone bring this up as a thing they need? I'd rather not do this carte-blanch. Maybe someone wants the AI to fetch all 20,000 rows? Is the issue that it blows up the LLM context?

PR #3478, opened 2025-08-29
Add support for bedrockdb

chuckdries:
[Bedrock](https://bedrockdb.com/) is Expensify's data platform. It wraps a SQLite database and implements MySQL's communication protocol. This PR adds a client `bedrock.ts` that extends the MySQL client, setting the dialect to `sqlite` and making a few other changes to get things working. Overall I'm pleased with how quick and easy it was to implement. Y'all have got a solid codebase here!

Apologies for not opening an issue about this first. I built this for myself to use, and I wanted to make sure it was feasible before asking about it, but then I prototyped it and it mostly Just Worked and like 2 hours later here we are!

Also, I had to upgrade electron-builder to v26 to build a working release bundle on MacOS. I was encountering https://github.com/electron-userland/electron-builder/issues/8973. Not sure what that's about but I'm happy to remove that and the related changes from the branch upon request.


apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
```diff
@@ -65,7 +65,7 @@ export interface BaseQueryResult {
 export abstract class BasicDatabaseClient<RawResultType extends BaseQueryResult> implements IBasicDatabaseClient {
   knex: Knex | null;
   contextProvider: AppContextProvider;
-  dialect: "mssql" | "sqlite" | "mysql" | "oracle" | "psql" | "bigquery" | "generic";
+  dialect: "mssql" | "sqlite" | "mysql" | "oracle" | "psql" | "bigquery" | "generic" | "bedrock";
```
not-night-but, line 68: This is actually meant to be the `sql-query-identifier` dialect. So unless identifier supports it, we shouldn't allow that type. I also think you're setting this to sqlite anyways.

apps/studio/src/lib/db/clients/bedrock.ts
```diff
@@ -0,0 +1,351 @@
...
+const SD = SqliteData;
+
+export class BedrockClient extends MysqlClient {
+  dialectData = SqliteData;
+
+  constructor(server: IDbConnectionServer, database: IDbConnectionDatabase) {
+    super(server, database);
+    // Override to use SQLite dialect for SQL syntax
+    this.dialect = "sqlite";
+  }
```
not-night-but, line 40: You may have to override the knex driver in the base class as well here. Just the same way you do the dialect. Idk exactly how this database functions, but if the query language is sqlite, then you'll definitely need to replace the knex driver (that provides us with an easy way to generate sql queries programmatically)

rathboma:
@chuckdries Thanks so much for implementing a new database!

Is it fully working for you?

Looks like some conflicts for us to resolve, and we'll have to add an integration test suite, but this is awesome.

chuckdries:
@rathboma It works well enough to be useful to me. I have noticed it has bugs, but I've been too busy to solve them lately 😅

How should we proceed? Do y'all wanna solve the conflicts then I'll try to work through the bugs in my implementation maybe this weekend?

not-night-but:
@chuckdries If you're up for continuing to work on this, I'll solve the conflicts for you and then shoot a review your way.

chuckdries:
@not-night-but sounds good to me!

apps/studio/src/lib/db/clients/bedrock.ts
```diff
@@ -0,0 +1,351 @@
...
+  TableChanges,
+} from "../models";
+import {
+  IndexColumn,
+  TableKey as SharedTableKey,
+} from "@shared/lib/dialects/models";
+import _ from "lodash";
+import { parseVersion } from "@/common/version";
+import { buildSelectTopQuery } from "./utils";
+import logger from "@/lib/log/mainLogger";
```
not-night-but, line 29: The workflow will probably tell you this, but we don't import this directly, instead you should do this
```suggestion
import logger from "@bksLogger";

const log = logger.scope('bedrock.ts');
```
That will resolve to the proper logger for the file.
And creating the log object just makes it simpler to use, eg
```ts
log.info('Hello')
```

apps/studio/src/lib/db/clients/bedrock.ts
```diff
@@ -0,0 +1,351 @@
...
+    // Use SQLite change builder for DDL operations
+    return new SqliteChangeBuilder(table);
+  }
+
+  async getVersion() {
+    const { rows } = await this.driverExecuteSingle(
+      "SELECT sqlite_version() as version"
+    );
+    const version = rows[0]["version"];
+    console.log("version", version);
```
not-night-but, line 52: If we want any logs to remain for debugging purposes, we should use the logger, otherwise remove any console.logs used for debugging

apps/studio/src/lib/db/clients/bedrock.ts
```diff
@@ -0,0 +1,351 @@
...
+import {
+  IndexColumn,
+  TableKey as SharedTableKey,
+} from "@shared/lib/dialects/models";
+import _ from "lodash";
+import { parseVersion } from "@/common/version";
+import { buildSelectTopQuery } from "./utils";
+import logger from "@/lib/log/mainLogger";
+
+const SD = SqliteData;
```
not-night-but, line 31: This is up to you, depends on how much Bedrock differs from sqlite feature wise, but it might be worth having dialect data specifically for bedrock.

apps/studio/src/lib/db/clients/bedrock.ts
```diff
@@ -0,0 +1,351 @@
...
+        throw ex;
+      }
+    });
+
+    return results;
+  }
+
+  private dataToColumns(data: any[], tableName: string): ExtendedTableColumn[] {
+    return data.map((row) => {
+      const defaultValue = row.dflt_value === "NULL" ? null : row.dflt_value;
```
not-night-but, line 238: If possible, it would be nice if the data here had a type.

apps/studio/src/lib/db/clients/bedrock.ts
```diff
@@ -0,0 +1,351 @@
...
+
+  // We also need to override parseTableColumn - let's use a simplified version
+  parseTableColumn(row: any): any {
+    return {
+      name: row.name,
+      type: row.type,
+      nullable: Number(row.notnull || 0) === 0,
+      defaultValue: row.dflt_value === "NULL" ? null : row.dflt_value,
+    };
+  }
```
not-night-but, line 350: This should just return a BksField. That is used for our transcoders system, the simplest form would just be
```suggestion
  // We also need to override parseTableColumn - let's use a simplified version
  parseTableColumn(row: any): any {
    return {
      name: row.name,
      bksType: 'UNKNOWN'
    };
  }
```

The other information you put here is actually on the ExtendedTableColumn struct, not this one

not-night-but (changes requested):
This looks good so far! I haven't tried to run/test it out yet, so this is purely a cursory glance over the code quality. Once I get a chance to test it out properly I will be back with more feedback

rathboma:
@claude take a look through this PR. It's labelled as `draft` which makes me think it's not ready to be merged. How much work is remaining? Can you make a clear list of remaining work.

rathboma:
Closing in favor of #4101 - I took the code from this PR and added a passing integration test and some UX tweaks. Thanks so much for the contribution!

PR #3489, opened 2025-09-02
Support dynamic menu bar

azmy60:
This PR aims to make the **client** menu bar to be a little bit extensible. After this, we should be able to add menu items from plugins. This is an example to add/remove a custom item to the "Tools" menu:

```js
export default {
  methods: {
    ...mapActions('menu', ['add', 'remove']),
  },
  mounted() {
    this.add({
      id: 'test-item', parentId: 'tools',
      label: 'Open a new query',
      action: {
        event: AppEvent.newTab,
        args: "SELECT * FROM test" // optional, depending on the event
      },
      enableWhenConnected: true, // optional
    })

    await new Promise(resolve => setTimeout(resolve, 3000))

    this.remove('test-item') // use the id
  },
}
```

NOTE: Make sure that the `AppEvent` is forwarded in AppEventHandler.js: https://github.com/beekeeper-studio/beekeeper-studio/blob/0a27f1ba56e5a6c58df63036949751aea1a00f73/apps/studio/src/lib/events/AppEventHandler.js#L19-L32

### DIFF
--- a/apps/studio/src/background/NativeMenuActionHandlers.ts
+++ b/apps/studio/src/background/NativeMenuActionHandlers.ts
@@ -9,6 +9,7 @@ import { IMenuActionHandler } from '@/common/interfaces/IMenuActionHandler'
 import { autoUpdater } from "electron-updater"
 import { DevLicenseState } from '@/lib/license';
 import { setAllowBeta } from './update_manager'
+import { CustomMenuAction } from '@/types'
 
 type ElectronWindow = Electron.BrowserWindow | undefined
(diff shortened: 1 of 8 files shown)


not-night-but (changes requested):
This looks good for the most part, but it does remove the reactivity to the settings. Which means the current theme is no longer shown:

Image
(Dark should have a checkmark next to it)

azmy60:
@not-night-but Good catch! I just pushed the fix.

PR #3506, opened 2025-09-10
add getTableKeys API

azmy60:


not-night-but:
@azmy60 Do we no longer want the ability to get the active tab in plugins?

azmy60:
@not-night-but No. It was not planned anyway.

azmy60:
It was used when AI Shell was a sidebar plugin. But then we drop the sidebar support so `getActiveTab` is kinda useless for now and removing it keep things a little clean. Maybe someday we can resurrect it.

EDIT: also `getAllTabs`. Maybe for 1.0, database interaction is more important than UI interaction.

not-night-but (approved):

not-night-but:
@azmy60 merge conflicts then go for it

PR #3507, opened 2025-09-10
add plain tab type for non shell plugins

azmy60:
Plugins can now have two options of view:
- `shell`: used by AI Shell. It has result table at the bottom, and other elements at statusbar. The result table can be in sync with JSON sidebar.
- `base`: No other elements exist. Just the plugin.

### DIFF
--- a/apps/studio/src/components/TabPluginBase.vue
+++ b/apps/studio/src/components/TabPluginBase.vue
@@ -1,0 +1,80 @@
+<template>
+  <div v-if="isCommunity && tab.context.pluginId === 'bks-ai-shell'" class="tab-upsell-wrapper">
+    <upsell-content />
+  </div>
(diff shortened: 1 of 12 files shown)


apps/studio/src/components/TabPluginBase.vue
```diff
@@ -0,0 +1,80 @@
+<template>
+  <div v-if="isCommunity && tab.context.pluginId === 'bks-ai-shell'" class="tab-upsell-wrapper">
+    <upsell-content />
+  </div>
```
not-night-but, line 4: This looks like it doesn't need to be here, since bks-ai-shell is a shell tab, right?

azmy60: Yeah, but just in case. It'd save us some effort in the future.

PR #3529, opened 2025-09-17
Update Redis query results

not-night-but:


apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -845,4 +846,26 @@ export class RedisClient extends BasicDatabaseClient<RedisQueryResult> {
   async getQueryForFilter(): Promise<string> {
     return "";
   }
+
+  private makeQueryResult(command: string, result: unknown): NgQueryResult {
+    log.info("RESULT: ", result)
```
azmy60: Might be useful to also include the command in the log?

not-night-but: Oh I meant to actually remove the log entirely lol. Whoops

azmy60 (approved):
Feel free to merge!

Issue #3536, opened 2025-09-18
BUG: Cursor getting stuck in mongo shell when autosuggestion pops up


itsamrit:
In the mongo shell, the cursor gets stuck when any autosuggestion pops up and then even you press anything on keyboard nothing gets typed in the shell. The only option left is to press `esc` and then start typing again.

<img width="1512" height="982" alt="Image" src="https://github.com/user-attachments/assets/f3117105-66bc-49a9-9968-6e7040bc5dde" />

itsamrit:
Also history doesn't show if you press arrow up or arrow down

Issue #3519, opened 2025-09-15
BUG: Unable to query non-primary ObjectId fields in MongoDB


itsamrit:
RAISED THE FIX FOR THIS: https://github.com/beekeeper-studio/queryleaf/pull/13 (please merge)

Describe the bug:
When using Beekeeper Studio with MongoDB (via QueryLeaf), I am unable to search on fields of type ObjectId if they are not the primary key (_id). It seems Beekeeper only supports querying _id as an ObjectId.

For example, the following query does not return any results, even though matching documents exist:

select * from soldinventories where transaction_id = '68c7199af2aae351fa1f2287'; 

If I query by _id, it works as expected. eg:
select * from soldinventories where _id = '68c719aff2aae351fa1f22d7';

Steps to reproduce

Connect to a MongoDB database in Beekeeper Studio.

Open a collection that has a field of type ObjectId other than _id (e.g., transaction_id).

Try running a query such as:
select * from soldinventories where transaction_id = '68c7199af2aae351fa1f2287';

itsamrit:
RAISED THE FIX FOR THIS: https://github.com/beekeeper-studio/queryleaf/pull/13

Issue #3504, opened 2025-09-10
BUG: Say something here


itsamrit:
Describe the bug
When opening a MongoDB collection and searching with an _id, the JSON view only shows the keys that are part of the schema inferred by Beekeeper. Any additional keys that exist in the document but are not in the schema are not displayed in the JSON view.

To Reproduce

Connect to a MongoDB database in Beekeeper Studio.

Open any collection.

Search/filter documents using an _id.

Switch to the JSON view of the fetched document.

Notice that only schema keys are shown, while other keys in the document are missing.

Expected behavior
All keys present in the actual MongoDB document should be displayed in the JSON view, regardless of whether they are part of the inferred schema.

Screenshots / Gifs
(Attach screenshots or screen recordings showing the missing fields in JSON view if possible)

Version Information (please complete the following information):

OS: [e.g. Windows 11 / MacOS Ventura / Ubuntu 22.04]

App Version: [Find under Help -> About Beekeeper Studio]

Database type and version: MongoDB [version number, e.g. 6.0]

Additional context
This issue makes it difficult to work with dynamic MongoDB documents since not all keys are visible in the JSON view.

itsamrit:
eg: Lets say in the left side i opened a collection of mongodb then in the dropdown its shows all the columns i,e all the root level keys of the collection which it found most common then when i a open a document i,e row of that collection in json viewer then the json viewer and the row shows only those column but since mongodb is schema less it can any new key as well and that key its not showing

itsamrit:
Fix for this bug:- https://github.com/beekeeper-studio/beekeeper-studio/pull/3544

Issue #2413, opened 2024-09-10
Implement JSON Column Filtering Feature


rathboma:
A user has moved to a competitor, TablePlus, because Beekeeper Studio lacks JSON column filtering functionality. To retain and attract users, we need to implement this feature. Additionally, we should ensure that it works properly once implemented.

**Action Items:**
1. Research JSON column filtering requirements.
2. Develop and integrate JSON column filtering into Beekeeper Studio.
3. Thoroughly test the functionality to confirm it works as expected.
*This issue was created from an email or slack message by [IssueWizard](https://issuewizard.com)*

PR #3543, opened 2025-09-21
feat: Unified search bar for JSON 🔥🔥🔥🔥

itsamrit:
✨ Search, but simpler!
Earlier we had two different search boxes 🤯 –
🔎 One just for keys
🔎 Another (opened with Cmd+F) with way too many options cluttering the UI
😕 And too be very honest, the second one was so cluttered and confusing, that no one in our team wanted to use that.

Explored `Monaco editor` but its integration is too complicated at this stage to replace `CodeMirror` with that.

Inspiration:- VSCode search bar

Now 👉 replaced all that with a unified, super simple search bar 🪄
🚀 Cleaner, easier, and way more user-friendly
🚀 By default, only find will be visible to make the UI simpler, you can toggle to use replace functionality.
📸 Attaching screenshots for a quick look 👀

🛠️ Built on CodeMirror only, with UI-level improvements 🎨
🔥 A lot less confusion, a lot more speed!
🔥 Sleeker and more fun – ready to go!

Screenshot 2025-09-21 at 8 42 48 PM
Screenshot 2025-09-21 at 8 42 43 PM
Screenshot 2025-09-21 at 8 42 35 PM

EASY BOX CONTAINS TOOLTIP FOR CLARITY:-

Screenshot 2025-09-21 at 10 05 53 PM

### DIFF
--- a/apps/studio/src-commercial/entrypoints/renderer.ts
+++ b/apps/studio/src-commercial/entrypoints/renderer.ts
@@ -40,6 +40,8 @@ import { ForeignCacheTabulatorModule } from '@/plugins/ForeignCacheTabulatorModu
 import { WebPluginManager } from '@/services/plugin/web'
 import PluginStoreService from '@/services/plugin/web/PluginStoreService'
 import * as UIKit from '@beekeeperstudio/ui-kit'
+// Import for side effects: enhances CodeMirror search panel UI (hides replace row, injects 3-dots menu, auto-opens search/focuses find box in JSON viewer)
+import '@/search-panel-ui-toggle'
 
 (async () => {
 
--- a/apps/studio/src/assets/styles/components/text-editor.scss
+++ b/apps/studio/src/assets/styles/components/text-editor.scss
@@ -1,5 +1,73 @@
 @import "./text-editor-plugins/partial-read-only.scss";
 
+// Shared search panel UI variables and mixins
+$bks-toggle-size: 20px !default;
+$bks-toggle-radius: 4px !default;
+$bks-toggle-gap-x: 2px !default;
+
+@mixin bks-toggle-label() {
+  position: relative;
+  display: inline-flex;
+  align-items: center;
+  gap: 0;
+  margin: 0 $bks-toggle-gap-x 0 0;
+  color: transparent;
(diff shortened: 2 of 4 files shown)


itsamrit:
@rathboma @not-night-but @azmy60 please review

itsamrit:
@azmy60 Currently, custom filter supports only key search. Shouldn't it search for both keys and values just like codemirror search??

azmy60:
@itsamrit Oh, sorry. I just realized CodeMirror search is different from the custom filter. I'll get back in a moment.

itsamrit:
@azmy60  Could you please share your suggestions on how we should handle this?

azmy60:
Okay, first of all, the current custom filter was a requested feature (#2413) and it's different from the built-in CodeMirror Search by functionality. They serve different purposes:

- We can remove unneeded data visually by **filtering**
- We can focus or highlight matching data by **searching** 

Second, I think the search feature itself is fine (I didn't realize it exists in JSON Viewer), but it's not obvious because it's hidden behind `Ctrl+F`. And I'm sure Matthew likes "non-hidden features" in this scenario.

With that said, I'm closing this PR because it drops our existing feature that's still needed. We might need to revisit the built-in CodeMirror search for the bugs (#3562, #3563). We discovered them thanks to this PR :)

azmy60:
@itsamrit Feel free to suggest your ideas on how we could make this feel better for your workflow!

itsamrit:
I had to switch back to MongoDB Atlas since the current MongoDB support feels very limited. At the moment, updating a key’s value leaves me with following workarounds and none of them are working:

Opening the collection in the JSON viewer, which unfortunately doesn’t display all the keys.  https://github.com/beekeeper-studio/beekeeper-studio/issues/3504
Using the query tab to write SQL-like queries, but it only works on primary ObjectId keys. https://github.com/beekeeper-studio/beekeeper-studio/issues/3519
Falling back to the Mongo shell, which often freezes whenever an autosuggestion pops up. https://github.com/beekeeper-studio/beekeeper-studio/issues/3536

I’ll be really happy to see as the support for mongodb improves over time — and I’d be excited to switch back once support improves 😊!

PR #3544, opened 2025-09-21
fix: JSON viewer not showing columns which are not there in left sidebar inside collection in MongoDB

itsamrit:
Description:
This PR makes column filtering configurable by connection type. For types like MongoDB, the full document is shown (no filtering).

### DIFF
--- a/apps/studio/src/mixins/data_converter.js
+++ b/apps/studio/src/mixins/data_converter.js
@@ -13,9 +13,14 @@ export default {
               ];
             */
           const columnNamesOnly = columns.map((c) => c.field)
-          return data.rows.map((row, idx) => {
-            return _.pick(row, columnNamesOnly)
-          })
+          // For MongoDB, do not filter fields at all, return full document
+          if (this.connectionType === 'mongodb') {
+            return data.rows
+          } else {
+            return data.rows.map((row, idx) => {
+              return _.pick(row, columnNamesOnly)
+            })
+          }
         },
         extractColumns(data) {
             // columns here
--- a/apps/studio/src/mixins/data_converter.js
+++ b/apps/studio/src/mixins/data_converter.js
@@ -13,8 +13,9 @@ export default {
               ];
             */
           const columnNamesOnly = columns.map((c) => c.field)
-          // For MongoDB, do not filter fields at all, return full document
-          if (this.connectionType === 'mongodb') {
+          // Set of connectionTypes that should bypass column filtering
+          const bypassColumnFilterConnectionTypes = new Set(['mongodb'])
+          if (bypassColumnFilterConnectionTypes.has(this.connectionType)) {
             return data.rows
           } else {
             return data.rows.map((row, idx) => {


itsamrit:
@azmy60  please review, minor bug fix

not-night-but:
@itsamrit I don't think this is enough to totally fix the issue unfortunately, as we don't use the automatic column generation set up for tabulator, so even if the data is technically there, it won't show up in the table view. I have most of the fix for this, so I'm going to close this PR (I'll open a draft PR at least for you to look at)

itsamrit:
@not-night-but Currently in mongodb we check random table for column generation. In that cases it won't be possible to show all columns. just curious, how will show all columns then??

not-night-but:
@itsamrit So tabulator has a mode that auto generated columns for us. Since we limit the results in the table table view to 100 at a time, I think it's feasible for us to enable this (so it will check all 100 objects for what fields exist and generate a column for each). We can then match up the columns that we've already made in the backend, with these auto-generated ones. Any that aren't already in our columns list will have to be generated, and I would like to figure out a way to update the table state to also include these generated columns, so we can display a broader picture of what the collection looks like. 

From what we're working with I feel like this is the best solution, as we're taking a NoSQL database and trying to display it using a UI that was originally designed for only SQL databases. If you have any ideas regarding this I'd love to hear it, and I am thankful for all of the feedback you've given us on the mongo integration already

itsamrit:
@not-night-but Seems like a good approach 👍 Auto-generating columns for the limited 100 results sounds both feasible and practical. This should cover almost all the  use cases nicely. Thanks 🙏🏻

itsamrit:
@not-night-but If you do above solution 1 problem will be there. The filteration in table will only show the columns name which are there in 100 rows. 
Solution, you can have an additional free text if column name is not present or something like that

not-night-but:
@itsamrit Oh good point! I think I may have to figure out a way to enable raw filters for mongo then. In our SQL databases, you can toggle to filter to be raw, where you would just write a SQL statement to filter the data. In mongo's case, I think we could probably just allow for key value pairs or something? I may look at how mongo's cloud service handles this.

itsamrit:
@not-night-but I feel there might be either of below 2 solutions:- 
Free-text filter (in fallback form) – let users type field names manually if not in 100 rows (like Atlas JSON editor).
Raw query mode (toggle or by default for mongodb)(Mongodb atlas uses this solution) – switch to raw JSON filter editor for advanced use cases. This will be more powerful as we can query nested fields as well.

attaching how mongodb atlas filters
Screenshot 2025-09-23 at 12 48 15 PM

itsamrit:
also checked tableplus they have a option to select raw sql where we can type raw query 
we can have a similar option like raw mongoquery for filteration. It would be very powerful and useful
in that case we can reuse the same code which we use to query in our mongoshell

Screenshot 2025-09-23 at 1 41 49 PM

itsamrit:
@not-night-but are you working on this. If not, can you assign this issue to me. i have made some of changes discussed here locally

not-night-but:
@itsamrit Sorry this kinda fell off my radar, I do have some changes if you want to see them, but there is still an issue with updating our list of columns in the frontend. I don't want them to be overridden by the next refresh, so ideally we would have a separate object with auto generated columns detected in the table table view, that are then added to any other places we list columns (structure, entity sidebar, and table table). The branch is `fix/mongo-schemaless` I believe

itsamrit:
No worries at all! Thanks. I’d love to take a look at the changes. The separate object for auto-generated columns makes sense.

PR #3546, opened 2025-09-23
Slim queries

not-night-but:
Idk how this got lost, but this PR changes saved and used queries to only load 250 character excerpts of the queries, rather than the whole query when listing. The full query is loaded when you open it. This should prevent the app from crashing if someone happens to somehow have a query that is too big. This also adds a MaxLength validation to the text column of used and saved queries that prevents them from being longer than 2_000_000 characters.

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -682,9 +682,6 @@
         this.tab.unsavedQueryText = this.unsavedText
         this.saveTab()
       },
-      fullQuery() {
-        this.initializeQueries();
-      },
       remoteDeleted() {
         if (this.remoteDeleted) {
           this.editor.readOnly = 'nocursor'
@@ -1145,7 +1142,8 @@
       } else if (this.tab.usedQueryId) {
         this.fullQuery = await this.$store.dispatch('data/usedQueries/findOne', this.tab.usedQueryId);
       }
-      
+      this.initializeQueries();
+
       if (this.shouldInitialize) {
         await this.$nextTick()
         this.initialize()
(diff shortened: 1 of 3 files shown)


apps/studio/src/common/appdb/models/favorite_query.ts
```diff
@@ -13,9 +14,13 @@ export class FavoriteQuery extends ApplicationEntity implements QueryLike, ISave
   @Column({type: "varchar", nullable: false})
   title!: string
 
-  @Column({type: "text", nullable: false})
+  @MaxLength(2_000_000, { message: `Queries have a max length of 2,000,000 characters.` })
```
azmy60, line 17: What happens if there is an existing record that exceeds the `MaxLength`?

apps/studio/src/components/sidebar/core/favorite_list/FavoriteListItem.vue
```diff
@@ -33,7 +33,7 @@ export default Vue.extend({
   computed: {
     ...mapState('data/queryFolders', {'folders': 'items'}),
     truncatedText() {
-      return _.truncate(this.item.text, { length: 100});
+      return _.truncate(this.item.excerpt, { length: 100});
```
azmy60, line 36: We might not need the `_.truncate` anyway here?

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -681,6 +682,9 @@
         this.tab.unsavedQueryText = this.unsavedText
         this.saveTab()
       },
+      fullQuery() {
+        this.initializeQueries();
+      },
```
azmy60: We might just call `initializeQueries` right after it's assigned in the `mounted()`? or we plan to have something that triggers this in the future?

rathboma:
@not-night-but looks like this is ready for a round of edits!

apps/studio/src/components/sidebar/core/favorite_list/FavoriteListItem.vue
not-night-but: I mean technically the excerpt is 250 characters, and we're just truncating to 100 here so it is kinda needed?

apps/studio/src/common/appdb/models/favorite_query.ts
not-night-but: Right now it will still exist, you just won't be able to resave it until it is below the max length:
image

PR #3553, opened 2025-09-24
Add new AI Shell features to the docs

azmy60:
- Custom instructions in `General`
- A new option: `Always allow query execution in read-only mode`
- Troubleshooting section for Ollama CORS issue


docs/user_guide/sql-ai-shell.md
```diff
@@ -52,6 +52,16 @@ You can configure the AI Shell using the Beekeeper Studio [configuration system]
...
 
+## Custom Instructions
+
+You can provide custom instructions to personalize how the AI Shell responds to your queries. To add custom instructions:
+
+1. Open the settings menu (⚙️).
+2. Go to the General tab.
+3. Add your custom instructions in the provided text field
+
+![AI Custom Instructions](../assets/images/ai/ai-custom-instructions.png)
```
rathboma, line 66: Questions this needs to answer:
1. It should tell the user it applies to ALL AI sessions, regardless of the connection or workspace
2. Does this get sent on every prompt, or just at the beginning of the session?

docs/user_guide/sql-ai-shell.md
```diff
@@ -79,6 +89,13 @@ There are some limited actions the AI shell can take without your permission, bu
 
 ![AI Asks permission](../assets/images/ai/ai-asks-permission.png)
 
+#### Skipping the prompt in read-only mode
```
rathboma: You call it a prompt here, but you don't refer to this as a 'prompt' in the previous paragraph (you say it `asks permission`. This could be confusing to the reader.

Considering this more, I don't think this option makes sense -- I think it the feature needs to be:
`"Allow execution of read-only queries without asking for permission [in this connection]"`. 

This is different -- I don't want to have to enable read-only mode to skip permissions.

Also: We also need to provide settings like this globally and per-connection. Think about how a user wants the AI to behave differently when connected to production vs test databases.

This is similar to Claude and VSCode -  permissions can be set globally and per-project.

docs/user_guide/sql-ai-shell.md
```diff
@@ -87,3 +104,56 @@ You can disable the AI shell entirely by adding a flag to either your personal o
 {% ini-include section="plugins.bks-ai-shell" %}
 
 Even if the application user has previously entered an API key, this setting will disable the feature and lock out any further usage.
+
+### Troubleshooting
```
rathboma: Troubleshooting should not be nested under 'privacy'. This should be h2, not h3.

docs/assets/images/ai/ai-custom-instructions.png
rathboma, line 1: This should mention that it appended for all connections, not just this connection

docs/user_guide/sql-ai-shell.md
rathboma: Something to think about when providing settings:

Users are editing settings while they're connected to a specific database, so it would be natural for them to think that the settings only related to that database. That means we need to be explicit about when settings are applied globally - both in the app, and in the docs.

Also - this config system is implemented totally differently to the main configuration? Shouldn't this use the configuration files? I know we added a way for plugins to get/set credentials, but should this be how all configuration is stored?

I think we should provide these options as configuration options -- plugins should probably get their own configuration section in the config files?

docs/user_guide/sql-ai-shell.md
azmy60: @rathboma Does that mean for now we scope this option per connection? or for all connections?

azmy60: @rathboma I think we should use `sql-query-identifier` to identify read queries, right?

rathboma: Yep, or you could provide an api from BKS for all plugins to use. Seems like something other plugins might need. Or even provide a `executeReadOnlyQuery` method that throws an error if not read only

rathboma:
@azmy60 this all ready for re-review?

azmy60:
@rathboma Yes

PR #3564, opened 2025-09-25
Switch from Ioredis to node-redis

mrsimb:
Turns out ioredis is no longer actively maintained, and it lacks support for RESP3 protocol, which many managed redis instances use. I refactored the client to use node-redis instead, and tested with my instance on could.redis.io (not an ad), works fine now.

Needs more testing after refactoring.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -17,7 +17,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: '3.x'  # or any specific version
+        python-version: '3.13.7'  # or any specific version
         cache: 'pip' # caching pip dependencies
 
     - name: Install dependencies
--- a/apps/studio/src-commercial/entrypoints/renderer.ts
+++ b/apps/studio/src-commercial/entrypoints/renderer.ts
@@ -114,17 +114,26 @@ import ProductTourPlugin from '@/plugins/ProductTourPlugin'
     Vue.mixin({
       methods: {
         ctrlOrCmd(key) {
-          if (this.$config.isMac) return `meta+${key}`
-          return `ctrl+${key}`
+          const keys = key.split('+');
+          // First letter is uppercase
+          key = keys.map(k => _.upperFirst(k)).join('+');
+          if (this.$config.isMac) return `Meta+${key}`
+          return `Ctrl+${key}`
         },
         // codemirror sytax
         cmCtrlOrCmd(key: string) {
+          const keys = key.split('-');
+          // First letter is uppercase
+          key = keys.map(k => k.charAt(0).toUpperCase() + k.slice(1)).join('-');
           if (this.$config.isMac) return `Cmd-${key}`
           return `Ctrl-${key}`
         },
         ctrlOrCmdShift(key) {
-          if (this.$config.isMac) return `meta+shift+${key}`
-          return `ctrl+shift+${key}`
+          const keys = key.split('+');
+          // First letter is uppercase
+          key = keys.map(k => k.charAt(0).toUpperCase() + k.slice(1)).join('+');
+          if (this.$config.isMac) return `Meta+Shift+${key}`
+          return `Ctrl+Shift+${key}`
         },
         selectChildren(element) {
           const selection = window.getSelection()
--- a/apps/studio/src/App.vue
+++ b/apps/studio/src/App.vue
@@ -85,11 +85,13 @@ import PluginController from '@/components/plugins/PluginController.vue'
 import LockManager from "@/components/managers/LockManager.vue";
 
 import rawLog from '@bksLogger'
+import { assignContextMenuToAllInputs } from './mixins/assignContextMenuToAllInputs'
 
 const log = rawLog.scope('app.vue')
 
 export default Vue.extend({
   name: 'App',
+  mixins: [assignContextMenuToAllInputs],
   components: {
     CoreInterface, ConnectionInterface, Titlebar, AutoUpdater, NotificationManager,
     StateManager, DataManager, UpgradeRequiredModal, ConfirmationModalManager, Dropzone,
--- a/apps/studio/src/components/TabPluginBase.vue
+++ b/apps/studio/src/components/TabPluginBase.vue
@@ -1,5 +1,5 @@
 <template>
-  <div v-if="isCommunity && tab.context.pluginId === 'bks-ai-shell'" class="tab-upsell-wrapper">
+  <div v-if="isCommunity && tab.context.pluginId.startsWith('bks-')" class="tab-upsell-wrapper">
     <upsell-content />
   </div>
   <div v-else class="plugin-base" ref="container">
--- a/apps/studio/src/components/TabPluginShell.vue
+++ b/apps/studio/src/components/TabPluginShell.vue
@@ -1,6 +1,6 @@
 <template>
   <div
-    v-if="isCommunity && tab.context.pluginId === 'bks-ai-shell'"
+    v-if="isCommunity && tab.context.pluginId.startsWith('bks-')"
     class="tab-upsell-wrapper"
   >
     <upsell-content />
--- a/apps/studio/src/mixins/assignContextMenuToAllInputs.ts
+++ b/apps/studio/src/mixins/assignContextMenuToAllInputs.ts
@@ -1,0 +1,132 @@
+import Vue, { ComponentOptions } from "vue";
+
+export const assignContextMenuToAllInputs: ComponentOptions<Vue> = {
+  data() {
+    return {
+      ctxMenu_lastEditable: null as
+        | HTMLInputElement
+        | HTMLTextAreaElement
+        | null,
+    };
+  },
+
+  async mounted() {
+    await this.$nextTick();
+    this.$el.addEventListener("contextmenu", this.ctxMenu_showContextMenu);
+  },
+
+  beforeDestroy() {
+    this.$el.removeEventListener("contextmenu", this.ctxMenu_showContextMenu);
+  },
+
+  methods: {
+    ctxMenu_showContextMenu(event: MouseEvent) {
+      if (
+        !(event.target instanceof HTMLInputElement) &&
+        !(event.target instanceof HTMLTextAreaElement)
+      ) {
+        return;
+      }
+
+      if (event.target.dataset?.disableContextMenu) {
+        return;
+      }
+
+      event.preventDefault();
+
+      // Remember what was right-clicked so handlers act on the correct field
+      if (this.ctxMenu_isEditable(event.target)) {
+        this.ctxMenu_lastEditable = event.target;
+        this.ctxMenu_lastEditable.focus(); // ensure undo/redo applies here
+      } else {
+        this.ctxMenu_lastEditable = null;
+      }
+
+      const selectionDepClass = this.ctxMenu_hasSelectedText(event.target)
+        ? ""
+        : "disabled";
+
+      this.$bks.openMenu({
+        event,
+        options: [
+          {
+            name: "Undo",
+            handler: () => {
+              const el = this.ctxMenu_getLastEditable();
+              if (!el) return;
+              // Uses Chromium's native undo stack for inputs/textareas
+              document.execCommand("undo");
+            },
+            shortcut: this.ctrlOrCmd("z"),
+            write: true,
+          },
+          {
+            name: "Redo",
+            handler: () => {
+              const el = this.ctxMenu_getLastEditable();
+              if (!el) return;
+              document.execCommand("redo");
+            },
+            shortcut: this.ctrlOrCmd("shift+z"),
+            write: true,
+          },
+          { type: "divider" },
+          {
+            name: "Cut",
+            handler: () => document.execCommand("cut"),
+            class: selectionDepClass,
+            shortcut: this.ctrlOrCmd("x"),
+            write: true,
+          },
+          {
+            name: "Copy",
+            handler: () => document.execCommand("copy"),
+            class: selectionDepClass,
+            shortcut: this.ctrlOrCmd("c"),
+          },
+          {
+            name: "Paste",
+            handler: () => document.execCommand("paste"),
+            shortcut: this.ctrlOrCmd("v"),
+            write: true,
+          },
+          { type: "divider" },
+          {
+            name: "Select All",
+            handler: () => document.execCommand("selectAll"),
+            shortcut: this.ctrlOrCmd("a"),
+          },
+        ],
+      });
+    },
+
+    ctxMenu_isEditable(el: HTMLElement) {
+      if (!el) return false;
+      if (el instanceof HTMLInputElement) return !el.readOnly && !el.disabled;
+      if (el instanceof HTMLTextAreaElement)
+        return !el.readOnly && !el.disabled;
+      if (el.isContentEditable) return true;
+      return false;
+    },
+
+    ctxMenu_hasSelectedText(target: HTMLInputElement | HTMLTextAreaElement) {
+      return (
+        typeof target.selectionStart === "number" &&
+        typeof target.selectionEnd === "number" &&
+        target.selectionStart !== target.selectionEnd
+      );
+    },
+
+    ctxMenu_getLastEditable() {
+      // Prefer the last right-clicked field; fall back to the currently focused one
+      const active =
+        this.ctxMenu_lastEditable ||
+        (document.activeElement as HTMLElement | null);
+      if (active && this.ctxMenu_isEditable(active)) {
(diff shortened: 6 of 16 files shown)


mrsimb:
@not-night-but I think i'm mostly done here. To format results, i opted for checking command types manually in the end. There's no other way to know how to guess each command's reply format as far as i know. Works fine for me. Handles RESP3 map responses too.

I also improved autocomplete by much. Your call with COMMAND LIST was a good idea, i used this command to generate info about all available commands and their arguments and saved it as a big object to use in autocomplete. It now suggest optional tokens (like WITHSCORES) too! Oh and i made all command suggestions lowercase, to avoid having to use CAPS all the time.

Now the most boring thing left - testing all that.

edit:

Nevermind, there's still stuff to handle:

image

mrsimb:
To check commands in RESP3 mode, you can send "hello 3" command like so:

image

mrsimb:
As far as i know, this changes HGETALL response from array of pairs to a map of key-values. Don't know if it changes other commands, need to investigate.

mrsimb:
Ill just leave some useful stuff for myself and whoever interested

https://github.com/uglide/redis-doc/blob/e68960e22c654b20ac05236d611e177cf4dec72a/docs/reference/resp2-to-resp3-replies.md

https://github.com/redis/node-redis/blob/master/packages/client/lib/commands/generic-transformers.ts

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -18,19 +18,28 @@ import {
...
 import _ from "lodash";
 import rawLog from "@bksLogger";
 import { RedisChangeBuilder } from "@shared/lib/sql/change_builder/RedisChangeBuilder";
+import fs from "fs/promises";
+import REDIS_COMMAND_DOCS from "@beekeeperstudio/ui-kit/lib/components/text-editor/extensions/redisCommands.json";
 
 type RedisQueryResult = BaseQueryResult;
 
+type RedisSocketConfig = Parameters<typeof createClient>[0]["socket"];
+
```
not-night-but, line 38: I don't love this (mainly because it seems to be resolved to any). It seems even within the redis package that the type they use is technically just an alias for any, as one of the types they use to construct it is unresolved, so it just defaults to any. I think realistically we should just figure out what fields we support and create our own type for it

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -40,6 +49,70 @@ const redisContext: AppContextProvider = {
   },
 };
 
+// function objectFromPairs(pairs: unknown[]) {
+//   return Object.fromEntries(_.chunk(pairs, 2));
+// }
```
not-night-but: Remove any commented code you have, or if we must leave it in for posterity for some reason, at least leave a comment as to why it is currently commented out (eg. "This is currently not supported because A, we will need to do B to support in the future" or something like that)

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -40,6 +49,70 @@ const redisContext: AppContextProvider = {
...
+      return {
+        name: key,
+        command: commandWithArgs.slice(0, keyLength),
+        args: commandWithArgs.slice(keyLength),
+      };
+    }
+  }
+
+  return null;
+}
```
not-night-but, line 84: I think this can be done without having to iterate through the keys like this.

Say our command is something like `cmd list all` where `all` is actually an argument, and `cmd list` is the full command, but we also have the command group `cmd`. That means that we have the array `['cmd', 'list', 'all']`, and we can use that to use field access on the `REDIS_COMMAND_DOCS` object. We start with the full joined command `cmd list all`, check if that key exists (can probably just do `REDIS_COMMAND_DOCS[maybeKey]` or whatever). This is on average O(1). If it doesn't exist, just remove the last word and check again. So maybeKey would be `cmd list`. That way we're iterating over the length of our passed in command, which realistically will be very small compared to the number of keys we have stored.

Does that make sense? I feel like this might be better and would save us having to store the keys separately on import.

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -40,6 +49,70 @@ const redisContext: AppContextProvider = {
...
+        : COMMANDS.zRangeByScore;
+    // The rest should be handled here
+    default: {
+      const commandMerged = command.replaceAll(" ", "").toLowerCase();
+      for (const [key, value] of Object.entries(COMMANDS)) {
+        if (key.toLowerCase() === commandMerged) {
+          return value;
+        }
+      }
+    }
```
not-night-but, line 108: A similar thing here, I don't think it's totally necessary to iterate over all the commands. We can take the command, lets say `acl cat` (because it's been identified previously). With the way it's formatted I think we can just transform that into a camel case string. lodash `_.camelCase(command)` should give you `aclCat` in this case, which is what the command is stored as. There do seem to be some screaming snake case commands in there as well that we might have to support? For example `acl genpass`. It has both an `aclGenPass` command, and an `ACL_GENPASS` command in the `COMMANDS` object. That seems to be the typical formatting, so I'd just either check both, or check the most consistent that you can convert to. Lodash also has a snake case method, so you'd just have to uppercase the result of that. I haven't looked super deeply into this so feel free to correct me.

Unrelated to that, we don't generally put scopes for our cases, so I'd rather we not have that here.

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -68,36 +147,180 @@ type RedisTableRow = {
...
+function makeObjectResult(result: unknown, command: string) {
+  const rows = [result];
+  return {
+    rows,
+    fields: Object.keys(result).map((key) => ({ name: key, id: key })),
+    rowCount: rows.length,
+    affectedRows: 0,
+    command,
+  };
+}
```
not-night-but, line 173: One issue with this is that we want certain things, like hashes, to be vertical in this instance. As in we don't want just one row to be displayed when we run `HGETALL "actor:64"` or whatever, as hashes can truly have an insane amount of fields. Ideally they should be split into a `key` column, and a `value` column.

This is how it appears in the results right now:
<table>
<tr>
 <th>first_name
 <th>last_name
 <th>date_of_birth
<tr>
 <td>Denzel
 <td>Washington
 <td>1954
</table> 

This is how it should look (and this is how it is in master right now)
<table>
<tr>
 <th>Key
 <th>Value
<tr>
 <td>first_name
 <td>Denzel
<tr>
 <td>last_name
 <td>Washington
<tr>
 <td>date_of_birth
 <td>1954
</table>

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -68,36 +147,180 @@ type RedisTableRow = {
...
+
+function makeCursorResult(result: unknown, command: string) {
+  const { cursor, ...rest } = result as Record<string, unknown>;
+  const [data] = Object.values(rest);
+  const rows = Array.isArray(data)
+    ? data.length && _.isPlainObject(data[0])
+      ? data.map((d) => ({ cursor, ...d }))
+      : data.map((value) => ({ cursor, value }))
+    : [{ cursor }];
```
not-night-but: I think we should keep ternary statements to one level, this just becomes much too difficult to read. A boring old if/else statement please :)

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -68,36 +147,180 @@ type RedisTableRow = {
...
+    //     const { summary, ...rest } = objectFromPairs(value);
+    //     const tokens: string[] = [];
+    //     extractArgumentTokens(rest.arguments, tokens);
+    //     return { summary, tokens };
+    //   })
+    //   .toPairs()
+    //   .sortBy(0)
+    //   .fromPairs()
+    //   .value();
+    // await fs.writeFile("commands.json", JSON.stringify(mapped));
```
not-night-but: If this is something we need to keep around, I'd rather we just have it as a separate script that can be run, much like how our bksConfig type generation works

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -68,36 +147,180 @@ type RedisTableRow = {
...
-    return connection
+    this.redis = createClient({
+      socket: socketConfig,
+      username: this.server.config.user || undefined,
+      password: this.server.config.password || "",
+      database: parseInt(this.database.database, 10) || 0,
+    });
+    await this.redis.connect();
+
+    this.getRespVersion();
```
not-night-but: Just await this to be safe

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -68,36 +147,180 @@ type RedisTableRow = {
...
+      }
+    }
 
-    return connection
+    this.redis = createClient({
+      socket: socketConfig,
+      username: this.server.config.user || undefined,
+      password: this.server.config.password || "",
+      database: parseInt(this.database.database, 10) || 0,
+    });
```
not-night-but, line 291: I think we should move to using pools. You should just be able to call `createPool` on the client and use that for all interactions with redis, rather than the client itself (I don't think you should even have to store the result from `createClient`, just the pool. It has all the same methods as the client, and it handles the pooling under the hood (which honestly annoys me, I'd rather manual pool management but that's just me)

https://github.com/redis/node-redis/blob/master/docs/pool.md

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -265,27 +483,77 @@ export class RedisClient extends BasicDatabaseClient<RedisQueryResult> {
...
 
     const results: NgQueryResult[] = [];
 
-    for (const command of commands) {
-      const [commandName, ...args] = splitargs(command);
-      const lowerCommandName = commandName.toLowerCase();
+    for (const line of lines) {
+      if (line.startsWith("#")) {
+        continue; // this is a comment
+      }
```
not-night-but, line 473: It looks like the editor is also expecting to be able to do comments on the same line as a command, is this somehow handled by splitargs? We don't want to accidentally input something as a value or something from a comment

apps/studio/src/lib/db/clients/redis.ts
```diff
@@ -265,27 +483,77 @@ export class RedisClient extends BasicDatabaseClient<RedisQueryResult> {
...
+          ? getTransformReply(knownCommandDef, this.respVersion)
+          : null;
+        const transformedResult = transform ? transform(rawResult) : rawResult;
+
+        // Convert sets to arrays
+        const result = JSON.parse(
+          JSON.stringify(transformedResult, (_key, value) =>
+            value instanceof Set ? [...value] : value
+          )
+        );
```
not-night-but, line 507: This should be possible with type mappings, no? It also looks like Array is the default for sets (at least in RESP3, RESP2 has no mention of Sets).
https://github.com/redis/node-redis/blob/master/docs/RESP.md

apps/studio/src/lib/db/sql_tools.ts
```diff
@@ -33,6 +33,10 @@ export function convertParamsForReplacement(placeholders: string[], values: stri
 }
 
 export function deparameterizeQuery(queryText: string, dialect: Dialect, params: ParamItems | string[], paramTypes: Options["paramTypes"]) {
+  if (dialect === 'redis') {
+    // formatting breaks redis multi-line command execution
+    return queryText;
+  }
```
not-night-but, line 39: I do wonder if we can possibly pass a no-op formatter that only runs the parameter replacement logic, that way we can keep the parameter functionality in the editor.

apps/ui-kit/lib/components/text-editor/extensions/redis.ts
```diff
@@ -5,75 +5,125 @@
...
+  const lowerText = text.toLowerCase();
+
+  for (const key of REDIS_COMMAND_DOCS_KEYS_REVERSED) {
+    if (lowerText.startsWith(key)) {
+      return key;
+    }
+  }
+
+  return null;
+}
```
not-night-but: We should be able to do that same algorithm here, no?

apps/ui-kit/lib/components/text-editor/extensions/redis.ts
```diff
@@ -84,45 +134,55 @@ const redisStreamParser = StreamLanguage.define({
...
     }
   }
 
-  return options.length
-    ? {
-        from: word.from,
-        options: options.slice(0, 50),
-      }
-    : null;
+  return { options, from: currentWord.from };
```
not-night-but, line 191: This doesn't seem to be working at the moment? I can't get autocomplete to trigger.

apps/studio/src/lib/db/clients/redis.ts
mrsimb: It does seem to be correctly typed for me. That's what it shows if i prettify the type:
image

not-night-but: Damn, the typescript lang server is really letting me down here

apps/studio/src/lib/db/clients/redis.ts
mrsimb: It's node-redis that converts some values to sets, when not asked for. It breaks rendering of such values and shows {} instead of them (because there's no proper .toString() for Set in js i guess)

apps/studio/src/lib/db/clients/redis.ts
mrsimb: Adjusted syntax highlighting to show that only if the whole line starts with #, it is considered a comment

apps/studio/src/lib/db/clients/redis.ts
mrsimb: I'm gonna resolve that then

apps/studio/src/lib/db/sql_tools.ts
mrsimb: I couldn't make sql-formatter work with redis code, so i added a generic parameter replacer instead.

apps/ui-kit/lib/components/text-editor/extensions/redis.ts
mrsimb: Can you try it now?

apps/studio/src/lib/db/sql_tools.ts
```diff
@@ -32,7 +32,31 @@ export function convertParamsForReplacement(placeholders: string[], values: stri
...
+    let result = queryText;
+    for (const [key, value] of Object.entries(params)) {
+      // Match :key, $key, or @key with word boundaries
+      const regex = new RegExp(`[:$@]${key}\\b`, 'g');
+      result = result.replace(regex, value);
+    }
+    return result;
+  }
+}
+
```
not-night-but: I think this is a bit dangerous, as I'm sure there are going to be cases where this matches something that isn't actually a parameter. The reason we use sql-formatter to replace params is to avoid issues like that.

Do we think it is worth it to specifically develop this feature for redis? If so we need a couple things:
- This doesn't use the paramTypes, which specifies what parameter syntax is enabled.
- This also doesn't support quoted or numbered parameters from what I can see?
- We're going to need a lot of tests.

I think realistically if we want to allow this for redis, it should be done in a separate PR and heavily tested. So maybe we should just make sure it is disabled for redis for right now?

mrsimb: I'm removing this then. Parameter replacement should definitely be a separate PR as we should focus on making redis client actually usable now.

mrsimb:
> There are still some commented out sections of code that need to be removed or broken out into scripts (Those comments are part of my last review). I think we probably need some better tests (I can get claude to generate some for us if we want) and then I think there are probably some issues with how params work for redis that we need to think about

I'll try to move these into a package.json script, but later. For now, i'll remove the unused code. The current .json file for redis covers all available commands from the latest version with all modules enabled, and i doubt we would need to update it soon.

And yes it would be very nice if you bully claude into making better tests :)

apps/studio/src/lib/db/sql_tools.ts
mrsimb: Removed.

not-night-but (approved):

Chetan11-dev:
Awesome Man, Beekeeper should pay mrsimb for his work.

lipedjow:
Kudos to @mrsimb 
Awesome job!!

dipnlik:
👏🏽👏🏽👏🏽👏🏽👏🏽

LasseMunk:
I started using redis with beekeeper it works great. Thanks a lot !

orphanedrecord:
@mrsimb I saw your contribution via the Beekeeper Studio newsletter … thank you!

Viva open source :)

Kjue:
The redis client was awesome to use. Thanks for that! ❤️

i-walk-away:
you are unironically my hero. thank you for your contribution

PR #3572, opened 2025-09-30
fix: studio's build watcher dying after first rebuild

mrsimb:
The dev script only ever rebuilds once, then the file watcher dies i guess, and it stops rebuilding on file changes. Very annoying issue.

### DIFF
--- a/apps/studio/index.html
+++ b/apps/studio/index.html
@@ -37,7 +37,17 @@
           -webkit-app-region: drag;
         }
         .preloader-wrap {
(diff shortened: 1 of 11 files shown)


not-night-but:
These changes only change the behaviour of watching the configs, not any of the code, which I believe works fine, so I'm going to close this

mrsimb:
Weird, on my macbook this makes it rebuild, while by default i have to kill & rerun the process each time i make changes

PR #3574, opened 2025-10-02
Fix missing context menus

### LINKED ISSUE #3522: BUG: Missing context menus on right click
The following places don't have context menus with standard options (eg - copy / paste / cut / etc ), but they should have at least some of them:

1. JSON sidebar, select any text, expect to be able to right click -> copy
2. Connection form inputs - can't right click -> paste host / port / etc
3. Connection interface -> left side filter input
4. table sidebar -> left side filter input
5. quick search input
6. TableTable -> filter value input
7. TableTable -> complex filter input
8. Edit connection -> rename -> name input
9. JSON Sidebar -> filter input
10. Export table -> filename
11. Export tab -> any inputs
12. Left sidebar -> saved queries -> filter input

I'm sure there's more!

Basic options all text entry should show (even if disabled):

Copy, Cut, Paste, Undo, Redo (these work with keyboard shortcuts already)

Looks like we need an `input` component that has context menus pre-set. VSCode has custom context menus, but has appropriate context menus in the right places (eg search input)

Image

azmy60:
close #3522

### DIFF
--- a/apps/studio/src-commercial/entrypoints/renderer.ts
+++ b/apps/studio/src-commercial/entrypoints/renderer.ts
@@ -113,17 +113,26 @@ import * as UIKit from '@beekeeperstudio/ui-kit'
     Vue.mixin({
       methods: {
         ctrlOrCmd(key) {
-          if (this.$config.isMac) return `meta+${key}`
-          return `ctrl+${key}`
+          const keys = key.split('+');
+          // First letter is uppercase
+          key = keys.map(k => k.charAt(0).toUpperCase() + k.slice(1)).join('+');
+          if (this.$config.isMac) return `Meta+${key}`
+          return `Ctrl+${key}`
         },
         // codemirror sytax
         cmCtrlOrCmd(key: string) {
+          const keys = key.split('-');
+          // First letter is uppercase
+          key = keys.map(k => k.charAt(0).toUpperCase() + k.slice(1)).join('-');
           if (this.$config.isMac) return `Cmd-${key}`
           return `Ctrl-${key}`
         },
         ctrlOrCmdShift(key) {
-          if (this.$config.isMac) return `meta+shift+${key}`
-          return `ctrl+shift+${key}`
+          const keys = key.split('+');
+          // First letter is uppercase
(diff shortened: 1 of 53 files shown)


not-night-but:
@azmy60 is this still a draft?

azmy60:
@not-night-but Yeah. But feel free to do a review. I'm just not sure that we like to use `document.execCommand`, cause mdn says avoid this. The thing is, we know for sure our code runs on a chrome version (electron) always, so maybe that's justifiable? Otherwise, we just build the undo/redo history ourselves?

rathboma:
> @not-night-but Yeah. But feel free to do a review. I'm just not sure that we like to use `document.execCommand`, cause mdn says avoid this. The thing is, we know for sure our code runs on a chrome version (electron) always, so maybe that's justifiable? Otherwise, we just build the undo/redo history ourselves?

I think this is fine -- we're locked to Electron, and presumably that's what happens when you hit ctrl-z anyway?

azmy60:
@rathboma yep, that's what happens.

azmy60:
I'll take a quick look next week to make sure I don't miss any major things, and then mark this as ready.

azmy60:
@rathboma This should also unify the shortcut labels. Some shortcuts are `ctrl+x` (all lowercase), and some are `Ctrl+X`. The latter feels better since most apps do that.

azmy60:
@rathboma @not-night-but This is ready for review.

apps/studio/src-commercial/entrypoints/renderer.ts
```diff
@@ -114,17 +114,26 @@ import ProductTourPlugin from '@/plugins/ProductTourPlugin'
     Vue.mixin({
       methods: {
         ctrlOrCmd(key) {
-          if (this.$config.isMac) return `meta+${key}`
-          return `ctrl+${key}`
+          const keys = key.split('+');
+          // First letter is uppercase
+          key = keys.map(k => k.charAt(0).toUpperCase() + k.slice(1)).join('+');
```
not-night-but: Lodash has a function for this, `_.upperFirst()`

azmy60:
@not-night-but I can trigger context menu from JSON sidebar.

2025-10-31-194940_360x241_scrot

PR #3578, opened 2025-10-03
Add AI Shell tooltip

azmy60:
2025-10-03-083951_915x387_scrot

This PR also refreshed the add-tab button group:

Recording 2025-10-03 at 08 46 08

### DIFF
--- a/apps/studio/src/plugins/ProductTourPlugin.ts
+++ b/apps/studio/src/plugins/ProductTourPlugin.ts
@@ -1,0 +1,144 @@
+import { PluginObject } from "vue";
+import { driver as createDriver, DriveStep, PopoverDOM } from "driver.js";
+import { UtilityConnection } from "@/lib/utility/UtilityConnection";
+import type store from "@/store";
+
+type Context = {
+  store: typeof store;
+  utility: UtilityConnection;
+};
+
+type FlowId = "connectedScreen";
+
+type FlowStep = DriveStep & {
+  shouldShow: (context: Context) => boolean | Promise<boolean>;
+  onFinished: (context: Context) => void | Promise<void>;
+  onRender?: (popover: PopoverDOM) => void;
+};
+
+const flows: Record<
+  FlowId,
+  {
+    steps: FlowStep[];
+  }
+> = {
+  connectedScreen: {
+    steps: [
+      {
+        element: "#add-tab-group",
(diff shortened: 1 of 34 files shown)


rathboma:
@azmy60 This looks great!

Content update:

title: keep as is
para: AI can explore your database and run { SQL | code } to answer your questions. Integrates with your favorite LLM.
buttons: [learn more - https://beekeeperstudio.io/features/sql-ai] [okay]

apps/studio/src/plugins/ProductTourPlugin.ts
```diff
@@ -0,0 +1,130 @@
...
+          }
+
+          const settings = context.store.getters["settings/settings"];
+          if (settings["tabDropdownAIShellHintShown"]) {
+            return false;
+          }
+
+          // Dont show it if the plugin is not in the dropdown menu (disabled or not installed)
+          if (
+            !context.store.getters["tabs/newTabDropdownItems"].find(
```
rathboma: Seems like you should add getters here to the store:

`if(context.store.getters.aiShellInstalled) ...`

`if(context.store.getters.AIShellHintShown)...`

apps/studio/src/plugins/ProductTourPlugin.ts
```diff
@@ -0,0 +1,130 @@
...
+              (t) => t.config.pluginId === "bks-ai-shell"
+            )
+          ) {
+            return false;
+          }
+
+          const tabs = await context.utility.send("appdb/tabs/find", {
```
rathboma: i think here we only care if they've EVER used the ai shell right? Do we care what workspace they used it in?

Like if I use a lot in dev, I don't need reminding in production. I know it exists.

rathboma: Do we have to explicitly look for closed tabs? Maybe the utility process can just return a count?

azmy60: Nice catch!

azmy60:
@rathboma not sure what you mean by `{ SQL | code }`. Is it like this?

2025-10-06-203111_321x201_scrot

rathboma:
@azmy60 it's not SQL when connected to Redis. So we need to know if we should write SQL or code

PR #3587, opened 2025-10-06
onboarding notification

### LINKED ISSUE #3583: FEAT: Onboarding notification + new video
Image

Goal: Make it easier for folks to get started with Beekeeper Studio.

Changes needed:
- Display a welcome noty on first run.
- Keep displaying this noty on the connectionInterface unless the user has successfully connected to a database at least once
- Record a new 60 second getting started video showing where things are (must be 60s)
- Don't show the 'upgrade' noty while the 'getting started' noty is active.

azmy60:
fix #3583 

2025-10-08-171227_381x210_scrot


apps/studio/src/components/NotificationManager.vue
```diff
@@ -51,16 +62,73 @@ export default Vue.extend({
...
       }, 1000 * 60 * 5)
-    }
+    },
+    notifyUpsell() {
+      if (!this.isShowingOnboardingNoty) {
+        new Noty(this.upsellNotificationOptions).show()
+      }
+    },
+    async notifyOnboarding() {
+      Noty.closeAll('onboarding');
```
rathboma, line 78: You shouldn't need to do this - you can tell noty you only want one notification from a specific queue

apps/studio/src/components/NotificationManager.vue
```diff
@@ -51,16 +62,73 @@ export default Vue.extend({
...
 
       this.timeoutID = setTimeout(() => {
-        new Noty(this.upsellNotificationOptions).show()
+        this.notifyUpsell()
       }, 1000 * 60 * 5)
-    }
+    },
+    notifyUpsell() {
+      if (!this.isShowingOnboardingNoty) {
+        new Noty(this.upsellNotificationOptions).show()
```
rathboma, line 74: Noty might have a way to do this too -- maybe put them in the same queue and the onboarding noty has priority or something?

apps/studio/src/migration/20251008_add_onboarding_noty_shown_setting.js
```diff
@@ -0,0 +1,29 @@
+export default {
+  name: '20251008_add_onboarding_noty_shown_setting',
+  async run(runner) {
+    const queries = [
```
rathboma, line 4: This looks fine, but we could probably do with some sort of helper for creating new settings like this.

rathboma:
FYI - this is blocked as we need an onboarding video. I'll try and create one soon. Or maybe Matthew can?

apps/studio/src/migration/20251008_add_onboarding_noty_shown_setting.js
azmy60: Oh I've been meaning to ask if we can add a setting without migration? I think the existing helper can do that:

```js
context.dispatch('settings/save', { key: 'newKey', value: "the new key's value" });
```

rathboma: I think we should still have a migration -- or some way of seeding the app with data on start, but a helper would be super useful so we don't have to write these super raw query migrations for it

apps/studio/src/common/menus/MenuItems.ts
```diff
@@ -117,6 +117,11 @@ export function menuItems(actionHandler: IMenuActionHandler, settings: IGroupedU
       label: 'Contact Support',
       click: actionHandler.contactSupport
     },
+    gettingStartedGuide: {
+      id: 'gettingStartedGuide',
+      label: 'Getting Started Guide',
+      click: actionHandler.openGettingStarted
```
rathboma, line 160: FYI - this can just link directly to the docs page, electron will intercept and open in the browser

apps/studio/src/components/NotificationManager.vue
```diff
@@ -4,13 +4,14 @@
...
+import { mapGetters, mapActions, mapState } from 'vuex'
 
 export default Vue.extend({
   data: () => {
     return {
       notificationInterval: null,
       timeoutID: null,
+      isShowingOnboardingNoty: false,
       upsellNotificationOptions: {
         text: "👋 Beekeeper Studio is run by a small team. Buy the full version of Beekeeper Studio to support development and get more features. Thank you ♥",
```
rathboma: Let's update this:

```
Upgrade for features like the JSON row viewer, AI shell, & NoSQL support. All purchases come with a <strong>lifetime usage license</license>.
```

One link - `Get Started` -> beekeeperstudio.io/pricing

apps/studio/src/components/NotificationManager.vue
```diff
@@ -51,16 +62,73 @@ export default Vue.extend({
...
+      }
+
+      const n = new Noty({
+        text: `<div class="noty-onboarding-title">
+                <img class="noty-onboarding-logo" src="/src/assets/logo.svg">
+                Welcome to Beekeeper Studio!
+              </div>
+              <div class="noty-onboarding-body">
+                Double click the demo database to explore app features,
+                <a class="link" href="https://docs.beekeeperstudio.io">watch our 60s quickstart video</a>,
```
rathboma: Only hyperlink `60s quickstart video` and `getting started guide`

also change: `...our getting started guide` to `the getting started guide`.

Don't want to say `our` twice in the same sentence.

apps/studio/src/components/NotificationManager.vue
azmy60: It can't. It will still show both notification at the same time.

Recording 2025-10-16 at 21 33 47

apps/studio/src/common/menus/MenuItems.ts
azmy60: Yeah I know that. We can't do it here though, apple uses native menu bar.

azmy60:
@rathboma I addressed your feedbacks except the noty ones. I tried the `queue` and `killer` properties, but they don't seem to work the way we want it. Or maybe I'm missing something.

apps/studio/src/background/NativeMenuActionHandlers.ts
```diff
@@ -93,6 +93,10 @@ export default class NativeMenuActionHandlers implements IMenuActionHandler {
     shell.openExternal("https://docs.beekeeperstudio.io/support/contact-support/")
   }
 
+  openGettingStarted(): void {
+    shell.openExternal("https://docs.beekeeperstudio.io/getting-started/")
```
not-night-but: This link doesn't exist, I believe it's https://docs.beekeeperstudio.io/getting-started-guide/

apps/studio/src/components/NotificationManager.vue
```diff
@@ -51,16 +62,74 @@ export default Vue.extend({
...
+
+      const n = new Noty({
+        text: `<div class="noty-onboarding-title">
+                <img class="noty-onboarding-logo" src="/src/assets/logo.svg">
+                Welcome to Beekeeper Studio!
+              </div>
+              <div class="noty-onboarding-body">
+                Double click the demo database to explore app features,
+                watch our <a class="link" href="https://docs.beekeeperstudio.io">60s quickstart video</a>,
+                or read the <a class="link" href="https://docs.beekeeperstudio.io">getting started guide</a>.
```
not-night-but: This is the wrong link

apps/studio/src/components/NotificationManager.vue
```diff
@@ -51,16 +62,74 @@ export default Vue.extend({
...
+      }
+
+      const n = new Noty({
+        text: `<div class="noty-onboarding-title">
+                <img class="noty-onboarding-logo" src="/src/assets/logo.svg">
+                Welcome to Beekeeper Studio!
+              </div>
+              <div class="noty-onboarding-body">
+                Double click the demo database to explore app features,
+                watch our <a class="link" href="https://docs.beekeeperstudio.io">60s quickstart video</a>,
```
not-night-but: @rathboma I believe we still need the video, yeah?

rathboma:
Just remove the video link for now, then happy to merge to master. Should just link to https://docs.beekeeperstudio.io/getting-started-guide/

PR #3600, opened 2025-10-11
Fix column reordering for mysql

not-night-but:
Looks like the change builder was mucking things up. If you modify a column, it resets the column to default state. So we have to specify whether it's nullable, the default value, and the extra column, as well as the new position so we don't lose it.
This was also happening with alter column for mysql, so I fixed that as well.
Also disabled editing the extra field in table structure because there's literally no reason as far as I can tell

### DIFF
--- a/apps/studio/src/shared/lib/sql/change_builder/MysqlChangeBuilder.ts
+++ b/apps/studio/src/shared/lib/sql/change_builder/MysqlChangeBuilder.ts
@@ -27,7 +27,7 @@ export class MySqlChangeBuilder extends ChangeBuilderBase {
     // string, already quoted
     if (defaultValue.startsWith("'")) return this.wrapLiteral(defaultValue)
     // is a generated expression, so don't quote
-    if (isGenerated) return `(${defaultValue})`;
+    if (isGenerated) return `(${defaultValue.replace(/\\'/g, "'")})`;
     // string, not quoted.
     return this.escapeString(defaultValue.toString(), true);
   }
@@ -203,8 +203,9 @@ export class MySqlChangeBuilder extends ChangeBuilderBase {
           columnDef += ' NOT NULL'
         }
 
-        if (defaultValue !== undefined && defaultValue !== null) {
-          columnDef += ` DEFAULT ${this.defaultValue(defaultValue)}`;
+        if (!_.isNil(defaultValue)) {
+          const isGenerated = /DEFAULT_GENERATED/gi.test(extra);
+          columnDef += ` DEFAULT ${this.defaultValue(defaultValue, isGenerated)}`;
         }
 
         if (extra && !generated) {
--- a/apps/studio/src/shared/lib/dialects/mysql.ts
+++ b/apps/studio/src/shared/lib/dialects/mysql.ts
@@ -55,7 +55,6 @@ export const MysqlData: DialectData = {
     alter: {
       multiStatement: true,
       renameSchema: true,
-      reorderColumn: true, // FIXME remove this once it's fixed
     },
     schema: true,
   },


rathboma:
This needs some serious tests

apps/studio/src/components/tableinfo/TableSchema.vue
```diff
@@ -276,11 +276,9 @@ export default Vue.extend({
           field: 'extra',
           tooltip: true,
           headerTooltip: 'eg AUTO_INCREMENT',
-          editable: this.isCellEditable.bind(this, 'alterColumn'),
+          editable: false,
```
azmy60, line 279: Is this intentional? Do we not support editing extras?

apps/studio/src/shared/lib/sql/change_builder/MysqlChangeBuilder.ts
```diff
@@ -124,19 +182,62 @@ export class MySqlChangeBuilder extends ChangeBuilderBase {
...
+
+        if (collation) {
+          columnDef += ` COLLATE ${collation}`;
+        }
+
+        if (nullable === false) {
+          columnDef += ' NOT NULL'
+        }
+
+        if (defaultValue !== undefined && defaultValue !== null) {
```
azmy60: This can be `if (!_.isNil(defaultValue))`. But this is perfectly fine too.

azmy60 (changes requested):
I only have one question. And overall the tests look really good and more than enough (in a good way). Obviously it's best if the tests are in the common tests instead of just for MySQL, but we'll want to do that in a different PR, or maybe it's specific to MySQL?

Also, If you update your branch, I disabled the reordering feature for MySQL in the [DialectData](https://github.com/beekeeper-studio/beekeeper-studio/commit/d397f83c067ca9fc91f8520aead8d115889fd5c8). Be sure to re-enable it if it's ready.

apps/studio/src/components/tableinfo/TableSchema.vue
not-night-but: @azmy60 Editing the extra column actually just makes no sense. It's a column that mysql uses for four values:
- AUTO_INCREMENT
- on update CURRENT_TIMESTAMP
- DEFAULT_GENERATED
- STORED GENERATED or VIRTUAL GENERATED

For the most part this is just metadata. DEFAULT_GENERATED is in the column when the default for a column is an expression, for example `DEFAULT (now())`. STORED/VIRTUAL GENERATED is there when the column is a generated column. Neither of these is valid sql, just informatino. AUTO_INCREMENT is there, but in my opinion we shouldn't be using the extra column to be modifying whether or not it's auto increment, that would just be hacky. and then the on update CURRENT_TIMESTAMP is also valid sql like auto increment, but again is just there as metadata. Really how we had the extra column before was just a way for people to randomly input SQL that we would execute, and I don't think that should be something we support

apps/studio/src/shared/lib/sql/change_builder/MysqlChangeBuilder.ts
not-night-but: I knew that existed somewhere I was just too lazy to do a search lol

not-night-but:
@azmy60 I'm not totally sure if the tests make sense for other dialects? Definitely not the reordering tests, as we only allow reordering for mysql and firebird apparently, and firebird doesn't behave the same way that mysql does in this case. And for the other tests, the reason that this issue even exists is just because of how mysql `CHANGE` and `MODIFY` statements work, in that they just completely redefine the column, essentially just deleting it and replacing it with whatever is in the above statements. That doesn't seem to be a standard across databases, as most other databases just allow `ALTER COLUMN` which just changes whatever attribute you include in the statement

PR #3622, opened 2025-10-21
fix sidebar widths

### LINKED ISSUE #3603: BUG & FEATURE: Sidebar problems
**Describe the bug**
Sidebar behaviour is inconsistent depending on open instances, window size, which one is closed first, etc.

**To Reproduce**
1. Open beekeeper, note that the sidebar has a certain width.
2. Open another instance, then the sidebar is twice as big
3. Maximize. Now the sidebar is huge
4. Depending on which instance you close first, one or the other might be the new default width
5. This is aggravated by the fact that width is set when the window is maximized, so the next time I open the app the size is ridiculous. Please give some love to #622

**Expected behavior**
The sidebar should have a fixed size independent of window size. This could also be dependent on which database connection you have.

**Screenshots / Gifs**
If applicable, add screenshots or GIFS to help explain your problem.

 - OS: Windows
 - App Version: 5.4.1
 - Database type and version: all of them

azmy60:
resolves #3603

### DIFF
--- a/apps/studio/src/common/LocalStorage.ts
+++ b/apps/studio/src/common/LocalStorage.ts
@@ -71,7 +71,7 @@ export const SmartLocalStorage = {
   },
   /** Returns `true` if the key exists */
   exists(key: string) {
-    return localStorage.getItem(key) !== null;
+    return localStorage.getItem(key) != null;
   },
   remove(key: string) {
     localStorage.removeItem(key);


apps/studio/src/common/LocalStorage.ts
```diff
@@ -68,6 +68,13 @@ export const SmartLocalStorage = {
         expiry: now.getTime() + ttl,
       }
     localStorage.setItem(key, JSON.stringify(item))
-  }
+  },
+  /** Returns `true` if the key exists */
+  exists(key: string) {
+    return localStorage.getItem(key) !== null;
```
wmontgomery: not a hill to die on, but I'd have the check as `!= null` just in case undefined gets added there for some reason.

wmontgomery (approved):

Issue #3362, opened 2025-07-14
FEAT: Add Editor Font and Font Size Settings


milonas115:
**Is your feature request related to a problem? Please describe.**  
Yes. Beekeeper Studio currently does not offer any way to change the editor font or font size. I'm forced to use zoom as a workaround, which scales *everything* (buttons, tabs, layout), making the interface look terrible and inefficient to use. It’s incredibly frustrating, especially on high-resolution monitors or for users with visual needs.

**Describe the solution you'd like**  
I want a settings section where I can customize the **editor font family**, **font size**, and optionally **line height** or **letter spacing**. This is a standard feature in nearly every SQL or code editor (DBeaver, TablePlus, VS Code, etc.) and critical for usability.

**Describe alternatives you've considered**  
- Using the zoom feature (`Cmd +` / `Ctrl +`), but it affects the entire UI—not just the editor text. This makes the application look and behave poorly.  
- Manually editing any config file if available (none exists now that Preferences are deprecated).  
- Switching to a different SQL client that allows this.

**Additional context**  
- The lack of font control is a basic accessibility issue.  
- Even something minimal like a setting for font size would go a long way.  

**Database type, if appropriate**  
This feature request applies globally and is not specific to any database engine.

not-night-but:
Yeah I think this is definitely something we should add. We already have a config system available, so we should just need to add those keys

jcicurel-pass:
Hello 👋 

I have a question regarding the font size keyboard shortcut
On macOS, when I press Cmd+/ (Cmd+Shift+:), this triggers the Editor Font Size increase. But this shortcut was also used to toggle a line comment I think?
Is it possible to change one shortcut or the other with the configuration? I did not find either one in the Configuration Reference

Thank you!

PR #3624, opened 2025-10-21
feat: allow query editor window font size to be customized

dustin-jones:
related to https://github.com/beekeeper-studio/beekeeper-studio/issues/3362

## Description

  The lack of adjustable SQL editor font size has been my biggest pain point when using Beekeeper Studio. I created this solution as a suggestion for a simple implementation to increase usability.

  ## Changes

  - Added three menu items to View menu:
    - Reset Editor Font Size (default: 14px)
    - Increase Editor Font Size (⌘⇧. / Ctrl+Shift+.)
    - Decrease Editor Font Size (⌘⇧, / Ctrl+Shift+,)
  - Font size increments by 2px per adjustment (bounded: 10px - 24px)
  - Setting persists across app restarts

output

### DIFF
--- a/apps/studio/src/lib/menu/ClientMenuActionHandler.ts
+++ b/apps/studio/src/lib/menu/ClientMenuActionHandler.ts
@@ -64,4 +64,7 @@ export default class ClientMenuActionHandler implements IMenuActionHandler {
   managePlugins = () => send("managePlugins")
   keyboardShortcuts = () => send("keyboardShortcuts")
   handleAction = (action: CustomMenuAction) => send('handleAction', action)
+  editorFontSizeReset = () => send('editorFontSizeReset')
+  editorFontSizeIncrease = () => send('editorFontSizeIncrease')
+  editorFontSizeDecrease = () => send('editorFontSizeDecrease')
 }


apps/studio/src/App.vue
```diff
@@ -130,6 +133,9 @@ export default Vue.extend({
       document.body.className = `theme-${this.themeValue}`
       this.trigger(AppEvent.changedTheme, this.themeValue)
     },
+    editorFontSize(newSize) {
+      document.documentElement.style.setProperty('--bks-editor-font-size', `${newSize}px`)
```
azmy60: Just a way to make this a little cleaner: instead of adding a new watcher and a condition in the `mounted()`, we could simply define the style in the template like this:

```diff
<template>
  <div class="style-wrapper">
    <div
      class="beekeeper-studio-wrapper"
      :class="{ 'beekeeper-studio-minimal-mode': $store.getters.minimalMode }"
++    :style="{ '--bks-editor-font-size': `${editorFontSize}px` }"
     >
```

apps/ui-kit/lib/components/text-editor/text-editor.scss
```diff
@@ -117,6 +117,7 @@
   // Common CodeMirror styles
   .cm-editor {
     font-family: $font-family-mono;
+    font-size: var(--bks-editor-font-size, 0.875rem);
```
azmy60: To be consistent with the other variables, can we name this `--bks-text-editor-font-size` instead of `--bks-editor-font-size`?

azmy60:
Hey @dustin-jones ! just bumping this. Are you still planning to work on this? There are two minor changes and some merge conflicts to resolve before we can merge. Thanks!

PR #3625, opened 2025-10-22
Remember maximized state

### LINKED ISSUE #622: FEAT: remember maximized state on startup
I always maximimize beekeeper window after starting.

Remember the window maximized state (true/false) and apply that on startup

Should this be handled by desktop environment?

azmy60:
Resolves #622

### DIFF
--- a/apps/studio/src/migration/helpers.ts
+++ b/apps/studio/src/migration/helpers.ts
@@ -1,20 +1,9 @@
+import { UserSettingValueType } from "@/common/appdb/models/user_setting";
 import type { QueryRunner } from "typeorm";
 
 /**
  * Helper functions for database migrations
  */
-
-/**
- * Value types for user settings
- */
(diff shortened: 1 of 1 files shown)


apps/studio/src/migration/helpers.ts
```diff
@@ -0,0 +1,89 @@
...
+ * Value types for user settings
+ */
+const UserSettingValueType = {
+  string: 0,
+  int: 1,
+  float: 2,
+  object: 3,
+  array: 4,
+  boolean: 5,
+}
```
not-night-but: Could ojust import this from user_setting.ts, no?

not-night-but (changes requested):
Just a note about redefining the user setting enum.

Should we maybe push this into 55?

azmy60:
@not-night-but I just rebased the branch to rc-55.

PR #3631, opened 2025-10-23
Add incoming keys in getTableKeys

azmy60:
```ts
const keys = connection.getTableKeys("orders");
```

Before, `getTableKeys` can only return the foreign keys of `orders`:
```
[
  {
    toTable: 'products',
    fromTable: 'orders',
    toColumn: 'product_id',
    fromColumn: 'product_id',
  }
]
```

Now, it can return the incoming keys, which are the keys from other tables that reference `orders`:
```diff
[
    {
        toTable: 'products',
        fromTable: 'orders',
        toColumn: 'product_id',
        fromColumn: 'product_id',
++      direction: 'outgoing',
++  },
++  {
++      toTable: 'orders',
++      fromTable: 'order_items',
++      toColumn: 'order_id',
++      fromColumn: 'order_id',
++      direction: 'incoming',
    },
]
```


apps/studio/src-commercial/backend/lib/db/clients/surrealdb.ts
```diff
@@ -359,9 +359,46 @@ export class SurrealDBClient extends BasicDatabaseClient<SurrealDBQueryResult> {
...
+        })
+        .map((fieldInfo) => ({
+          columnName: fieldInfo.name,
+          dataType: fieldInfo.kind
+        }));
+    };
+
+    // Outgoing keys: columns in THIS table that reference OTHER tables
+    const currentTableFields = tableFieldsMap.get(table) || [];
+    const columns = getColumnsFromFields(currentTableFields, table);
```
not-night-but: Just be aware that some tables could be schemaless, so by not using `listTableColumns` we are missing out on those tables

not-night-but (changes requested):
Looks good! Just a note about surreal, if you don't feel it's important go ahead and dismiss and merge

azmy60:
I'm addressing Day's feedback, but do ya'll have any thoughts too @wmontgomery @rathboma? Do we want to put incoming keys and outgoing keys in the same array?

apps/studio/src-commercial/backend/lib/db/clients/surrealdb.ts
azmy60: It ain't me. Claude did it.  🤖

not-night-but:
@azmy60 I guess the other thing is should we always pull this information back? In this implementation this is the default, but it's filtered out in the table keys info. And in some cases this could be a pretty hefty query to run. I almost wonder if it should just be a completely separate method on the client?

azmy60:
@not-night-but valid point. It’s the reason why I’m not sure this is a great approach. I mean we’d use all keys in table info eventually. But it’s probably wiser to make a separate method.

EDIT: misclicked the close PR button lol.

PR #3633, opened 2025-10-25
fix: reset read-only mode to false for existing connection when premium expires

vimbing:
When a user's ultimate subscription expires, they lose access to the read-only toggle input (it becomes disabled). However, if read-only mode was enabled before expiration, it remained active, preventing them from making any changes to their connection.

This commit automatically resets read-only mode to false when it detects that subscription is not active, but read-only mode is enabled for selected connection, ensuring users which don't renew subscriptions can still use their existing connections normally.

I'm not sure how often such situations happen, but it was pretty annoying for me, so I've decided to contribute :)).

### DIFF
--- a/apps/studio/src/components/ConnectionInterface.vue
+++ b/apps/studio/src/components/ConnectionInterface.vue
@@ -345,10 +345,21 @@ export default Vue.extend({
         this.config = conn;
       })
     },
-    edit(config) {
+    async edit(config) {
       this.config = _.clone(config)
       this.errors = null
       this.connectionError = null
+
+       // If user didn't renew ultimate subscription, default read-only mode to false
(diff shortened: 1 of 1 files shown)


not-night-but:
@vimbing I think this is actually a regression. I'm like 90% sure we used to have a safeguard in the backend to make sure this scenario couldn't happen

not-night-but:
@vimbing So sorry I thought I had sent another message. I'm fixing the regression (which just overrides readonly mode entirely if you don't have a license, so even if the connection is in readonly mode, it won't throw an error). But I do think we should also update the connection so it's not in readonly. I just think it shouldn't be done here, so I think I'm going to close this and add some checks to the save logic and the loading logic so that the connections can't be in readonly. Again so sorry for keeping you in suspense

PR #3636, opened 2025-10-27
Manual commit mode

not-night-but:
## NOTE
- This needs [this pr](https://github.com/coresql/sql-query-identifier/pull/82) to be merged and released (and then identifier to be updated in this pr) before this is merged

Add the ability to manage transactions through the UI. Changing the transaction mode from `Auto` to `Manual` will reserve a connection for the current tab.
 
ScreenRecord_12-Nov_16-20-05_12306
image

There is a configurable maxReservedConnections now in the config file, which will prevent users from reserving too many connections from the pool:
image

As well as a configurable transaction timeout, which will auto switch tabs and present a modal if a transaction is left open for too long, allowing the user to continue the transaction, or just roll it back (default set to 10 minutes)
image

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -91,8 +91,8 @@
                 @click.prevent="toggleCommitMode('manual')"
                 v-tooltip="getCommitModeVTooltip({
                   title: 'Manual commit mode',
-                  description: 'This is new! It helps you running transactions.',
-                  learnMoreLink: 'https://beekeeperstudio.io/',
+                  description: 'Write actions will require you to manually commit your changes',
+                  learnMoreLink: 'https://docs.beekeeperstudio.io/user_guide/sql_editor/manual-transaction-management',
                 })"
               >
                 <span class="togglebutton-content">
@@ -109,7 +109,7 @@
                 ...getCommitModeVTooltip({
                   title: `<i class='material-icons'>commit</i><span>Transaction active</span>`,
                   description: 'Once committed or rolled back, it will be deactivated.',
-                  learnMoreLink: 'https://beekeeperstudio.io/',
+                  learnMoreLink: 'https://docs.beekeeperstudio.io/user_guide/sql_editor/manual-transaction-management',
                   className: 'transaction-active',
                   show: showTransactionActiveTooltip,
                   onClose() {


azmy60:
@not-night-but Can we use the `run` button in manual mode? If not, might worth to disable it and add a little tooltip that says something like "Can't use this button in manual mode. Please switch to auto commit mode."

not-night-but:
@azmy60 yes you can use run in manual commit mode

azmy60:
Oh mb. I mean ofc lol. I'm just thinking if the commit buttons should be using "primary" color in this case? meaning they are white buttons like `run`. I don't know. Will do a proper review in a minute.

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -142,6 +142,24 @@
               </x-menu>
             </x-button>
           </x-buttons>
+          <x-buttons v-show="canManageTransactions && isManualCommit" class="">
```
azmy60: I would reorder these buttons so they look like this:

Image

In this scenario, the primary button is `run`. So I think that it'll be more consistent to keep the `run` button on the right. There's an [article](https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/) that talks about this similar issue if you are interested. Please feel free to disagree or add your own thoughts :)

apps/studio/src/components/TransactionTimeoutModal.vue
```diff
@@ -0,0 +1,65 @@
+<template>
```
azmy60: What happens if there are active transactions in multiple tabs?

azmy60 (dismissed):
I left a question and pointed out one small detail. Overall this looks good!

apps/studio/src/components/TabQueryEditor.vue
azmy60: One thing to think about, no need to do it now: we could use icons in there based on the vscode commit/rollback icons.

apps/studio/src/components/TransactionTimeoutModal.vue
not-night-but: There is one modal for every tab, named with the tabId. I guess one issue might be if two transactions timeout at the same time... Also I'm not convinced this *should* be a modal. Perhaps we make it a noty that doesn't go away unless you press continue or rollback?

azmy60: Suppose there are multiple opened tabs, how do you tell which noty/modal belongs to which tab?

It'd probably make sense to show a noty or a modal, but it'd be better if the UI is scoped to the tab rather than global. I think it'll feel less confusing.

Here's my idea:

1. Add a alert box at the top of the editor (i used online ms paint so excuse the pixelation lol)

image

2. And if we are working on a different tab, and the other tab is timed out:

image

2 can be solved with a noty as well, that says "Active transaction in some tab has been timed out" Or we probably don't even need 2 at all.

I'm not sure how much effort for 1, but that's what I'd do personally.

At the very least, I think modal should be avoided if we can. I'm also worried that it'll look like an error modal. Like is it that bad to be shown a modal? I'm worried it'd feel like something serious happen (in reality, its just a timeout).

Noty is probably OK, but see my earlier question. And at least it's better than a modal (does not block you from doing something else).

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -1096,6 +1216,60 @@
...
+        if (this.warningNoty) {
+          this.warningNoty.close();
+          this.warningNoty = null;
+        }
+      },
+      async toggleCommitMode() {
+        if (!this.canManageTransactions) return
+        if (this.isManualCommit) {
+          if (this.hasActiveTransaction) {
+            this.connection.rollbackTransaction(this.tab.id);
```
azmy60: @not-night-but should we `await` this?

PR #3646, opened 2025-10-31
Improve JSON Field Readability in Table View

marcosstefani:
## Summary
Adds lightweight inline syntax highlighting and safer formatting for JSON values displayed in table cells. Keys, strings, numbers, booleans, and `null` are visually distinct, making structured data faster to scan.

## Motivation
Before this change JSON appeared as dense, monochrome text. That made it hard to spot important fields, differentiate value types, or quickly understand nested objects. Highlighting improves human parsing speed without adding heavy dependencies.

## Changes
- JSON detection (quick structural check + safe `JSON.parse`) added to the cell formatter in `data_mutators.ts`.
- Introduced a lightweight regex-based tokenizer `buildJsonHtml` to wrap:
  - Keys (`"key":`)
  - Strings
  - Numbers
  - Booleans
  - `null`
- Added monospace and color classes (`.json-key`, `.json-string`, `.json-number`, `.json-boolean`, `.json-null`) in `TableTable.vue`.
- Refactored logic into reusable helper `maybeHighlightJson`.
- Truncation safeguards for very large JSON values.
- Non-JSON values continue to render exactly as before.

## Implementation Notes
- Detection heuristic: starts with `{` or `[` and ends with `}` or `]`; only then attempts `JSON.parse`.
- Tokenizer uses ordered regex replacements to avoid interference between key and string matching.
- Formatting collapses multi-line pretty output to a single line for cell density.
- No external parser added—keeps bundle size and render cost low.

## Performance
- Overhead per cell is minimal (single parse + regex replaces for detected JSON).
- Large payloads truncated to prevent DOM bloat.
- Fallback path (invalid JSON) is a simple escaped string.

## Testing / Verification
- Manually validated with nested objects, arrays, numbers, booleans, null values.
- Confirmed malformed or partial JSON strings fall back without throwing.
- Ensured non-JSON columns unaffected.
- Existing unit tests touched/updated where necessary (e.g., `data_mutators.spec.js`).

## Related Files
- `apps/studio/src/mixins/data_mutators.ts`
- `apps/studio/src/components/tableview/TableTable.vue`
- `apps/studio/tests/unit/mixins/data_mutators.spec.js`

## Checklist
- [x] Inline JSON highlighting
- [x] Reusable tokenizer helper
- [x] Safe detection & fallback
- [x] Styling classes added
- [x] Large value truncation
- [x] Existing non-JSON unaffected

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -303,30 +303,16 @@
 .material-icons.menu-icon {
   margin-left: 10px !important;
 }
-/* JSON syntax highlighting */
-.json-cell { white-space: pre; font-family: var(--font-monospace, monospace); }
-.json-tooltip { 
-  font-family: var(--font-monospace, monospace); 
-  max-width: 600px; 
-  max-height: 400px;
-  overflow: auto; 
-  background: #1e1e1e; 
-  color: #d4d4d4; 
-  padding: 8px 10px; 
-  border: 1px solid #333; 
-  border-radius: 4px; 
-  line-height: 1.25; 
-  box-shadow: 0 2px 6px rgba(0,0,0,.4);
-}
-/* Light theme fallback (detect if body has light class if such convention exists) */
-body.light .json-tooltip { background: #fafafa; color: #333; border-color: #ccc; box-shadow: 0 2px 6px rgba(0,0,0,.2); }
-.json-type .json-key, .json-key { color: #c792ea; }
-.json-type .json-string, .json-string { color: #ecc48d; }
-.json-type .json-number, .json-number { color: #d8dee9; font-weight: bold; }
-.json-type .json-boolean, .json-boolean { color: #82aaff; font-weight: bold; }
-.json-type .json-null, .json-null { color: #ff5370; font-style: italic; }
-/* Keep pre tags selectable */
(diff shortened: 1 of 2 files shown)


azmy60:
Hey there, thanks for the PR! I really like the idea, but I don’t think this is the best approach. Since we use codemirror in JSON sidebar, I’m pretty sure that the text is already tokenized as JSON, so we only need to add/modify the css.

marcosstefani:
> Hey there, thanks for the PR! I really like the idea, but I don’t think this is the best approach. Since we use codemirror in JSON sidebar, I’m pretty sure that the text is already tokenized as JSON, so we only need to add/modify the css.

done. thanks.

not-night-but:
@azmy60 can you review?

azmy60:
@marcosstefani Sorry for the late reply. I think I misunderstood the goal of the PR. I was expecting the JSON Sidebar to be affected, but it seems like we want to colorize the table cells that are JSON.

First of all, I couldn't run the PR because it's missing a file:

2025-11-12-210629_716x222_scrot

Second, it'd be really helpful for us if there is a screenshot of how it works. That way, we understand how this PR affects the table view.

But looking at the motivation, it could be easily solved by using [JSON Sidebar](https://www.beekeeperstudio.io/features/json-sidebar), no?

PR #3648, opened 2025-11-01
Feature/2137 super formatter

### LINKED ISSUE #2137: FEAT: Super Query Formatter 🚀🚀
Right now we only have a single query formatting preset, but the library supports a whole bunch of formatting options.

Let's make a cool formatting modal & let users format to their hearts content.

This should be available in the community edition, but saving formatting presets should be ultimate-only

Configuration options - https://github.com/sql-formatter-org/sql-formatter?tab=readme-ov-file#configuration-options

Super Format

wmontgomery:
super-formatter-1
resolves #2137 

EDIT: Moved open formatter option to more actionable spot that's less in the way
image

EDIT: Buttons and sliders have been updated
image

EDIT: Add formats and open formatter to context menu
image

### DIFF
--- a/apps/studio/src/common/transport/index.ts
+++ b/apps/studio/src/common/transport/index.ts
@@ -78,13 +78,9 @@ export interface TransportHiddenEntity extends Transport {
   workspaceId: number
 }
 
-/**
- * Types below from apps/studio/src/common/appdb/models/FormatterPreset.ts because
- * "importing from appdb models in anything that is imported in the frontend (this can cause issues with production builds unfortunately)"
- */
 type CaseOption = "preserve" | "upper" | "lower";
 type LogicalOperatorNewlineOption = "before" | "after";
-type FormatterPresetConfig = {
+export type FormatterPresetConfig = {
   tabWidth: number;
   useTabs: boolean;
   keywordCase: CaseOption;
--- a/apps/studio/src/common/appdb/models/FormatterPreset.ts
+++ b/apps/studio/src/common/appdb/models/FormatterPreset.ts
@@ -2,27 +2,9 @@ import { Column, Entity } from "typeorm"
 import { ApplicationEntity } from "./application_entity"
 import _ from "lodash"
 import rawLog from "@bksLogger"
-import { TransportFormatterPreset } from "@/common/transport"
+import { FormatterPresetConfig, TransportFormatterPreset } from "@/common/transport"
 
 const log = rawLog.scope("FormatterPresets")
-
-/**
- * If changing the below cases and types, also update them in apps/studio/src/common/transport/index.ts
- */
-type CaseOption = "preserve" | "upper" | "lower";
-type LogicalOperatorNewlineOption = "before" | "after";
-export type FormatterPresetConfig = {
-  tabWidth: number;
-  useTabs: boolean;
-  keywordCase: CaseOption;
-  dataTypeCase: CaseOption;
-  functionCase: CaseOption;
-  logicalOperatorNewline: LogicalOperatorNewlineOption;
-  expressionWidth: number;
-  linesBetweenQueries: number;
-  denseOperators: boolean;
-  newlineBeforeSemicolon: boolean;
-};
 
 export type FormatterPresetValues = {
   name: string,


apps/ui-kit/lib/components/super-formatter/props.ts
```diff
@@ -0,0 +1,32 @@
+import { PropType } from "vue"
+import { FormatOptions } from "sql-formatter"
+
+export default {
+  value: String,
+  canAddPresets: Boolean,
+  clipboard: Object,
```
azmy60: I would use the [Clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard) interface for typing the `clipboard`. We also use this type in the [TextEditor](https://github.com/beekeeper-studio/beekeeper-studio/blob/ef6a2bff9286435a403617fe997c8b73fd7b586b/apps/ui-kit/lib/components/text-editor/props.ts#L103) component, so it would be more consistent if we use the same type. :)

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -221,6 +224,49 @@
...
+            >
+              X
+            </button>
+          </div>
+          <bks-super-formatter
+            :value="unsavedText"
+            :formatter-dialect="formatterDialect"
+            :identifier-dialect="identifierDialect"
+            :can-add-presets="true"
+            :clipboard="$native.clipboard.writeText"
```
azmy60: And then here, we can do `:clipboard="$native.clipboard"`.

apps/ui-kit/docs/api/super-formatter.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Properties
+
+| Name               | Type                     | Description                                                                                                                                       | Default                  |
+| ------------------ | ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------ |
+| `value`            | `string`                 | The SQL text to format and preview.                                                                                                               | `''`                     |
+| `canAddPresets`    | `boolean`                | Enable preset management features (add, save, delete). When `false`, only formatting controls are shown.                                          | `false`                  |
+| `clipboard`        | `object`                 | Clipboard handler for copy functionality. Must implement `writeText(text: string)` method.                                                        | `undefined`              |
```
azmy60: If the clipboard type is updated, be sure to update this too :)

apps/ui-kit/lib/components/super-formatter/types.ts
```diff
@@ -0,0 +1,7 @@
+export type SuperFormatterValueChangeEvent = CustomEvent<{
+  value: string;
+}>
+
+export interface SuperFormatterEventMap extends HTMLElementEventMap{
+  "bks-query-selection-change": SuperFormatterValueChangeEvent;
```
azmy60: I don't see `bks-query-selection-change` in `<SuperFormatter>`. It might've been incorrect? Maybe here we want to type `bks-create-preset`, `bks-delete-preset`, `bks-apply-preset`, and `bks-save-preset`?

azmy60 (changes requested):
Having the preview text updates in realtime is really nice!

My concern would mostly be on the UI part. For the config, I would suggest looking at the AI Shell config screen. I think it just needs some tweaking on the "spacing" and button colors (which buttons are primary and secondary?).

Also, it seems like we are adding a new "settings" button at the top of the editor. That means we have 2 "settings" buttons, one at the top, and one at the statusbar. That might be a bit confusing for users since it's not immediately clear what each one is about. And, I do realize that the status bar settings is way too far from the text editor (this could be bad because of [proximity](https://www.nngroup.com/articles/gestalt-proximity/#:~:text=Summary%3A%C2%A0%20Design%20elements%20near%20each%20other%20are%20perceived%20as%20related%2C%20while%20elements%20spaced%20apart%20are%20perceived%20as%20belonging%20to%20separate%20groups.)).

This will sound like killing two birds with one stone, and I might be wrong, but I think it might be better if both settings buttons are combined and placed right under the text editor like this:

Image

So that will:
1. Solve the proximity issue (of the settings in the status bar)
2. Make the text-editor related settings in one group, instead of two
3. Not hide the first line of the editor because there is a button on top of it

apps/ui-kit/lib/components/super-formatter/types.ts
wmontgomery: Oh this doesn't actually do anything to the superformatter. I'm going to take it out but add the other ones

wmontgomery:
> Having the preview text updates in realtime is really nice!
> 
> My concern would mostly be on the UI part. For the config, I would suggest looking at the AI Shell config screen. I think it just needs some tweaking on the "spacing" and button colors (which buttons are primary and secondary?).
> 
> Also, it seems like we are adding a new "settings" button at the top of the editor. That means we have 2 "settings" buttons, one at the top, and one at the statusbar. That might be a bit confusing for users since it's not immediately clear what each one is about. And, I do realize that the status bar settings is way too far from the text editor (this could be bad because of [proximity](https://www.nngroup.com/articles/gestalt-proximity/#:~:text=Summary%3A%C2%A0%20Design%20elements%20near%20each%20other%20are%20perceived%20as%20related%2C%20while%20elements%20spaced%20apart%20are%20perceived%20as%20belonging%20to%20separate%20groups.)).
> 
> This will sound like killing two birds with one stone, and I might be wrong, but I think it might be better if both settings buttons are combined and placed right under the text editor like this:
> Image
> 
> So that will:
> 
>     1. Solve the proximity issue (of the settings in the status bar)
> 
>     2. Make the text-editor related settings in one group, instead of two
> 
>     3. Not hide the first line of the editor because there is a button on top of it

What I'll do is drop the setting gear in the top right and add a button to open the formatter next to the run and save pieces. It's still largely format heavy and moving the gear with VIM and all that feels like a heavy move for this. Would be happy to tackle in another PR as its own thing. If people are using that gear, I have some qualms moving it since users are used to where it is for now.

azmy60:
Yeah that seems fair. Just be aware that there'll be new buttons there as well from the manual commit PR.

azmy60 (approved):
LGTM!

azmy60:
If I'm understanding this correctly: another idea is instead of a button that opens the Super Formatter, it could be a button that says "Format Query" that will **format** the query with the default format, and it also has a dropdown menu so you can select which preset to use, and the last item will be "Open Query Formatter". I don't know how much effort that'll take, but I think the current one is good regardless. :)

apps/studio/src/common/transport/index.ts
```diff
@@ -1,3 +1,4 @@
+import { FormatterPresetConfig } from "../appdb/models/FormatterPreset";
```
not-night-but: We shouldn't be importing from appdb models in anything that is imported in the frontend (this can cause issues with production builds unfortunately)

wmontgomery: Oops! Didn't know that. I'll handle this on Friday

wmontgomery: Oh main site docs! Drat I forgot about those entirely. Going to add them.

PR #3649, opened 2025-11-01
Feature/ilike to filters

wmontgomery:
ilike


apps/studio/src-commercial/backend/lib/db/clients/mongodb.ts
```diff
@@ -1058,6 +1059,7 @@ export class MongoDBClient extends BasicDatabaseClient<QueryResult> {
       "=": "$eq",
       "!=": "$ne",
       "like": "$regex", // special case for regex
+      "ilike": "$regex", // special case for regex
```
not-night-but, line 1062: We may need to look into this a bit further, as ideally they should behave how they behave in sql (ie like is case sensitive, ilike is case *in*sensitive

apps/studio/src/components/tableview/filter/BuilderFilter.vue
```diff
@@ -80,6 +78,23 @@ export default {
...
+          iLikeFilterSymbol,
+          ...secondHalf
+        ]
+      } 
+
+      return [
+        ...firstHalf,
+        ...secondHalf
+      ]
+    },
```
not-night-but, line 86: Something about this is bugging me, just some random spidey senses or something. I wonder if it would be better to just put a dialect support array on the filter symbol, and then just do a v-if in the template? I'm really not sure where the support filter should really be, I can just imagine this getting really annoying if we end up with a few more filters that aren't supported on every db

apps/studio/src/lib/db/types.ts
```diff
@@ -33,19 +33,27 @@ export const keymapTypes = [
   { name: "Vim", value: "vim" }
 ] as const
 
+export const iLikeFilterSymbol = {
+  value: 'ilike', label: 'ilike'
+}
+
 // if you update this, you may need to update `translateOperator` in the mongodb driver
 export const TableFilterSymbols = [
```
not-night-but, line 37: Okay actually we could just add an array here like `requiresFeature` which is just the name of a property on supported features. So it could be undefined/empty on everything else, but `[ 'iLike' ]` on the ilike filter, and then just use that to determine what to show if that makes sense?

not-night-but (changes requested):
Looks good, just some questions. I'm not sure how maintainable the filtering of the filter symbols is, I made a couple suggestions, I think having a `requiresFeature` field is probably the best option but if you disagree, feel free to flame me

apps/studio/src-commercial/backend/lib/db/clients/mongodb.ts
wmontgomery: I can get behind that. If I remember it right, it's an option in mongo to say case insensitive or not. I can add that as part of it. I guess with that, might have to look into equals or others like that if they are case insensitive by default? Not totally sure if Mongo wants people going like or ilike.

apps/studio/src/components/tableview/filter/BuilderFilter.vue
wmontgomery: I was thinking something similar but decided for the time being this was ok. My thought was to have the full list of the symbols, then maybe have each dialect build its own list, but then was getting worrying about consistent order, but then a possible solution for that too. I can add a solution to this

apps/studio/src-commercial/backend/lib/db/clients/mongodb.ts
wmontgomery: I make like and ilike case sensitive and case insensitive respectively. I'm not sure what the right way to go is with it. I'm still not sure how many people use Mongo with SQL vs the query language built in 🤷

apps/studio/src/components/tableview/filter/BuilderFilter.vue
wmontgomery: OK, went a route where filter types are either standard or different type (like "ilike") so new ones can be grouped and consistently ordered. All clients will have a "standard" type they use and add different types to the supportedFeatures.

not-night-but (changes requested):
I think we need some tests and then this should be good to go

PR #3657, opened 2025-11-07
try catch deparameterize

### LINKED ISSUE #3654: BUG: Clickhouse (maybe others) fail with string-quoted table syntax
**Describe the bug**
Putting any query in the query UI which has this "string quoted table" syntax, like the following:
```
SELECT * FROM "default"."test"
```

fails to parse.

Image

**To Reproduce**

1. Connect to clickhouse (not sure if this applies to other databases universally or not)
2. Enter the above query, like in the screenshot
3. Run

**Expected behavior**
Should not error, as this is valid clickhouse syntax

**Screenshots / Gifs**
See above

 - OS: Arch Linux
 - App Version: 5.4.9
 - Database type and version: Clickhouse 25.10.1.3832

**Additional context**

azmy60:
fixes #3654 

The error was thrown by `sql-formatter`. Simply adding try-catch solves the linked issue but I'm not sure if there are parts that depend on deparameterized query. Two things that we probably want ask ourselves IMO:

1. Can we fix this from `sql-formatter`?
2. If deparameterizing fails, do we have a way to anticipate it? does it still work normally?

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -566,6 +566,9 @@
 
         try {
           const placeholders = this.individualQueries.flatMap((qs) => qs.parameters);
+          if (_.isEmpty(placeholders)) {
+            return query;
+          }
           const values = Object.values(this.queryParameterValues) as string[];
           const convertedParams = convertParamsForReplacement(placeholders, values);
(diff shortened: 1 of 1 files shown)


not-night-but:
@azmy60 good to go, but maybe we should merge this into rc since a few people have had issues with this? Idk up to you

azmy60:
@not-night-but Yeah, that make sense.

PR #3660, opened 2025-11-09
Feature/record count on filter

### LINKED ISSUE #3666: when we filter a table we should show the filtered record count
Currently we show the total record count, regardless of filters.

Probably we only enable this for specific databases (that support paging). It should only populate on click - just like the total record count.

wmontgomery:
Resolves #3666 
filter-count

### DIFF
--- a/apps/studio/src/common/menus/MenuItems.ts
+++ b/apps/studio/src/common/menus/MenuItems.ts
@@ -151,7 +151,7 @@ export function menuItems(actionHandler: IMenuActionHandler, settings: IGroupedU
     },
     importSqlFiles: {
       id: 'import-sql-files',
-      label: "Import SQL Files",
+      label: "Import Saved Queries",
       accelerator: "CommandOrControl+I",
       click: actionHandler.importSqlFiles,
       showWhenConnected: true,
@@ -183,7 +183,7 @@ export function menuItems(actionHandler: IMenuActionHandler, settings: IGroupedU
       id: 'menu-secondary-sidebar',
       label: 'Toggle Secondary Sidebar',
       // accelerator: "Alt+S",
-      click: actionHandler.toggleSecondarySidebar,  
+      click: actionHandler.toggleSecondarySidebar,
       enabled: false,
     },
     themeToggle: {


apps/studio/src/components/common/TableLength.vue
```diff
@@ -35,9 +37,19 @@ export default Vue.extend({
...
       this.fetchingTotalRecords = true
+      
       try {
+        const allFilters = []
         this.error = null
-        this.totalRecords = await this.connection.getTableLength(this.table.name, this.table.schema);
+        if (Array.isArray(this.filters) && this.filters.length > 0 ) {
+          for (const filter of this.filters) {
+            allFilters.push(await this.connection.getQueryForFilter(filter))
+          }
```
not-night-but, line 47: Could we not do this in the backend? Just pass the filters in getFilteredDataCount and do it there?

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
```diff
@@ -596,6 +596,30 @@ export abstract class BasicDatabaseClient<RawResultType extends BaseQueryResult>
...
+  async getFilteredDataCount(table: string, schema: string | null, filter: string ): Promise<string> {
+    if (!this.knex) {
+      return ''
+    }
+
+    try {
+      const query = await this.knex(schema ? `${schema}.${table}` : table)
+        .count('*')
+        .whereRaw(filter)
+        .toString()
```
not-night-but, line 608: We do have a couple dialects that don't have working knex dialects (surreal, and mongo for sure). So we probably need a backup for this.

apps/studio/src/components/common/TableLength.vue
```diff
@@ -35,9 +37,19 @@ export default Vue.extend({
   methods: {
     async fetchTotalRecords() {
```
not-night-but, line 38: I personally think this should be separate from total records. Maybe that's stupid?

wmontgomery: I guess that's a question for what " total" is in this case. Is it the full table's amount or whatever the filter is. I think it's more clear to have it as totalrecords because I'm seeing it as "total records I'm showing from this here query"

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
wmontgomery: There is a check for if knex exists or not above to make show "unknown" if I had to guess. I can add a `getFilteredDataCount` function to each of those that don't have knex

apps/studio/src/components/common/TableLength.vue
wmontgomery: I grabbed this from the [rowfilterbuilder](https://github.com/beekeeper-studio/beekeeper-studio/blob/84b6cae1d19ade16627fd2bce6b343f14bd50b2f/apps/studio/src/components/tableview/RowFilterBuilder.vue#L253). Probably could be on the backend. Maybe as a todo or you think I should tackle it in both places with this PR?

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
not-night-but: Doh, What was I on when I wrote this

apps/studio/src/components/common/TableLength.vue
not-night-but: Lets add it as a todo methinks. As long as this works for now we should be good

not-night-but (approved):

PR #3662, opened 2025-11-09
[3661] Added ScyllaDB docs and runner

andrii-kysylevskyi:


not-night-but:
@rathboma If this is a drop in, should we subclass it like we did with `TIDB`? Could add tests for it as well

azmy60:
@not-night-but If it's a drop in, then I don't see why SycllaDB is an exception. At least, there is a clear way (in the UI) for users to discover that SycllaDB is supported, e.g. via connection type dropdown. I don't think adding this to the docs or readme is enough.

andrii-kysylevskyi:
Hi @azmy60 and @not-night-but - thank you for your input!

Just to confirm, I understand correctly - you'd like me to:

1. Create a ScyllaDB subclass that extends the Cassandra client (similar to how TiDB extends MySQL)
2. Add ScyllaDB to the connection type dropdown so users can discover it in the UI rather than just through documentation
3. Add tests for the ScyllaDB integration

If that's right, I'll implement those changes and update this PR with a link to the commit. Otherwise, happy to adjust based on your suggestions.

Thanks!

andrii-kysylevskyi:
Hi @not-night-but and @azmy60,

Thanks for the feedback! I've implemented the changes as mentioned in my previous comment:

1. Created a ScyllaDB subclass that extends the Cassandra client (following the TiDB/MySQL pattern)
2. Added ScyllaDB to the connection type dropdown so users can discover it in the UI
3. Updated UI components to show Cassandra-specific terminology (e.g., "Keyspace") for ScyllaDB connections
4. Added an integration test for ScyllaDB

**Test-related changes:**

There were no Cassandra tests before, so I had to make some additional changes to the existing setup.

- tests/integration/lib/db/clients/scylladb.spec.js - New integration test using scylladb/scylla Docker image
- tests/lib/db.ts - Added cassandra and scylladb to KnexTypes mapping, and set knex = null for these clients (Cassandra-based DBs don't use the standard knex setup like SQL databases)

Note on tests: The integration test spins up a ScyllaDB Docker container via testcontainers, so it takes a while to run locally. It should run automatically in CI based on your existing integration test infrastructure. The test connects to ScyllaDB, verifies the connection, and checks that system keyspaces are listed.

All unit tests pass. Let me know if you'd like any adjustments!

[The commit](https://github.com/beekeeper-studio/beekeeper-studio/pull/3662/commits/bf152f3574aedaf87c375657c2eeef4655e63be4) with the changes just in case

rathboma:
Pushed a fix on top of your changes — the integration test had the jest timeout dropped to 5s right before a 15s sleep, so `beforeAll` was timing out before it ever connected. Replaced the sleep with a `Wait.forLogMessage` strategy (matching the pattern the other DB integration tests use), pinned the image to 2025.1.4 to match docker-compose, and added `--developer-mode 1` so Scylla is happy on CI runners. Passes locally in ~12s.

One note on CI: the `E2E tests` job is also currently red on master, so that failure isn't coming from this PR — safe to ignore for this review.

rathboma (approved):

PR #3674, opened 2025-11-13
fix: fix for mongoshell getting stuck when autocompletion pops up

itsamrit:
Issue: https://github.com/beekeeper-studio/beekeeper-studio/issues/3536

### DIFF
--- a/apps/studio/src/lib/editor/extensions/mongoHint.ts
+++ b/apps/studio/src/lib/editor/extensions/mongoHint.ts
@@ -37,10 +37,16 @@ async function completionSource(
 
   const cmd = context.state.doc.sliceString(start, end);
 
-  const completions = await getCompletionsFn(cmd);
+  let completions: string[];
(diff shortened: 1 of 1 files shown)


itsamrit:
@azmy60 please review

apps/studio/src/lib/editor/extensions/mongoHint.ts
```diff
@@ -39,6 +39,10 @@ async function completionSource(
 
   const completions = await getCompletionsFn(cmd);
```
azmy60: While we're at it, could we wrap `getCompletionsFn` in a try-catch and return `null` if it fails? Don't want this to block the shell too. Thanks!

itsamrit: sure, have done the changes..

azmy60 (approved):

PR #3678, opened 2025-11-14
Fix readonly mode

not-night-but:
This does a few things:
1. Re-adds the readonly mode guards we used to have (idk where they went) so even if a user somehow has a connection with readonly mode enabled, but they are on the community version, it will still run all queries.
2. Add a validator and a transformer that will detect if the user is on community and a connection has readonly mode enabled and flip it. So any connection saved will just silently flip it off, as well as when fetching from the db it will turn it off if the user can't use it.
3. Update the many places in the UI that do write actions so that they just aren't accessible in read only mode (why even allow them to add a row and click apply if it's just gonna yell at them for it)

### DIFF
--- a/apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
+++ b/apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
@@ -85,8 +85,9 @@ export abstract class BasicDatabaseClient<RawResultType extends BaseQueryResult>
   }
 
   async checkAllowReadOnly() {
+    if (platformInfo.testMode) return true;
     const status = await LicenseKey.getLicenseStatus()
-    return status.isUltimate || platformInfo.testMode;
+    return status.isUltimate;
   }
 
   set connectionHandler(fn: (msg: string) => void) {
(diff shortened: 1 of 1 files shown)


apps/studio/src/common/appdb/validators/ReadOnlyOrDefault.ts
```diff
@@ -0,0 +1,31 @@
...
+        async validate(value: any, args: ValidationArguments) {
+          const status = await LicenseKey.getLicenseStatus();
+          const canSaveReadOnly = status.isUltimate || platformInfo.testMode;
+
+          if (canSaveReadOnly) {
+            return true;
+          }
+
+          if (value === true) {
+            (args.object as any).readOnlyMode = false;
```
azmy60, line 25: Isn't this hacky? It'd be awesome if we put this somewhere else for predictable outcome. I think TypeORM should have a way to do that? Like a setter or something. Otherwise, I guess it's fine for now.

azmy60 (dismissed):

PR #3681, opened 2025-11-17
test front end plugin manager and fix getTables and getColumns

azmy60:


azmy60:
`getTables` and `getColumns` do not respect `schema` parameter. This causes AI Shell to fail to load columns for some tables.

PR #3703, opened 2025-11-27
doc: AUR (now Arch Linux) and Flatpak installation instructions

103sbavert:
- changed AUR section to clarify package format
- changed the title to "Arch Linux (and derivatives)
- fixed a typo where Flatpak was packages were called "AUR" packages

Note: While I don't know exactly which package format is an "AUR package", especially one ending in file extension `.pacman` (AUR packages typically come as PKGBUILD manifests), looking into the actual `.pacman` file in the releases tells me these are simply 'local' packages as the [Wiki calls them](https://wiki.archlinux.org/title/Pacman#Additional_commands). Local packages are not specific to AUR, they're simply local packages and AUR helps as a centralized repository for commonly needed "local" (or custom) packages. They are not maintained by the Arch Linux team, so they are largely [chaotic](url).

### DIFF
--- a/docs/installation/linux.md
+++ b/docs/installation/linux.md
@@ -68,15 +68,15 @@ sudo dnf install beekeeper-studio
 sudo yum install beekeeper-studio
 ```
 
-## AUR
+## Arch Linux (and derivatives)
 
-AUR files are provided for both x86_64 and ARM64 systems, you can download them from [the latest release](https://github.com/beekeeper-studio/beekeeper-studio).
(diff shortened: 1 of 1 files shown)


not-night-but:
Oh yeah I didn't realize we had that in there. Thanks for the correction!

not-night-but (approved):

103sbavert:
Thanks, also just realized we're using org.freedesktop.Platform/x86_64/23.08 for Flatpak which is now EOL.

Issue #1403, opened 2022-10-18
FEAT: horizontal scroll on tabs using mouse


mcanepa:
If there are too many tabs, I would expect two things

1) Each of them to have a minimum with so at least part of it's name gets shown. See #1401

2) To see a certain number of tabs and scroll left or right with mouse wheel in order to access the rest

loomchild:
Yeah, it would be great. Otherwise, the workspace eventually becomes unusable and the only solution is to close all tabs and start from scratch:

![Image](https://github.com/user-attachments/assets/434378ea-9ecd-4547-8048-8702a691d436)

mcanepa:
Please see [https://github.com/beekeeper-studio/beekeeper-studio/pull/3706#issuecomment-5269483713](https://github.com/beekeeper-studio/beekeeper-studio/pull/3706#issuecomment-5269483713)

@rathboma @not-night-but @azmy60

Issue #1401, opened 2022-10-18
FEAT: minimum width for tabs


mcanepa:
If I open too many tabs, I reach a point where I no longer know what they are about (see screenshot)

They shoud have a minimum width so they can display it's name

![Captura de pantalla de 2022-10-18 12-34-04](https://user-images.githubusercontent.com/25961439/196495841-796ff3ba-8128-464e-b308-0785fc30b6b2.png)

rathboma:
Yes, this is something we have to deal with. We previously had an overflow area, but then it wasn't obvious how many tabs were open, so this was the compromise for the time being.

We're probably not going to substantially change this anytime soon, but if someone creates a PR with a solution they think is better I'd be happy to consider it.

mcanepa:
Thanks for your feedback

> but then it wasn't obvious how many tabs were open

I think this should be done with #1403 in mind, so the whole thing looks something like this:

![65159595-07c93d80-da3d-11e9-9e47-159c9fc3693f](https://user-images.githubusercontent.com/25961439/196511866-59333022-1889-4f19-b597-fa45c9886429.gif)

vespadj:
`tabs-header > nav`, tabs should be in more than one line.
This can be rapid tested by edit 
`<ul class="nav-tabs nav">`
to
`<ul class="nav">`

at

`#page-content > div > div.tabs-header > ul`
XPath: `/html/body/div/div[1]/div[2]/div[1]/div[4]/div/div[1]/ul`

some cosmetic fixes of the CSS class will be required.

<img width="1914" height="488" alt="Image" src="https://github.com/user-attachments/assets/6cadc624-7d2a-417b-acdc-817a3cb88594" />

mcanepa:
the multiline idea is horrible UX

PR #3706, opened 2025-11-29
Allow scrolling on tabbar

lutostag:
When more tabs are created than fit in the tabbar, allow scrolling and hide the scrollbar on the tabstrip.

Also bumps up minimal tab width so icons can be seen (and hovered for knowing which tab to jump to).

Addresses #1403 and #1401.

To apply to an already running beekeeper instance for testing (gets close, does not include the webkit-scrollbar hiding):
`Help > Show Developer Tools > Console`
`allow pasting`
```css
document.querySelector('.nav-tabs').style.overflowY = 'hidden';
document.querySelector('.nav-tabs').style.overflowX = 'auto';
document.querySelector('.nav-tabs').style.whiteSpace = 'nowrap';

[...document.querySelectorAll('.nav-tabs .nav-item-wrap')]
  .forEach(el => el.style.minWidth = '50px');
```

### DIFF
--- a/apps/studio/src/assets/styles/app/core-tabs.scss
+++ b/apps/studio/src/assets/styles/app/core-tabs.scss
@@ -64,13 +64,18 @@
   }
   .nav-tabs {
     position: relative;
-    overflow: hidden;
+    overflow-y: hidden;
+    overflow-x: auto;
+    white-space: nowrap;
     flex-wrap: nowrap;
     margin: 0;
     height: $tab-height;
     -webkit-app-region: no-drag;
+    &::-webkit-scrollbar {
+      display: none;
+    }
     .nav-item-wrap {
-      min-width: 28px;
+      min-width: 50px;
       overflow: hidden;
     }
     .nav-item-wrap-chosen {


azmy60 (changes requested):
Thanks for the PR! Could we make this as an opt-in feature for now? We have [configuration](https://docs.beekeeperstudio.io/user_guide/configuration/) feature that can support that. Please see [this](https://github.com/beekeeper-studio/beekeeper-studio/commit/0574c69064dde6cb51d0a5458749386de25a0d96) to learn how to implement it. Thanks!

rathboma:
So one thing this doesn't do is keep the active tab in view by default when you open a new tab or zoom to the tab when you switch to it. I think stuff like this is the real outstanding work. @lutostag you still game for working on this?

mcanepa:
I've tried the proposed solution but I get the UI shown in the picture

imagen

50px is not enought, it should be at least 100px.... 125px would be even better, but that horizontal scroll is very annoying!! I guess it should behave like vscode, where I can scroll with the mouse wheel with no srollbar visible.


[vs_tabs_scroll.webm](https://github.com/user-attachments/assets/51d93d2f-4a49-4bed-994b-2ea5b4844233)

azmy60:
@mcanepa Having an invisible scroll bar is a common practice so that's fair.

Issue #1342, opened 2022-08-13
BUG:tryed to install key on ubuntu 22.04.1


billykd4dpb:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots / Gifs**
If applicable, add screenshots or GIFS to help explain your problem.

**Version Information (please complete the following information):**
 - OS: [e.g. Windows/Mac/Ubuntu]
 - App Version [e.g. help -> About Beekeeper Studio]
 - Database type and version [eg Posgresql 9.3]

**Additional context**
Anything else you think might be helpful

leogm360:
Hi billykd4dpb, I had the same problem on Ubuntu 22.04 LTS. The problem is that Ubuntu deprecated managing integrity keys directly by apt, now any key should be gpg and be managed at trusted keyring (at /usr/share/keyrings/). I've made a solution that work while there's no official solution:


```shell
 #!/usr/bin/env bash

# create a gpg keyring file for beekeeper in trusted gpg keyring (you can choose whatever name)
sudo touch /usr/share/keyrings/apt.beekeeper.io.gpg

# download ascii beekeeper key, dearmor ascii beekeeper key with gpg and send it to trusted keyring (here you use the file you created)
curl https://deb.beekeeperstudio.io/beekeeper.key | gpg --dearmor | sudo tee /usr/share/keyrings/apt.beekeeper.io.gpg >/dev/null

# add beekeeper repository to apt list index, but specify that the signature comes from trusted gpg kering
echo "deb [signed-by=/usr/share/keyrings/apt.beekeeper.io.gpg arch=amd64] https://deb.beekeeperstudio.io stable main" | sudo tee /etc/apt/sources.list.d/beekeeper-studio-app.list

# update apt index and install beekeeper
sudo apt update && sudo apt install beekeeper-studio -y
```

I hope it helps.

rathboma:
Thanks! We'll fix this using the code you provided! It's already updated in the docs, but needs to be updated in the `postinstall.js` file also

PR #3713, opened 2025-12-03
switch apt source to deb822 .sources and specify signed-by key inside it rather…

Skgland:
… than using deprecated apt-key

The follow-up of https://github.com/beekeeper-studio/beekeeper-studio/pull/1785#issuecomment-1839185386 that never was. 
Related to https://github.com/beekeeper-studio/beekeeper-studio/issues/1342 and https://github.com/beekeeper-studio/beekeeper-studio/issues/1218

### DIFF
--- a/apps/studio/build/deb-postinstall
+++ b/apps/studio/build/deb-postinstall
@@ -19,7 +19,7 @@ if [ -f $OLD_FILE ]; then
   rm -f $OLD_FILE.save
 fi
 
-if [ ! -f /usr/bin/beekeeper-studio ]; then
+if [ ! -e /usr/bin/beekeeper-studio ]; then
   ln -s /opt/Beekeeper\ Studio/beekeeper-studio /usr/bin/beekeeper-studio
 fi
 


rathboma:
Hey! This is interesting. Are .sources files widely supported?

rathboma:
If not, how do we detect that?

Looks like this is available in apt 1.1+, so I guess we could just check for that?

1.1 was released in 2015, but not sure when distros added support. Probably this is fine, but wouldn't hurt to have both old and new, with a switch based on apt version

Skgland:
@rathboma something like this?

Skgland:
Ups, just read in the man pages I linked that embedded keys aren't supported prior to apt 2.4 so switched to always use a separate key file even for deb822 .sources files.

apps/studio/build/deb-postinstall
```diff
@@ -23,10 +23,8 @@ if [ ! -f /usr/bin/beekeeper-studio ]; then
   ln -s /opt/Beekeeper\ Studio/beekeeper-studio /usr/bin/beekeeper-studio
```
Skgland, line 23: When testing these script changes I noticed that I get a warning from `ln` that the file already exists. I think the check should use `-e` instead of `-f`.

rathboma: oh, I guess /usr/bin/beekeeper-studio is a symlink now. Yes that will need to use `-e`. Good find! Feel free to update in this PR also.

Legend :star:

Skgland: Done

Skgland:
superseded by #3917

Issue #3607, opened 2025-10-15
DOCS: help link in beekeeper for redshift cluster credentials is broken - macOS


mtesch-um:
Documentation requests will be triaged separately from features, so someone should get to it pretty quickly!

The button `Beekeeper Docs` in the amazon redshift database setup to get help leads to a 404:

<img width="527" height="132" alt="Image" src="https://github.com/user-attachments/assets/ea02d422-cca0-49fb-a9ce-f89f9d4869c5" />

points at https://docs.beekeeperstudio.io/user_guide/connecting/amazon-rdb?ref=bks-app <- which doesn't exist

not sure what it should be - still looking for the docs :)

davidkaufman:
https://docs.beekeeperstudio.io/user_guide/connecting/amazon-rds/

mtesch-um:
@davidkaufman close... that's for RDS though - my error was for a Redshift connection (which uses a different API than RDS called `GetClusterCredentials`) -- the left link in that image `Amazon Docs` correctly points at https://docs.aws.amazon.com/redshift/latest/mgmt/generating-user-credentials.html?ref=bks-app - but the Beekeeper Docs 404 ...

not-night-but:
@mtesch-um I actually think Dave is right here. I think that is the page that was intended to be linked. Not saying it's right, I think we should probably just have a docs page specifically for IAM auth that takes you through the process. 

edit: I am like 100% sure there was supposed to be a docs page for this, whoever ended up implementing that feature must've forgotten it. I'm doing a bit of a rework of our IAM auth right now, so I'll make sure to add some better docs for this. In the meantime I'm not sure if we actually have a page other than the rds page that meantions IAM at all 🤦

mtesch-um:
@not-night-but yeah, I'm just looking for directions on how to setup IAM for Redshift, not RDS 👍  - let me know if there are new instructions!  We would buy seats (5+) if the redshift iam authentication would work for us.  Use case is iam identity center SSO authentication with an aws policy allowing individual users to GetClusterCredentials for a redshift user matching their iam user.  The GetClusterCredentials are only valid for 60 mins, but allow a refresh, so auto-refresh should be built in.

PR #3716, opened 2025-12-05
DRAFT: aws sso rds redshift

mtesch-um:
Adds a workflow to log into AWS via SSO, then get the db credentials using the various DB credential retrieval using the auth from the SSO login.

https://github.com/beekeeper-studio/beekeeper-studio/issues/3607

DRAFT because I cant build it locally - seems to require node20?  or for whatever reason during `yarn install`, node-gyp fails with 

```
    from distutils.version import StrictVersion
ModuleNotFoundError: No module named 'distutils'
```


not-night-but:
@mtesch-um I'm so sorry I totally missed this. I think we might have this all set up in the app now for the next release. It should be in master, to build if you want to test it out you should need node 22.21.1

not-night-but:
I'm going to close it just because if there is something missing in the aws sso  I think it'll probably need to be a new branch anyways?

Issue #3266, opened 2025-06-11
FEAT: Better fuzzy search in Quick Search


valtism:
**Is your feature request related to a problem? Please describe.**

When opening a table, such as `job_positions`, I will naturally want to search for it by typing something like `jobpos`. Right now that does not work with the current Flexsearch implementation, because there is an underscore between `job` and `pos`. 

**Describe the solution you'd like**

I would like the fuzzy search to work more like `fzf` does, or the menubar of VSCode, where it will match even if the search text is non-sequential.

manudoni-hubcore:
Hello, i don't know if this is closed because it used to work, but now the search has the same old problem.
Like the previous example, if there is an underscore between `jobs` and `position` i cannot search by `jobpos`, no results are returned.
(version 5.8.1)

PR #3725, opened 2025-12-10
Fuzzy find in command palette

valtism:
Replace flexsearch with match-sorter to allow fuzzy finding in command palette

Closes https://github.com/beekeeper-studio/beekeeper-studio/issues/3266

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 dev/**/*.sql linguist-language=SQL
+*.surql linguist-generated
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -128,7 +128,7 @@
     "sqlanywhere": "beekeeper-studio/node-sqlanywhere#54d1ef2052ccdfe963f0956a3e4d024ee6f1fd8d",
     "ssh2": "^1.14.0",
     "surrealdb": "^1.3.2",
-    "tabulator-tables": "beekeeper-studio/tabulator#137018661b71291b18f19324ef39484496da281b",
+    "tabulator-tables": "beekeeper-studio/tabulator#f9d3c0cdf0933a9a1bc5056b773c02a68a309fa1",
     "tinyduration": "^3.2.4",
     "trino-client": "^0.2.7",
     "typeface-roboto": "^0.0.75",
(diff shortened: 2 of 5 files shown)


valtism:
I don't code in Vue, so please double check my work here. I've tested this locally and it all seems to work.

I've used match-sorter instead of flexsearch since the latter does not support substring matching because it's architecture means it processes char-by-char.

not-night-but:
@valtism Thanks so much for this! I'm going to mention this to the team just to make sure no one has any issues with switching, but I've given this a test with one of my load test dbs and it performed well!

The only issue I can see right now is that quicksearch doesn't highlight the matches properly for fuzzyfinding. Here's a fuzzy match:
image
Whereas this is how I'd want it to highlight:
image
Fully sequential matches still highlight:
image

If this isn't something you'd want to work on, no worries. Just want to make sure it's at least documented.

valtism:
That's a good point. I'll look into implementing that now

valtism:
@not-night-but I've switched out `match-sorter` to `microfuzz` because it supports match ranges and have used them to highlight characters in the search titles. Let me know if that looks good

not-night-but:
@valtism How did you choose this library? I'm not totally sold on microfuzz, as it doesn't seem to be super popular and hasn't been updated in a while. What about something like uFuzzy?

valtism:
@not-night-but Yeah, hadn't really used it so just picked it as something that supported highlights, but `uFuzzy` seems better. I've replaced it with that and also moved the highlighting to the search module so it returns the `highlight` formatted (and escaped) string directly. Hope that works

valtism:
@not-night-but gentle nudge on this. anything else you need?

not-night-but:
@valtism So sorry for the late reply!

The code looks good for the most part, but in testing it I have come across a rather interesting issue:
image

Granted this is an extreme case, but it does work with the old search method so idk what's going on there. (if you want to test it yourself, there's a container in the docker compose called `psql15-load-test`. It is a terrifying amount of data, but I tend to use it to test stuff like this)

valtism:
Thanks for the info! Looks like uFuzzy sets a default threshold of 1000 at which point it will bail out. 

I was able to run search on the database perfectly well with the limit being set to `Infinity` so I think that should work well for all use-cases

not-night-but:
@valtism Okay so sorry for the long turnaround on this!

Code looks good, and testing with the load db it all seems to work well! I'm just gonna let the workflows run and then we should be good to merge! Thanks for the contribution! (sorry again for how long it took)

PR #3726, opened 2025-12-10
Super Formatter UI Tweaks

azmy60:
- Improve super formatter UI
image

- Hide the "Open query formatter" button. This keeps it organized because, unlike the other buttons, open formatter is not a quick action button.
image

NOTE:
I'd say that this is not an easy UX problem, but not the most difficult one probably. I noticed that the flow of "adding/editing/removing preset" feels a little confusing IMO, and it'll be awesome if we can improve that at some point. Maybe I'll revisit this again later and come up with a solution.

### DIFF
--- a/apps/studio/src/assets/styles/app/modals.scss
+++ b/apps/studio/src/assets/styles/app/modals.scss
@@ -140,7 +140,7 @@
       display: flex;
       container-type: inline-size;
 
-      @container (max-width: 550px) {
+      @container (max-width: 600px) {
         .BksSuperFormatter {
           grid-template-columns: auto;
           grid-template-rows: 1fr 1fr 1.75rem;
--- a/apps/ui-kit/lib/components/super-formatter/SuperFormatter.vue
+++ b/apps/ui-kit/lib/components/super-formatter/SuperFormatter.vue
@@ -389,7 +389,7 @@ export default Vue.extend({
     flex: 1;
     min-height: 0;
     display: grid;
-    grid-template-columns: 18rem auto;
+    grid-template-columns: 20rem auto;
     grid-template-rows: auto 1.75rem;
     grid-row-gap: 1rem;
     align-items: stretch;


not-night-but:
Okay so I like the addition of the editor background in the modal and not having the extra button in the editor, but honestly I think it feels squished now? Like I think having multiple columns of options before made it look better? 

@rathboma what do you think?

wmontgomery:
I'm with @not-night-but on this. The modal looks a little too squished together. I like that it's condensed, but it could benefit from being wider as a whole. Maybe something more like this?

image

css changes are

max-width changed to 70vw
image

(Check the element.style for changes on .BksSuperFormatter
The gap might be a bit much, but was just playing around with getting some whitespace in there to give it some breathing room
image

azmy60:
Hmm I think I'd disagree with y'all for it looking squishy lol. But hey, I might be wrong. By the time I squished it, I wasn't sure if it's better. And even now, it still feels bit off so maybe y'all are right!

@wmontgomery And thanks for the snippets! To be fair, I'd make it wider, but probably not relative to the screen size because I'm worried about users with bigger monitor (that's me! :D). It'll make you tilt your head from left to right if it's too wide, which isn't the best feeling. It's about finding the balance.

@not-night-but For multi-column, that's the first thing I tried. In UI, it probably takes much less space. But it's hard to justify the cost of the UX of it, e.g. eye movement is zigzaggy, form groups, etc.. So I stick with single column cause it's simple (and safe). Most apps use single column for settings anyway.

PS: Love the sharp feedbacks! Please don't hold back lol. :)

azmy60:
Update:

[screen-recording-1765453494168.webm](https://github.com/user-attachments/assets/ce8a6c9b-5ac5-44c9-b78f-3077bc412db6)

Not much difference, honestly. I don't really have a reference whether I prefer condensed or looser space. It's still condensed because that should be consistent with how our apps look (especially if you look at the Connection Interface).

not-night-but:
I think the extra width looks good and I will relent on separating into columns for now. The main one that annoys me is just the `Use Tabs` checkbox. The extra space just makes me feel weird lol

PR #3728, opened 2025-12-11
Fix switch connection menu

not-night-but:
The switch connection menu was not properly passing the config to the store action. It also looks like we weren't doing license checks in those methods so it could be possible for users to circumvent license restrictions.

### DIFF
--- a/apps/studio/src/components/quicksearch/QuickSearch.vue
+++ b/apps/studio/src/components/quicksearch/QuickSearch.vue
@@ -265,6 +265,7 @@ export default Vue.extend({
           break;
         case 'connection':
           if (!this.isUltimate && isUltimateType(result.item.connectionType)) {
+            this.$noty.error('Cannot switch to Ultimate only connection.')
             return
           }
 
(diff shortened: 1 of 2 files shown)


apps/studio/src/components/quicksearch/QuickSearch.vue
```diff
@@ -262,6 +264,10 @@ export default Vue.extend({
           this.$root.$emit('favoriteClick', result.item)
           break;
         case 'connection':
+          if (!this.isUltimate && isUltimateType(result.item.connectionType)) {
+            return
```
azmy60, line 269: Would be best if we put a noty or something to tell user that "this action requires a license?" kinda thing? A follow-up PR is fine if it's too much.

apps/studio/src/components/sidebar/core/ConnectionButton.vue
```diff
@@ -266,9 +268,15 @@ export default {
       }
     },
     async selectConnection(config) {
+      if (!this.isUltimate && isUltimateType(config?.connectionType)) {
+        return;
```
azmy60, line 273: Same comment 👆

PR #3735, opened 2025-12-14
Bugfix/3668 currency query magic

### LINKED ISSUE #3668: BUG: Currency query magic halts display of results
**Describe the bug**
The __format__money query magic appears to prevent Beekeeper Studio from displaying the rest of the results table.

**To Reproduce**
1. In BKS, run the following query:
`-- Works everywhere; no tables needed
SELECT 1 AS id, 'Vaccination' AS service, 49.99 AS amount__format__money__USD FROM (SELECT 1) x
UNION ALL SELECT 2, 'Neutering', 120.00
UNION ALL SELECT 3, 'Dental Cleaning', 89.50
UNION ALL SELECT 4, 'Microchipping', 25.00;
`
3. View the result and see that only the first two columns of the first row appear, with the rest of the result not displaying at the point that the query magic should appear.

**Expected behavior**
Would expect to see four results with the numberic values rendered as USD>

 -  MacOS
 - 5.4.9
 - Tested with SQLite and Postgres

wmontgomery:
resolves #3668 
3668_currency-query-magic

### DIFF
--- a/apps/studio/src/common/platform_info/mainPlatformInfo.ts
+++ b/apps/studio/src/common/platform_info/mainPlatformInfo.ts
@@ -33,8 +33,6 @@ export function mainPlatformInfo(): IPlatformInfo {
 
   // eslint-disable-next-line @typescript-eslint/no-var-requires
   const e = require('electron')
-  // defaulting applocale to US (sorry friends, but that's where I am). Maybe add it to the ini file at some point?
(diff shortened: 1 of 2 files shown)


not-night-but:
I think we could probably also look into loading it in the preload script and seeing if that fixes the issue (future fix)

PR #3751, opened 2025-12-21
Bugfix/3559 blinking dialog

wmontgomery:


apps/studio/src/store/index.ts
```diff
@@ -517,8 +517,8 @@ const store = new Vuex.Store<State>({
 
       context.commit('clearConnection')
       context.commit('newConnection', null)
-      context.dispatch('updateWindowTitle')
-      context.dispatch('refreshConnections')
+      await context.dispatch('updateWindowTitle')
```
azmy60, line 520: this doesn't need `await` maybe?

azmy60 (approved):

azmy60:
Also, might worth to consider merging this to `rc-55`.

apps/studio/src/store/index.ts
wmontgomery: context.commits are synchronous whereas dispatches are async. The thought with the blinking is there was something getting in the way and maybe refreshing state in a really weird manner causing some flickering so made the parts that need to be sequential sequential.

wmontgomery:
@azmy60 or @not-night-but , should I merge this into master and rc-55 or just rc-55? I can reset the base

wmontgomery:
Since this is going to be for RC-55, I'm going to base this off of the rc-55 branch and resubmit. Hold please

PR #3766, opened 2025-12-26
Bundle first-party plugins

### LINKED ISSUE #3763: FEAT: Bundle specific first-party plugins with the app download
For key first-party plugins (erds, ai shell), we want to bundle them with the app so they are always available.

How to do this:
1. Make a script that can re-use the plugin list/installation code that: a) reads the plugin.json for first party plugins, and b) downloads them to the resources directory during the build process
2. Make sure they are bundled in the resources directory on build. Copy how we do this for the demo db
3. On start, the app needs to know that plugins can be in several places: resources and config. Updates and everything can work the same as it does now.
4. Remove any logic that auto-downloads plugins on start.

azmy60:
fix #3763 

In this PR, AI Shell and ER Diagram are registered as npm dependencies and included as `extraResources`. 

`BundledPluginModule` is responsible for handling bundled plugins, and it acts as an extension to the plugin system rather than being coupled with it.

```ts
import { BundledPluginModule } from "@src-commercial/plugin-system/modules";
const manager = new PluginManager({ ... });
manager.registerModule(BundledPluginModule);
await manager.initialize();
```

Shout out to the Tabulator's module system!

I'm going to stick with this approach to extend the plugin system from now on.


apps/studio/src-commercial/backend/plugin-system/modules/BundledPluginInstaller.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    const manifest = JSON.parse(fs.readFileSync(manifestPath, "utf-8"));
+    const pluginId = manifest.id;
+
+    // Have installed before?
+    if (this.manager.pluginSettings[pluginId]) {
+      throw new Error(
+        `Plugin "${pluginId}" was previously installed and cannot be installed again.`
+      );
```
not-night-but: This doesn't seem like an error to me? Like maybe just a warn and return? I just don't see a reason to throw

apps/studio/src-commercial/backend/plugin-system/modules/BundledPluginInstaller.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    const dst = path.join(pluginsDirectory, pluginId);
+    if (fs.existsSync(dst)) {
+      // This must be set, otherwise the plugin will be copied again
+      await this.manager.setPluginAutoUpdateEnabled(pluginId, true);
+      throw new Error(
+        `Plugin "${pluginId}" installation directory already exists on disk.`
+      );
```
not-night-but: Same thing as above

azmy60:
@not-night-but I addressed the feedback and I also moved the new tests to a new file to clean it a little bit.

azmy60:
@not-night-but Jean made a PR (#3869) to test the ERD plugin so we might want to merge rc-56 to master as well.

not-night-but (approved):

PR #3767, opened 2025-12-27
MSVC redist auto install

### LINKED ISSUE #3202: BUG: Running in the background and not executing
**Describe the bug**
The app after installing version 5.2.9 for the first time on my computer runs in the background and does not open the Beekeeper home screen on Windows.

Running in the background and not executing

**To Reproduce**
1. Go to 'Advanced format your computer by wiping everything'
2. Click on 'install version 5.2.9'
4. See error: It doesn't run, it just runs in the background, you can see this in the task manager.

**Expected behavior**
Open Beekeeper .

Solution: 
Clean files all cache beekeeper and a install version 5.0.9.

not-night-but:
fix #3202

### DIFF
--- a/apps/studio/build/win/msvc-redist.nsh
+++ b/apps/studio/build/win/msvc-redist.nsh
@@ -28,7 +28,7 @@
     ;download installer
     inetc::get "$VCRedistDownload" $TEMP\bks-setup\vcppredist.exe
     ;exec installer
-    ExecWait "$TEMP\bks-setup\vcppredist.exe"
+    ExecWait '"$TEMP\bks-setup\vcppredist.exe" /passive /norestart'
   ${EndIf}
 
 


apps/studio/electron-builder-config.js
```diff
@@ -208,7 +208,8 @@ module.exports = {
     "artifactName": "${productName}-${version}-portable.exe",
   },
   nsis: {
-    oneClick: false
+    oneClick: false,
+    include: './build/win/msvc-redist.nsh'
```
rathboma, line 212: let's download the installer during builds and bundle it with the windows version. Then we can download the right version based on the arch.

That way there are fewer things that can go wrong here. This is the same problem as downloading plugins.

apps/studio/build/win/msvc-redist.nsh
```diff
@@ -0,0 +1,37 @@
+; https://github.com/GitCommons/cpp-redist-nsis
+
+!include LogicLib.nsh
+
+!macro customInit
+  Var /GLOBAL VCRedistDownload
+  Var /GLOBAL isInstalled
+  ${If} ${RunningX64}
+    ;check H-KEY registry
+    ReadRegDWORD $isInstalled HKLM "SOFTWARE\Wow6432Node\Microsoft\VisualStudio\14.0\VC\Runtimes\x64" "Installed"
```
rathboma, line 10: When does it change from version 14? Is this specified by DuckDB Node? How do we know when we need to change this?

How should we manage this?

rathboma: Maybe we could have a script that checks this script vs the requirements somehow? Is that in node_modules somewhere?

apps/studio/electron-builder-config.js
rathboma: Ok, nevermind. I just read the license and we can't redistribute. So you're approach is the best.

apps/studio/build/win/msvc-redist.nsh
```diff
@@ -0,0 +1,37 @@
...
+      to function properly.$\r$\n$\r$\n\
+      Download and install now?" /SD IDYES IDNO VSRedistInstalled
+
+    ;if no goto executed, install vcredist
+    ;create temp dir
+    CreateDirectory $TEMP\bks-setup
+    ;download installer
+    inetc::get "$VCRedistDownload" $TEMP\bks-setup\vcppredist.exe
+    ;exec installer
+    ExecWait "$TEMP\bks-setup\vcppredist.exe"
```
rathboma: I think you should use the /passive flag here. See the command line flags:
https://learn.microsoft.com/en-us/cpp/windows/redistributing-visual-cpp-files?view=msvc-170#command-line-options-for-the-redistributable-packages

rathboma: maybe also /norestart?

apps/studio/build/win/msvc-redist.nsh
not-night-but: 14.0 Is the latest version and is for visual studio 2017-2026. DuckDB just requires the latest version, which is 14

PR #3769, opened 2025-12-29
fix: add numpad plus key support for Zoom In on macOS

### LINKED ISSUE #3765: BUG: `Cmd` + `Keypad +` (Numpad Plus) does not trigger Zoom In on macOS
I've noticed that the **Zoom In** shortcut does not work when using the **Numpad Plus key** (`Cmd` + `Keypad +`) on macOS.

-   **Zoom Out** works correctly with both the standard `-` key and the Numpad `-` key.
-   **Zoom In** works correctly with the standard `=` (`+`) key.
-   **Zoom In** fails when using the Numpad `+` key.

It seems the application is not listening to the `keypad_plus` key code for the Zoom In action.

### **Steps to Reproduce**
1. Open Beekeeper Studio on macOS.
2. Press `Cmd` + `Keypad -` (It zooms out correctly).
3. Press `Cmd` + `Keypad +`.
4. **Result:** Nothing happens (The font size/interface scale does not increase).
5. Press `Cmd` + `+` (on the main keyboard area).
6. **Result:** It zooms in correctly.

### **Expected Behavior**
Pressing `Cmd` + `Keypad +` should increase the zoom level, identical to pressing `Cmd` + `+` on the main keyboard.

### **Environment**
- **OS:** macOS Tahoe 26.1 (apple sillicon)
- **Beekeeper Studio Version:** 5.5.3

### **Additional Context (Input Logs)**
I have verified my key inputs using an event viewer. The system correctly registers the `left_command` + `keypad_plus` down events, but Beekeeper Studio ignores them.

Here is the log comparison between the working standard key and the non-working keypad key:

**1. Non-working `Cmd` + `Keypad +` log:**
```json
{
  "type": "down",
  "name": {"key_code":"keypad_plus"},
  "usagePage": "    7 (0x0007)",
  "usage": "   87 (0x0057)",
  "misc": "flags left_command"
}
```

**2. Working `Cmd` + `+` (Equal Sign) log:**
```json
{
  "type": "down",
  "name": {"key_code":"equal_sign"},
  "usagePage": "    7 (0x0007)",
  "usage": "   46 (0x002e)",
  "misc": "flags left_command"
}
```

line1029:
## Summary
- Add hidden menu item with `CommandOrControl+numadd` accelerator to enable Zoom In via numpad plus key
- The menu item is set to `visible: false` so it doesn't appear in the View menu but the accelerator still works

## Related Issue
Fixes #3765


line1029:
I tested the build on Linux OS and found that numpad plus now works after the fix, but numpad minus doesn't work. Are you experiencing the same issue?

azmy60:
@line1029 Yeah, Ctrl+- using numpad didn't work for me.

apps/studio/src/common/menus/MenuItems.ts
```diff
@@ -73,6 +73,13 @@ export function menuItems(actionHandler: IMenuActionHandler, settings: IGroupedU
       accelerator: 'CommandOrControl+=',
       click: actionHandler.zoomin
     },
+    zoominNumpad: {
+      id: 'zoom-in-numpad',
+      label: "Zoom In",
+      accelerator: 'CommandOrControl+numadd',
```
azmy60: Could we use `CommandOrControl+Plus` instead of `CommandOrControl+numadd`? Maybe that can fix the linux key capture.

apps/studio/src/common/menus/MenuItems.ts
```diff
@@ -73,6 +73,13 @@ export function menuItems(actionHandler: IMenuActionHandler, settings: IGroupedU
...
+      click: actionHandler.zoomin,
+      visible: false,
+    },
```
azmy60, line 82: Maybe we should add new similar items for zoom out and zoom reset. This is an example: https://github.com/Automattic/simplenote-electron/pull/2404/changes

line1029:
I've addressed the feedback. The shortcuts weren't working on Windows and Linux because the menu bar on those platforms is rendered through Vue. I've fixed this. Please review when you have a chance!

Issue #3742, opened 2025-12-16
BUG: Query Magics `money` format fails with `locale` error


rjenningsFH:
**Describe the bug**
When using [query magics](https://www.beekeeperstudio.io/features/query-magics) to format a numeric column as money the results fail to render.  The developer console shows:

<img width="517" height="162" alt="Image" src="https://github.com/user-attachments/assets/0a36376c-873a-4495-9c4d-05060736a5e6" />

**To Reproduce**
1. Connect to beekeeper's database
2. Open a new query tab
3. Try the following query:
```sql
SELECT
  i.version stars__format__stars,
  i.version money__format__money,
  i.*
FROM
  installation_ids as i;
```
5. Running the query shows the column names but no table beneath it:

<img width="1218" height="299" alt="Image" src="https://github.com/user-attachments/assets/8bc8a106-19cb-4cdb-ba76-e912fcbdfa91" />

6. The developer console will show something like:

```
Uncaught (in promise) 
RangeError: Incorrect locale information provided
    at new NumberFormat (<anonymous>)
    at I5.formatter (index-NaxMtlZB.js:3914:30218)
    at I5.formatValue (index-NaxMtlZB.js:3170:90530)
    at index-NaxMtlZB.js:3186:23732
    at Array.forEach (<anonymous>)
    at InternalEventBus._chain (index-NaxMtlZB.js:3186:23699)
    at Cell.chain (index-NaxMtlZB.js:3166:4573)
    at Cell._generateContents (index-NaxMtlZB.js:3168:4844)
    at Cell.layoutElement (index-NaxMtlZB.js:3168:6398)
    at Cell.getElement (index-NaxMtlZB.js:3168:5464)
```

7. Comment out the `money` format and you'll see that the `stars` magic does work:

<img width="566" height="488" alt="Image" src="https://github.com/user-attachments/assets/e073b5f6-0179-404c-8cb3-2f7e63180b8d" />

Note: I minimized the installation ID column width so it wouldn't be in the screenshot, but that did size itself correctly.

**Expected behavior**
Using the `money` query magic should format the column as a monetary value.

**Version Information (please complete the following information):**
 - OS: Mac 26.2
 - App Version: 5.4.12
 - Database type and version: MySQL 8.x but also recreate on sqlite with Beekeeper's Database

not-night-but:
This actually should be fixed in version 5.5 (which I am pushing out today)

rjenningsFH:
[...]

PR #3782, opened 2026-01-04
Bugfix/3742 big int money

wmontgomery:
resolves [#3742](https://github.com/beekeeper-studio/beekeeper-studio/issues/3742)
3742_big-int-money

### DIFF
--- a/apps/studio/src/background/WindowBuilder.ts
+++ b/apps/studio/src/background/WindowBuilder.ts
@@ -55,7 +55,8 @@ class BeekeeperWindow {
         spellcheck: false,
         sandbox: false,
       },
-      icon: getIcon()
+      icon: getIcon(),
+      show: false,
     })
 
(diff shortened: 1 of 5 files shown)


not-night-but:
@wmontgomery There's no shot it was that simple.... really? :sob: 

I think we should probably look into writing some tests for magics, but we can probably push this into 55 if you're happy with it and do that in another PR

wmontgomery:
> @wmontgomery There's no shot it was that simple.... really? 😭
> 
> I think we should probably look into writing some tests for magics, but we can probably push this into 55 if you're happy with it and do that in another PR

The internationalization stuff takes either numbers, strings, or big int thingers. Some of our drivers returns big numbers as strings (mysql specifically) and having the "isNumber/toNumber" stuff that was there would fail if the number in question was bigint. It was just easier to have it try and when it failed say "hey, comrade, this isn't a number". I was just as shocked.

PR #3786, opened 2026-01-07
fix saving multiple models in appdb handlers

not-night-but:
The saving was happening, but the return value was bad. Not sure if this actually affected anything, but it was throwing errors.


azmy60:
@not-night-but curious what error? this one?
image

not-night-but:
@azmy60 yeah. It's been annoying me lol

azmy60 (approved):
LGTM.

Obviously, this affects the return value of `appdb/[something]/save` if the argument is an array. But from what I saw, all of them don't use the return values so I think it's safe.

PR #3812, opened 2026-01-21
feat(studio): add DROP CASCADE option for table drops

### LINKED ISSUE #3762: FEAT: Add DROP CASCADE syntax when dropping tables
Whenever I am dropping tables in Beekeeper, I have to drop them in the correct sequence because there isn't an option to drop them with cascade.

In Intellij, there is a confirmation dialog that pops up when you are dropping a table, there is a checkbox there that allows you to tick to use DROP CASCADE syntax.

Image

The solution should be made for all relational databases.

For SQL Server, there's not really an option to DROP CASCADE so I was thinking of having some sort of script that queries for the table details and drop the constraints one by one.

rathboma:
## Summary
- Add a CASCADE checkbox to the drop table confirmation dialog
- When checked, tables are dropped with CASCADE syntax, removing dependent objects (foreign keys, views)
- Checkbox is disabled with tooltip for databases that don't support CASCADE

## Database Support
| Database | CASCADE Support | Syntax |
|----------|-----------------|--------|
| PostgreSQL | Yes | `DROP TABLE name CASCADE` |
| CockroachDB | Yes | Inherits from PostgreSQL |
| Redshift | Yes | Inherits from PostgreSQL |
| Oracle | Yes | `DROP TABLE name CASCADE CONSTRAINTS` |
| DuckDB | Yes | `DROP TABLE name CASCADE` |
| MySQL/MariaDB | No | Checkbox disabled |
| SQL Server | No | Checkbox disabled |
| SQLite | No | Checkbox disabled |
| Others | No | Checkbox disabled |

## Test plan
- [ ] Connect to PostgreSQL, create two tables with foreign key relationship
- [ ] Try dropping parent table without CASCADE (should fail or warn)
- [ ] Try dropping parent table with CASCADE checked (should succeed)
- [ ] Connect to MySQL, verify CASCADE checkbox is disabled with tooltip
- [ ] Verify checkbox only appears for tables, not views

Closes #3762

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


apps/studio/src/components/CoreTabs.vue
```diff
@@ -187,6 +187,25 @@
               /> {{ this.dbElement }}</span>?
             </div>
             <p>This change cannot be undone</p>
+            <div
+              v-if="showCascadeOption"
+              class="form-group checkbox-form-group"
+            >
```
not-night-but, line 193: Why even show the checkbox if cascade isn't supported for the dialect?

not-night-but (changes requested):
I don't think we should show the checkbox if cascade is not supported. Also I think we should probably have some tests for drop cascade for schemata.

PR #3837, opened 2026-02-03
Wire plugin menu items to configurable keybindings

azmy60:
- Add a default keybinding for AI Shell

```ini
[keybindings.plugins.bks-ai-shell]
new-tab-dropdown-item = ctrlOrCmd+l
```

- Add support for plugin keybindings (currently limited to `newTabDropdown` and `menubar.tools`). So any plugins can attach keybindings through config, not just AI Shell.

| Placement        | Preview |
|------------------|---------|
| `menubar.tools`  | image (For demonstration. ER plugin isn't attached to a default keybinding.) |
| `newTabDropdown` | image |

I picked `Ctrl+L` because that's how you open the AI sidebar in Cursor. Feel free to suggest a better keybind!


apps/studio/src/common/bksConfig/BksConfigProvider.ts
```diff
@@ -100,6 +100,24 @@ const vHotkeyModifierMap: ModifierMap = {
...
+  COMMANDORCONTROL: "Control",
+  SHIFT: "Shift",
+  ALT: "Alt",
+  OPTION: "Alt",
+  ALTGR: "AltGraph",
+  SUPER: "Super",
+  META: "Meta",
+  WINDOWS: "Meta",
+} as const;
+
```
not-night-but: I think this is kinda just a copy of the context-menu map? I made that for xel because our context-menus use xel

not-night-but (dismissed):
I think this looks good, I just don't think we should duplicate the modifier map stuff (`context-menu` vs `xel`)

not-night-but (approved):

PR #3839, opened 2026-02-04
fix codemirror dependency

azmy60:


not-night-but:
@azmy60 The codemirror dep is still causing issues. (Locally it works with 1.15.13)

azmy60:
@not-night-but I only saw a type issue. Were there other issues besides that? In any case, I downgraded to the latest patch version that has the fork update. https://github.com/marimo-team/codemirror-languageserver/pull/7

not-night-but:
No the only issue I had was the type issue preventing it from building as well

PR #3841, opened 2026-02-04
Fix missing plugin menu in the native menu (macOS)

azmy60:


rathboma:
looks like we could also use this to show/hide different menus on the connection vs core interface screens? How annoying do you think this is to maintain?

not-night-but:
@rathboma I think that's probably a possibility. I think we probably want to get this merged as soon as we can though (maybe into 55?)

wmontgomery:
Any concern of this menu getting too large as more plugins are added to it? I know that's a later problem, but something to think about as this scales?

azmy60:
Apologies for the late reply.

> looks like we could also use this to show/hide different menus on the connection vs core interface screens? How annoying do you think this is to maintain?

@rathboma Yeah, I think we could use this to show/hide menus in general. I don't think it's hard to maintain if the flow stays the way it is (`renderer` -> sends `AppEvents` to ->  `main/utility`) and not the other way around. Will be happy to make a separate PR for that if you want.

> @rathboma I think that's probably a possibility. I think we probably want to get this merged as soon as we can though (maybe into 55?)

@not-night-but Yeah, I'm rebasing this to rc-55 now.

> Any concern of this menu getting too large as more plugins are added to it? I know that's a later problem, but something to think about as this scales?

@wmontgomery Good point! I just checked Figma and it seems like they just show all plugins if you go `Right click` > `Plugins` > `Saved Plugins`. So the easiest solution would probably be avoiding adding plugins to the `Tools` immediately, and we can "hide" them in `Tools > Plugins > Installed Plugins` or something like that. It doesn't actually fix the menu for getting too large, but I feel like that's better. Also, I think it'll be clear that "tools in plugins are plugins" where as other tools are not plugins. But that's just an idea.
 
image

Issue #3840, opened 2026-02-04
BUG: Unable to import from URL with spaces in password


Elgirhath:
**Describe the bug**

**To Reproduce**
1. Open a new connection
2. Select Import from URL
3. Paste a url similar to this: `postgres://postgres:some%20password@localhost:5432/postgres`

**Expected behavior**
The %20 should be treated as space within the password.

**Actual behavior**
`some%20password` is treated as a string literal. I need to edit the `%20` and replace it with spaces

<img width="652" height="786" alt="Image" src="https://github.com/user-attachments/assets/5d1be523-2ed1-4284-8b67-93f0191d6473" />

**Version Information (please complete the following information):**
 - OS: Mac
 - App Version: 5.5
 - Database type and version: PostgreSQL 16

**Additional context**
A password with literal spaces cannot be parsed. 
`postgres://postgres:some password@localhost:5432/postgres`

<img width="552" height="237" alt="Image" src="https://github.com/user-attachments/assets/f32d72dd-feeb-4b83-aaa0-77128c0ef485" />

Issue #1495, opened 2023-02-05
FEAT: Build a better connection string parser library


rathboma:
Right now we use a generic db url parser to parse out db connection URLs, but this doesn't work right with all the different databases with all the different options they present.

We should build our own db connection parsing library. Features it should include:
1. Make use of [PG Connection String](https://www.npmjs.com/package/pg-connection-string) for postgres URLs
2. Be able to parse JDBC connection strings
3. Be able to parse ODBC connection strings
4. Be able to parse SQLite file paths
5. Be able to parse SQLite file:// protocol paths
6. Be able to figure out specific default driver options for special databases like - RDS, Heroku, Aurora, Redshift.
7. Be able to fall back to a default implementation and identify the database

PR #3852, opened 2026-02-06
fix: decoding connection password from url

pvlvld:
Simply wrapped password in the [decodeURIComponent()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent).

[connection-string](https://github.com/vitaly-t/connection-string) by default doesn't parse them, and [provides a .toString method](https://github.com/vitaly-t/connection-string?tab=readme-ov-file#api) for that.

Fixes: #3840 
Useful info: vitaly-t/connection-string#48
Related TODO: #1495


rathboma:
Hey! this is good. I've taken this code into another branch where I also added tests. thanks for your help!

PR #3855, opened 2026-02-06
feat: add configuration option to disable automatic update checks

### LINKED ISSUE #3854: FEAT: Add a way to disable automatic update checks entirely
Some users might not want to allow automatic update checking, eg enterprise users or folks with old perpetual licenses.

1. Add a new configuration key to disable update checks in default.config.ini

@claude please implement

rathboma:
Add checkForUpdatesEnabled boolean flag to default.config.ini that defaults to true.
When set to false, the update manager will skip all automatic update checking,
including the periodic interval checks and manual update requests.

This addresses the needs of enterprise users and those with old perpetual licenses
who may not want automatic update checking enabled.

Closes #3854

### DIFF
--- a/apps/studio/src/background/update_manager.ts
+++ b/apps/studio/src/background/update_manager.ts
@@ -34,10 +34,6 @@ function shouldSkipUpdater() {
 }
 
 function checkForUpdates() {
-  if (BksConfig.general.checkForUpdatesDisabled) {
-    log.info('update checks are disabled, skipping')
-    return
-  }
   log.info('checking for updates right now')
   try {
     autoUpdater.checkForUpdates()


azmy60 (changes requested):
Should this be `checkForUpdatesDisabled` instead of `checkForUpdatesEnabled`? your other [PR](https://github.com/beekeeper-studio/beekeeper-studio/pull/3858/changes) uses `disabled`. We also use `disabled` for plugins like this:

```ini
[plugins.bks-ai-shell]
disabled=true
```

rathboma:
Yeah I think so, we should keep it consistent

rathboma:
@claude change this flag to be `checkForUpdatesDisabled` rather than Enabled, then invert the checks

not-night-but:
@claude I don't think the first guard within checkForUpdates is necessary? As that's only called within manageUpdates after the second guard you put in

not-night-but:
@claude yeah fix that

not-night-but (approved):

PR #3862, opened 2026-02-09
docs: remove anchor/span wrapping from code blocks for clipboard compatibility

### LINKED ISSUE #3836: DOCS: Copying Linux commands breaks the syntax
At [https://docs.beekeeperstudio.io/installation/linux](https://docs.beekeeperstudio.io/installation/linux) whenever I copy a CLI command like `sudo apt update && sudo apt install beekeeper-studio -y` and try to paste it into the terminal, it paste with a line break right after `sudo`, resulting in something like:

```bash
sudo
 apt update && sudo apt install beekeeper-studio -y
```

**That happens to every command on this page and even if you paste it into a regular text editor.**

It seems to be related to the `<span class="w"> </span>` that is append after each term. Replacing them by regular spaces solves the problem.

EDIT: if it wasn't clear, the way the command it's pasted doesn't work, at least on zsh.

scrlkx:
Closes #3836

As we are not showing line numbers anywhere this settings seems to be useless.


not-night-but:
@scrlkx Apologies for the wait, thanks for the fix!

PR #3874, opened 2026-02-12
Basic support for community-plugins.json

azmy60:


rathboma:
only issue I see is that ID isn't unique official vs community plugin. Community user has no way of knowing if they're overriding one of our plugins, but that's fairly minor

azmy60:
@rathboma I think Obsidian repo has a bot that reviews PRs to prevent something like that, so I'm thinking if we can do that for our repo too.

PR #3875, opened 2026-02-12
fixed query rename issue #3822

### LINKED ISSUE #3822: BUG: Can't rename saved query
**Describe the bug**
When saving a new query I can give it the name I want, it works very well. After saving, if I try to rename it, I get the following error:

> An instance of FavoriteQuery has failed the validation: - property text has failed the following constraints: Queries have a max length of 2,000,000 characters.

**To Reproduce**
1. Go to **Saved Queries**
2. Right click **on any saved query**
3. Click on **Rename**
4. **Type another name** on the name field
5. Press **Save**

**Expected behavior**
It should rename the query without giving any errors.

**Screenshots / Gifs**
Image

 - OS: Fedora 43
 - App Version: 5.5.5
 - Database type and version: PostgreSQL 18

**Additional context**
Nothing else.

aniketsaurav18:
Fixes #3822 

Updated appdb save handling to load full entity data (including columns normally excluded by default selection) before merge + validation on updates.

This fixes rename/save failures where partial payloads (like { id, title }) were validated against an entity missing required hidden fields (for example, FavoriteQuery.text), causing false validation errors.

### DIFF
--- a/apps/studio/src/components/plugins/PluginManagerModal.vue
+++ b/apps/studio/src/components/plugins/PluginManagerModal.vue
@@ -50,7 +50,7 @@ import PluginPage from "./PluginPage.vue";
 import _ from "lodash";
 import ErrorAlert from "@/components/common/ErrorAlert.vue";
 import type { PluginContext, PluginRegistryEntry } from "@/services/plugin";
-import { mapState } from "vuex";
+import { mapGetters, mapState } from "vuex";
 
 const log = rawLog.scope("PluginManagerModal");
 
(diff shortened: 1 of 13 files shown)


not-night-but (changes requested):
@aniketsaurav18 Thanks so much for the fix! This is perfect, but I think we probably also need a similar change on the other side of that branch for when we're saving multiple objects at once just to keep everything consistent

aniketsaurav18:
@not-night-but I have made changes according to your suggestion. Please check.

not-night-but:
@aniketsaurav18 Cool, thanks for this!

PR #3881, opened 2026-02-16
Move truncation logic for results to the backend

not-night-but:
Move the truncation logic for results to the backend so that we aren't sending stupid amounts of data to the frontend. 

Before this, if you selected 300_000 rows, we would select all of that, send it over IPC to the frontend, and then discard 250_000 rows. Not only is this a waste, it could actually crash the app. There are other issues to be fixed to address the OOME that can happen with big queries like this, but this is one of them.

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -1352,7 +1352,6 @@
           results.forEach((result, idx) => {
             result.rowCount = result.rowCount || 0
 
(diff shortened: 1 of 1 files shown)


apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -1353,13 +1353,7 @@
             result.rowCount = result.rowCount || 0
 
             // TODO (matthew): remove truncation logic somewhere sensible
```
azmy60: should we remove the todo comment as well?

PR #3882, opened 2026-02-17
Support table view history (Undo/Redo)

### LINKED ISSUE #3843: Need a way of undoing a single change in tabletable
1. Make 5 time consuming and complicated changes
2. Oops, I made a mistake on one of them, how do I undo a single change and not lose the rest of my work?

azmy60:
Resolves #3843

### DIFF
--- a/apps/studio/config-metadata.json
+++ b/apps/studio/config-metadata.json
@@ -12,7 +12,9 @@
         { "key": "copySelection", "label": "Copy Selection" },
         { "key": "pasteSelection", "label": "Paste Selection" },
         { "key": "cloneSelection", "label": "Clone Selection" },
-        { "key": "deleteSelection", "label": "Delete Selection" }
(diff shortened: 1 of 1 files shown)


apps/studio/src/common/bksConfig/types.ts
```diff
@@ -0,0 +1 @@
+export type KeybindingTarget = "electron" | "v-hotkey" | "codemirror" | "tabulator";
```
not-night-but: Looks like this is missing the ui target? I could have sworn that there was another type similar to this in the keymap modal

not-night-but: Yeah I just checked. There's another definition of that in `apps/studio/src/types.ts`

azmy60: Fixed it!

PR #3883, opened 2026-02-17
Fix duckdb exports

not-night-but:
I believe the cursor wasn't reimplemented after we switched drivers. This method makes the cursor use duckdbs stream and just manually fetches chunks when we need them

### DIFF
--- a/apps/studio/src/lib/data/tools.ts
+++ b/apps/studio/src/lib/data/tools.ts
@@ -1,6 +1,6 @@
 import _ from 'lodash'
 import { Dialect } from '@shared/lib/dialects/models'
-import { friendlyJsonObject } from '@/common/utils';
+import { friendlyJsonObject, stringifyWithBigInt } from '@/common/utils';
 import rawLog from '@bksLogger';
 
 const log = rawLog.scope('data/tools')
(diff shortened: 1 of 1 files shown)


apps/studio/src-commercial/backend/lib/db/clients/duckdb/DuckDBCursor.ts
```diff
@@ -1,50 +1,40 @@
...
-      this.options.filters,
-      this.options.schema
-    );
-    this.cursorPos++;
-
-    return result.result.map((row) => Object.values(row));
+  async read(): Promise<any[][]> {
+    // We can't set the chunk size for duckdb, so we just buffer it
+    if (this.rowBuffer.length < this.chunkSize) {
+      const chunk = await this.stream.fetchChunk();
```
azmy60, line 28: Looks like we can kinda set the chunk size with this?

```ts
const reader = await connection.streamAndRead('from range(5000)');
reader.readUntil(2000);
// reader.currentRowCount === 2048 (Rows are read in chunks of 2048.)
// reader.done === false
reader.readUntil(4000);
// reader.currentRowCount === 4096
// reader.done === false
reader.readUntil(6000);
// reader.currentRowCount === 5000
// reader.done === true
```

This is from their readme.

apps/studio/src/lib/data/tools.ts
```diff
@@ -56,9 +56,11 @@ export const Mutators = {
     if (ArrayBuffer.isView(value)) {
       return value
     }
+    if (typeof value === 'bigint') return Number(value)
     if (_.isDate(value)) return value.toISOString()
-    if (_.isArray(value)) return preserveComplex? value.map((v) => mutate(v, preserveComplex)) : JSON.stringify(value)
-    if (_.isObject(value)) return preserveComplex? _.mapValues(value, (v) => mutate(v, preserveComplex)) : JSON.stringify(value)
+    const bigIntReplacer = (_k: string, v: any) => typeof v === 'bigint' ? Number(v) : v
```
azmy60: There's a similar helper function for this in `src/common/utils.ts` called `stringifyWithBigInt`. Can we use that instead? or does this need to be pure `number`?

apps/studio/src-commercial/backend/lib/db/clients/duckdb/DuckDBCursor.ts
not-night-but: This doesn't actually set the chunk size though, you can see that the chunks are still 2048 sized, so we'd still need to buffer the way that we're doing with the fetchChunk above. I honestly feel like the reader would complicate it a bit

azmy60: Oh I didn't notice that.

PR #3893, opened 2026-02-20
add version header for cloud requests

not-night-but:


apps/studio/src/store/modules/CredentialsModule.ts
```diff
@@ -31,7 +31,7 @@ interface State {
 
 async function credentialToBlob(c: TransportCloudCredential): Promise<CredentialBlob> {
   const clientOptions: CloudClientOptions = {
-    app: c.appId, email: c.email, token: c.token, baseUrl: window.platformInfo.cloudUrl
+    app: c.appId, email: c.email, token: c.token, baseUrl: window.platformInfo.cloudUrl, version: window.platformInfo.appVersion
```
azmy60: This doesn't match the type?

Shouldn't it be `clientVersion: window.platformInfo.appVersion`?

azmy60: Also, I would name it `appVersion` since it's an `appVersion`. But that's probably a personal preference. Not that I care about it too much.

PR #3902, opened 2026-02-24
Fix Pinned Entities

ThiagoCComelli:
## Summary

This PR fixes two issues related to the table pinning functionality.

### Bug 1 – Duplicated Pin Records in the UI

Two different instances were handling the same event. As a result, a single pin action was processed twice, creating duplicate records in the database.

**Root cause**
- Multiple listeners responding to the same pin event.

**How to reproduce** 

https://github.com/user-attachments/assets/d5e4c171-a79f-49df-84b5-5e42566e342c

**Fix** 

https://github.com/user-attachments/assets/25125269-c77e-4737-a551-a92b0deda4e2

---

### Bug 2 – Unpin Not Removing Records From DB

This issue occurred when unpinning a table that did not yet have any records in the database.

**What happened**
- When the table was pinned, the record was saved in the store **without an ID**.
- When attempting to unpin, the system tried to remove the record by checking its ID.
- Since the record had no ID, it was never removed from the database.

**Root cause**
- The pinned record is being added to storage without an ID when saving the pin to the database.

**How to reproduce** 

https://github.com/user-attachments/assets/1dc4289d-84fb-4f84-9ab4-bab7eef45f19

**Fix** 

https://github.com/user-attachments/assets/39cf07ec-972f-47ba-9512-da3abd90e4d8

### DIFF
--- a/apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
+++ b/apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
@@ -282,7 +282,6 @@ export default Vue.extend({
         pinned = !item.pinned;
       }
       item.pinned = pinned;
-      this.generateDisplayItems();
     },
     handleScrollEnd() {
       this.updateTableColumnsInRange(true);
@@ -321,10 +320,6 @@ export default Vue.extend({
   },
   watch: {
     schemaTables() {
-      this.generateItems();
-      this.generateDisplayItems();
-    },
-    pins() {
       this.generateItems();
       this.generateDisplayItems();
     },


ThiagoCComelli:
@not-night-but This PR should fix some bugs related to pinned entities. You've also made some [fixes](https://github.com/beekeeper-studio/beekeeper-studio/pull/3743) in in the past in this context, could you take a look?

not-night-but:
@ThiagoCComelli Yup I'm reviewing right now! Thanks so much

apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
```diff
@@ -320,6 +324,10 @@ export default Vue.extend({
       this.generateItems();
       this.generateDisplayItems();
     },
+    pins() {
+      this.generateItems();
+      this.generateDisplayItems();
+    },
```
not-night-but: I don't believe this needs to happen? `handleTogglePinned` will update the item

apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
```diff
@@ -272,13 +276,13 @@ export default Vue.extend({
...
+
       if (typeof pinned === "undefined") {
         pinned = !item.pinned;
       }
-      item.pinned = !item.pinned;
-
-      if (pinned) this.$store.dispatch('pins/add', entity)
-      else this.$store.dispatch('pins/remove', entity)
+      item.pinned = pinned;
+      this.generateDisplayItems();
```
not-night-but: I'm actually not sure we need to call generateDisplayItems here? Technically the display item is just a reference to the item itself, and we don't have to recalculate height (I just checked and it seems to work without regenerating the display items. This should work right @azmy60 ?)

azmy60: Yeah, I don't think calling it here seems necessary.

apps/studio/src/components/sidebar/core/table_list/VirtualTableList.vue
ThiagoCComelli: Yep, you're right! These are code snippets that I thought were necessary when I was debugging to understand the flow.

not-night-but:
@ThiagoCComelli It looks like there's one unrelated commit still in this branch, then I think this will be good to go

ThiagoCComelli:
@not-night-but I have no idea where that commit came from during the rebase. It's done now.

not-night-but:
@ThiagoCComelli No worries, git rebases are weird lol

PR #3904, opened 2026-02-25
Entra Fixes

not-night-but:
- Fix for execution on windows
- Add missing SSL & user form to entraid form
- Make CLI forms filepickers 
- Remove unnecessary accountname logic from top level forms (that stuff is done in the entra form)

### DIFF
--- a/apps/studio/src/components/connection/CommonSsl.vue
+++ b/apps/studio/src/components/connection/CommonSsl.vue
@@ -98,15 +98,17 @@
   </div>
 </template>
 
-<script>
+<script lang="ts">
 import FilePicker from '@/components/common/form/FilePicker.vue'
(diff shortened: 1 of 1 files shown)


apps/studio/src/components/connection/CommonSsl.vue
```diff
@@ -0,0 +1,156 @@
...
+
+<script>
+import FilePicker from '@/components/common/form/FilePicker.vue'
+import ExternalLink from '@/components/common/ExternalLink.vue'
+import ToggleFormArea from '../common/ToggleFormArea.vue'
+import { findClient } from '@/lib/db/clients'
+
+export default {
+  props: {
+    config: Object,
```
azmy60: I would add a ts type here using `PropType`.

PR #3906, opened 2026-02-25
FIX: Adding SSL settings to connection string parser

smallbenji:
Adding SSL settings to connection string parser.

File paths must be URL encoded so it doesn't break the connection string URL


rathboma:
@smallbenji This is great! Could you add a unit or integration test to this to make sure it works?

PR #3910, opened 2026-02-26
customize all keybinds

azmy60:
# Keybinding Syntax

Adding a keybinding can be as simple as this:

```ini
refresh = f5

```

Or very explicit to the platform:

```ini
refresh = win: f5, mac: cmd+r, linux: ctrl+r

```

## Multiple Keybindings

For multiple keybinds, you can use `[]`:

```ini
refresh[] = f5
refresh[] = ctrlOrCmd+r

```

To be more explicit (when things need to get complex):

```ini
refresh[] = f5, mac: cmd+r
refresh[] = mac: cmd+shift+r

```

> [!NOTE]
> The example above means Windows and Linux will use **f5**, while macOS will use **cmd+r** and **cmd+shift+r**.

## Overriding Defaults

Specifying the platform tags probably make sense if you are defining new keybinds for default configs as a developer. As a user, you probably don't need platform tags when overriding a default config. While they work, a plain keybind is easier to read and will replace the default on all platforms:

**default.config.ini**

```ini
[keybindings.general]
quit = mac: cmd+q

```

**user.config.ini (or system.config.ini)**

```ini
[keybindings.general]
quit = ctrl+q

```

The example above shows that `ctrl+q` will now apply on **all platforms**, including macOS, replacing the default.

## Tag

Previously, a value can be a string or number. This PR introduces a new type called `TaggedString`, (I don't know if that's a good name lol but let me know if we want to change that).

A plain value (string or number) looks like this:

```ini
key = value
```

When parsed, it'll become a string:

 ```json
"value"
```

If a value contains a "tag":

```ini
key = tag1: value
```

It'll be parsed as a TaggedString and become an object with `$$type` property:

```json
{
  "$$type": "taggedString",
  "tag1": "value"
}
```

Another example:

```ini
key = val0, tag1: val1, tag2: val2
```

The "untagged" value is going to be assigned to a "default" property when parsed:

```json
{
  "$$type": "taggedString",
  "default": "val0",
  "tag1": "val1",
  "tag2": "val2"
}
```

If we take one of the keybinding examples:

```
refresh[] = f5, mac: cmd+r
refresh[] = mac: cmd+shift+r
```

That will be parsed into an array of tagged objects:

```json
{
  "refresh": [
    { "$$type": "taggedString", "default": "f5", "mac": "cmd+r" },
    { "$$type": "taggedString", "mac": "cmd+shift+r" }
  ]
}
```

This is applied to all keys in keybindings section only and only affects to values that contain tags. Other values will be normally parsed as strings.

### DIFF
--- a/apps/studio/src/common/bksConfig/mainBksConfig.ts
+++ b/apps/studio/src/common/bksConfig/mainBksConfig.ts
@@ -3,7 +3,7 @@ import platformInfo from "@/common/platform_info";
 import * as path from "path";
 import _ from "lodash";
 import { existsSync, readFileSync, copyFileSync, accessSync, constants } from "fs";
-import { parseIni, processRawConfig } from "@/config/helpers";
+import { parseIni, processRawConfig, recurse } from "@/config/helpers";
 import {
   BksConfigProvider,
   ConfigEntryDetailWarning,
@@ -29,34 +29,44 @@ export function checkUnrecognized(
   sourceName: "system" | "user"
 ): ConfigEntryDetailWarning[] {
   const results: ConfigEntryDetailWarning[] = [];
+  const sections = new Set<string>();
+  const keys = new Map<string, string>();
 
-  function traverse(obj: Record<string, any>, parentPath = "") {
-    for (const key of Object.keys(obj)) {
-      const path = parentPath ? `${parentPath}.${key}` : key;
+  for (const { path } of recurse(newConfig)) {
+    // Skip validation for plugins section
+    if (path[0] === 'plugins') {
+      continue;
+    }
 
-      // Skip validation for plugin configurations (plugins and plugins.[plugin-id])
-      if (path === 'plugins' || /^plugins\.[^.]+/.test(path)) {
-        continue;
-      }
+    const section = path.slice(0, -1);
+    if (!_.has(defaultConfig, section)) {
+      sections.add(section.join("."));
+      continue;
+    }
 
-      const unrecognized = !_.has(defaultConfig, path);
-      const value = obj[key];
-
-      if (unrecognized) {
-        const section = typeof value === "object" ? path : parentPath;
-        results.push({
-          type: "unrecognized-key",
-          sourceName,
-          section,
-          path,
-        });
-      } else if (typeof value === "object") {
-        traverse(value, path);
-      }
+    if (!_.has(defaultConfig, path)) {
+      keys.set(path.join("."), section.join("."));
+      continue;
     }
   }
 
-  traverse(newConfig);
+  for (const [path, section] of keys) {
+    results.push({
+      type: "unrecognized-key",
+      sourceName,
+      section,
+      path,
+    });
+  }
+
+  for (const section of sections) {
+    results.push({
+      type: "unrecognized-key",
+      sourceName,
+      section,
+      path: section,
+    });
+  }
 
   return results;
 }
--- a/apps/studio/tests/unit/configuration/config.spec.js
+++ b/apps/studio/tests/unit/configuration/config.spec.js
@@ -63,14 +63,14 @@ minRes = 10
       {
         sourceName: "user",
         type: "unrecognized-key",
-        section: "generall",
-        path: "generall",
+        section: "ui.general",
+        path: "ui.general.load",
       },
       {
         sourceName: "user",
         type: "unrecognized-key",
-        section: "ui.general",
-        path: "ui.general.load",
+        section: "generall",
+        path: "generall",
       },
       {
         sourceName: "user",


apps/studio/tests/unit/configuration/keybindings.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+keys[] = mac: i
+keys[] = mac: j, linux: k
+keys[] = linux: l, m
+keys[] = n
+      `);
+
+      parseTags(config);
+
+      expect(config.test.keys).toStrictEqual([
+        { default: "a", mac: "b", $$type: "taggedString" },
```
rathboma, line 45: I think this output format is confusing. There can be multiple objects, each with repeating keys. Is the app supposed to treat each object differently?

I think something like this would be more standard and easier to deal with in code:

```js
[
  { scope: "default", values: ["a", "c"] },
  { scope: "mac", values: ["b", "d"] },
  ...
```

Or even an object with a key per-platform:

```js
{
  mac: ["b", "d"],
  default: ["a", "c"]
}
```
I would expect multiple objects each with repeated keys but no indication of heirarchy.

apps/studio/tests/unit/configuration/keybindings.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+        { default: "c", mac: "d", win: "e", $$type: "taggedString" },
+        { linux: "f", mac: "g", win: "h", $$type: "taggedString" },
+        { mac: "i", $$type: "taggedString" },
+        { mac: "j", linux: "k", $$type: "taggedString" },
+        { linux: "l", default: "m", $$type: "taggedString" },
+        "n",
+      ]);
+    });
+
+    //     it("should keep base key when platform qualifier does not match", () => {
```
rathboma, line 55: Do these need uncommenting?

apps/studio/tests/unit/configuration/keybindings.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+keys[] = a, mac: b
+keys[] = c, mac: d, win: e
+keys[] = linux: f, mac: g, win: h
+keys[] = mac: i
+keys[] = mac: j, linux: k
+keys[] = linux: l, m
+keys[] = n
+      `);
+
+      parseTags(config);
```
rathboma, line 42: I'm concerned about adding a loose configuration language (almost-json objects) inside an existing configuration language (ini).

Maybe this means we shouldn't be using ini, but rather something like yaml? If we truly need another level of structure then I'm not sure introducing our own sub-language is the most reliable way to achieve platform-specific configurations.

Have you thought about other ways to accomplish cross-platform configurations? Just spitballing:
- <platform>.yml config files
- `key_<platform>` keys, eg `quit_mac`
- using raw JSON as a value? `quit: { "mac": "a", "linux": "ctrl+q" }`

azmy60: My initial approach was similar to the second one, which was `key(platform)`, e.g. `quit(mac)`.

It was not really an issue until I jumped into customizing the CodeMirror keybindings. So the current structure follows how you'd define keybindings in cm:

https://github.com/codemirror/commands/blob/dbae3a14705d265ff13431fec6c5ca226fa600c5/src/commands.ts#L1086-L1091

There is a doc [here](https://codemirror.net/docs/ref/#view.KeyBinding) of how it works, and I personally think this structure makes a lot of sense. It answers problems like, how to make a keybind that applies to macOS only, or all platform except macOS in a much easier way compared to scoping the keybinds per platform.

Cause imagine if you do this:

```ini
startCompletion = ctrl+space
startCompletion_linux = alt+`
```

One would assume that `startCompletion_linux` will replace the base `startCompletion` if the platform is linux. But what should we do if we want both on linux? With that structure, we'd probably duplicate the base key:

```ini
startCompletion = ctrl+space
startCompletion_linux[] = ctrl+space
startCompletion_linux[] = alt+`
```

Instead of that, the cm-like structure would look nicer IMO:

```ini
startCompletion[] = ctrl+space
startCompletion[] = linux: alt+`
```

Also, I guess you're right when you call it json-like. If we want to use the raw JSON values, then I think it makes sense if we move away from ini. But I would go with toml instead of yaml to be honest cause it's much much similar to ini and probably makes migration easier.

azmy60: Not just look nicer. It's easier to handle too, because in code, we would have only one key (which is `startCompletion`), instead two (`startCompletion` and `startCompletion_linux`). That's what I've discovered before I finally use the current structure. And if we move away from ini to lets say toml, we probably don't need the extra parser cause it supports an inline table:

```toml
startCompletion = [
  "ctrl+space",
  { linux: "alt+`" }
]
```

That's probably not bad? not as verbose as raw json.

PR #3918, opened 2026-03-01
Add MySQL server status tab (CPU/RAM/Swap/Query cache-related metrics)

sirrodgepodge:
## Summary
This PR adds a new **Server Status** tab for MySQL/MariaDB connections so users can quickly inspect server-level metrics.

## What was implemented
- Added new core tab type: `server-status`
- Added new tab component: `TabServerStatus.vue`
  - Fetches stats via `connection.getServerStatistics()`
  - Supports manual refresh
  - Supports 5-second auto-refresh while tab is active
  - Displays:
    - Performance (connections, uptime, threads, slow queries, QPS)
    - Memory/buffer data (key buffer, InnoDB pool size/used)
    - Query cache stats (size/limit/hits/inserts/low memory prunes)
- Added menu entry under new-tab dropdown: **View Server Status**
  - Only shown for MySQL/MariaDB connections
- Added tab icon for server status tabs
- Added renderer client method `getServerStatistics()` -> `conn/getServerStatistics`
- Added tab matching behavior so only one server-status tab opens per connection

## Notes
- Metrics are sourced from existing backend support that reads `SHOW GLOBAL STATUS` and `SHOW GLOBAL VARIABLES`.
- Some values such as CPU/Swap can vary by MySQL distribution/plugin availability; this tab surfaces available status-oriented server metrics in Beekeeper UI.

## Testing
- Could not run project lint/tests in this environment because `yarn` is not installed in the execution host.

### DIFF
--- a/apps/studio/src/components/CoreTabs.vue
+++ b/apps/studio/src/components/CoreTabs.vue
@@ -92,6 +92,12 @@
           :tab="tab"
           :tab-id="tab.id"
         />
+        <ServerStatus
+          v-if="tab.tabType === 'server-status'"
+          :active="activeTab?.id === tab.id"
+          :tab="tab"
+          :tab-id="tab.id"
+        />
         <PluginBase
           v-if="tab.tabType === 'plugin-base'"
           :tab="tab"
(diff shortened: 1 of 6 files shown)


apps/studio/src/components/CoreTabs.vue
```diff
@@ -672,6 +680,13 @@ export default Vue.extend({
         this.createQuery()
       } else if (config.type === "shell") {
         this.createShell()
+      } else if (config.type === 'server-status') {
+        const tab = {
+          tabType: 'server-status',
+          title: 'Server Status',
+          unsavedChanges: false,
+        } as TransportOpenTab;
```
not-night-but: I'd agree with the AI on this one, we probably would want only one of these tabs to be able to be open, so we should check if there is a server-status tab already open first

not-night-but:
Hi @sirrodgepodge , would you be able to provide some screenshots of this PRs functionality?

not-night-but:
Closing as there has been no headway on this in a while

PR #3919, opened 2026-03-01
[CHORE] Add cross-platform E2E smoke tests

JeanPaiva42:
## Summary                                                                                             
  Adds a new CI workflow to verify the app launches and runs basic functionality on macOS, Windows, and Linux.
                                                                                                         
  ## Changes                                                                                             
  - New `e2e-smoke-test.yml` workflow runs a single basic E2E test across all three platforms            
  - Tests that the app opens and can execute a simple Postgres query                                     
  - Includes platform-specific setup (Python 3.11, Playwright browsers)                                  
  - Runs on PRs to catch platform-specific issues early                                                  
  - Full E2E suite continues running only on Linux                                                       
                                                                                                         
  ## Benefits                                                                                            
  - Catches platform-specific launch issues before merge                                                 
  - Minimal CI cost (single test vs full suite)                                                          
  - Early detection of breaking changes on macOS/Windows


.github/workflows/e2e-smoke-test.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Serve + Run Tests (Linux/macOS)
+        if: runner.os != 'Windows'
+        shell: bash
+        run: |
+          if [ "$RUNNER_OS" == "Linux" ]; then
+            xvfb-run --auto-servernum --server-args="-screen 0 1024x768x24" bash -c '
+            yarn run electron:serve > server.log 2>&1 &
+              sleep 15
+              cat server.log
```
not-night-but, line 59: Not sure why we need to view the `server.log` here? We could realistically just redirect the command to `/dev/null`

.github/claude/issue-templates.md
```diff
@@ -0,0 +1,129 @@
+# GitHub Issue Templates for Email-to-Issue Creation
```
not-night-but: I think you included this file by accident?

not-night-but (changes requested):
Two quick questions but otherwise looks good!

not-night-but (approved):

Issue #3832, opened 2026-01-31
Ability to righ click a saved query (in a workspace) and click 'share' to create a link to the query that can be opened in the app if the user has access


wmontgomery:
Have the share url be something like:

https://cloud.beekeeperstudio.io/queries/slug-name-of-query-xxxxx

where xxxxx a random set of characters

PR #3920, opened 2026-03-01
Feature/3832 share saved query

wmontgomery:
related to #3832 
3832-generate-share-link

3832-generate-share-link-2-x

Note: The share query modal that is on the ConnectionInterface is purely for dev reasons to not have to build the application to test the path


apps/studio/src/components/sidebar/ConnectionSidebar.vue
```diff
@@ -384,6 +384,34 @@ export default {
...
+        }
+        const { workspaceId, databaseId } = this.sharedQueryLink
+        const connectionConfig = cList.find(c => c.workspaceId === workspaceId && c.id === databaseId )
+
+        if (connectionConfig != null) {
+          this.connect(connectionConfig)
+        }
+      },
+      deep: true
+    },
```
not-night-but, line 400: I'm not totally sure what the purpose of this is? I'm getting vibes

apps/studio/src/components/sidebar/ConnectionSidebar.vue
```diff
@@ -403,6 +432,9 @@ export default {
       foldersLoading: 'loading',
       foldersError: 'error',
     }),
+    ...mapState(['workspaceId', 'sharedQueryLink', 'connected']),
+    ...mapGetters('credentials', { 'availableWorkspaces': 'workspaces'}),
+    ...mapState('settings', ['privacyMode']),
```
not-night-but: Looks like this is a merge kerfuffle. We get privacy mode down below in the getters

apps/studio/src/handlers/appDbHandlers.ts
```diff
@@ -181,6 +182,24 @@ export const AppDbHandlers = {
...
+      baseUrl: platformInfo.cloudUrl,
+    });
+    const response = await client.axios.get('/queries/open-from-share', {
+      params: { database: db, query },
+    });
+    if (response.status !== 200) {
+      throw new Error(`Cloud error (${response.status}): ${response.data?.message ?? 'Unknown error'}`);
+    }
+    return response.data;
+  },
```
not-night-but: I don't think this should be an appdb handlner, it doesn't really hvae anything to do with the appdb. Maybe add it to queryHandlers instead?

apps/studio/src/store/index.ts
```diff
@@ -51,6 +51,12 @@ const tablesMatch = (t: TableOrView, t2: TableOrView) => {
 }
 
 
+export interface SharedQueryLink {
+  databaseId: Nullable<string|number>,
+  queryId: Nullable<string|number>,
+  workspaceId: Nullable<string|number>
+}
+
```
not-night-but, line 78: I also don't think this belongs here. I think we move this to a multi process friendly file so we can also use it to properly type the share handler

apps/studio/src/store/index.ts
```diff
@@ -432,6 +443,15 @@ const store = new Vuex.Store<State>({
       await context.dispatch('connect', { config: conn, auth })
     },
 
+    async openSharedQuery(context, { db, query }: { db: string, query: string }) {
+      try {
+        const { databaseId, queryId, workspaceId } = await Vue.prototype.$util.send('appdb/share-query', { db, query });
```
not-night-but: Now that I'm thinking about this further, I don't know that we need to necessarily even have a handler for this? The cloud client should be automatically authenticated in the frontend, we can just this there maybe?

apps/studio/src/App.vue
```diff
@@ -179,7 +179,16 @@ export default Vue.extend({
       try {
         const { auth, cancelled  } = await this.$bks.unlock();
         if (cancelled) return;
-        await this.$store.dispatch('openUrl', { url: this.url, auth })
+
+        const match = this.url.match(/open-query\/([^/]+)\/([^/]+)/)
+        if (match) {
+          log.info('open shared query', this.url)
```
not-night-but: debug log :)

not-night-but (changes requested):
This looks good! I did have a couple questions, and I'm not sure that some of the logic needs to go in the backend, but it looks good to me other than that

apps/studio/src/components/sidebar/ConnectionSidebar.vue
wmontgomery: Yeah there were some gnarly merge conflicts. I'll clean this up. Oops!

apps/studio/src/components/sidebar/ConnectionSidebar.vue
wmontgomery: The workspace gets committed in the sharedQueryLink watch method below it. This was needed because they the data/connection items get set, for the workspace and I can get the configs out of that to then connect it for next steps

apps/studio/src/handlers/appDbHandlers.ts
wmontgomery: I was having a hard time making a guess where this should go and even this set up as a whole.

apps/studio/src/store/index.ts
wmontgomery: Cool. I can add it to `apps/studio/src/types.ts`

apps/studio/src/store/index.ts
wmontgomery: Perfect! I know how the cloudclient stuff works much better! Was able to just call it via cloudclient ezpz

PR #3923, opened 2026-03-02
Add ssh bastion port

### LINKED ISSUE #938: FEAT: allow setting a port for the Bastion Host (Jump Host) currently I can only use it on port 22
Most times our Jump host is actually not on port 22, only the machines on the internal network listen on 22 and the jump host is generally listening on some non standard port. Currently we can only use the ssh tunnel if the jump host is listening on port 22:

azmy60:
fix #938


azmy60:
This needs more than just bastion port. I could connect to the jump host but not the ssh endpoint because they use different username. The current form doesn't allow separate username and auth method. What if we want to upload private key for the ssh endpoint and use agent for the jump host? That's not possible right now, and it should be fairly easy to add.

PR #3945, opened 2026-03-05
feat: implement greengage integration

v-astakhov:


apps/studio/src-commercial/backend/handlers/connHandlers.ts
```diff
@@ -219,10 +220,13 @@ export const ConnHandlers: IConnectionHandlers = {
     const server = ConnectionProvider.for(config, osUser, settings);
     const abortController = new AbortController();
     state(sId).connectionAbortController = abortController;
-    await server?.createConnection(config.defaultDatabase || undefined).connect(abortController.signal);
+    const connection = server.createConnection(config.defaultDatabase || undefined);
```
remizov-arena: in original row ` await server?.createConnection` server can be undefined. Look like that it should be support in our changes

apps/studio/src/common/appdb/models/saved_connection.ts
```diff
@@ -134,7 +135,7 @@ export class DbConnectionBase extends ApplicationEntity {
   public get defaultSocketPath(): Nullable<string> {
     if (['mysql', 'mariadb'].includes(this.connectionType || '')) {
       return '/var/run/mysqld/mysqld.sock'
-    } else if (this.connectionType === 'postgresql') {
+    } else if (this.connectionType === 'postgresql' || this.connectionType === 'greengage') {
```
remizov-arena: may be combine to array `['postgresql', 'greengage']`? By anlogy with `['mysql', 'mariadb'].includes(this.connectionType || '')`

apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -67,6 +67,11 @@
                   :config="config"
                   :testing="testing"
                 />
+                <postgres-form
+                  v-else-if="config.connectionType === 'greengage'"
```
remizov-arena: may be combine by analogy with:
```
<mysql-form
  v-else-if="['mysql', 'mariadb', 'tidb'].includes(config.connectionType)"
```

apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -516,7 +521,12 @@ export default Vue.extend({
       try {
         this.testing = true
         this.connectionError = null
-        await this.$store.dispatch('test', this.config)
+        const result = await this.$store.dispatch('test', this.config)
+        console.log(result)
```
remizov-arena: remove pls

apps/studio/src/shared/lib/dialects/models.ts
```diff
@@ -28,6 +28,8 @@ export function dialectFor(s: string): Dialect | null {
   switch (s) {
     case 'cockroachdb':
       return 'postgresql'
+    case 'greengage':
```
remizov-arena, line 31: Is this not work in default fallback? `Dialects.find((d) => d === s) || null`

apps/studio/src/shared/lib/dialects/models.ts
```diff
@@ -78,7 +82,7 @@ export function FormatterDialect(d: Dialect): FormatterDialect {
   if (d === 'sqlserver') return 'tsql'
   if (d === 'sqlite') return 'sqlite'
   if (d === 'oracle') return 'plsql'
-  if (d === 'postgresql') return 'postgresql'
+  if (d === 'postgresql' || d === 'greengage') return 'postgresql'
```
remizov-arena: may be separate to two conditions? By analogy with cassandra, duckdb and surrealdb

apps/ui-kit/lib/components/sql-text-editor/customDialects.ts
```diff
@@ -0,0 +1,11 @@
+import {PostgreSQL, SQLDialect} from "@codemirror/lang-sql";
+
+const GREENGAGE_EXTRA_KEYWORDS = [
+  "distributed", "exchange", "inclusive", "list", "protocol", "resource",
+  "retrieve", "subpartition", "text", "replicated",
+].join(" ");
+
+export const GreengageSQL = SQLDialect.define({
+  ...PostgreSQL.spec,
+  keywords: (PostgreSQL.spec.keywords || "") + GREENGAGE_EXTRA_KEYWORDS,
```
remizov-arena: space between PostgreSQL.spec.keywords and GREENGAGE_EXTRA_KEYWORDS ?

apps/studio/src-commercial/backend/handlers/connHandlers.ts
v-astakhov: return initial logic

apps/studio/src/components/ConnectionInterface.vue
v-astakhov: removed

apps/studio/src/common/appdb/models/saved_connection.ts
v-astakhov: fixed

apps/studio/src/components/ConnectionInterface.vue
v-astakhov: fixed

apps/studio/src/lib/db/clients/greengage.ts
```diff
@@ -0,0 +1,9 @@
+import { PostgresClient } from "./postgresql";
+import { IDbConnectionServer } from "../backendTypes";
+import { IDbConnectionDatabase } from "../types";
+
+export class GreengageClient extends PostgresClient {
```
v-astakhov, line 5: not actual

apps/studio/src/shared/lib/dialects/models.ts
v-astakhov: fixed

apps/studio/src/shared/lib/dialects/models.ts
v-astakhov: I think it will work, but I just put additional condition to switch case

apps/ui-kit/lib/components/sql-text-editor/customDialects.ts
v-astakhov: add space beetween

not-night-but:
@v-astakhov Would you be able to provide a docker container in `docker-compose.yml` So that we can verify that the client works? Integration tests would also be super helpful, you should hopefully be able to just hook it up to the common tests if it uses postgres syntax

v-astakhov:
Updated the `docker-compose.yml `file and added integration tests for Greengage

not-night-but:
@v-astakhov It looks like you're just running the postgres container for the greengage test? We'd want the greengage container to run

apps/studio/src/lib/editor/CodeMirrorDefinitions.ts
```diff
@@ -13,6 +13,18 @@ CodeMirror.defineMIME("text/x-pgsql", {
...
+  // @ts-ignore
+  ...CodeMirror.resolveMode("text/x-pgsql"),
+  identifierQuote: '"',
+  hooks: {
+    // eslint-disable-next-line
+    // @ts-ignore
+    '"': CodeMirror.resolveMode("text/x-sqlite")?.hooks['"'],
+  },
+});
+
```
not-night-but: We don't need a specific mime type for this, you can just use the `x-pgsql` one, no?

apps/studio/src/shared/lib/dialects/models.ts
```diff
@@ -28,6 +28,8 @@ export function dialectFor(s: string): Dialect | null {
   switch (s) {
     case 'cockroachdb':
       return 'postgresql'
+    case 'greengage':
+      return 'greengage'
```
not-night-but, line 32: Yeah I'd probably just not have a dialect for greengage and just have it return postgresql like cockroach does

apps/studio/tests/integration/lib/db/clients/greengage.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+      ...PostgresTestDriver.config,
+      client: 'greengage' as const,
+    }
+    util = new DBTestUtil(config, 'banana', {
+      dialect: 'greengage',
+      defaultSchema: 'public',
+      ...PostgresTestDriver.utilOptions,
+    })
+    await util.setupdb()
+  })
```
not-night-but: Like I mentioned in my ping earlier, we should be using a greengage container here.

I'd also look into what knex client is being used here. It should be postgres

apps/studio/src/lib/editor/CodeMirrorDefinitions.ts
v-astakhov: For Greengage we keep a separate MIME (text/x-ggsql) and dialect data because Greengage extends PostgreSQL with constructs like DISTRIBUTED BY, REPLICATED, and other Greenplum-specific keywords. Without a custom dialect, these would not be highlighted correctly. The same reasoning applies to GreengageData and dialectFor('greengage') they are required for the editor to use the custom syntax highlighting.

v-astakhov:
I’ve implemented the Greengage container for integration tests and the dev migration for Greengage 6 and 7 as suggested.

apps/studio/tests/integration/lib/db/clients/greengage.spec.ts
```diff
@@ -0,0 +1,73 @@
+import fs from 'fs'
+import path from 'path'
+import { DockerComposeEnvironment, Wait } from 'testcontainers'
+import { DBTestUtil } from '@tests/lib/db'
+import { runCommonTests } from './all'
+
+const greengageTimeout = 600_000
+const composeDir = path.resolve(__dirname, '../../../../docker')
```
not-night-but: This looks like the wrong docker directory? We should just do this how we do the cockroach setup

apps/studio/tests/integration/lib/db/clients/greengage.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+describe('Greengage client (PostgreSQL-compatible)', () => {
```
not-night-but: If we're going to support both greengage 6 and 7, then we should be testing both of them. Take a look at the postgres test file for how we do that, perhaps just point to different compose files?

apps/studio/src/lib/editor/CodeMirrorDefinitions.ts
not-night-but: I get that, but here we are just mapping it to postgres, so it's not doing anything special? At the moment we don't really define custom mime types are handle any of them differently. Unless codemirror does anything differently or we are actually implementing something for greengage, I don't see this as being necessary

apps/studio/tests/integration/lib/db/clients/greengage.spec.ts
v-astakhov: Implement  greengage 6 support

apps/studio/tests/integration/lib/db/clients/greengage.spec.ts
v-astakhov: Update to cockroach setup

apps/studio/src/shared/lib/dialects/greengage.ts
```diff
@@ -0,0 +1,6 @@
+import { PostgresData } from "./postgresql";
+
+export const GreengageData = {
+  ...PostgresData,
+  textEditorMode: "text/x-ggsql" as const,
```
not-night-but: The mime type was removed so this should just use the default from postgres. We can keep this file for the future if we ever need to do something specific for greengage though

PR #3948, opened 2026-03-06
feat: support redis:// and rediss:// protocol schemes

yogasw:
Registered 'redis://' and 'rediss://' protocol schemes in the electron builder configuration. 

This allows the desktop application to be opened directly via deep links from browsers or terminals. I have manually verified this by updating the Info.plist in my local environment and confirmed that running 'open redis://localhost:6379' successfully triggers the application.


not-night-but (approved):

not-night-but:
Thanks for the contribution!

Issue #2370, opened 2024-08-20
SSH Tunnel does not re-establish itself after losing connection


rathboma:
Here’s a GitHub issue based on the email:

---

**Title:** SSH Tunnel Reconnection Issue After Network Change

**Description:**

**Issue:**  
When using SSH tunnels to connect to DB servers, Beekeeper Studio does not automatically re-establish the tunnel if the connection breaks (e.g., due to a change of Wi-Fi). As a result, the app continues to show the disconnected state, and actions like tapping on `# unknown` or reloading the page do not trigger a reconnection. Eventually, the app times out with a `connect ETIMEDOUT` error.

**Steps to Reproduce:**

1. Establish a connection to a DB server via an SSH tunnel.
2. Change the network (e.g., switch Wi-Fi).
3. Attempt to interact with the DB connection in Beekeeper Studio (e.g., tap on `# unknown` or reload a page that requires DB access).

**Expected Behavior:**  
Beekeeper Studio should detect the broken SSH tunnel and attempt to re-establish the connection before interacting with the DB.

**Actual Behavior:**  
The app does not reconnect automatically, leading to a timeout error. The only workaround is to close the DB window and reconnect manually, which results in losing the UI state.

**Environment:**

- Beekeeper Studio version: [insert version]
- Operating System: [insert OS]

**Additional Information:**  
This issue affects all DB servers connected via SSH tunnels. Any insight into an easy fix or further assistance would be appreciated.

**Contact:**  
Ulf (Original Reporter)

---

This issue description captures the essence of the email while structuring it in a way that's clear and actionable for a developer.

PR #3958, opened 2026-03-09
Add reconnect flow with connection-lost notification and reconnect prompt

azmy60:
1. Connection-lost notification - When conn/connection-lost fires from the backend, show a noty asking the user if they'd like to reconnect

    image

2. New "Reconnect" menu item

    image image]

## Limitation:
This issue is related to #2370, but this PR doesn't fix it entirely. In scenarios like switching WiFi networks, the OS-level TCP connection doesn't immediately fail; it hangs for roughly 60 seconds (at least in Mac) before the socket error is thrown. Only at that point the app can detect the loss and prompt the user.

I haven't tested using "ping" because it's not supported on many db drivers we use (only Postgres and ClickHouse have "ping"), but I believe it wouldn't solve the "hang" since it's an OS-level issue, and we probably won't get "pong" after a while too.

Adding a shorter timeout could solve the long wait, but that wouldn't allow running queries that actually take long.

I believe this is a common issue in basically any apps, and from what I've tested (using DataGrip, and even the `ssh` command) would have this "hang" issue.


azmy60:
@rathboma @not-night-but I know the tests aren't working yet, but I want to get your thoughts first about what this PR solves and the overall implementation of it.

rathboma:
Heads up on a conflict with #4755: this PR adds `state(sId).connection.connection.removeAllListeners()` to `conn/clearConnection`. #4755 makes `conn/clearConnection` handle a window whose state (or connection) is already gone after a utility restart — that `removeAllListeners()` call null-derefs in exactly that case. Whichever of these two PRs lands second should guard it with `?.` on both hops (`state(sId)?.connection?.connection?.removeAllListeners()`).

---

PR #3964, opened 2026-03-12
Feat/editable results

### LINKED ISSUE #2429: Feature Request: Enable In-Place Editing of Query Results
Users have expressed a desire to edit query result fields directly within the results view. This feature would be especially useful if there is a Primary Key (PK) in the result set. The proposed functionality includes:

1. In-place field editing.
2. An auto-update mode for immediate application of changes or an option to highlight changes and apply them later with a button.

This enhancement would greatly improve user experience and efficiency when working with query results.
***

### LINKED ISSUE #3191: FEAT: allow edit in selected rows
It's a bit annoying that after running a select I can't edit the rows returned, making me write a update query or go directly in the table and use the filters to get the row and then update it.

On the select result, allow editing and commit changes in the returned rows.

Just using the table and filters, very annoying when you have many different clauses.

**Additional context**
N/a

N/a

not-night-but:
Fix #3191
Fix #2429 

Alrighty, this is a big one! First of all, the workflows wont work, and neither will your local build if you want to test it unless you clone and build the corresponding sql-query-identifier branch [here](https://github.com/coresql/sql-query-identifier/pull/89). That also needs to be merged/released before we can do anything with this. 

If you want to test:
- Pull down the above branch
- Run `yarn build` in the root of that repository
- Update the package.json for both ui-kit and studio for sql-query identifier to be the path to the lib directory in the identifier repo :)

## Things this PR does:
Edit results of your queries in the query editor! We do this be analyzing the query and mapping any parsed columns/tables to their corresponding objects that we get from our `listTableColumns` and `listTables` functions.
image
image
image
image

## Things this PR does NOT do:
- Integration with manual commit mode, this will be added in a separate PR before `5.7` ideally
- Adding/removing rows. I think this would be super cool, but it's hard to wrap my mind around how that will work UI/UX wise with the possibility of multiple tables per row.
- Any sort of FK identification at the moment
- It currently also does not do anything special with data types, this may be something we should think about, but I haven't run into any issues with it yet?


apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1375,13 +1378,15 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
     })
   }
 
-  parseRowQueryResult(data: QueryResult, command: string, rowResults: boolean): NgQueryResult {
+  async parseRowQueryResult(data: QueryResult, command: IdentifyResult, rowResults: boolean): Promise<NgQueryResult> {
```
azmy60: Maybe this isn't supposed to use async?

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -798,9 +800,10 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
     const arrayMode: boolean = options?.arrayMode;
     const data = await this.driverExecuteMultiple(queryText, { arrayMode, tabId: options?.tabId });
 
-    const commands = this.identifyCommands(queryText).map((item) => item.type);
+    const commands = this.identifyCommands(queryText);
+    log.info("COMMANDS: ", commands)
 
-    return data.map((result, idx) => this.parseRowQueryResult(result, commands[idx], arrayMode));
+    return await Promise.all(data.map(async (result, idx) => this.parseRowQueryResult(result, commands[idx], arrayMode)));
```
azmy60: Same here.

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1784,7 +1789,7 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
 
   private identifyCommands(queryText: string) {
     try {
-      return identify(queryText);
+      return identify(queryText, { identifyColumns: true, identifyTables: true });
```
azmy60: What does the return data look like after the new arguments?

Also, I think these are the new features you added in sql-query-identifier, right? Just a thought: why not just make `identifyColumns` and `identifiyTables` true by default?

azmy60 (commented):
I haven't tested this locally yet, but at a glance, it seems like `getResultEditData` is the core part of this implementation. Could we have a dead simple unit test for the `getResultEditData` if possible? It's probably easier to grasp the code from reading the tests. Though, by the time I finish writing this I feel like I know what's going on lol.

apps/studio/src/lib/db/clients/postgresql.ts
not-night-but: Oop that's from before a refactor

apps/studio/src/lib/db/clients/postgresql.ts
not-night-but: Yup same here, whoops

apps/studio/src/lib/db/clients/postgresql.ts
not-night-but: Oh we actually don't need these here. It's the same return type, just also with a `columns[]` and a `tables[]`. I'm not sure why we originally had `identifyTables` (I believe that was you lol), but now I think we should have it be an opt in feature because it's more processing that isn't always necessary?

rathboma:
Using the feature feedback:
1. 'edit results' is present even if the results are not editable (no primary keys). Maybe that's fine, but clicking it shows no feedback that editing isn't possible
2. It's not super clear when I'm in 'editing results mode'. 
3. When I click 'edit results', the button turns to 'reset'. Maybe it should say 'stop editing'
4. Clicking 'reset' when in edit mode with pending changes shouldn't exit editing mode, simply reset the changes. Only clicking 'stop editing' should exit.

Quick video walkthrough of my thoughts: 

https://github.com/user-attachments/assets/08d0480c-2257-4f3d-94e6-dd9005b55246

rathboma:
Couple more points:

1. I think it should say 'Edit data' rather than edit results -- you're not editing the results, you're editing the live table data.
2. There should be a tooltip when users hover over it saying 'Edit table data directly from the query results'
3. First time users should get a dismissable notification, similar to Azmi's AI Shell notification that points to the button and says something like `You can edit table data directly from query results. [Docs]()`, the first time they run a query and the results are editable. Once dismissed it doesn't pop up again.

rathboma:
If they click it and they're not on a paid plan - it should show the upgrade modal

apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -316,6 +348,144 @@
             },
           },
         });
+
+        this.tabulator.on('cellEdited', this.cellEdited);
+      },
+      cellEdited(cell: CellComponent) {
```
rathboma, line 470: this is a very long function and is a little scary, possible to break out a little and test at all?

apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -316,6 +348,144 @@
...
+        // Should this be id?
+        const field: FieldDescriptor = this.result.fields.find((f) => f.id === cell.getField());
+        const fieldEditData: FieldEditData = this.editData?.get(cell.getField());
+        if (!field) {
+          log.warn('Could not find matching field for', cell.getField());
+          return;
+        }
+
+        const pkFields: Map<string, FieldEditData> = new Map(
```
rathboma, line 477: should this be a computed property?

apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -316,6 +348,144 @@
...
+            .entries()
+            .filter(([_id, editData]) => editData?.isPK && editData?.linkedTable === fieldEditData?.linkedTable && editData?.linkedSchema === fieldEditData?.linkedSchema)
+        );
+        const pkCells: CellComponent[] = cell.getRow().getCells().filter((c) => pkFields.has(c.getField()));
+
+        if (cell.getOldValue() == cell.getValue()) {
+          return;
+        }
+
+        log.info('PK CELLS: ', pkCells)
```
rathboma: debug, also check for other log statements

apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -316,6 +348,144 @@
...
+
+        cell.getElement().classList.add('edited');
+        const currentEdit = _.find(this.pendingChanges.updates, { key: key });
+
+        if (typeof currentEdit?.oldValue === 'undefined' && cell.getValue() == null) {
+          // don't do anything because of an issue found when trying to set to null, undefined == null so was getting rid of the need to make a change\
+        } else if (currentEdit?.oldValue == cell.getValue()) {
+          this.$set(this.pendingChanges, 'updates', _.without(this.pendingChanges.updates, currentEdit));
+          cell.getElement().classList.remove('edited');
+          this.propogateChanges(pkCells, cell, key, true);
```
rathboma: There's a few branching paths in this function that all lead to calling `propogateChanges`, could these paths be broken out as individual functions?

Why break them out -- right now it's hard to reasonabout where the function will return or not return, having some plain english names for the states / conditions, plus a clear action (another function) would help I think.

```
 if (inStateA) {
  doX()
} else if (inStateB) {
  doY()
}
```

So this function becomes a sort of dispatcher

apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
```diff
@@ -193,6 +209,98 @@ export abstract class BasicDatabaseClient<RawResultType extends BaseQueryResult,
     return Promise.resolve([]);
   }
 
+  async getResultEditData(queryText: string, fields: FieldDescriptor[]): Promise<FieldEditData[]> {
```
rathboma, line 212: In contrast this function is pretty neat and more clearly advertises what it does

rathboma (dismissed):
Minor comment about code legibility, otherwise looks great.

I think this needs more tests before going live. If you break out some of the logic in the components into their own functions it should be possible to unit test them without having to rely on e2e tests. We have some component tests already.

My particular concern is: If this feature misbehaves it could have very bad consequences on a user's database, so let's do some due diligence right now and add tests where we are able to find edge cases and ensure there are no regressions from future development.

apps/studio/src/components/editor/ResultTable.vue
not-night-but: I don't see how it could be? We have to match it to the `fieldEditData` declared above

azmy60 (commented):
I have an idea that could be for v2 if we want:

So I always want to double click/press enter to edit the data (it's the muscle memory lol). It'd be great if later we have a configuration for enabling this. If we're aiming for spreadsheet like experience, I personally think this should be enabled by default, just like any spreadsheet app.

not-night-but:
@azmy60 I agree that would be nice. I guess the main thing would be just communicating the first double click well. It might lag a bit because of the queries that have to be run to get the edit info. And then it's also possible that the cell you clicked isn't actually editable.

apps/studio/tests/unit/components/editor/ResultTable.spec.ts
```diff
@@ -0,0 +1,1055 @@
...
+// Tests
+// ===========================================================================
+
+describe('ResultTable - Cell Editing Methods', () => {
+  // -----------------------------------------------------------------------
+  // cellEdited
+  // -----------------------------------------------------------------------
+  describe('cellEdited', () => {
+    it('should return early when editData has no matching field', () => {
+      const ctx = createContext()
```
azmy60, line 221: Just an opinion: I'm not sure if I like the way the tests work. If this requires a lot of mocks, then it might be suitable as integration tests rather than unit tests. And then, I would avoid using mocks  at all.

azmy60: I know I wrote a lot of mocks in Tabulator repo as well lol (not me but Claude), but I'd want to refactor those one day.

apps/studio/src/components/editor/QueryEditorStatusBar.vue
```diff
@@ -21,8 +21,12 @@
             <select
               name="resultSelector"
               id="resultSelector"
-              @change="selectedResult = parseInt($event.target.value);"
+              @change="selectedResult = parseInt($event.target.value)"
               class="form-control"
+              @mouseover="showSwitch = editing && changesCount > 0"
+              @mouseleave="showSwitch = false"
+              :disabled="editing && changesCount > 0"
+              v-tooltip="{ content: 'Discard or apply your changes to switch result sets', trigger: 'manual', show: showSwitch }"
```
azmy60, line 29: @rathboma For tooltips, should we use `v-tooltip` or `title` (native tooltip)?

azmy60 (approved):

apps/studio/src/components/editor/QueryEditorStatusBar.vue
azmy60: It's just more native in mac if we use `title`.

PR #3973, opened 2026-03-17
Use ~/.ssh/config in ssh agent auth type

azmy60:
The following ssh options can be derived from `~/.ssh/config`:

- Hostname
- Port
- IdentityFile
- User

For example, suppose this is the ssh config that we have:

```config
Host dockertest
  Hostname 192.168.1.20
  Port 2222
  IdentityFile ~/.ssh/dockertest_key
  User beekeeper
```

You should be able to use the ssh tunnel by entering `dockertest` as the host and your user name to the ssh tunnel form.


azmy60:
Note: Proxy host is not added here but I'm working on a separate PR that suports multiple proxy / jump hosts and each host can use different auth method (including the ssh agent).

not-night-but:
@azmy60 Any chance we can add some tests for the parsing at least?

Issue #1186, opened 2022-04-22
FEAT: Support pass-through windows authentication for SQL Server without re-entering user/pass/domain


rathboma:
**Is your feature request related to a problem? Please describe.**
With SQL Server you should be able to 'pass through' your windows credentials to authenticate to the database without having to provide a username, password, or domain (the OS should give those to Beekeeper Studio).

I'm not sure if this is possible with the node driver. Needs investigation.

Ripped from #534

PandaSITT:
.NET supports "Trusted Connection" with their default [SQL Connection](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlconnection?view=dotnet-plat-ext-6.0). Implementing this for windows could be the solution.

rklec:
Indeed a good idea, [DBGate e.g. supports this](https://github.com/dbgate/dbgate).

sfscgjohnson:
Is there a roadmap of when features are planned or a prioritization vote?  This would be high on my list.

LucGendre2023:
A most needed feature, indeed! Please prioritize this issue.

Clifra-Jones:
With the eventual demise of Azure Data Studio and the SQL Server Extension for VS Code not being up to speed (and the refusal of Microsoft to add in features like the management extension from ADS to VS Code) This application could be a very appealing replacement for ADS. But without an Integrated Security option it isn't an option for corporate DBAs or Devs.
It's a simple parameter to be added to the connection.

And, at this time ADS locks up on Ubuntu 24.04 so it is now essentially dead for users of the most popular Linux desktop,

Issue #1009, opened 2022-02-01
FEAT: Multiple Filters


PHLAK:
**Is your feature request related to a problem? Please describe.**

When filtering a table I often wish to filter by multiple columns.

**Describe the solution you'd like**

The ability to add multiple filters. Ideally this would be a button to let me add another filter that would be applied in addition to the first.

**Additional context**

This is inspired by TablePlus, you can see mutliple filters in their documentation [here](https://docs.tableplus.com/gui-tools/filter).

rathboma:
So there is a button that switches the filter to 'sql mode', that allows you to enter as many filters as you like.

For now this is the 'multi filter' solution. Going to close the issue, not because I don't care about this, but with my limited time there are just so many other things to prioritize.

rathboma:
Ok, I'm going to re-open this. I think we should do it.

Some notes for implementation:
- When switching from multiple filters -> SQL it should try to convert the filters to SQL
- When switching back, it should keep the filters if the SQL wasn't modified (so if you wanna just see what the button does), but discard the filters if the SQL is edited
- It should 'look nice' on the table view and feel natural, no weird scroll areas

azmy60:
I came across some good examples from google images lol. This is one of them. @rathboma Do you think something like this would work?

![googled it](https://blog.coupler.io/wp-content/uploads/2023/06/image-15-1024x536.png)

It has AND / OR conditions. It's a popup that sits right after the filter button. For us, we could probably use a dialog that shows in the center of the screen like Column Filter.

rathboma:
@azmy60 Make sure you `git pull` before working on this, as I changed that part of the tabletable last week to accomodate row selection.

I think for now just keep things simple and functional. I don't want to totally redesign that part of the view right now.

In the future we can do a full redesign -- I have an idea of how it should look when we get there.

Issue #660, opened 2021-05-20
QUESTION: Run in portable mode with a switch


naderi:
**Details:**
Is there a possibility to run the file "Beekeeper Studio.exe" in portable mode. Something like:
```
Beekeeper Studio.exe --portable
```

**Version Information**
Operating System: Windows 10
App Version: 1.11.5

rathboma:
I have to package the portable mode differently I'm afraid -- but there is a portable version of Beekeeper Studio you can download from the website!

Issue #534, opened 2021-01-30
BUG: can't connect SQL Server (improve documentation and tooltip help for Windows Authentication integration, don't ask port, don't ask user / password)


aisbergde:
**Describe the bug**
it should be clear and documented how to connect to MS SQL Server using Windows Authentication
user and password are required, but if the are entered, they are in conflict with Windows Authentication

For a new user it is not clear what he should enter to connect with MS SQL Server using Windows Integrated Authentication
There is a very poor instruction in the tooltip after domain:

>Set 'domain' to be logged in using Windows Integrated Authentication

I tried to enter the name of my PC as domain, I also tried the string "domain" and also "'domain'", but anyway I get some error that the username property must be a string

But If I want to use Windows Integrated Authentication, then I will not enter a user or password. It should be possible to connect without a user and password required in this case.

So I entered some dummy user and password. But now I get the error that a login for an untrusted domain can't be used with Integrated authentication.

Next issue is a required port.
Normally I don't care about the port of SQL Server instances because name or name\instance is fine.
How a "normal" user should know the port if it is not the standard port?

First I need to search the internet, then I need to connect to the server using another frontend.
And then I can execute:

```sql
USE master
GO
xp_readerrorlog 0, 1, N'Server is listening on' 
GO
```

Or something else.
But it looks like the port I get this way doesn't work.

I have several instances
- localhost (standard port, should be fine)
- localhost\sql2019 (no idea about the port, the query above gives me 53152 or 50365)  
  but whatever I enter I get the error "Port for sql2019 not found in localhost"
- there are also some localDB instances, but I did not try to connect

It should be possible to connect without knowing the port but only based on name or name\instance

**To Reproduce**
see above

**Expected behavior**
- It should be possible to connect without knowing the port but only based on name or name\instance
- To use Windows Authentication there should be a checkbox, not a "domain" entry and if the option is selected, then user and password should be allowed to be empty.
- It should be possible to omit the port because server or server\instance should be enough

**Screenshots / Gifs**

**Version Information (please complete the following information):**
 - OS: Windows 10
 - App Version 1.9.1

**Additional context**
My goal to evaluate beekeeper is to see, if and how I can enter data, especially multi line data.
[...]

PR #3982, opened 2026-03-23
Add Windows Authentication auth option for SQL Server and dynamic switch to msnodesqlv8 if chosen

mattpcanales:
With Microsoft ending support for Azure Data Studio in Feb 2026, I am left without a clean, DB focused tool that I can use to keep my SQL queries separate from VSCode at work. Beekeeper Studio seems like a great fit, but the requirement of storing credentials isn't ideal for my corporate windows environment, and we aren't fully on Entra ID yet. Unfortunately the pure javascript Tedious driver cannot leverage Microsoft's SSPI bindings required for kerberos without a wrapper of some kind.

There are several open issues on beekeeper-studio that are requesting a solution to this:
#1186, #48, and #534 
as well as more on tedious github repo to show discussion on tedious limitations/roadmap.
[#660](https://github.com/tediousjs/tedious/issues/660), [#415](https://github.com/tediousjs/tedious/issues/415), [#1009](https://github.com/tediousjs/tedious/issues/1009), and [#306](https://github.com/tediousjs/tedious/issues/306)

So I know I am not alone in hoping for this to be addressed, and with ADS end of support I assume this is only increasing recently

The cleanest solution I could come up with involved an added optional dependency for msnodesqlv8 which does offer this support. On top of this node module, the SQL driver on the local system must be ODBC 17 or 18 to fully support TLS1.2
The basic SQL Server driver works without TLS 1.2, but is no longer officially supported by Microsoft past SQL 2012.

I do not notice any difference in the experience between connecting with integrated auth with msnodesqlv8 and using password with mssql + the tedious driver, but I haven't tested every edge case.

I ran the unit test in /app/studio and all passed.

If you have any feedback, even harsh criticism please share. I am happy to consider this just a learning experience since it is my first work with electron and first time contributing to a project. I would love to learn if there is a plan to include this feature some other way, and what architecture is planned to do so, or if the plan is to simply support Azure auth only. 

Thank you for all your work!

### DIFF
--- a/apps/studio/electron-builder-config.js
+++ b/apps/studio/electron-builder-config.js
@@ -37,7 +37,8 @@ module.exports = {
   ],
   afterPack: "./build/afterPack.js",
   asarUnpack: [
-    'package.json'
+    'package.json',
+    '**/msnodesqlv8/build/Release/*.node'
   ],
   extraResources: [
     {
--- a/apps/studio/src/components/connection/SqlServerForm.vue
+++ b/apps/studio/src/components/connection/SqlServerForm.vue
@@ -5,11 +5,12 @@
       <!-- need to take the value -->
       <select name="" v-model="authType" id="">
         <option value="default">Username / Password</option>
+        <option value="windows">Windows Authentication</option>
         <option :key="`${t.value}-${t.name}`" v-for="t in authTypes" :value="t.value">{{t.name}}
         </option>
       </select>
     </div>
-    <common-server-inputs v-show="!azureAuthEnabled" :config="config">
+    <common-server-inputs v-show="!azureAuthEnabled" :config="config" :hide-credentials="windowsAuthEnabled">
       <div class="advanced-connection-settings">
         <h4 class="advanced-heading">
           SQL Server Options
@@ -68,12 +69,18 @@
     components: {CommonEntraId, CommonServerInputs, CommonAdvanced },
     props: ['config'],
     mounted() {
-      this.authType = this.config?.azureAuthOptions?.azureAuthType || 'default'
-      this.azureAuthEnabled = this.config?.azureAuthOptions?.azureAuthEnabled || false
+      if (this.config?.windowsAuthEnabled) {
+        this.authType = 'windows'
+        this.windowsAuthEnabled = true
+      } else {
+        this.authType = this.config?.azureAuthOptions?.azureAuthType || 'default'
+        this.azureAuthEnabled = this.config?.azureAuthOptions?.azureAuthEnabled || false
+      }
     },
     data() {
       return {
         azureAuthEnabled: false,
+        windowsAuthEnabled: false,
         authType: 'default',
         authTypes: AzureAuthTypes,
       }
@@ -82,6 +89,13 @@
       async authType() {
         if (this.authType === 'default') {
           this.azureAuthEnabled = false
+          this.windowsAuthEnabled = false
+          this.config.windowsAuthEnabled = false
+          this.config.azureAuthOptions.azureAuthType = undefined
+        } else if (this.authType === 'windows') {
+          this.azureAuthEnabled = false
+          this.windowsAuthEnabled = true
+          this.config.windowsAuthEnabled = true
           this.config.azureAuthOptions.azureAuthType = undefined
         } else {
           if (this.$store.getters.isCommunity) {
@@ -89,6 +103,8 @@
             this.authType = 'default'
           } else {
             this.azureAuthEnabled = true
+            this.windowsAuthEnabled = false
+            this.config.windowsAuthEnabled = false
             this.config.azureAuthOptions.azureAuthType = this.authType
           }
         }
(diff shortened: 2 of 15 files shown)


rathboma:
Hey! Wow, this is super cool. Thanks for the contribution. We'll review the code and get back to you.

Off the top of my head I bet there are a couple of changes needed in esbuild.js to externalize the dependency, and might not work when using the full 'build' (vs in dev mode) without that change.

I also need to think about how we properly integration test this. Would be good to have some way to automated-test this with a windows VM / container of some sort.

apps/studio/src/lib/db/clients/sqlserver.ts
```diff
@@ -1072,7 +1072,50 @@ export class SQLServerClient extends BasicDatabaseClient<SQLServerResult, Transa
     await super.connect();
 
     this.dbConfig = await this.configDatabase(this.server, this.database, signal)
-    this.pool = await new ConnectionPool(this.dbConfig).connect();
+
+    if (this.server.config.windowsAuthEnabled) {
+      if (process.platform !== 'win32') {
+        throw new Error('Windows Authentication is only supported on Windows.')
+      }
```
not-night-but, line 1079: We should use platform info here instead.
```suggestion
      import platformInfo from "@/common/platform_info"
      if (!platformInfo.isWindows) {
        throw new Error('Windows Authentication is only supported on Windows.')
      }
```

apps/studio/src/lib/db/clients/sqlserver.ts
```diff
@@ -1072,7 +1072,50 @@ export class SQLServerClient extends BasicDatabaseClient<SQLServerResult, Transa
...
+      // defaulting to "SQL Server Native Client 11.0"), passes
+      // Trusted_Connection as a boolean (should be yes/no), and doesn't include TrustServerCertificate.
+      // beforeConnect also preserves mssql's instance name and SSH tunnel server handling.
+      const trustCert = this.dbConfig.options?.trustServerCertificate
+      this.pool = await new sqlWindows.ConnectionPool({
+        ...this.dbConfig,
+        beforeConnect: (cfg: any) => {
+          cfg.conn_str = cfg.conn_str
+            .replace(/Driver=[^;]*/i, `Driver={${driver}}`)
+            .replace(/Trusted_Connection=[^;]*/i, 'Trusted_Connection=yes')
```
not-night-but, line 1112: Looks like we can pass the driver if we update to at least [v12.2.0](https://github.com/tediousjs/node-mssql/releases/tag/v12.2.0)
Usage: 
https://github.com/tediousjs/node-mssql#windows-authentication-example-using-msnodesqlv8

It also looks like you would be able to pass trusted connection in the config there as well

apps/studio/src/components/connection/SqlServerForm.vue
```diff
@@ -5,11 +5,12 @@
...
       <select name="" v-model="authType" id="">
         <option value="default">Username / Password</option>
+        <option value="windows">Windows Authentication</option>
         <option :key="`${t.value}-${t.name}`" v-for="t in authTypes" :value="t.value">{{t.name}}
         </option>
       </select>
     </div>
-    <common-server-inputs v-show="!azureAuthEnabled" :config="config">
+    <common-server-inputs v-show="!azureAuthEnabled" :config="config" :hide-credentials="windowsAuthEnabled">
       <div class="advanced-connection-settings">
```
not-night-but, line 14: We should think about cleaning up this form for Windows Auth. Looking at how SSMS handles this, we probably don't need these sql server options here? Correct me if I'm wrong please :)

apps/studio/src/components/connection/SqlServerForm.vue
```diff
@@ -5,11 +5,12 @@
       <!-- need to take the value -->
       <select name="" v-model="authType" id="">
         <option value="default">Username / Password</option>
+        <option value="windows">Windows Authentication</option>
```
not-night-but, line 8: We just shouldn't show this option if we're not on windows. In the frontend you should just be able to use `window.platformInfo.isWindows` to check (`apps/studio/src/common/IPlatformInfo.ts` will show you what all is available on that object)

apps/studio/electron-builder-config.js
```diff
@@ -37,7 +37,8 @@ module.exports = {
   ],
   afterPack: "./build/afterPack.js",
   asarUnpack: [
-    'package.json'
+    'package.json',
+    '**/msnodesqlv8/build/Release/*.node'
```
not-night-but, line 41: I'm not sure this is necessary?

not-night-but (changes requested):
@mattpcanales Okay so I think the code mostly looks good, but I can't actually seem to connect with windows auth. It just hangs indefinitely, so we should probably make sure there's at least a connection timeout there (also if you know of anything that can cause that kind of issue please let me know. We should upgrade the `mssql` package so we can actually use the configuration object for the drivers. And I don't love how many changes there are to the `yarn.lock` It should not be that big of a diff. It looks like something has caused issues for yarn installs on other platforms

rathboma:
Closing this as support shipped with Beekeeper Studio v6 for both windows auth and kerberos

PR #3985, opened 2026-03-24
Feature/1679 run all to run current

wmontgomery:
image

image


rathboma:
@wmontgomery so just FYI we also need to add a configuration to default.config.ini to allow users to change the default from run current to run all if they want to. We have one right now that changes the keyboard shortcut for each, but that's silly, we should have a primary/secondary action and bind that to the keyboard shortcut.

so like:

primaryAction: submitCurrent
primaryActionKeybinding: ctrl+enter
secondaryAction: submitAll
secondaryActionKeybinding: ...etc

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -198,37 +198,86 @@
...
               class="btn btn-primary btn-small"
               :disabled="this.tab.isRunning || running"
               menu
             >
               <i class="material-icons">arrow_drop_down</i>
-              <x-menu>
+              <x-menu v-if="isPrimaryRunCurrentQuery">
+                <x-menuitem @click.prevent="submitCurrentQuery">
+                  <x-label>Run Current</x-label>
+                  <x-shortcut value="Control+Enter" />
```
azmy60: This shortcut is still harcoded.

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -198,37 +198,86 @@
...
+                </x-menuitem>
                 <x-menuitem @click.prevent="submitTabQuery">
-                  <x-label>{{ hasSelectedText ? 'Run Selection' : 'Run' }}</x-label>
+                  <x-label>{{ hasSelectedText ? 'Run Selection' : 'Run All' }}</x-label>
+                  <x-shortcut value="Control+Shift+Enter" />
```
azmy60: Same here.

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -198,37 +198,86 @@
...
+                    class="material-icons menu-icon"
+                  >
+                    stars
+                  </i>
+                </x-menuitem>
+              </x-menu>
+              <x-menu v-else>
+                <x-menuitem @click.prevent="submitTabQuery">
+                  <x-label>{{ hasSelectedText ? 'Run Selection' : 'Run All' }}</x-label>
                   <x-shortcut value="Control+Enter" />
```
azmy60: Harcoded shortcut

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -198,37 +198,86 @@
...
                 </x-menuitem>
                 <x-menuitem @click.prevent="submitCurrentQuery">
                   <x-label>Run Current</x-label>
                   <x-shortcut value="Control+Shift+Enter" />
```
azmy60: Harcoded 🚑

azmy60 (changes requested):
Some hardcoded labels could be misleading if the settings are changed.

apps/studio/src/components/TabQueryEditor.vue
wmontgomery: Oops, missed those ones. Thanks for catching them!

wmontgomery: I updated all these to pull from the bksconfig. Also updated what I could in the ShortcutHints component.

apps/studio/src/components/editor/ShortcutHints.vue
```diff
@@ -144,12 +109,32 @@
...
     props: ['isMongo'],
     data() {
       return {
 
       }
     },
+    methods : {
+      getShortcut(shortcut) {
+        console.log('!!!')
+        console.log(shortcut)
```
not-night-but: Remove

apps/studio/src/components/editor/ShortcutHints.vue
```diff
@@ -144,12 +109,32 @@
   </div>
 </template>
 <script type="text/javascript">
+  import { convertKeybinding } from '@/common/bksConfig/BksConfigProvider'
```
not-night-but: I think we should probably just add this to the bks plugin or something so we can access it anywhere in the frontend. Could even shorten it because we should probably always be calling it with `'ui'` and `this.$bksConfig.platformInfo.platform` and have it also handle whether or not it's an array. So we can just call it with `this.$getUiKeybind(shortcut)` or something.

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -198,37 +198,86 @@
...
+            <x-button
+              v-else
+              class="btn btn-primary btn-small"
+              :v-tooltip="displayShortcut(this.$bksConfig.keybindings.queryEditor.secondaryQueryAction)"
               @click.prevent="submitTabQuery"
               :disabled="this.tab.isRunning || running"
             >
-              <x-label>{{ hasSelectedText ? 'Run Selection' : 'Run' }}</x-label>
+              <x-label>Run All</x-label>
             </x-button>
```
not-night-but: I feel like we can just have one button here. We should just have `primaryActionText`, `primaryActionShortcut`, `secondaryActionText`, `secondaryActionShortcut`, etc. That makes the markup much easier to read imo. And looking at this I feel like the tooltip for run all might be wrong anyways? Because if it's being displayed, shouldn't it be the primaryAction?

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -198,37 +198,86 @@
...
+                  <x-label>Run All to File</x-label>
+                  <i
+                    v-if="isCommunity"
+                    class="material-icons menu-icon"
+                  >
+                    stars
+                  </i>
+                </x-menuitem>
+              </x-menu>
+              <x-menu v-else>
```
not-night-but: And if we do the above, we don't have to have these separate menus. We an just have a menu with Primary action, secondary action, run current to file and run all to file.

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -1569,6 +1679,9 @@
         clearInterval(this.timerInterval);
         this.timerInterval = null;
       },
+      displayShortcut(shortcut: string|string[]) {
+        return convertKeybinding('ui', Array.isArray(shortcut) ? shortcut[0] : shortcut, this.$bksConfig.platformInfo.platform).join('')
+      },
```
not-night-but: Why do we allow for an array if we don't actually really handle the array case?

not-night-but (changes requested):
A few thoughts. I don't love that we're having to add so much markup, and I think we can simplify it a bit. Feel free to yell at me :)

Also, I'm worried about just making this change quietly. I do wonder if we can somehow add a tooltip that shows up for users that already had the app installed or something? If someone is very used to using `ctrlOrCmd+shift+enter` to submit their current query, and then they do it and it runs their entire tab I think they may get annoyed? Idk @rathboma , thoughts?

apps/studio/src/components/TabQueryEditor.vue
wmontgomery: This is more for the display the shortcut. The shortcuts respect all keybindings.

apps/studio/src/components/editor/ShortcutHints.vue
wmontgomery: OK, have that plugin added.

apps/studio/src/components/editor/ShortcutHints.vue
```diff
@@ -22,33 +22,19 @@
...
-        v-if="$config.isMac"
       >
-        <span>⌘</span><span>⇧</span><span>Enter</span>
-      </div>
-      <div
-        class="shortcut"
-        v-else
-      >
-        <span>Ctrl</span><span>Shift</span><span>Enter</span>
+        <span v-for="sc in this.getShortcut(this.$bksConfig.keybindings.queryEditor.secondaryQueryAction)" :key="sc">{{ sc }}</span>
```
azmy60: I would use `$bksConfig.getKeybindings`, for example:

```vue
<span v-for="sc in getShortcut('queryEditor.secondaryQueryAction')" :key="sc">{{ sc }}</span>

<script lang="ts">
import { KeybindingPath } from "@/common/bksConfig/BksConfigProvider";

export default {
  methods: {
    getShortcut(path: KeybindingPath) {
      const result = this.$bksConfig.getKeybindings("ui", path)
      return Array.isArray(result) ? result[0] : result;
    }
  }
}
</script>
```

apps/studio/src/plugins/UiKeybind.ts
```diff
@@ -0,0 +1,13 @@
+import Vue from 'vue';
+import { convertKeybinding } from '@/common/bksConfig/BksConfigProvider'
+
+export function UiKeybind(shortcut:string|string[]): string[] {
```
azmy60: I think this isn't needed. We have `$bksConfig.getKeybindings('ui', path)` for this exact scenario.

apps/studio/default.config.ini
```diff
@@ -274,14 +274,21 @@ altOpen = ctrlOrCmd+enter
...
+[settings.queryEditor]
+; primary and secondary query actions must be one of the following 2 options
+; "submitCurrentQuery" will only run the active query
+; "submitTabQuery" will run all queries or what is highlighted
+primaryQueryAction=submitCurrentQuery
+secondaryQueryAction=submitTabQuery
+
 [keybindings.queryEditor]
 selectEditor = ctrlOrCmd+l
-submitTabQuery[] = ctrlOrCmd+enter
```
azmy60, line 290: This could be dangerous. If a user modified this before (let's say they switched run all with run current), their modified shortcut will be reverted. They also get a warning dialog when they start the app, because this key is now missing.

How should we handle that? Ideally we'd want to migrate these but we don't have a feature that overwrites config files yet. @rathboma How would you address this situation? I personally would not change these for now.

But if we do, we probably don't want to replace them but just introduce the new keys and keep the old keys. The logic is not straight forward, but it's safer, it might look like this:

1. Keep the old keys undeleted (submitTabQuery, submitCurrentQuery, etc.)
2. Use the old keys if defined in the config like usual
3. If the new keys (primaryQueryAction, etc.) are defined OR the old keys are NOT defined, use the new keys.
4. If the user is using the old keys, tell them that to review their configs with a less disruptive warning like a noty, for example:
- "Hey we are deprecating some keys, please review"
- [Remind me later] [Review]

apps/studio/config-metadata.json
```diff
@@ -56,10 +56,10 @@
       "label": "Query Editor",
       "properties": [
         { "key": "selectEditor", "label": "Focus Editor" },
-        { "key": "submitTabQuery", "label": "Run All Queries" },
```
azmy60, line 60: I would add a deprecated flag here, and a reason text about why it's deprecated.

azmy60 (changes requested):
This is good cause it's really straight forward. The only real issue is users can lose their "submit query" keybindings because they are removed/replaced. It'll be dangerous if users have modified the "submit current query" keybinding, but all of a sudden, it changes to "submit all queries".

apps/studio/src/plugins/UiKeybind.ts
wmontgomery: Does it also convert them to be shown on the app? All for using that instead of adding a new plugin to the system.

azmy60: Wait, I thought `getKeybindings('ui', path)` was already used somewhere lol. But yeah it uses `convertKeybinding` with the right `platform` internally so you can use it like:

```ts
const keybinds = this.$bksConfig.getKeybindings('ui', 'general.openQuickSearch');
```

This will return either `string[]` or `string[][]`. If there are two shortcuts for `'general.openQuickSearch'`, it will return `string[][]`. (yeah, I know, it should've just sticked to returning one type -_-)

azmy60: Feel free to modify the code. I don't think it's used anywhere right now. I thought I've used it for making the shortcut modal.

wmontgomery: OK cool! I probably didn't use it too much because I misread what it was doing :D

I can use it for the shortcut modal and also this too. Happy to use something existing before jumping into a new thing. I'll switch this up then.

apps/studio/default.config.ini
wmontgomery: @azmy60 I get what you're saying. Do you think there would be a time we officially remove these keys from the ini file and if someone doesn't have it setup, then that's kind of on them?

My concern comes in now that there are 2 methods to run, there's some UI implementation concerns just around how it shows up. Either the buttons remain the same IF the original keybindings are there, or it goes to the new format which can change. Might be worth sitting on this for a bit and having a note that a change is coming, then on first site of this feature when upgraded, give the warning of "keybindings have changed to run current as default" or something like that? I'm all for being backward compatible and not breaking people's experience, however also need to keep it moving forward 😄

azmy60: @wmontgomery Absolutely. On top of all that, maybe we should have a doc page that tells them "what should be removed and what you need to replace them with". Like a tutorial page that goes "If you have this setup, the new format looks like ... ".

azmy60: This is why auto-migration is great I think.

rathboma:
@not-night-but let's get this in but NOT change the default action. Rather get the code in and document it

apps/studio/default.config.ini
rathboma: Yes this PR needs to address these problems in some way. Either some warning, an auto-migration, or some other way to give the user the information. Without that extra attention to detail we can't really merge this.

not-night-but:
@rathboma @azmy60 Added a `deprecation.config.ini` as well as a section to the config warning modal for deprecation warnings. Also added a deprecation section to the docs for us to list these things out as well as upgrade paths: 
image
image

apps/studio/src/common/bksConfig/mainBksConfig.ts
```diff
@@ -16,7 +16,8 @@ type ConfigFileName =
   | "default.config.ini"
   | "system.config.ini"
   | "user.config.ini"
-  | "local.config.ini";
+  | "local.config.ini"
+  | "deprecated.config.ini";
```
azmy60, line 20: Oh this is nice.

azmy60 (changes requested):
@not-night-but We are missing the old keys check. What if a user already customize `submitCurrentQuery` key to be swapped with `submitTabQuery` key?

If I have this in my config:

```ini
[keybindings.queryEditor]
submitTabQuery = ctrlOrCmd+shift+enter
submitCurrentQuery = ctrlOrCmd+enter
```

The app will pop up a warning (which is good), but the keys in my config are not applied. I should still be able to press ctrl+shift+enter even though it's deprecated.

not-night-but:
@azmy60 I think that was the original plan but @rathboma said we should just ignore the old keys in this instance

apps/studio/src/components/editor/ShortcutHints.vue
```diff
@@ -144,12 +90,25 @@
...
     props: ['isMongo'],
-    data() {
-      return {
+    components: {
+      ShortcutDisplayHelper
+    },
+    methods : {
+      isPrimaryRunCurrentQuery() {
+        const { settings: configSettings } = this.$bksConfig
+        return configSettings.queryEditor?.primaryQueryAction.toLowerCase() === 'submitcurrentquery'
```
azmy60: Ideally, we want to abstract this somehow in the future.

azmy60: I've seen this duplicated multiple times.

apps/studio/default.config.ini
```diff
@@ -285,14 +285,21 @@ altOpen = ctrlOrCmd+enter
 openInBackground = right
 altOpenInBackground = ctrlOrCmd+right
 
+[settings.queryEditor]
```
azmy60: I'm not really sure if we want to categorize this as `[settings.queryEditor]`. The config file itself is already a settings file and we already have a `[ui.queryEditor]` section already. Could we use that instead?

azmy60 (changes requested):
This is awesome. But we still need one more change before we are ready to merge!

azmy60 (approved):
Nice!

PR #3987, opened 2026-03-24
Fix #3732: Prioritize table names over columns in Table/View contexts

### LINKED ISSUE #3732: BUG: Inaccurate autocomplete order for columns
**Describe the bug**
We always prioritize the column options to be at the top of the list in almost all situations.
Image

**To Reproduce**

1. Type something like:

```sql
select * from actor join
```

2. And then trigger the autocomplete (`Ctrl+Space`) after the `join`.

The first options are the columns of the `actor`, which is not accurate because after `join`, we type a table, not a column. You can see this in any position as long as the table `actor` is there, so it's not only `join`.

**Expected behavior**
Don't make columns first priority all the time. Put them at the top only when it makes sense, for example, after `SELECT` keyword.

HINT:
See `columnsToCompletions()` in `apps/ui-kit/lib/components/sql-text-editor/utils.ts`,
and `sqlCompletionSource()` in `apps/ui-kit/lib/components/sql-text-editor/extensions/sqlContextComplete.ts`.

Martimborgss:
## Description
Fixes #3732 

This PR updates the autocomplete logic to prioritize table and view names over column names when the user is typing after specific SQL clauses that expect a Table/View context. 

**Changes:**
- Modified the autocomplete behavior to show tables instead of columns immediately following `FROM`, `JOIN`, `UPDATE`, and `INTO` clauses.
- Prevented columns from incorrectly applying a high-priority boost in these specific contexts.
- Added comprehensive unit tests to verify the correct autocomplete priority.

## Testing
- [x] Added unit tests for table, column, and multi-line contexts.

## Screenshot
image

Signed-off-by: Martim Borges <martim.borges@tecnico.ulisboa.pt>

### DIFF
--- a/apps/ui-kit/lib/components/sql-text-editor/extensions/sqlContextComplete.ts
+++ b/apps/ui-kit/lib/components/sql-text-editor/extensions/sqlContextComplete.ts
@@ -159,12 +159,17 @@ async function loadColumnsFromQueryContext(
  * Check if cursor is at a position where column completion makes sense
  */
 function isColumnCompletionPosition(textBeforeCursor: string): boolean {
+  // After FROM, JOIN, UPDATE, INTO should suggest tables, not columns
+  if (/\b(FROM|JOIN|UPDATE|INTO)\s+[a-zA-Z0-9_]*$/i.test(textBeforeCursor)) {
+    return false;
+  }
+
   const completionPatterns = [
-    /\bSELECT\s+/i, // After SELECT (with or without trailing content)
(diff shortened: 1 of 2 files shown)


rathboma (approved):
looks great, thanks so much!

Martimborgss:
Hi @rathboma @not-night-but, bumping this for review for the workflows when you get a chance! No rush

not-night-but:
@Martimborgss it looks like this change was already made and merged in a different PR, so I'm going to close this one

PR #4006, opened 2026-03-26
suppress deprecation warnings from sass

rathboma:


not-night-but:
@rathboma I'm still seeing deprecation warnings with these changes

azmy60:
Yeah I'm also still seeing some warnings

PR #4018, opened 2026-03-27
Add driver dependency auto-download framework

### LINKED ISSUE #4003: FEAT: Auto-download framwork for additional driver files + Oracle Instant Client implementation
Some drivers require additional files that we can't bundle with the application because of licensing reasons, for example:
- ODBC bundles
- JDBC Jars
- Oracle Instant client shared libraries

Beekeeper Studio doesn't have many of these connections right now (Oracle is the only one I think), but more are coming.

We need a simple framework for easily adding new 'sources' of additional driver files. Some sort of standard interface that we can implement.

This interface will contain stuff like:
1. A way to resolve what file to download based on platform-specific inputs
2. A way to tell the application what databases might require the files
3. A set of information to present to the user before starting the download (eg links to licenses to accept, or documentation)
4. A way for a database driver to say "I need X", and a way for the file provider to say "I provide X".
5. It should have a method to check for updates to the installed version of a library that can be triggered from the UI

The app then needs a bunch of standard code that can:
1. Download the package, put it in the user config directory, save it's path, and provide that to any drivers that need it.
2. Show progress to the user during download
3. When a file provider is available for something required by a driver, next to the `choose the file` picker, it should provide a `Get it for me` button
4. When user asks Beekeeper to download the file, it presents any required information in a modal before starting the download (for the user to accept), eg "By downloading this driver you agree to the XYZ terms available at the url below" /Cancel/ /OK/
5. If download fails, it should track the failure and point users to the correct place to download the files manually

The goal of this issue is to establish the building blocks for the library/file provider and to implement the first one: A provider for the Oracle Instant Client.

For testing we already download the instant client from oracle in the `get-instant-client.sh` shell script, so we can use this as a starting point. Downloads for each OS are listed here: https://www.oracle.com/database/technologies/instant-client/downloads.html

Note: For many third party libraries we CANNOT distribute ourselves

rathboma:
## Summary

Adds a framework for auto-downloading driver files that can't be bundled (licensing restrictions), with Oracle Instant Client as the first implementation. Closes #4003.

- **Supply/demand model**: Drivers declare requirements (`DriverRequirement`), providers resolve platform-specific downloads (`DriverDepProvider`). New drivers just register a provider — no form changes needed.
- **Generic UI integration**: `SettingsInput` auto-detects when a provider exists for its setting key and shows the download button. Filled state shows a compact card with the path and a clear button.
- **License & details modal**: Shows what's being downloaded, from where, install destination, platform-specific notes (e.g. glibc requirement on Linux), and license terms before proceeding.
- **Progress via Noty**: Download progress streams from the utility process to a toast notification so users can keep working.
- **Configurable user agent**: `downloadUserAgent` in `default.config.ini` — defaults to a Chrome-compatible string so CDNs like Oracle's don't block requests.

### Also fixes
- `FilePicker` now opens at the current value's directory (was always home dir)
- `OracleForm` restart noty no longer crashes (`$native.app.restart()` never existed)
- `SettingsInput` filled state is now a compact card instead of showing the raw picker

### Adding a new driver dependency
1. Create a provider in `src/services/driverDeps/providers/`
2. Register it in `createDefaultRegistry()` with the setting key and connection types
3. Any `SettingsInput` using that setting key automatically gets the button

## Test plan
- [ ] Open Oracle connection form, verify auto-download button appears next to Instant Client Location
- [ ] Click auto-download, verify license modal shows with download URL, install path, and glibc note (Linux)
- [ ] Accept and verify progress noty updates through downloading/extracting/installing
- [ ] Verify Instant Client installed to `~/.config/beekeeper-studio/driver-deps/oracle-instant-client/`
- [ ] Verify setting populated and card view shows with path and clear button
- [ ] Clear value, re-download, verify restart prompt fires
- [ ] Verify non-Oracle `SettingsInput` fields (e.g. TNS_ADMIN) don't show download button
- [ ] Build: `yarn run electron:build --linux AppImage`


apps/studio/src/services/driverDeps/index.ts
```diff
@@ -0,0 +1,27 @@
...
+export { default as DriverDepRegistry } from "./DriverDepRegistry";
+export * from "./types";
+
+import DriverDepRegistry from "./DriverDepRegistry";
+import OracleInstantClientProvider, {
+  ORACLE_INSTANT_CLIENT_REQUIREMENT_ID,
+} from "./providers/OracleInstantClientProvider";
+import type { DriverRequirement } from "./types";
+
+const ORACLE_REQUIREMENT: DriverRequirement = {
```
rathboma: On the process of registering a dependency for a driver:

1. Document in docs/development how to add a register a new external need for a driver.
2. Simplify the amount of work required to add an external dependency request - should be a) the provider file, but then only a line or two of code in this file

rathboma (commented):
Looks good, but needs clarity on how developers are supposed to use it.

Also needs master merging in, and removal of claude.yml

PR #4024, opened 2026-03-30
Support multiple jump hosts for SSH Tunnel

### LINKED ISSUE #3986: FEAT: Multiple SSH Jump Hosts
The current SSH Tunnel feature can contain up to two endpoints; the target and the jump host (bastion).

But some advanced users might need multiple nested jump hosts setup.

The plan to implement this feature is:

1. [ ] Create a "ssh config" table that each row will contain host, port, username, auth mode, etc..
2. [ ] Migrate the existing ssh columns (`sshHost`, `sshPort`, `sshUsername`, etc..) in saved/used connections into a new row of the ssh config table
3. [ ] Create a join table for the ssh configs and connections (Many-to-Many) with an extra column named `order` to determine the order of the jump host chain.
    - [ ] The first order of the ssh config will be the first jump host. The last one will be the target host.
4. [ ] At this point, the ssh columns in the connection (`sshHost`, `sshPort`, `sshUsername`, etc..) become obsolete.
5. [ ] Use tabulator with "movable row" enabled to show the ssh configs (image below*)
    - [ ] The table uses the same style as the SchemaBuilder component
    - [ ] The add jump host button is not a part of the tabulator. It's a separate button element with identical style.
    - [ ] If you focus on a row by clicking it, it should update the form below it.

Image

Image

*The ssh configs are stacked vertically. It starts with the first row (jump host), and it goes down to the final row (target host). Think of like adding multiple destinations in Google Map or Apple Map!

> [!NOTE]
> In the future, the new ssh config table can help us implement reusable ssh configs for multiple connections.

azmy60:
Fix #3986 

Screenshot 2026-04-13 at 19 49 17


azmy60:
Blocked by cloud update

apps/studio/src/common/appdb/models/SshConfig.ts
```diff
@@ -0,0 +1,40 @@
...
+  connectionSshConfigs?: any[]
+
+  @Column({ type: 'varchar', nullable: false })
+  host: string
+
+  @Column({ type: 'int', nullable: false, default: 22 })
+  port = 22
+
+  @Column({ type: 'varchar', length: 8, nullable: false, default: 'agent' })
+  mode: import('@/common/interfaces/IConnection').SshMode = 'agent'
```
not-night-but: Just import at the top of the file

apps/studio/src/common/appdb/models/SshConfig.ts
```diff
@@ -0,0 +1,40 @@
...
+@Entity({ name: 'ssh_config' })
+export class SshConfig extends ApplicationEntity implements TransportSshConfig {
+  withProps(props?: Partial<TransportSshConfig>): SshConfig {
+    if (props) SshConfig.merge(this, props)
+    return this
+  }
+
+  // Back-reference to the join table; not used directly in app logic
+  @OneToMany('ConnectionSshConfig', 'sshConfig')
+  connectionSshConfigs?: any[]
```
not-night-but: Then do we need it?

apps/studio/src/common/appdb/models/saved_connection.ts
```diff
@@ -165,18 +166,23 @@ export class DbConnectionBase extends ApplicationEntity {
   @Column({ type: 'boolean', nullable: false, default: false })
   sshEnabled = false
 
+  /** @deprecated use `sshConfigs` instead */
```
not-night-but: Why not just remove these columns instead of deprecating if we are migrating all of the connections anyways?

apps/studio/src/common/utils.ts
```diff
@@ -367,3 +369,62 @@ export function isNumericDataType (dataType) {
 
   return numericStarts.some(t => base.startsWith(t))
 }
+
+/**
+ * Should we use the legacy ssh config format or use the {@link IConnection.sshConfigs}
+ *
+ * This makes sure we consume array all the time instead of `null`.
+ **/
+export function resolveSshConfigs(config: IConnection): TransportConnectionSshConfig[] {
```
not-night-but: See I don't love using the legacy format here. It feels like the sshConfigs should be being built in the frontend by the modal, and that the old root level config values shouldn't do anything

apps/studio/src/components/connection/CommonAdvanced.vue
```diff
@@ -56,270 +90,326 @@
...
-      >
         <div class="form-group">
-          <label for="sshUsername">SSH Username</label>
+          <label>Username</label>
           <masked-input
-            :value="config.sshUsername"
+            :value="selectedConfig.username"
             :privacyMode="privacyMode"
-            @input="val => config.sshUsername = val"
+            @input="updateSelectedSsh('username', $event)"
```
not-night-but: Completely unrelated but ugh idk why we don't have these masked inputs setup to just allow for v-model.

apps/studio/src/handlers/appDbHandlers.ts
```diff
@@ -159,7 +173,40 @@ async function transformConn(obj: SavedConnection, cls: any): Promise<IConnectio
 }
 
 export const AppDbHandlers = {
-  ...handlersFor<IConnection>('saved', SavedConnection, transformConn),
+  ...handlersFor<IConnection>('saved', SavedConnection, transformConn, {
+    preSave: async (obj: Partial<IConnection>, em) => {
```
not-night-but: I don't love this. I feel like we should maybe just manually call save for the ssh configs from the frontend if they are changed. And I almost feel like refcounting them is too much. I think realistically you should be able to manage your ssh connections separately from db connections, but I can just see this system breaking down real quick.

not-night-but (changes requested):
I'm not sure how I feel about the new models. Idk if it's just the way that typeorm handles relations like this, but it feels like it is complicating some systems unnecessarily. I feel like we almost need to communicate to the user that the SSH Config is separate from the db connection, and allow it to be properly managed separately.

apps/studio/src/components/connection/CommonAdvanced.vue
azmy60: I think we do actually

not-night-but:
Whoops, before I forget: 

We need to think about how this is going to function for older applications as well. I haven't looked at the cloud PR at all, but if someone creates a connection in their workspace with 3 jump hosts, how does someone that is back on version 5.0 or something see that connection? Do we just not send the ssh information? Do we attempt to flatten it back into the legacy format? Just a thought

apps/studio/src/common/appdb/models/SshConfig.ts
azmy60: Yes. Required by TypeORM for back-referencing.

apps/studio/src/components/connection/CommonAdvanced.vue
azmy60: Masked inputs can accept v-model. It's just that it needs to go through `updateSelectedSsh` in this case.

PR #4025, opened 2026-03-30
fix #3629: fix delete row command on Mac

### LINKED ISSUE #3629: BUG: Allow deleting rows with a keyboard on a Mac (DELETE doesn't work)
Allow deleting rows with a keyboard on a Mac (Delete key doesn't work)

I'd like to use Apple keyboard's Delete key to delete rows in a table.

I tried Cmd+Delete, like it usually works for eg. deleting files in Finder, but it doesn't work.

**Additional context**
When I right-click on the row number in a table, the context menu offers eg. "duplicate row" (Cmd+D) which works and "delete row" (DELETE) which doesn't work.

I use Postresql.

antoniogmello:
The delete row command was not working on Mac because the delete key maps to a different key code than on Windows. This patch updates the convertKeybinding function to remap delete to backspace on Mac for the v-hotkey target, with unit tests verifying this correct mapping


rathboma:
@azmy60 is this how we should handle this do you think? Delete and backspace are different right?

bmoleda:
@rathboma

> @azmy60 is this how we should handle this do you think? Delete and backspace are different right?

On a mac, <kbd>backspace</kbd> is called delete and has the same effect as backspace. I guess we also expect the same action - deleting a row - for both <kbd>delete</kbd> and <kbd>backspace</kbd> anyway.  Also the change seems to be for macs only.

Just my two cents - thanks for looking into this!

azmy60:
@rathboma yes, backspace is delete in Mac keyboard.

azmy60 (approved):
@antoniogmello thanks for the contribution!

PR #4027, opened 2026-03-30
Fix JSON array columns splitting into separate rows on copy as SQL

rathboma:
## Summary

- When copying a JSON column containing an array (e.g. `[1, 2, 3]`) as SQL from the table view, knex's `.insert()` received the raw JS array and expanded it into separate values — producing `values (1, 1, 2, 3)` instead of `values (1, '[1,2,3]')`.
- Added `JSON.stringify` for array and plain object column values in `buildInsertQuery` before they reach knex. This fixes the copy-as-SQL path and protects all other callers of `buildInsertQuery`.
- Added unit tests covering: single-row JSON arrays/objects, multi-row JSON arrays, nested structures, empty arrays/objects, null preservation, mixed JSON + scalar columns, and a regression test for plain scalar data.

## Test plan

- [x] Wrote failing test first, confirmed it reproduced the bug (`values (1, 1, 2, 3)`)
- [x] Applied fix, confirmed test passes (`values (1, '[1,2,3]')`)
- [x] All 283 unit tests pass
- [ ] Manual: connect to a database with a JSON column containing an array, select the cell, copy as SQL, verify the INSERT treats the array as a single value

### DIFF
--- a/apps/studio/tests/integration/lib/db/clients/libsql.spec.ts
+++ b/apps/studio/tests/integration/lib/db/clients/libsql.spec.ts
@@ -64,6 +64,9 @@ function testWith(options: typeof TEST_VERSIONS[number]) {
         )
           .withName(`libsql${options.readOnly ? "-readOnly" : ""}`)
           .withExposedPorts(8080)
+          .withEnvironment({
+            SQLD_CONNECTION_CREATION_TIMEOUT_SEC: "30",
+          })
           .withStartupTimeout(dbtimeout)
           .start();
         jest.setTimeout(timeoutDefault)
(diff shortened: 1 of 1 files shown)


apps/studio/src/lib/db/clients/BasicDatabaseClient.ts
```diff
@@ -439,6 +439,16 @@ export abstract class BasicDatabaseClient<RawResultType extends BaseQueryResult,
...
+        // JSON objects/arrays must be stringified so knex treats them as
+        // single values. Skip native DB array columns (e.g. PostgreSQL
+        // text[]) which knex handles natively.
+        if (_.isPlainObject(row[key]) || Array.isArray(row[key])) {
+          const col = columns.find((c) => c.columnName === key);
+          const isNativeArray = col?.dataType && (col.dataType.endsWith('[]') || col.dataType.toUpperCase() === 'ARRAY');
+          if (!isNativeArray) {
+            row[key] = JSON.stringify(row[key]);
+          }
+        }
```
not-night-but, line 451: I think this is mostly okay, but I'm not sure I love having this logic here. I feel like it should be a transcoder. Maybe we also pass column definitions to the transcoder value checker? I mostly worry that this is a bit of a hack rather than fixing the underlying issue, and this will only fix this issue in this specific instance.

not-night-but (changes requested):

Issue #3165, opened 2025-05-13
FEAT: change cursor while hovering drag handle


mcanepa:
I think a better UX would be for the mouse cursor to change to `move` when I move the mouse into the drag handle to alter the order of table columns

rathboma:
Agreed.

iamsatriya:
Hi @rathboma , I’m interested in contributing to this issue and would be happy to work on it.

rathboma:
@iamsatriya please feel free to implement and open a PR!

baguse:
Is it still available to work on?

PR #4045, opened 2026-04-02
fix: change cursor to grab on drag handle hover

boergeson:
Adds `cursor: grab` to `.drag-handle.draggable` in `entity-list.scss` so the cursor changes when hovering the reorder handle on pinned tables. Also adds `cursor: grabbing` on `:active` for visual feedback while dragging.


not-night-but:
@boergeson Thanks for this! I'm going to remove the tag for #3165 as that is talking about dragging table columns in the TableTable view I believe

boergeson:
It's a pleasure, have a nice weekend

PR #4049, opened 2026-04-03
Add right-click copy to structural table cells

### LINKED ISSUE #1458: FEAT: Copy column value of structural table (indexes, relations, triggers)
I need a function that copies the values of columns of structural tables (indexes, relations, triggers).

It should be possible to copy columns:
- Indexes: id, name, columns
- Relationships: name, column, FK schema, FK table, FK column, on update, on delete
- Triggers: name, time, manipulation, action, condition

**Implementation**
Create a function that copies the value of a column to the clipboard (like as a function that copies an error message).

Screenshot 2022-12-20 at 12 44 25

boergeson:
Adds a "Copy" option to the right-click context menu on cells in the indexes, relations, and triggers tables.

Previously there was no way to copy a value from these tables without double-clicking the cell to make it selectable and then manually selecting the text. Now you can just right-click and hit Copy.

**Changes**
- New `copyCellMenu` helper in `tableMenu.ts` that grabs the cell value and writes it to clipboard
- Added `contextMenu: copyCellMenu` to all column definitions in `TableIndexes.vue`, `TableRelations.vue`, and `TableTriggers.vue`

Closes #1458

### DIFF
--- a/apps/studio/src/components/tableinfo/TableIndexes.vue
+++ b/apps/studio/src/components/tableinfo/TableIndexes.vue
@@ -128,6 +128,7 @@ const log = rawLog.scope('TableIndexVue')
 import { escapeHtml } from '@shared/lib/tabulator'
 import { parseIndexColumn as mysqlParseIndexColumn } from '@/common/utils'
 import { SelectableCellMixin } from '@/mixins/selectableCell';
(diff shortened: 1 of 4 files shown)


not-night-but:
@boergeson This looks good, thanks so much!

PR #4050, opened 2026-04-03
update vim docs

austinwilcox:
Turns out you have great docs on the vim and vimrc implementation. The only tweak I added (outside of my markdown lsp triggering some changes on save) was that the cog is now in the bottom right of the query editor and not on the left.


austinwilcox:
@not-night-but You already have great docs that I just failed to read. I made one tweak with just the location of the cog, which is now in the bottom right instead of the bottom left.

not-night-but:
@austinwilcox apparently I just don't even know what docs I have written and what docs I haven't lol. Thanks!

PR #4057, opened 2026-04-03
Fix Copy Name on sidebar columns copying table name

### LINKED ISSUE #3981: BUG: Copying column names doesn't work - it copies table name instead
**Describe the bug**
In the left tree of appears the tablenames when I expand a node, I find the name of columns. So when I rigth click over any node and select "Copy Name"  from context menu, I hope the copy was the name of column, but  I get the name of table.

**To Reproduce**
1. In the left tree of appears the tablenames  I expand a tablename node, 
2. When columns names are visualized,   
3.  Select any column and then put rigth click with the mouse 
4.  Select "Copy Name" element of the contextual menu

**Expected behavior**
The Copy Name expected in the clipboard is for the  -Column name of the table, but the value copied is a -Table name.

**Screenshots / Gifs**

Image

 - OS: Windows
 - App Version 5.6.2
 - Database type and version MariaDB

**Additional context**
Im trying this product, I like it. Id like this behaviour like other database handlers.

boergeson:
Right-clicking a column in the sidebar and selecting "Copy Name" was copying the parent table's name instead of the column name. This happened because column items didn't have their own context menu handler, so the click bubbled up to the table's handler which copies `item.name`.

Fixed by adding a `@contextmenu.prevent.stop` handler on column sub-items in both `TableListItem.vue` and `StatelessTableListItem.vue`. The handler opens a small context menu with "Copy Name" that copies `item.columnName`.

Closes #3981

### DIFF
--- a/apps/studio/src/assets/styles/app/sidebar/sidebar.scss
+++ b/apps/studio/src/assets/styles/app/sidebar/sidebar.scss
@@ -516,6 +516,10 @@
     padding: ($gutter-h * 0.5);
     padding-right: $gutter-h * 0.5;
     padding-left: $gutter-w * 2;
+    border-radius: 4px;
+    &:hover {
+      background: rgba($theme-base, 0.035);
+    }
   }
   .btn-fab {
     min-width: 18px;


not-night-but:
@boergeson This looks good, but I do wonder if we should have some hover styling for the columns? Just to make it clear which column is being hovered

boergeson:
> @boergeson This looks good, but I do wonder if we should have some hover styling for the columns? Just to make it clear which column is being hovered

I definitly think we can do that, I'll look into it later

boergeson:
Added a hover highlight on the column items, should make it more obvious which one you're about to interact with

PR #4060, opened 2026-04-05
feat/windows: Deep Link Protocol Support 

yogasw:
I tested deep linking on macOS and it works well, but when switching to Windows the protocol handler is not available, so URLs like `redis://` cannot open the app automatically. This PR adds Windows deep link support so users can open database connections directly from a URL without creating a manual configuration first.

This PR also adds a `yarn run generate:nsis-protocols` script to auto-generate `apps/studio/build/win/protocol-handler.nsi` from `electron-builder-config.js`, removing the need for manual edits and preventing configuration drift.

**How to test (Windows)**

1. Build and install the app
2. Press **Windows + R**
3. Run the following deep link:

```id="h4p0n9"
redis://localhost:6379
```

**Expected**

* Beekeeper Studio opens
* The connection is created automatically


rathboma:
Hey, thanks for the PR! This is a great new feature add that I thought we already had.

Couple of quick questions:
- Looks like we need to run `generate:nsis-protocols` before packaging the windows binary, is that right?
- If the *.nsi is basically just a build artifact, do we have to commit it, or should we just generate it on-demand?
- After installing Beekeeper Studio, does the user need to move this *.nsi file somewhere?

Thanks for your help! I know little about this part of Windows development

yogasw:
Hey @rathboma, thanks for the review! Let me answer each question:

**1. Do we need to run `generate:nsis-protocols` before packaging?**

Only when the supported protocols in [electron-builder-config.js](https://github.com/beekeeper-studio/beekeeper-studio/blob/cd75fb08797b737e81d96d499057d4d3402e075d/apps/studio/electron-builder-config.js) change — not on every build. Since the protocol list rarely changes, regenerating every time would unnecessarily update the file's last-modified timestamp and could trigger auto-formatters, creating noise in the diff.

**2. Should the `*.nsi` be committed, or generated on-demand?**

Agreed — it's just a build artifact and shouldn't be committed. I initially tried wiring the generation into [electron-builder-config.js](https://github.com/beekeeper-studio/beekeeper-studio/blob/cd75fb08797b737e81d96d499057d4d3402e075d/apps/studio/electron-builder-config.js) directly but haven't gotten it working yet. Do you think there's a `beforeBuild` hook or similar we could hook into? If so, the generation could happen automatically before the Windows build, and the `*.nsi` file could then be safely added to `.gitignore`.

**3. Does the user need to move the `*.nsi` file after install?**

No — the placement is already handled [here](https://github.com/beekeeper-studio/beekeeper-studio/pull/4060/changes#diff-e8320dc36f168a11aa2a1cad4989647e4b2dac114464ec8295b25757e427a849R6). One thing worth noting though: `electron-builder` only allows a [single NSI include file](https://github.com/beekeeper-studio/beekeeper-studio/blob/cd75fb08797b737e81d96d499057d4d3402e075d/apps/studio/electron-builder-config.js#L238). 
I’m thinking of creating a small wrapper file in `.nsh` format, such as `beekeeper.nsh`, which includes (using the `!include` command) both `msvc-redist.nsh` and the generated protocol file — so that electron-builder still has a single entry point.

rathboma:
@yogasw Thanks for the answers!

For 1/2 - yeah let's generate the file on build, we can use the electron-builder `beforePack` or `beforeBuild` callbacks to do this. If you have trouble with this, let us know I'm sure we can add that last part for the PR pretty easily.

3: Sounds like a great plan!

Thanks again :-D

PR #4066, opened 2026-04-07
Code Style Guide: Input

azmy60:
Code style guide for input elements to ensure consistent style.

TODO
- [ ] Make sure <input> is styled by default
- [ ] Make sure the default style doesn't interfere with elements that aren't meant to be 'default styled'. E.g. Tabulator's default input.


apps/studio/src/components/common/form/PasswordInput.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <div class="password-input">
+    <input
```
rathboma: Should the password input also use `masked-input` under the hood?

Also - should we still call it `masked-input`?

CONTRIBUTING.md
```diff
@@ -21,6 +21,43 @@ It's usually best to open an issue before spending a lot of time working on the
...
+#### Form Inputs
+
+Use the `.form-group` class when an input has a label, hint, or both. It provides standard flex column layout, label styling, and hint text styling. A standalone input without a label or hint does not need `.form-group`.
+
+```html
+<!-- Input only (no .form-group needed) -->
+<input type="text" v-model="value">
+
+<!-- With label -->
+<div class="form-group">
```
rathboma, line 35: does grouping work with `masked-input` or `password-input`? It probably wouldn't given the extra dom elements it introduces?

rathboma (commented):
questions

apps/studio/src/components/common/form/PasswordInput.vue
azmy60: > Should the password input also use masked-input under the hood?

Yeah, that's a good idea.

> Also - should we still call it masked-input?

I think masked-input is still needed for other types of input like email, hostname, etc.

CONTRIBUTING.md
azmy60: Yes, it works! We already did this in some places. And it shouldn't affect the group since a component will typically contain one root element.

rathboma (approved):

PR #4081, opened 2026-04-09
Fix flaky e2e tests

not-night-but:


rathboma:
@not-night-but doesn't look fixed :eyes:

not-night-but:
@rathboma yeah there are some issues with the playwright command not using the proper config for some reason, along with some xserver things (I think). This is the next thing on my to-do list after the mac issue

PR #3991, opened 2026-03-24
Use OS keychain for encryption key storage

rathboma:
## Summary

- Migrates encryption key storage from a file encrypted with a hardcoded bootstrap key to the OS keychain via Electron's `safeStorage` API (macOS Keychain, Windows DPAPI, Linux libsecret)
- Falls back to the existing file-based approach when `safeStorage` is unavailable (e.g., Linux without a keyring daemon), with a dismissable warning banner
- Automatically migrates existing users' `.key` files to keychain-protected `.encryption-key` files on first launch after upgrade
- Adds a documentation page with per-platform keyring setup instructions

Closes #943

## What changed

| File | Change |
|------|--------|
| `apps/studio/src/common/encryption_key.ts` | Rewritten: safeStorage with fallback, exports `isEncryptionKeyInsecure()` |
| `apps/studio/src-commercial/entrypoints/main.ts` | Loads key early with error dialog; passes key to utility process via env var; IPC handler for renderer |
| `apps/studio/src-commercial/entrypoints/preload.ts` | Exposes `keychainInsecure` to renderer via `contextBridge` |
| `apps/studio/src-commercial/entrypoints/renderer.ts` | Calls `requestKeychainInsecure()` during bootstrap |
| `apps/studio/src/@types/bridge.d.ts` | Adds `keychainInsecure` to Window type |
| `apps/studio/src/components/common/KeychainWarning.vue` | New dismissable warning banner |
| `apps/studio/src/App.vue` | Includes KeychainWarning component |
| `docs/user_guide/encryption-key-security.md` | New docs page with keyring setup instructions |
| `apps/studio/tests/unit/common/encryption_key.spec.ts` | 12 unit tests for all code paths |

## How it works

- **Keychain available**: Master key encrypted by `safeStorage.encryptString()`, stored as binary in `.encryption-key`. Only the OS can decrypt it.
- **Keychain unavailable**: Falls back to existing `.key` file with hardcoded bootstrap key (same as before). Shows a warning banner linking to docs.
- **Migration**: On first launch, reads old `.key` file, re-encrypts with safeStorage, verifies, then deletes the old file.
- **Utility process**: Receives the decrypted key via `BKS_ENCRYPTION_KEY` env var (safeStorage is main-process only).
- **No DB changes**: Same master key, same `simple-encryptor` transformers, same encrypted data in SQLite.

## Test plan

- [x] 12 unit tests covering: env var path, safeStorage mode (fresh/existing/migration), fallback mode, error handling
[...]

PR #4090, opened 2026-04-14
Feature/4065 privacy mode

### LINKED ISSUE #4065: Make it clear when a user is in privacy mode

wmontgomery:
resolves #4065 

## app
**UPDATED SCREENSHOTS**
image

image
The link takes the user to the docs about privacy mode

**OLD SCREENSHOTS** (to show color used through themes)
4065_privacy-mode

## docs
image


azmy60 (changes requested):
I agree with Day that the code is good enough. But, I think you missed the opportunity to give a little more clarity to the UI:

1. I understand that the eye icon acts as a button for turning off the privacy mode because the gif shows it, but when there are plenty of space, generally we'd want to make it explicit by adding text. So, I would add a text to the button rather than just an icon, so something like "<eye_icon> Disable", or just "Disable" without the icon.

> [!NOTE] 
> Also, buttons with "icons only" make more sense if it's for actions that are often used, like the search button in TableTable, eye icon in password field, close icon in a modal, etc.. Those examples are super common and we can assume users already know what they mean.

2. Same principle, but for the main text. Since we have a lot of space, I think It'd be better if it says something like "Privacy mode is enabled. Some information is redacted." or something like that, instead of just "In Privacy Mode". So I would add some more helpful stuff there and keep it concise.

As for the styling, I would reduce the font size to match the title bar, reduce the paddings a little bit, and increase the font-weight a little too:

Image

Image

wmontgomery:
> I agree with Day that the code is good enough. But, I think you missed the opportunity to give a little more clarity to the UI:
> 
>     1. I understand that the eye icon acts as a button for turning off the privacy mode because the gif shows it, but when there are plenty of space, generally we'd want to make it explicit by adding text. So, I would add a text to the button rather than just an icon, so something like "<eye_icon> Disable", or just "Disable" without the icon.
> 
> 
> Note
> 
> Also, buttons with "icons only" make more sense if it's for actions that are often used, like the search button in TableTable, eye icon in password field, close icon in a modal, etc.. Those examples are super common and we can assume users already know what they mean.
> 
>     2. Same principle, but for the main text. Since we have a lot of space, I think It'd be better if it says something like "Privacy mode is enabled. Some information is redacted." or something like that, instead of just "In Privacy Mode". So I would add some more helpful stuff there and keep it concise.
> 
> 
> As for the styling, I would reduce the font size to match the title bar, reduce the paddings a little bit, and increase the font-weight a little too:
> Image Image

All good suggestions! I'll make the updates

azmy60 (approved):
Good for merge!

Looking at the screenshot, I thought the link was too bright so I pulled the branch to test it a little bit, but it turns out fixed(?). But if the link is actually colored white, I think we should avoid that cause it's not eligible with the orange background.

Also, @rathboma made an interesting banner by reusing the `alert` style. https://github.com/beekeeper-studio/beekeeper-studio/pull/3991/changes#r3101056720

Not telling you to do it now, I think it's fine to keep the PR as-is, and this already looks way better. But I think moving forward, we will want to choose which style to pick.

Personally: the alert style is better because it's less distracting. It's not super loud, good for cases where it's not super duper important but worth to be noticed.

PR #4093, opened 2026-04-14
fix: Windows Pageant SSH agent forwarding silently fails since v3.9.0

### LINKED ISSUE #4092: BUG: Windows Pageant SSH agent forwarding silently fails since v3.9.0
**Describe the bug**

On Windows, SSH agent forwarding via Pageant does not work. The connection succeeds but agent forwarding is silently disabled — the Pageant agent is never passed to ssh2.

In `apps/studio/src/vendor/node-ssh-forward/index.ts` line 196, the class `ElectronFriendlyPageantAgent` itself is assigned to `agentDefault` instead of an instance. ssh2's `client.js` validates agent with `isAgent(cfg.agent)` (`val instanceof BaseAgent`), which returns `false` for a class reference, so agent is silently set to `undefined`.

**To Reproduce**

1. On Windows, have Pageant running with an SSH key loaded
2. Create a connection with SSH tunnel enabled and "SSH Agent" as the auth method
3. Enable "Agent Forwarding"
4. Connect to the server, would always fail

**Expected behavior**

Agent forwarding should work. The remote server should be able to use the keys from the local Pageant agent.

**Version Information:**
- OS: Windows 11
- App Version: affected since v3.9.0 through current master
- Database type and version: any (SSH tunnel related, not database specific)

**Additional context**

Introduced in commit ba97de543 (v3.9.0), which replaced `agentDefault = 'pageant'` (a string that ssh2 internally handled via `createAgent('pageant')` → `new PageantAgent()`) with the bare class reference `ElectronFriendlyPageantAgent`. Instantiation was missed.

Fix is one line:

```diff
- agentDefault = ElectronFriendlyPageantAgent
+ agentDefault = new ElectronFriendlyPageantAgent()

### PR DESCRIPTION
**Pull request by jeffcaiz**
Fixes #4092

## Summary

Pass an instance of `ElectronFriendlyPageantAgent` instead of the class itself when setting up Pageant agent forwarding on Windows.

## Problem

Since v3.9.0 (ba97de543), the code passes the class reference `ElectronFriendlyPageantAgent` as the `agent` config to ssh2. ssh2's `client.js` validates this with `isAgent()` (`val instanceof BaseAgent`), which returns `false` for a class — causing agent to silently become `undefined`.

The original code used the string `'pageant'`, which ssh2 handled via `createAgent('pageant')` → `new PageantAgent()`. When it was replaced with the custom `ElectronFriendlyPageantAgent`, instantiation was missed.

## Change

One-line fix in `apps/studio/src/vendor/node-ssh-forward/index.ts`:

```diff
- agentDefault = ElectronFriendlyPageantAgent
+ agentDefault = new ElectronFriendlyPageantAgent()
```

## Testing

Tested on Windows 11 with Pageant running — SSH agent forwarding now works correctly.

### DIFF
--- a/apps/studio/src/vendor/node-ssh-forward/index.ts
+++ b/apps/studio/src/vendor/node-ssh-forward/index.ts
@@ -193,7 +193,7 @@ class SSHConnection {
         if (this.isWindows) {
           // null or undefined
           if (agentDefault == null) {
-            agentDefault = ElectronFriendlyPageantAgent
+            agentDefault = new ElectronFriendlyPageantAgent()
           }
         }
 


jeffcaiz:
ssh2 `client.js:219-224` handles the `agent` parameter as follows:

```js
  if (typeof cfg.agent === 'string' && cfg.agent.length)
    this.config.agent = createAgent(cfg.agent);    // string path → create agent
  else if (isAgent(cfg.agent))                      // instanceof BaseAgent check
    this.config.agent = cfg.agent;                  // already an agent instance → use directly
  else
    this.config.agent = undefined;                  // anything else → discard
```

  isAgent() (agent.js:1110) is simply val instanceof BaseAgent.

rathboma:
Thanks for contributing! I'm going to see if I can get claude to add an integration test for this (to help us catch regressions). If not will merge anyway

rathboma (approved):
Simple fix, hard to find, thank you!!

PR #4101, opened 2026-04-15
Pr bedrock support

rathboma:


rathboma:
Built a GithubAction to publish bedrockdb to dockerhub automatically - https://github.com/rathboma/Bedrock/actions/runs/24473028940/job/71518021448. Once that's done we can add the integration tests

PR #4120, opened 2026-04-17
Add Amazon DynamoDB driver

### LINKED ISSUE #481: FEAT: Support for dynamodb

Want to gauge how applicable this feature is or how much interest there is for nosql connections 

A dynamodb plugin that supports queries, filters, data browsing  
Select dynamodb from the connections list just like any other rdbms  
continue to authenticate aws credentials and establish connection  

Login to AWS to manage/query dynamodb on their website  

**Additional context**

rathboma:
## Summary

Closes #481. Adds DynamoDB as an Ultimate-tier commercial driver, following the MongoDB/Cassandra pattern.

- Full driver in `src-commercial` on AWS SDK v3 (`@aws-sdk/client-dynamodb` + `@aws-sdk/lib-dynamodb`). Full CRUD, GSI/LSI listing, table create/drop, alterIndex, streaming scan cursor, PartiQL via `ExecuteStatementCommand`, read-only-mode enforcement at the driver level.
- AWS auth reuses the existing `IamAuthOptions` + `resolveAWSCredentials()` used by RDS/Redshift — IAM key / profile / CLI all wired. Custom endpoint field for DynamoDB Local. No new auth migration — `iamAuthOptions` is already there.
- PartiQL syntax highlighting **and** formatter support, not just Mongo-style "reuse Postgres":
  - CM6 `SQLDialect` in `@beekeeperstudio/ui-kit` (`customDialects.ts`) registered as `text/x-partiql`, plus a CM5 `defineMIME` fallback for cell editors
  - Custom `DialectOptions` for `sql-formatter` v15, modeled on Trino with `extraParens: ['[]', '{}']` so document literals round-trip intact
  - `formatOptionsFor(dialect)` returns `{ language } | { dialect }`; `safeSqlFormat` dispatches appropriately and falls back to raw input on parse error
  - `SqlTextEditor` (Ctrl+Shift+F) and `SuperFormatter` (preview dialog) both gained an optional `formatterDialectOptions` prop

## Test plan

- [x] 36 integration tests pass via `testcontainers` against `amazon/dynamodb-local`: schema introspection, all filter operators (`=`, `!=`, `<`, `>`, `IN`, `LIKE`/contains, `IS`/`IS NOT`, `OR` chaining, projection), composite-key CRUD on a `(userId, ts)` table, batch chunking over 25 items, offset pagination across 60 rows, PartiQL `SELECT`/`INSERT`/`UPDATE`/multi-statement + error paths, cursor read/filter/cancel, `queryStream`, `alterIndex` with `ACTIVE` status polling, read-only mode blocks, table lifecycle.
- [x] 6 unit tests for the PartiQL formatter (`tests/unit/shared/lib/dialects/partiqlFormatter.spec.ts`): dialect shape, SELECT/UPDATE formatting with identifier quoting preserved, INSERT-with-document-literal round-trip, unparseable-garbage fallback, non-leakage to other dialects.
- [x] `yarn build:esbuild` + `yarn build:vite` clean.
- [x] `yarn lib:build` clean.
- [x] Lint clean on all new/modified files.
- [x] Manual smoke against real AWS DynamoDB (key auth + profile auth). *Local DynamoDB is covered by the integration suite; real AWS requires an account to smoke.*
- [ ] Manual smoke of the query editor: paste a PartiQL `SELECT`, verify keyword highlighting, format with Ctrl+Shift+F, open the Super Formatter dialog to see the preview.

## Notes for reviewers

- Driver lives under `src-commercial` + is gated by `isUltimateType`, matching Mongo/Cassandra.
- `used_connection.ts` gets the `dynamoDbOptions` field forwarded alongside the other option blobs in `withProps`.
- `docker-compose.yml` gains a `dynamodb-local` service for local dev; the integration suite uses testcontainers and doesn't rely on it.
- I considered expanding `queryleaf` as the old issue comment suggested, but PartiQL is natively supported by the DynamoDB SDK and covers SELECT/INSERT/UPDATE/DELETE without a translator — simpler and more reliable. Happy to revisit if desired.


apps/studio/src-commercial/backend/lib/db/clients/dynamodb/DynamoDBCursor.ts
```diff
@@ -0,0 +1,109 @@
+import { BeeCursor } from "@/lib/db/models";
+import { DynamoDBDocumentClient, ScanCommand, ExecuteStatementCommand } from "@aws-sdk/lib-dynamodb";
+import rawLog from '@bksLogger';
+
+const log = rawLog.scope('dynamodb-cursor');
+
+// Default timeout for cursor fetch operations (30 seconds)
+const CURSOR_FETCH_TIMEOUT_MS = 30000;
```
rathboma: This should be in default.config.ini, as should any other magic numbers in the implementation

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
```diff
@@ -0,0 +1,970 @@
...
+
+    const items: any[] = [];
+    let lastKey: Record<string, any> | undefined;
+    const hasOrder = orderBy && orderBy.length > 0;
+    const target = (offset || 0) + (limit || 0);
+    // Memory protection: limit items buffered for ORDER BY operations
+    const MAX_ORDER_BY_ITEMS = 100000;
+    // Scan page-by-page. DynamoDB has no SQL-style offset and no server-side
+    // ORDER BY for Scan, so:
+    //   - without orderBy: stop once we've buffered enough to satisfy offset+limit
```
rathboma: Does this mean that to sort by a particular column we need to pull the WHOLE table into Beekeeper Studio, then sort manually?

If so - instead we should disable custom sorting, and be sure to only sort by default order in the tabletable.

Only support sorting by whatever the server supports natively and isn't going to produce a full table scan (maybe id is reverse sortable, if not, support no sorting

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
```diff
@@ -0,0 +1,970 @@
...
+    //   - with orderBy: we must scan the full result set so the in-memory sort
+    //     produces a globally-correct slice
+    do {
+      try {
+        const result = await this.doc.send(new ScanCommand({
```
rathboma, line 549: I don't think a user will ever want beekeeper to issue a full table scan while they are simply navigating around and clicking buttons

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
```diff
@@ -0,0 +1,970 @@
...
+      });
+      projection = tokens.join(', ');
+    }
+
+    const items: any[] = [];
+    let lastKey: Record<string, any> | undefined;
+    const hasOrder = orderBy && orderBy.length > 0;
+    const target = (offset || 0) + (limit || 0);
+    // Memory protection: limit items buffered for ORDER BY operations
+    const MAX_ORDER_BY_ITEMS = 100000;
```
rathboma: magic number. We should kill this anyway because we shouldn't support sorting.

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
rathboma: If there's a quick way to check the size of the table, we could support sorting when the table has fewer than 50k records. This size limitation should be in the config system though.

We should disable entirely for larger tables though. Never want to do a full table scan. Make sure this limitation is documented in the docs.

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
```diff
@@ -0,0 +1,970 @@
...
+
+  protected async rawExecuteQuery(_q: string, _options: any): Promise<DynamoQueryResult | DynamoQueryResult[]> {
+    // Not used — executeQuery goes through the PartiQL SDK directly.
+    throw new Error('DynamoDB does not support rawExecuteQuery');
+  }
+}
+
+// Split a PartiQL string on ';' boundaries but ignore semicolons inside single
+// quotes. `sql-query-identifier` doesn't speak PartiQL, so we do it by hand.
+function splitPartiQL(text: string): string[] {
```
rathboma: Will sql-query-identifier not properly split partiql queries anyway given that they have similar rules for quoting and query termination?

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
```diff
@@ -0,0 +1,970 @@
...
+  async duplicateTableSql(): Promise<string> {
+    throw new Error('DynamoDB does not support duplicating tables');
+  }
+
+  async getInsertQuery(): Promise<string> {
+    throw new Error('DynamoDB does not support generating SQL');
+  }
+
+  wrapIdentifier(value: string): string {
+    return `"${value}"`;
```
rathboma: This should be part of a Dynamo DialectData object, then referenced from there

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
```diff
@@ -0,0 +1,970 @@
...
+  }
+  if (buf.trim()) out.push(buf);
+  return out;
+}
+
+function firstWord(s: string): string {
+  return (s.trim().split(/\s+/, 1)[0] || '').toUpperCase();
+}
+
+function isMutation(s: string): boolean {
```
rathboma: again, isn't sql-query-identifier good enough for identification of alteration queries with ANSI SQL support?

apps/studio/src/components/TabPluginBase.vue
```diff
@@ -2,6 +2,18 @@
   <div v-if="isCommunity && tab.context.pluginId.startsWith('bks-')" class="tab-upsell-wrapper">
     <upsell-content />
   </div>
+  <div v-else-if="isUnsupportedErDiagram" class="plugin-base plugin-unsupported">
```
rathboma: Hmm, this doesn't seem like the place to put this. Won't this show the message for any plugin tab?

Let's remove this UI change. We can keep the unsupported key, and maybe pass that through to the plugin for evaluation?

apps/studio/src/migration/20260417_add_dynamodb_options.js
```diff
@@ -0,0 +1,12 @@
+export default {
+  name: "20260417_add_dynamodb_options",
+  async run(runner) {
+    const queries = [
+      `ALTER TABLE saved_connection ADD COLUMN dynamoDbOptions TEXT NOT NULL DEFAULT '{}'`,
```
rathboma, line 5: Please make a note in the readme of migrations that need adding to the cloud app (rails)

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
```diff
@@ -0,0 +1,958 @@
...
+    return desc.KeySchema?.find((k) => k.KeyType === 'HASH')?.AttributeName || null;
+  }
+
+  async getTableLength(table: string): Promise<number> {
+    const desc = await this.describeTable(table);
+    return desc.ItemCount || 0;
+  }
+
+  // Map a subset of the BKS TableFilter vocabulary to a DynamoDB FilterExpression.
+  private buildFilter(filters: TableFilter[]) {
```
rathboma, line 440: any string / literal escaping needed here to prevent injection attacks?

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
```diff
@@ -0,0 +1,958 @@
...
+
+    // DynamoDB does not support server-side ORDER BY for Scan operations.
+    // Custom sorting requires a full table scan which is expensive and can cause
+    // memory/performance issues on large tables. Sorting is disabled by default
+    // (maxSortableTableSize = 0) and can optionally be enabled for small tables.
+    let hasOrder = orderBy && orderBy.length > 0;
+    if (hasOrder && maxSortableTableSize > 0) {
+      const tableSize = await this.getTableLength(table);
+      if (tableSize > maxSortableTableSize) {
+        log.warn(`DynamoDB selectTop: Sorting disabled for table ${table} (${tableSize} items exceeds maxSortableTableSize ${maxSortableTableSize})`);
```
rathboma: Is this information presented to the user in any way in the status bar? It should be

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
```diff
@@ -0,0 +1,958 @@
...
+    const maxOrderByItems = BksConfig.db.dynamodb.maxOrderByItems;
+    const maxSortableTableSize = BksConfig.db.dynamodb.maxSortableTableSize;
+
+    // DynamoDB does not support server-side ORDER BY for Scan operations.
+    // Custom sorting requires a full table scan which is expensive and can cause
+    // memory/performance issues on large tables. Sorting is disabled by default
+    // (maxSortableTableSize = 0) and can optionally be enabled for small tables.
+    let hasOrder = orderBy && orderBy.length > 0;
+    if (hasOrder && maxSortableTableSize > 0) {
+      const tableSize = await this.getTableLength(table);
```
rathboma: is this operation O(1)? If it is O(n) we should just disable sorting entirely

apps/studio/src-commercial/backend/lib/db/clients/dynamodb.ts
```diff
@@ -0,0 +1,958 @@
...
+      }
+    } else if (hasOrder && maxSortableTableSize === 0) {
+      // Sorting is disabled entirely
+      hasOrder = false;
+    }
+
+    // Scan page-by-page. DynamoDB has no SQL-style offset and no server-side
+    // ORDER BY for Scan, so without sorting we stop once we've buffered enough
+    // to satisfy offset+limit.
+    do {
```
rathboma, line 547: we only this this if the table is small enough, correct?

apps/studio/src/shared/lib/dialects/models.ts
```diff
@@ -204,12 +223,14 @@ export interface DialectData {
     compositeKeys?: boolean    // Whether composite keys are supported
     schemaValidation?: boolean  // Whether schema validation features are disabled
     readOnlyPrimaryKeys?: boolean  // Whether primary keys are read-only
+    erd?: boolean              // Whether ERD view is disabled
```
azmy60: Interesting. I personally believe the app isn't responsible for whether a plugin or a feature of a plugin should be disabled this way. The plugins themselves should decide whether they are compatible with the generic information that the app gives. This gives more flexibility for plugins.

Instead of `erd`, I would add a `noSQL` boolean flag, or like a string union `databaseType: "sql" | "no-sql" | "graph"`. And then add `getDialectData` as a new plugin API that returns this object to the plugin.

apps/studio/src/shared/lib/dialects/dynamodb.ts
```diff
@@ -0,0 +1,68 @@
+import { ColumnType, DialectData } from "./models";
+
+// DynamoDB attribute types. B = binary, S = string, N = number, plus document
+// types (M, L) and scalar types that PartiQL surfaces. These are shown in the
+// column-type picker when creating/editing schema elements.
+const types = [
+  'S', 'N', 'B', 'BOOL', 'NULL', 'L', 'M', 'SS', 'NS', 'BS'
```
azmy60: Why do we use a single letter instead of a word? Is it a DynamoDB thing?

azmy60: Image

It looks unclear. I can guess what "S" means, probably. Even if the word is >10 letters, shortening it to a single letter is not good I think. :)

apps/studio/src/lib/editor/CodeMirrorDefinitions.ts
```diff
@@ -13,6 +13,43 @@ CodeMirror.defineMIME("text/x-pgsql", {
...
+// PartiQL — DynamoDB's SQL-compatible query language. Double-quoted words are
+// identifiers (reuse the sqlite hook for correctness), single-quoted are
+// strings. Keywords cover the DynamoDB subset plus PartiQL extensions.
+function partiqlSet(str: string) {
+  const out: Record<string, boolean> = {};
+  str.split(" ").forEach((w) => { if (w) out[w] = true; });
+  return out;
+}
+
+CodeMirror.defineMIME("text/x-partiql", {
```
azmy60: This doesn't work in cm6. It should be done by using the `language` package https://codemirror.net/examples/lang-package/.

apps/ui-kit/lib/components/sql-text-editor/customDialects.ts
```diff
@@ -9,3 +9,38 @@ export const GreengageSQL = SQLDialect.define({
...
+  "begin", "commit", "rollback",
+  "contains", "begins_with", "attribute_exists", "attribute_not_exists", "size",
+].join(" ");
+
+const PARTIQL_TYPES = [
+  "s", "n", "b", "bool", "null", "l", "m", "ss", "ns", "bs",
+  "string", "number", "binary", "list", "map", "boolean",
+].join(" ");
+
+export const PartiQL = SQLDialect.define({
```
azmy60, line 37: Yes, this is the correct way to add a PartiQL to cm6!

mkdocs.yml
```diff
@@ -139,6 +139,7 @@ nav:
     - ClickHouse: user_guide/connecting/clickhouse.md
     - user_guide/connecting/cockroachdb.md
     - user_guide/connecting/duckdb.md
+    - DynamoDB: user_guide/connecting/dynamodb.md
```
azmy60, line 142: Just be careful that this will be 'indexed' once merged to master, right?

azmy60 (changes requested):
Some minor feedback and these three bugs I found:

1. The database dropdown shows nothing. Can we show the database name and hide the + button? If DynamoDB doesn't support multiple databases then the dropdown shouldn't be clickable too. 

Image

Image

2. I think adding a shell tab is a mistake. What's the difference between that and the query tab?
Image

3. This always shows `localhost:null` whether I'm connected to a local db or an actual aws db. Maybe show the project name and the region?

Image

azmy60 (approved):

PR #4122, opened 2026-04-19
fix(mssql): bracket-wrap qualified names so dotted table names aren't mis-split

### LINKED ISSUE #3722: BUG: Incorrect parsing of SQL Server table names containing dots
**Describe the bug**

Beekeeper Studio incorrectly parses SQL Server table names that contain dots (periods) in the actual table name. When a table name includes a dot as part of its name (not as a schema separator), Beekeeper treats it as if it were a schema delimiter, causing issues with table operations.

**To Reproduce**
1. Connect to a SQL Server database
2. Create or navigate to a table with a dot in its name (e.g., Schema: dbo, Table name: Common.Companies)
3. Try to perform operations like editing records through the interface or generating queries
4. Observe that Beekeeper interprets the table as dbo.Common.Companies (treating "Common" as a schema) instead of dbo.[Common.Companies] (correct format with brackets)

**Expected behavior**
Beekeeper should properly handle table names containing dots by enclosing them in square brackets.

- Current behavior: dbo.Common.Companies (interpreted as schema "dbo.Common", table "Companies")
- Expected behavior: dbo.[Common.Companies] (schema "dbo", table "Common.Companies")

**Screenshots / Gifs**

Image

Image

 - OS: Windows
 - App Version: 5.4.12 Ultimate
 - Database type and version: SQL Server 2019
 
**Additional context**
I'm currently migrating from DBeaver to Beekeeper Studio, and the experience has been excellent overall. However, this issue with parsing table names that contain dots significantly impacts core DBMS functionalities such as inline record editing, query generation, and other table operations. This is one of the few pain points preventing a completely smooth transition.

rathboma:
sp_helptrigger, sp_spaceused, and sp_rename received 'schema.table' as a
single quoted string, so a table named 'Common.Companies' in schema 'dbo'
would be parsed as schema 'dbo.Common', table 'Companies'. Wrap each part
with SqlServerData.wrapIdentifier first so the emitted form is
'[dbo].[Common.Companies]'.

Closes #3722

### DIFF
--- a/apps/studio/tests/integration/lib/db/clients/sqlserver.spec.ts
+++ b/apps/studio/tests/integration/lib/db/clients/sqlserver.spec.ts
@@ -1,7 +1,7 @@
 import { GenericContainer, Wait } from 'testcontainers'
 import { DBTestUtil, dbtimeout } from '../../../../lib/db'
 import { runCommonTests, runReadOnlyTests } from './all'
-import { IDbConnectionServerConfig } from '@/lib/db/types'
+import { DatabaseElement, IDbConnectionServerConfig } from '@/lib/db/types'
 import fs from 'fs';
 import path from 'path';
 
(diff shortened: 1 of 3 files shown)


apps/studio/tests/integration/lib/db/clients/sqlserver.spec.ts
```diff
@@ -274,6 +274,63 @@ function testWith(dockerTag: string, readonly: boolean) {
       })
     })
 
+    // Regression test for #3722 -> dotted table names fail in sp_helptrigger /
+    // sp_spaceused / sp_rename because the schema+table were concatenated as a
+    // single quoted string, so SQL Server split on the first dot.
+    describe("Dotted table names (#3722)", () => {
```
rathboma, line 280: make sure the integration tests also list columns so you can be sure they're returning data for the correct table

rathboma: Done in bc66e4b. The fixture now has a unique column signature (`id`, `company_name`, `fingerprint`) and a fingerprint row; the tests assert `listTableColumns` returns exactly that signature and `selectTop` returns the fingerprint row — so the dotted-name path is provably hitting the right table. The rename test does the same post-rename to confirm `sp_rename` actually moved the intended table.

---

PR #4130, opened 2026-04-21
docs: update greengage docs

v-astakhov:


docs/includes/supported_databases.md
```diff
@@ -11,6 +11,7 @@
 | [TiDB](https://pingcap.com/products/tidb/)               | ⭐ Full Support              |    ✅     |    ✅    |        [Features](https://beekeeperstudio.io/db/tidb-client) |
 | [Google BigQuery](https://cloud.google.com/bigquery)     | ⭐ Full Support             |    ✅      |    ✅    |    [Features](https://beekeeperstudio.io/db/google-big-query-client), [Docs](https://docs.beekeeperstudio.io/user_guide/connecting/bigquery) |
 | [Redis](https://redis.io/)                               | ⭐ Full Support               |    ✅    |    ✅    |       [Features](https://www.beekeeperstudio.io/db/redis-client/), [Docs](https://docs.beekeeperstudio.io/user_guide/connecting/redis) |
+| [GreengageDB](https://greengagedb.org/)                  | ⭐ Full Support              |    ✅     |    ✅    |   [Docs](https://docs.beekeeperstudio.io/user_guide/connecting/greengage) |
```
rimakan, line 14: is `greengage` fully supported? it has not been released yet as far as i remember. shouldn't its support status be `Coming Soon`?

not-night-but: It's in 5.7, which is in beta right now and should be released next week

docs/user_guide/connecting/greengage.md
```diff
@@ -0,0 +1,75 @@
...
+summary: "Connect Beekeeper Studio to GreengageDB (PostgreSQL-compatible) — self-hosted and Docker"
+description: "GreengageDB is PostgreSQL-compatible, so connecting from Beekeeper Studio is very similar to Postgres. This guide covers Greengage-specific notes and common settings, including Greengage 6 and 7."
+icon: material/database
+---
+
+# How To Connect to GreengageDB
+
+GreengageDB is **PostgreSQL-compatible** (it speaks the PostgreSQL wire protocol), so the basics — host, port, database, user, password, SSL, and SSH tunneling — work the same way as a Postgres connection.
+
+In Beekeeper Studio, GreengageDB uses the same connection form as PostgreSQL.
```
not-night-but: No need to say that it uses the same connection form, especially as this may change in the future

docs/user_guide/connecting/greengage.md
```diff
@@ -0,0 +1,75 @@
...
+### GreengageDB 6 (PostgreSQL 9.4–compatible)
+
+GreengageDB 6 is based on an older PostgreSQL version, so some newer PostgreSQL features are not available.
+
+Known example:
+- **GENERATED columns** (PostgreSQL 12+) are not supported.
+
+### GreengageDB 7
+
+GreengageDB 7 is generally recommended when you have a choice.
```
not-night-but: This part seems like more of a guide for how to use Greengage than it is how to use beekeeper. I think we should probably just link to their docs. Like "To learn more about Greengage and how to set up your own development environment with it, visit the docs here"

not-night-but (changes requested):
I'm not sure we need to provide a guide on how to set up greengage as that is outside of our scope, but the docs in general are greatly appreciated

docs/user_guide/connecting/greengage.md
v-astakhov: updated

docs/user_guide/connecting/greengage.md
v-astakhov: updated

not-night-but (approved):
Thanks!

PR #4136, opened 2026-04-22
Some editing fixes

not-night-but:
Add an icon to the header of columns that are read-only in result editing, and also add a tooltip that gives a reason for why the column is read only (Couldn't map table, couldn't map column, missing PK, or column is generated).

Add the boolean dropdown for bool columns

image

image

### DIFF
--- a/apps/studio/src/components/editor/ResultTable.vue
+++ b/apps/studio/src/components/editor/ResultTable.vue
@@ -53,7 +53,7 @@
   import dateFormat from 'dateformat'
   import Converter from '../../mixins/data_converter'
   import Mutators from '../../mixins/data_mutators'
-  import { escapeHtml } from '@shared/lib/tabulator'
+  import { escapeHtml, FormatterParams } from '@shared/lib/tabulator'
   import { dialectFor, FormatterDialect } from '@shared/lib/dialects/models'
   import { FkLinkMixin } from '@/mixins/fk_click'
   import MagicColumnBuilder from '@/lib/magic/MagicColumnBuilder'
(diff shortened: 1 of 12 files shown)


azmy60 (changes requested):
So the goal is to tell users "this column isn't editable", right? If so, I'd do these to make it look better, for example:

1. Like you said before, double clicking the cells would show a tooltip that says "This cell is not editable". I think that's great.
2. If we are adding icons in the header, there are icons that are more recognizable. Something like "error_outlined" is going to be recognized easily because we use it quite a lot. Just don't show these icons when we stop editing. It's probably gonna be confusing if we stop editing, and we see error icons in the header, idk. Other icons like pk and fk are fine to keep.

Image

Image

apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -714,6 +846,8 @@
       async copyToSql() {
         this.saveError = null;
 
+        if (!this.editingData) return;
```
azmy60, line 849: We should at least log this as a warning, or tell users what's happening, like "I click this and why is this button not working?". Just don't leave them hanging. (just look at the rhyming man, I can't...)

azmy60: Or just disable/hide the button altogether.

apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -758,6 +892,8 @@
       async saveChanges() {
         this.saveError = null;
 
+        if (!this.editingData) return;
```
azmy60, line 895: Maybe a log here is helpful too.

apps/studio/src/components/editor/ResultTable.vue
not-night-but: The button is hidden, this was just added as a fallback if they try the keyboard shortcut when not in edit mode

azmy60 (approved):
Looks solid!

PR #4137, opened 2026-04-22
Fix read-only editor context menu

azmy60:
The context menu shows "write" actions even though the editor is in read only mode.

When read-only is enabled:

Screenshot 2026-04-22 at 13 21 59

When read-only is disabled:

Screenshot 2026-04-22 at 13 22 17


not-night-but:
I don't think we want to put the editor into read only mode if we're editing results.

not-night-but (approved):

PR #4138, opened 2026-04-22
make it clear that "editing result mode" is activated

azmy60:
So one thing in common that I noticed in many apps is that when a "mode" is activated, specifically professional apps, beside adding a label "This mode is activated", they usually disable/hide/swap elements.

For example,

- In Google Docs, if you click version history, all tools at the top are removed except print and zoom. So it shows you the tools that are usable only, cause you can't edit while navigating the version history.
- In Unity3D (used to play with this a while ago), if you click play, all elements are dim, except the camera view, which gives the effect of "hey, your game has started, you should focus on this view". You still can edit parameters/properties and all that, but the focus is now on the camera view.

The current solution hides the entire query editor, including the toolbar, with an overlay that shows a label and a "stop editing" button:

https://github.com/user-attachments/assets/77cbacdc-9fc2-439a-ae90-f7733808821a

## Alternative solution

We dim the query editor and the toolbar, and then we add an alert box with the same label and button:

Screenshot 2026-04-22 at 18 31 02

In this one, the editor is in read-only, so we can still select & copy the text. But toolbar buttons are disabled.

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -185,36 +185,6 @@
             >
           </x-button>
           <x-button
-            class="btn btn-flat btn-small settings-btn"
-            menu
-          >
-            <i class="material-icons">settings</i>
-            <i class="material-icons">arrow_drop_down</i>
-            <x-menu>
-              <x-menuitem disabled togglable>
-                <x-label>Editor keymap</x-label>
-              </x-menuitem>
-              <x-menuitem
-                :key="t.value"
-                v-for="t in keymapTypes"
-                togglable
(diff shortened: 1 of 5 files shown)


not-night-but:
@azmy60 The only thing I would say is maybe disable the editor only when there actually are changes?

rathboma:
For discoverability:

1. First time the user runs a query, add a tour tooltip to the edit button saying "Click here to edit query results and save the changes back to your database"
2. First time they click the button, tooltip on the whole result table saying "Double click any cell to edit it's value and (optionally) save the value back to your database"

rathboma:
Don't think we need to disable the query editor or have a banner about it

not-night-but (approved):

PR #4139, opened 2026-04-22
Fixes bug #4068 for Postgres + 'contains' filter in SQL queries

### LINKED ISSUE #4068: BUG: unable to edit array of enums
**Describe the bug**
we have a field that has array of enums, trying to edit it gives JSON errors

**To Reproduce**
Create a simple table with a field containing an array of enums and attempt to edit it.

**Expected behavior**
Should save correctly

**Screenshots / Gifs**

Image

 - OS:Mac
 - Version 5.6.5 (5.6.5)
 - PostgreSQL 16

Rovalx:
## Summary
This basically fixes https://github.com/beekeeper-studio/beekeeper-studio/issues/4068 for Postgres.

## Problem
Postgres arrays of enum values are represented as `{VALUE_1, VALUE_2}`, not as JSON arrays. The current implementation attempts to parse these values as JSON, which results in errors such as:
`Expected property name or '}' in JSON at position (line 1 column 2)`

This was driving me crazy and resulted in using other GUIs just to change enum array columns.

## Solution

•	Detect when a column is an array and avoid parsing its value as JSON.
•	Add a simple check to see if the value starts with `{` or `[` before deciding how to handle it.
•	Apply similar handling in the JSON Viewer to prevent parsing errors when displaying enum arrays.

## Additional Changes

•	Added a Postgres-only contains filter. I guess you may not want to have it in the same PR, but I still give it a shot :)
•	For array columns: uses `ANY()` to check if the provided value exists in the array.
•	For non-array columns: casts the value to text and checks if the resulting string contains the value.

This is extremely useful when working a lot with arrays in Postgres, and it was something I was really missing.

### DIFF
--- a/apps/studio/src/lib/data/jsonViewer.ts
+++ b/apps/studio/src/lib/data/jsonViewer.ts
@@ -169,7 +169,10 @@ export function parseRowDataForJsonViewer(data: Record<string, any>, tableColumn
 
       if (isColumnHasStringAndNotEmpty) {
         const trimmedValue = columnValue.trim()
-        const isJsonObjectString = trimmedValue.startsWith('{') && trimmedValue.endsWith('}')
+        const isJsonObjectString =
+          trimmedValue.startsWith('{') &&
+          trimmedValue.endsWith('}') &&
+          (trimmedValue === '{}' || /^\{\s*"/.test(trimmedValue))
         const isJsonArrayString = trimmedValue.startsWith('[') && trimmedValue.endsWith(']')
 
         if (isJsonObjectString || isJsonArrayString) {
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -125,7 +125,7 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
       restore: true,
       indexNullsNotDistinct: this.version.number >= 150_000,
       transactions: true,
-      filterTypes: ['standard', 'ilike']
+      filterTypes: ['standard', 'ilike', 'contains']
     };
   }
 
@@ -1431,6 +1431,13 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
           })
           paramIdx += values.length
           return `${wrapIdentifier(item.field)} ${item.type.toUpperCase()} (${values.join(',')})`
+        } else if (item.type === 'contains') {
+          const value = options.inlineParams
+            ? knex.raw('?', [item.value]).toQuery()
+            : `$${paramIdx}`
+          paramIdx += 1
+          const field = wrapIdentifier(item.field)
+          return `(CASE WHEN jsonb_typeof(to_jsonb(${field})) = 'array' THEN EXISTS (SELECT 1 FROM jsonb_array_elements_text(to_jsonb(${field})) AS element(value) WHERE element.value = ${value}) ELSE ${field}::text ILIKE '%' || ${value} || '%' END)`
         } else if (item.type.includes('is')) {
           return `${wrapIdentifier(item.field)} ${item.type.toUpperCase()} NULL`
         }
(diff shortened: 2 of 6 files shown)


apps/studio/src/lib/data/jsonViewer.ts
```diff
@@ -169,7 +169,10 @@ export function parseRowDataForJsonViewer(data: Record<string, any>, tableColumn
 
       if (isColumnHasStringAndNotEmpty) {
         const trimmedValue = columnValue.trim()
-        const isJsonObjectString = trimmedValue.startsWith('{') && trimmedValue.endsWith('}')
+        const isJsonObjectString =
+          trimmedValue.startsWith('{') &&
+          trimmedValue.endsWith('}') &&
+          (trimmedValue === '{}' || /^\{\s*"/.test(trimmedValue))
```
not-night-but, line 175: I don't think quotes are required around keys in JSON so this may not work consistently (please correct me if I'm wrong't. I wonder if we just try to match a colon? Like something like `/^\{\w*:/i`?

not-night-but (changes requested):
Thanks for the contribution ! Yes that enum array bug is quite annoying. I think this looks good, but we may want to refine the regex a bit in the json viewer. Also if you can retarget this to the `rc-57` branch we will be able to get it out much sooner

apps/studio/src/lib/data/jsonViewer.ts
Rovalx: Honestly, I didn’t put a lot of thought into it, so I had to do a little bit of digging.
According to this thread:
https://stackoverflow.com/questions/4201441/is-there-any-practical-reason-to-use-quoted-strings-for-json-keys
and this:
https://stackoverflow.com/questions/8758715/using-number-as-index-json
JSON keys have to be strings.

And according to RFC 8259, JSON objects are defined as collections of name/value pairs where the name (key) must be a string :)

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1773,17 +1780,20 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
...
-
-
-  // If a type starts with an underscore - it's an array
+    // If a type starts with an underscore - it's an array
   // so we need to turn the string representation back to an array
   private normalizeValue(value: string, column?: ExtendedTableColumn) {
     if (column?.array && _.isString(value)) {
-      return JSON.parse(value)
+      const trimmedValue = value.trim();
+      if (trimmedValue.startsWith("{")) {
```
rathboma, line 1788: Seems like we should start using `https://github.com/bendrucker/postgres-array` to parse postgres array values rather than doing it by hand.

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1431,6 +1431,13 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
           })
           paramIdx += values.length
           return `${wrapIdentifier(item.field)} ${item.type.toUpperCase()} (${values.join(',')})`
+        } else if (item.type === 'contains') {
+          const value = options.inlineParams
+            ? knex.raw('?', [item.value]).toQuery()
+            : `$${paramIdx}`
+          paramIdx += 1
+          const field = wrapIdentifier(item.field)
+          return `(CASE WHEN jsonb_typeof(to_jsonb(${field})) = 'array' THEN EXISTS (SELECT 1 FROM jsonb_array_elements_text(to_jsonb(${field})) AS element(value) WHERE element.value = ${value}) ELSE ${field}::text ILIKE '%' || ${value} || '%' END)`
```
rathboma, line 1440: definitely feels like we could smplify this, do we not have column type information when we get to this point? I think we do have the type, so we should be able to handle this in JS-land I think>

apps/studio/src/lib/data/jsonViewer.ts
rathboma: Should we be just using `JSON.parse` and asserting that the resulting value is an object, or do we think that's too heavy here? We do that elsewhere

rathboma: I'm fine with this change so long as we have some detailed unit tests to back it up (I bet claude/codex could easily generate some)

apps/studio/tests/integration/lib/db/clients/postgres.spec.ts
```diff
@@ -260,6 +260,29 @@ function testWith(dockerTag: TestVersion, socket = false, readonly = false) {
 
     })
 
+    it("Should allow me to update enum array values using postgres array literal", async () => {
+      const columns = await util.connection.listTableColumns("extra_moody_people")
+      const moodsColumn = columns.find((c) => c.columnName === "current_moods")
+
+      const updates = [{
+        value: "{ok,happy}",
```
rathboma, line 268: I think our drivers should accept js-native types, so in this case an array of strings, and then turn this into the correct enum array literal, rather than having the front-end handle them.

In the same vein, when editing this value in the app, presumably right now you're editing a string value, not an array? Really you should be editing an array.

I guess from my side if we're dealing with arrays, the conversion to/from db-specific types should happen at the driver, then the front-end can be focused on editing of the primitives?

That said - I recognize that this is an improvement over the status quo, I just wonder if it's worth the effort to do it the 'right' way. What do you think @Rovalx ?

PR #4155, opened 2026-04-22
fix: add empty state to quick switcher panel (#4150)

### LINKED ISSUE #4150: BUG: The switch connection pop-out needs an empty state when there are no connections you can switch to
Also (more) should be disabled if there are not more.

Image

ChihebBENCHEIKH1:
## Summary

The switch connection pop-out shows a blank list when there are no other
connections available. This adds:

- An empty state message ("No recent connections" / "No saved connections")
  when `displayedConnections` is empty
- Disables the more/less toggle button when the alternate view also has
  no connections to display

## Changes

- **ConnectionButton.vue** — added empty state div, `:disabled` binding
  on more/less button, and `moreDisabled` computed property
- **sidebar.scss** — added `.quick-switcher-empty` style and `&:disabled`
  state for `.quick-switcher-item`

Closes #4150


rathboma:
@ChihebBENCHEIKH1 any chance you could add a screenshot?

ChihebBENCHEIKH1:
@rathboma Here's the empty state when there are no recent connections:
Screenshot from 2026-04-23 12-44-18


The "more" button is still enabled here because there are saved connections available. When both views are empty, the button gets disabled.

azmy60 (changes requested):
First of all, I think the empty label should be aligned with the text above it like this:

Image

Second, I know this is not from your PR, but I think the chevron (or the angle bracket) icon doesn't need to be flipped if the sub menu is opened. It should just be > without change. It'd be awesome if you could also help us on fixing that.

Image

 Thank you @ChihebBENCHEIKH1 !

azmy60:
@ChihebBENCHEIKH1 Oh also, please hide the "more" item as well if it's empty. Thanks!

azmy60 (approved):

PR #4157, opened 2026-04-23
feat: add overwrite config API with renderer IPC wiring

azmy60:
How to edit the content of user config:

```js
// 1. read it
const rawContent = $bksConfig.getContent("user");

// 2. overwrite it
$bksConfig.overwrite("user", rawContent + "; LOOK me in the comment 👀")
```

NOTE: It only supports user type. The system type config isn't wired up yet (be aware of the permission stuff cause it sits in an admin folder). Please add later when needed.


azmy60:
This can be used for the built-in config editor UI.

PR #4174, opened 2026-04-27
Improve idle detection by combining system and app input tracking

rathboma:
## Summary
Enhanced the idle detection mechanism to be more reliable by combining system-level idle time with application-level input tracking. This addresses issues where the system idle monitor (particularly on Linux with Wayland and tiling window managers) would incorrectly report users as idle while they were actively using the application.

## Key Changes
- Added per-window input event tracking via a `WeakSet` to monitor user interactions within Beekeeper windows
- Introduced `lastAppInputAt` timestamp to track the most recent user input across all app windows
- Created `trackInput()` function to attach input event listeners to web contents
- Created `appIdleSeconds()` helper function to calculate idle time based on app input
- Modified idle check logic to use the minimum of system idle time and app idle time (more conservative approach)
- Enhanced logging to show both system and app idle times for better debugging
- Automatically track input events for all existing and newly created web contents

## Implementation Details
- Uses `WeakSet` for tracking to avoid memory leaks when web contents are destroyed
- Leverages Electron's `input-event` signal for reliable per-window input detection
- Registers input tracking on app startup and for dynamically created windows via `web-contents-created` event
- The effective idle time is the minimum of system and app measurements, ensuring the app disconnects when either indicates idleness

https://claude.ai/code/session_013tYKhmakJqfMB4JnGfJvJN


apps/studio/src/backend/lib/security.ts
```diff
@@ -2,27 +2,58 @@ import { getActiveWindows } from "@/background/WindowBuilder";
...
+const trackedContents = new WeakSet<WebContents>();
+
+const recordInput = _.throttle(() => {
+  lastAppInputAt = Date.now();
+}, 1000);
+
+function trackInput(contents: WebContents) {
+  if (trackedContents.has(contents) || contents.isDestroyed()) return;
+  trackedContents.add(contents);
+  contents.on("input-event", (_event, input) => {
```
rathboma: can't we limit the listener itself to only keydown/mousedown events?

PR #4175, opened 2026-04-28
Docs for editable query results

### LINKED ISSUE #4142: CHORE: Tweak wording when clicking 'edit data' for query results and there are no primary keys
Should say:

"Editing results cannot be enabled because no primary keys are included in the query"

Also should show a link to `learn more` which links to the section of the docs that applies to this.

not-night-but:
image

And fix #4142


azmy60:
@not-night-but Unrelated to the docs, and I don't know if we already handle this but I think we'll need a loading spinner on the top of the table or something to indicate that it's currently fetching extra information if the fetch takes like > 1 second.

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -1184,7 +1184,13 @@
           this.$refs.table.rebuildColumns()
 
           if (!resultEditData.some((e) => e.editable)) {
-            this.$noty.warning("There is not enough information in the result set to generate an update query. Make sure all primary keys are present.")
+            this.$noty.warning("There is not enough information in the result set to generate an update query. Make sure all primary keys are present.", {
```
azmy60: Should this be merged to rc-57?

azmy60 (approved):

PR #4180, opened 2026-04-28
fix(table): show loading spinner instead of 'No Data' while fetching

### LINKED ISSUE #4178: Oracle tables show “No Data” instead of “Loading” before rows appear
**Title:** Oracle tables show “No Data” instead of “Loading...” before rows appear

**Describe the bug**
When opening a table in Beekeeper Studio, the data grid initially shows **“No Data”** for an extended period before rows eventually appear. This happens even though the tables contain data and  is being loaded. The issue occurs across tables regardless of whether they contain small or large numbers of rows.

**To Reproduce**

1. Open Beekeeper Studio 5.6.5
2. Connect to an Oracle database
3. Browse tables in a connection containing ~12,285 entities
4. Click a table to view data
5. Table view displays **“No Data”** for a prolonged time before rows eventually load

**Expected behavior**
Instead of showing **“No Data”** while the table is still being fetched, the UI should show a **Loading...** state until rows are retrieved and rendered.

**Screenshots / Gifs**
Image

**Version Information**

* OS: Windows 11
* App Version: Beekeeper Studio 5.6.5
* Database type and version: Oracle Database 19c Enterprise Edition (19.0.0.0.0)

**Additional context**

* Happens on both small and large tables, so it does not appear related to row count.
* Connection contains approximately 12k entities.
* Tables do load, but show "No Data" instead of "Loading".
* Developer tools sometimes show initialization-related warnings/errors after loading (for example “Table Not Initialized...”, a `verticalFillMode` null error, and column definition warnings), though it’s unclear whether these are related or separate issues.

NianJiuZst:
## Summary

Fixes #4178 — Oracle tables show 'No Data' instead of 'Loading...' before rows appear.

### Problem

When opening a table in Beekeeper Studio (especially Oracle with large number of entities), the UI shows **'No Data'** during the extended fetch period. Users see this misleading message even though the table contains data and is still loading.

### Root Cause

The `TableTable.vue` component's `dataFetch()` function (Tabulator's ajaxRequestFunc callback) was fetching data asynchronously but there was no loading state tracking. The `isEmpty` computed property only checked if `this.data` was empty without distinguishing between:
1. Initial state (no data fetched yet)
2. Loading state (fetch in progress)
3. Empty result (fetch complete, table is truly empty)

### Solution

- Added `dataLoading` reactive property to track fetch state
- Set `dataLoading = true` at the start of `dataFetch()`
- Set `dataLoading = false` when fetch completes (success or error)
- Show loading spinner with 'Loading...' text when `dataLoading === true`
- Only show 'No Data' when `isEmpty && !dataLoading`

### Changes

- `apps/studio/src/components/tableview/TableTable.vue`:
  - Added `LoadingSpinner` import and component registration
  - Added `dataLoading: false` to component data
  - Added loading state tracking in `dataFetch()` lifecycle
  - Added conditional rendering logic in template

### Testing

- Open Oracle database with many entities
- Click on a table
- Should see loading spinner (not 'No Data') while rows are being fetched
- After data arrives, spinner disappears and rows display
- If table is empty after fetch completes, 'No Data' appears as expected


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -317,6 +323,7 @@ import Statusbar from '../common/StatusBar.vue'
 import RowFilterBuilder from './RowFilterBuilder.vue'
 import ColumnFilterModal from './ColumnFilterModal.vue'
 import EditorModal from './EditorModal.vue'
+import LoadingSpinner from "@/common/loading/LoadingSpinner.vue"
```
not-night-but: This is the wrong import path, should be this
```suggestion
import LoadingSpinner from "@/components/common/loading/LoadingSpinner.vue"
```

not-night-but (changes requested):
Looks like the import path is wrong, so the app isn't launching

rathboma (approved):

PR #4186, opened 2026-04-28
disable in read only mode, don't kick out of edit mode on apply

not-night-but:


apps/studio/src/components/editor/QueryEditorStatusBar.vue
```diff
@@ -352,6 +351,15 @@ export default {
     },
     canEdit() {
       return !this.dialectData?.disabledFeatures?.resultEditing;
+    },
+    editButtonTooltip() {
+      if (this.usedConfig.readOnlyMode) {
+        return "Read Only Mode is enabled for this connection. Editing is disabled.";
```
azmy60, line 357: Maybe the flow will be better if it starts with "Editing is disabled"? So something like: 
```
Editing result is disabled because we are in read only mode
```

not-night-but: This is just the tooltip we have everywhere else

azmy60 (approved):
Just a nitpick on the tooltip writing. But ALGETM (Actually Looks Good Enough To Me).

PR #4195, opened 2026-04-30
Support SSH config file resolution for host aliases

rathboma:
## Summary
This PR adds support for resolving SSH host aliases from `~/.ssh/config` files. Users can now enter a `Host` alias in SSH connection fields, and the application will automatically resolve the corresponding `HostName`, `Port`, `User`, and `IdentityFile` values from their SSH config.

## Key Changes

- **SSH Config Resolution**: Extended `connection-provider.ts` to read and merge SSH config values for both primary and bastion hosts across all authentication modes (keyfile, agent, userpass), not just agent mode
- **Precedence Logic**: Implemented proper precedence where user-entered values always take priority over SSH config values, which in turn take priority over SSH defaults
- **Host Alias Resolution**: The resolved hostname always replaces the alias, while other fields (port, user, identity file) only fill in missing values
- **Mode-Aware Identity File Handling**: Identity files from SSH config are only applied for keyfile and agent modes, never for userpass mode
- **Input Sanitization**: Added `.trim()` calls for bastion host and username to handle whitespace
- **SSH Config Reader Enhancement**: Updated `sshConfigReader.ts` to properly handle optional user fields
- **UI Documentation**: Added comprehensive help text and a new info alert in `CommonAdvanced.vue` explaining:
  - How to use SSH config aliases in host fields
  - The precedence order for value resolution
  - Fallback behavior for optional fields (username, port, identity file)
  - SSH agent availability warnings for different platforms
- **Test Coverage**: Added comprehensive unit tests covering alias resolution, precedence rules, mode-specific behavior, and edge cases

## Implementation Details

- SSH config resolution now works for all authentication modes, not just agent mode
- The `readSshConfig()` function is called for both primary and bastion hosts whenever they are specified
- Identity files are only pulled from SSH config for keyfile and agent modes, respecting the authentication method
- User-supplied values in form fields always take precedence over SSH config values
- The resolved hostname from SSH config always replaces the alias to ensure proper connection

https://claude.ai/code/session_01S8EoEHhdyZyYnfqESiauJW


apps/studio/src-commercial/backend/lib/connection-provider.ts
```diff
@@ -26,36 +26,50 @@ export default {
...
+      }
       if (fileConfig.user && !ssh.user) {
         ssh.user = fileConfig.user
       }
+      if (
+        fileConfig.identityFile &&
+        !ssh.privateKey &&
+        (config.sshMode === 'keyfile' || config.sshMode === 'agent')
+      ) {
+        ssh.privateKey = fileConfig.identityFile
```
rathboma: shouldn't the agent provide the authentication key if the user selected agent mode?

apps/studio/src-commercial/backend/lib/connection-provider.ts
```diff
@@ -26,36 +26,50 @@ export default {
...
+      }
       if (fileConfig.user && !ssh.bastionUser) {
         ssh.bastionUser = fileConfig.user
       }
+      if (
+        fileConfig.identityFile &&
+        !ssh.bastionPrivateKey &&
+        (config.sshBastionMode === 'keyfile' || config.sshBastionMode === 'agent')
+      ) {
+        ssh.bastionPrivateKey = fileConfig.identityFile
```
rathboma: shouldn't the agent provide the key for authentication?

apps/studio/src-commercial/backend/lib/connection-provider.ts
rathboma: Good catch — you're right. The agent should be the source of truth for the key in agent mode; pulling `IdentityFile` here muddies the intent (and could let `ssh2` pick the file-based key over the agent depending on order). Fixed in 0cd0bc7b: `IdentityFile` from `~/.ssh/config` is now only consumed for `keyfile` mode (target and bastion). Added a regression test and updated the docs.

---

apps/studio/src-commercial/backend/lib/connection-provider.ts
rathboma: Same fix applied for the bastion in 0cd0bc7b — `IdentityFile` from `~/.ssh/config` is now only consumed when `sshBastionMode === 'keyfile'`. Added a `sshBastionMode === 'agent'` regression test alongside the target-host one.

---

apps/studio/src/components/connection/CommonAdvanced.vue
rathboma, line 23: for host and bastion host you can add the little help icon with the help tooltip to talk about host aliases in the ssh config. Same thing with username.

Also - we no longer show that an ssh agent was found, or a ssh config was found, can we add little (check) agent found and (check) ssh config found in the help text underneath the auth mode dropdown when it's `automatic`?

rathboma: So it would be symbols like this:
- (check) ssh agent (tooltip - we found your ssh-agent socket - path/to/socket)
- (x) ssh agent (tooltip - we couldn't find a running ssh-agent to use)
- (check) ssh config (tooltip - config found at path/to/config)
- (x) ssh config (tooltip - we couldn't find an ssh config at path/to/expected/config

azmy60:
UI update:

Screenshot 2026-05-04 at 19 57 41


Clicking the help icon should open a modal (it should also mask the content if privacy mode is enabled):

Screenshot 2026-05-04 at 20 00 57

apps/studio/src-commercial/backend/lib/connection-provider.ts
```diff
@@ -26,36 +26,48 @@ export default {
...
       if (fileConfig.host) {
         ssh.host = fileConfig.host
       }
+      if (fileConfig.port && !ssh.port) {
+        ssh.port = fileConfig.port
+      }
       if (fileConfig.user && !ssh.user) {
         ssh.user = fileConfig.user
       }
+      if (config.sshMode === 'agent') {
```
azmy60, line 44: Aren't we supposed to remove 'agent' here and replaced by 'auto'?

rathboma: Good catch on the inconsistency. We reverted the value rename in `a8a5e252` after auditing cloud + SQLite compat — keeping `'agent'` as the persisted/wire value avoids breaking older clients that read connections from the cloud (they don't recognize `'auto'` and would silently disable agent mode). Only the UI label changed to "Automatic"; the dropdown still stores `'agent'`, so this `=== 'agent'` check is intentional.

---

azmy60 (approved):

PR #4198, opened 2026-04-30
Disable write action shortcuts in readonly mode

not-night-but:
Some of our actions had their buttons disabled but not their shortcuts in readonly mode, so added some guards to make sure they can't be triggered by either.

These write actions should be caught in the backend anyways (I think), so it's not a huge deal, just not super clean

### DIFF
--- a/apps/studio/src/components/tableview/TableTable.vue
+++ b/apps/studio/src/components/tableview/TableTable.vue
@@ -1432,6 +1432,7 @@ export default Vue.extend({
       }
     },
     cloneSelection(range?: RangeComponent) {
+      if (!this.editable) return;
       const rows = range && range.getRows ? range.getRows() : this.getSelectedRows()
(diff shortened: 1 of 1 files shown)


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1468,9 +1469,10 @@ export default Vue.extend({
       })
     },
     cellAddRow() {
-      if (this.dialectData.disabledFeatures?.tableTable) {
```
azmy60, line 1471: Just to understand: why is this removed?

azmy60 (approved):

apps/studio/src/components/tableview/TableTable.vue
not-night-but: Editable checks for disabled features already

PR #4211, opened 2026-05-05
fix: preserve column layout when refreshing table

azmy60:
Refreshing a table after hiding/resizing the columns resets the column states. This saves and restores the column layout around `setColumns`.


azmy60:
We can still reset the layout by right click a column > "Reset Layout".

rathboma (approved):

PR #4213, opened 2026-05-05
Add security exploit tests for shell execution vulnerabilities

rathboma:
This PR adds a comprehensive suite of security exploit tests that demonstrate and validate fixes for critical vulnerabilities across the application.

Note - all of these exploits require preexisting physical access to the user's machine, or control over the machine or the user's data in some way.

## Summary
Added 9 new security-focused test files that serve as both vulnerability demonstrations and regression tests. Each test file contains live exploit demonstrations (where feasible) that prove the vulnerability exists, along with assertions that verify the secure invariant once fixes are applied.

## Key Changes

**Backup & Command Execution:**
- `backupHandlers.exploit.spec.ts` - Demonstrates RCE via shell injection in backup command handlers through `shell: true` in spawn calls

**Database & Extensions:**
- `sqlite.exploit.spec.ts` - Demonstrates arbitrary code execution through SQLite's `loadExtension` API loading attacker-controlled shared libraries

**Authentication:**
- `cliAuth.exploit.spec.ts` - Demonstrates RCE in AWS and Azure CLI token helpers via shell injection in spawn calls

**File System Access:**
- `fileHandlers.exploit.spec.ts` - Demonstrates unrestricted file read/write access outside intended sandbox boundaries
- `pluginFileManager.exploit.spec.ts` - Demonstrates plugin sandbox escape via path traversal using `../` sequences

**Protocol Handlers:**
- `protocolBuilder.exploit.spec.ts` - Demonstrates file disclosure via dev-mode fallback in `app://` protocol and validates containment in `plugin://` protocol

**IPC Bridge Exposure:**
- `preloadOpenPath.exploit.spec.ts` - Validates that `openPath` bridge doesn't expose unguarded `shell.openPath` calls
- `webPluginLoader.exploit.spec.ts` - Validates that plugin postMessage uses specific origin targeting instead of wildcard `"*"`

## Implementation Details

- Tests use live spawning and file system operations to prove RCE primitives where possible
- Structural/source-level assertions validate fixes for vulnerabilities that are difficult to exploit in test environments
- All tests clean up temporary files in finally blocks regardless of pass/fail
- Tests are skipped on Windows where platform-specific exploits don't apply
- Each test file includes detailed comments explaining the vulnerability, the attack vector, and the expected secure behavior
- Tests currently fail (assertions expect secure behavior), validating that the vulnerabilities exist and will pass once fixes are applied

https://claude.ai/code/session_016QHRkDapNMnvRCDP89cpoc


apps/studio/src/handlers/fileHandlers.ts
```diff
@@ -1,33 +1,58 @@
-import { PathLike, WriteFileOptions } from "fs";
 import { promises as fs } from 'fs'
 import path from 'path';
+import platformInfo from '@/common/platform_info';
+
+const VIMRC_FILENAME = '.beekeeper.vimrc';
+const MAX_SQL_FILE_BYTES = 50 * 1024 * 1024; // 50 MB — sanity cap on a single saved query.
```
rathboma, line 6: these belong in the globals file

apps/studio/src/background/lib/electron/ProtocolBuilder.ts
```diff
@@ -77,9 +92,18 @@ export const ProtocolBuilder = {
       const url = new URL(request.url);
       const pluginId = url.host;
       const pathName = path.join(pluginId, url.pathname);
-      const normalized = path.normalize(pathName)
-      const fullPath = path.join(platformInfo.userDirectory, "plugins", normalized)
+      const pluginsRoot = path.resolve(path.join(platformInfo.userDirectory, "plugins"))
```
azmy60: We should use `platformInfo.pluginsDirectory` instead.

apps/studio/src/lib/db/clients/sqlite.ts
```diff
@@ -628,12 +629,28 @@ export class SqliteClient extends BasicDatabaseClient<SqliteResult> {
...
-        try {
-          connection.loadExtension(extension)
-        } catch (err) {
-          log.error(`Unable to load extension file ${extension}`)
-          throw err
+      // Loading SQLite runtime extensions executes arbitrary native code from
+      // the extension path via dlopen()/LoadLibrary. Require the user to
+      // explicitly opt in through bksConfig.security.allowRuntimeExtensions
+      // before honouring any extension paths from the connection config.
+      const allowRuntimeExtensions = !!_.get(
```
azmy60: You should be able to do `bksConfig.security.allowRuntimeExtensions` instead without `_.get`.

apps/studio/tests/unit/security/pluginFileManager.exploit.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    );
+  });
+
+  it("readAsset refuses to read files that traverse out of the plugin's own directory", () => {
+    const fm = new PluginFileManager({ pluginsDirectory: pluginsDir });
+    const manifest = { id: "evil-plugin" } as any;
+
+    let result: string | null = null;
+    try {
+      result = fm.readAsset(manifest, "../victim-plugin/SECRET.txt");
```
azmy60: Can we use expect(...).toThrow() instead?

apps/studio/tests/unit/security/webPluginLoader.exploit.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+ * We do this as a source-level structural check rather than a runtime test
+ * because the WebPluginLoader module references PointerEvent at module-load
+ * time, which is not present in jsdom and brings in a long dependency tail.
+ */
+
+import * as fs from "fs";
+import * as path from "path";
+
+describe("WebPluginLoader postMessage exploit", () => {
+  it("postMessage targets plugin://<id> rather than '*' (else any embedded frame can intercept)", () => {
```
azmy60: This isn't really a test? It just reads the code file.

I think the test should run two plugins, and make one plugin exploits the other, if it's possible.

apps/studio/default.config.ini
```diff
@@ -12,6 +12,10 @@ lockMode = disabled ; disabled, pin, future: google, ldap, oauth; applies when c
 idleThresholdSeconds = 300 ; time before user is considered 'idle'
 idleCheckIntervalSeconds = 30 ; time between idle checks
 minPinLength = 6
+; Loading SQLite runtime extensions executes arbitrary native code from the
+; extension path. Off by default - users must explicitly opt in via their
+; user.config.ini before any sqliteExtensionFile setting will be honoured.
+allowRuntimeExtensions = false
```
azmy60, line 19: This will look like a regression for users who use SQLite extensions because we're being silent with this change. We should add an alert right next to the extension form and hide it if the config is changed / enabled.

azmy60 (commented):
Overall looks good for plugin system!

apps/studio/tests/unit/security/webPluginLoader.exploit.spec.ts
rathboma: Rewrote it in c0a00433 — it now stands up two `WebPluginLoader` instances ("alpha" and "beta"), gives each a fake iframe, has `alpha.broadcast(...)` fire, and asserts the message reaches only alpha's own iframe and is addressed to `plugin://alpha` (never `"*"`, never `plugin://beta`). So it exercises the real `postMessage` path now rather than grepping the source.

One caveat: it can't literally run two real `plugin://` origins and prove the browser drops the cross-origin delivery — jsdom doesn't let you set arbitrary iframe origins, and `WebPluginLoader.ts` references `PointerEvent` at module-load time which jsdom lacks (aliased before `require()`). So the assertion is "the target origin is locked to the sender's own plugin origin" rather than a live interception attempt. If you'd prefer a full Playwright-level test with real origins, that's doable but heavier — happy to follow up if you want it.

---

apps/studio/default.config.ini
rathboma: Added in c0a00433 — `SqliteForm.vue` now shows a warning in the Runtime Extensions area when `bksConfig.security.allowRuntimeExtensions` is off ("Configured extensions will be ignored until you set `allowRuntimeExtensions = true` under `[security]`…"), and it disappears once the config is enabled.

---

PR #4217, opened 2026-05-06
Query Edit History

azmy60:
https://github.com/user-attachments/assets/9497c2b4-b7c3-4a4a-ad77-efb375b65ee5


rathboma:
I get an error trying to start this branch:

image

rathboma:
Same error in the e2e tests

rathboma:
This looks great. Works perfectly.

Couple of questions about discoverability and UX:

1. The only place this is available is the query right-click menu, is that right?
2. Can we add to both the file menu, or maybe we add a new `query` menu when the user is on a query tab (maybe a separate PR for this) and then it can be listed in there too.
3. What about the right click menu for a query in the sidebar?
4. What about the right click menu for the tab header

apps/studio/src/common/date.ts
```diff
@@ -3,3 +3,84 @@ export function monthAgo(): Date {
   date.setMonth(date.getMonth() - 1);
   return date;
 }
+
+/** Human-readable group label for `when` relative to `now`, used by the query
+ * edit history sidebar (e.g. "Today", "Yesterday", "Last month"). */
+export function relativeDateGroup(when: Date, now: Date): string {
```
rathboma: I believe we already have a library for this - `javascript-time-ago` we use it already in the historylist, tabletable, and favorite list item

apps/studio/src/components/editor/QueryEditHistory.vue
```diff
@@ -0,0 +1,646 @@
...
+      showDiff: true,
+      split: null as ReturnType<typeof Split> | null,
+    };
+  },
+  watch: {
+    open: {
+      immediate: true,
+      async handler(isOpen: boolean) {
+        if (isOpen) {
+          this.loadHistory();
```
rathboma, line 175: Does this need an `await`?

apps/studio/src/common/date.ts
rathboma: `this.$bks.timeAgo(someDate)`

apps/studio/src/components/editor/QueryEditHistory.vue
```diff
@@ -0,0 +1,646 @@
...
+        return false;
+      }
+      return target.id === this.queryAudits[0].id;
+    },
+    formatTime(createdAt: number): string {
+      // createdAt is float seconds since epoch (cloud API convention).
+      const d = new Date(createdAt * 1000);
+      const month = d.toLocaleString("en-US", { month: "long" });
+      const day = d.getDate();
+      const time = d.toLocaleString("en-US", {
```
rathboma, line 326: shouldn't we format according to the user locale?

apps/studio/src/components/editor/QueryEditHistory.vue
```diff
@@ -0,0 +1,646 @@
...
+    },
+    async selectAudit(auditId: number | "unsaved"): Promise<void> {
+      const queryId = this.queryId;
+      if (queryId == null) {
+        return;
+      }
+      if (auditId === this.selectedAuditId) {
+        return;
+      }
+      this.selectedAuditId = auditId;
```
rathboma, line 377: we don't reset this on error, is that ok?

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -1785,8 +1796,18 @@ import { KeybindingPath } from '@/common/bksConfig/BksConfigProvider'
           {
             label: "Open Query Formatter",
             id: "formatter",
-            handler: this.formatterPreset
+            handler: this.formatterPreset,
           },
+          ...(this.query?.id && this.isCloud
```
rathboma, line 1806: So if I'm in a cloud connection and I have a new unsaved query, this menu option doesn't show at all. We should just disable the menu entry in that situation (with a title explaining why), rather than making it not appear at all

rathboma (changes requested):
Overal A+, would be fine to ship.

Review only contains minor questions, maybe one bug? There's a lib we already use that might be appropriate, either way we should be consistent with `time ago` wording in different parts of the app

rathboma:
Also meta question -- doesn't seem that crazy to enable this for local workspaces in a future PR?

azmy60:
@rathboma Awesome! These are good questions. Some of them I stumbled upon but didn't get there because I was probably too focus on coding and designing (took me a couple of trials) and just want to get it done, but yeah I'll do a follow up when I can, probably next monday!

apps/studio/src/components/editor/QueryEditHistory.vue
azmy60: Sure. But other components still use this format, so changing this would make it look inconsistent. Probably it's best to plan a little bit about all time elements (including the timestamp tooltip) which is out of scope of this feature I think.

apps/studio/src/components/editor/QueryEditHistory.vue
azmy60: I'm not feeling quite sure about that. I'm imagining clicking an audit, and then it's throwing an error after 200ms, which resets the selected audit, and then I do it again 5x and it keeps resetting to the same previous audit. Isn't it a liiiittle bit more annoying?

azmy60: Like if there's a rope attached to a device, and you pull it towards you, but the device doesn't like it so it pulls back, like tug of war.

azmy60: But maybe it's just my preference

azmy60:
@rathboma Yes, not too crazy for local workspaces. It'll be verbose due to the nature of the codebase, but the concept is straight forward.

azmy60:
@rathboma Also answering the other questions:

1. We can now open the history from right clicking the query from sidebar too.
2. I'll think about a separate PR. I'm actually leaning towards a button below the editor or somewhere.
3. Yes, already pushed it.
4. I think tab header is best for tab related only.

rathboma (approved):

PR #4220, opened 2026-05-06
fix(cassandra): convert UUIDs in set and map collections to strings

andreasarf:


rathboma (commented):
This looks great, but def needs some tests adding

rathboma (approved):

PR #4226, opened 2026-05-08
fix shortcut for switching tabs on mac

gruberlu:
This pull request fixes a bug for the current tab switching shortcuts on mac. For some keyboard layouts `alt+<0-9>` binds to regularly used symbols for writing SQL.


not-night-but:
@gruberlu I'm so sorry! I thought this got pulled into Matthew's branch, but I guess that got closed. I think the best way to do this is to actually not map `ALTORCMD` to `ALT` but map it to a function like how the uiModifierMap is handling it. That way we don't need to do any special handling in the conversion function

gruberlu:
@not-night-but i did map all of the special bindings to functions now

apps/studio/src/common/bksConfig/BksConfigProvider.ts
```diff
@@ -76,6 +80,10 @@ const electronModifierMap = {
   COMMANDORCONTROL: "CommandOrControl",
   SHIFT: "Shift",
   ALT: "Alt",
+  ALTORCMD: "CommandOrAlt",
+  CMDORALT: "CommandOrAlt",
+  ALTORCOMMAND: "CommandOrAlt",
+  COMMANDORALT: "CommandOrAlt",
```
not-night-but: This doesn't exist in electron, so it'll have to be a function

apps/studio/src/common/bksConfig/BksConfigProvider.ts
```diff
@@ -56,6 +56,10 @@ const codeMirrorModifierMap = {
   COMMANDORCONTROL: "Mod",
   SHIFT: "Shift",
   ALT: "Alt",
+  ALTORCMD: "Alt",
+  CMDORALT: "Alt",
+  ALTORCOMMAND: "Alt",
+  COMMANDORALT: "Alt",
```
not-night-but: Probably want a function here too

apps/studio/src/common/bksConfig/BksConfigProvider.ts
```diff
@@ -112,6 +125,10 @@ export const tabulatorModifierMap = {
   COMMANDORCONTROL: "ctrl",
   SHIFT: "shift",
   ALT: "alt",
+  ALTORCMD: "alt",
+  CMDORALT: "alt",
+  ALTORCOMMAND: "alt",
+  COMMANDORALT: "alt",
```
not-night-but: Tabulator doesn't handle switching between these so you'll want a function here too

not-night-but (changes requested):
Just need to update the other maps

gruberlu:
i've adjusted the electron and code mirror mappings and removed the mapping for tabulator as it does not seem to handle `command` anyways

not-night-but:
@gruberlu Thanks!

not-night-but (approved):

PR #4231, opened 2026-05-08
add altOrCmd modifier alias and fix tabulator case fall-through

rathboma:
## Summary

- Adds an `altOrCmd` modifier (plus `cmdOrAlt`/`altOrCommand`/`commandOrAlt` aliases) for keybindings that should bind to Alt on Linux/Windows but Cmd on Mac. The motivating case: `alt+<digit>` collides with SQL-relevant symbols (`@`, `#`, `[`, `]`, etc.) on several Mac keyboard layouts, so tab-switching shortcuts like `alt+1..9` are unusable for those users — see #4226.
- Wires the new alias through every modifier map (codemirror, electron, v-hotkey, tabulator, ui, context-menu) as a platform-aware function value, so each subsystem gets the right native token (`Cmd` vs `Alt` for CodeMirror, `Command` vs `Alt` for Electron accelerators, `meta` vs `alt` for v-hotkey, etc.). The function form is already supported by `convertKeybinding`, so no special-case branches are needed in the conversion logic.
- Fixes a pre-existing missing `break` in the `tabulator` case of `convertKeybinding`'s target switch, which was silently falling through into `case "ui"` and overwriting `modifierMap` with `uiModifierMap`. Tabulator keybindings were being resolved against UI symbols.

With this in place, #4226 collapses down to just the `default.config.ini` change.

## Test plan

- [x] `yarn test:unit --testPathPattern=config.spec` — existing tests still pass
- [x] new test: `altOrCmd+1` resolves correctly per-platform across `v-hotkey`, `electron`, `codemirror`, `tabulator`, and `ui` targets
- [x] new test: tabulator target produces tabulator-style output (`ctrl + shift + c`), confirming the fall-through fix
- [x] alias spellings (`cmdOrAlt`, `altOrCommand`, `commandOrAlt`) resolve identically to `altOrCmd`


apps/studio/src/common/bksConfig/BksConfigProvider.ts
```diff
@@ -44,7 +44,14 @@ interface IBksConfigDebugInfo {
 
 const log = rawLog.scope("BksConfigProvider");
 
-const codeMirrorModifierMap = {
+const codeMirrorAltOrCmd = (isMac: boolean) => (isMac ? "Cmd" : "Alt");
+const electronAltOrCmd = (isMac: boolean) => (isMac ? "Command" : "Alt");
+const vHotkeyAltOrCmd = (isMac: boolean) => (isMac ? "meta" : "alt");
+const tabulatorAltOrCmd = (isMac: boolean) => (isMac ? "ctrl" : "alt");
+const uiAltOrCmd = (isMac: boolean) => (isMac ? "⌘" : "Alt");
+const contextMenuAltOrCmd = (isMac: boolean) => (isMac ? "Meta" : "Alt");
```
not-night-but: I would rather these be inline like the rest of the mod maps so that  we don't have to jump around to read it

not-night-but (changes requested):

apps/studio/src/common/bksConfig/BksConfigProvider.ts
```diff
@@ -145,6 +165,10 @@ const contextMenuModifierMap: ModifierMap = {
   COMMAND: "Control",
   CONTROLORCOMMAND: "Control",
   COMMANDORCONTROL: "Control",
+  ALTORCMD: (isMac) => (isMac ? "Meta" : "Alt"),
+  CMDORALT: (isMac) => (isMac ? "Meta" : "Alt"),
+  ALTORCOMMAND: (isMac) => (isMac ? "Meta" : "Alt"),
+  COMMANDORALT: (isMac) => (isMac ? "Meta" : "Alt"),
```
not-night-but, line 171: I believe this is wrong. Looking at the code, it looks like xel displays meta as `^` and Control as `⌘`. Either way we should just be consistent

apps/studio/src/common/bksConfig/BksConfigProvider.ts
```diff
@@ -91,16 +99,20 @@ const vHotkeyModifierMap: ModifierMap = {
   COMMAND: "cmd",
   CONTROLORCOMMAND: "ctrlOrCmd",
   COMMANDORCONTROL: "ctrlOrCmd",
+  ALTORCMD: (isMac) => (isMac ? "meta" : "alt"),
+  CMDORALT: (isMac) => (isMac ? "meta" : "alt"),
+  ALTORCOMMAND: (isMac) => (isMac ? "meta" : "alt"),
+  COMMANDORALT: (isMac) => (isMac ? "meta" : "alt"),
```
not-night-but, line 105: We should use `cmd` here to be consistent

not-night-but (changes requested):

apps/studio/src/common/bksConfig/BksConfigProvider.ts
rathboma: oh yeah, my bad

PR #4238, opened 2026-05-10
Fix: unable to edit PostgreSQL array of enums (JSON error)

### LINKED ISSUE #4068: BUG: unable to edit array of enums

rathboma:
$(cat <<'EOF'
## The Bug

Fixes #4068 — "BUG: unable to edit array of enums"

When a PostgreSQL table has a custom enum array column (e.g. `this_is_a_mood[]`), editing the cell throws a JSON error and the update never reaches the database.

### Root cause

`node-postgres` (`pg`) only has registered type parsers for **built-in** array types (`_text`, `_int4`, etc.). For **custom** enum array types it has no registered parser, so it returns the raw PostgreSQL array literal as a plain string — e.g. `{sad,happy}` instead of `["sad","happy"]`.

`normalizeValue()` in `postgresql.ts` is responsible for turning stringified arrays back into JS arrays before building the SQL. It called `JSON.parse(value)` unconditionally, which throws a `SyntaxError` for PostgreSQL array literal syntax (`{sad,happy}` is not valid JSON). That error propagated up as the "JSON error" the user saw.

```typescript
// Before — throws for PostgreSQL array literals like {sad,happy}
private normalizeValue(value: string, column?: ExtendedTableColumn) {
  if (column?.array && _.isString(value)) {
    return JSON.parse(value)   // ← SyntaxError for {sad,happy}
  }
  return value
}
```

## The Fix

Wrap `JSON.parse` in a try/catch. If parsing fails (PostgreSQL array literal format), return the value as-is. PostgreSQL can parse its own array literal syntax on assignment, so the `UPDATE`/`INSERT` succeeds:

```typescript
// After — handles both JSON ["sad","happy"] and PostgreSQL literal {sad,happy}
private normalizeValue(value: string, column?: ExtendedTableColumn) {
  if (column?.array && _.isString(value)) {
    try {
      return JSON.parse(value)
    } catch {
      // pg returns custom enum arrays as PostgreSQL array literals (e.g. {sad,happy})
      // rather than JSON. Pass the string through so PostgreSQL can parse it on assignment.
      return value
    }
  }
  return value
}
```

This is safe because both paths produce a value PostgreSQL can accept:
- JSON array `["sad","happy"]` → parsed to JS array → knex serialises as `ARRAY['sad','happy']`
- PostgreSQL literal `{sad,happy}` → passed through as string → knex quotes it as `'{sad,happy}'`, which PostgreSQL parses natively on assignment to the enum array column

## Test

Added a regression test in `postgres.spec.ts`:
- Seeds a row into `extra_moody_people` (table with `current_moods this_is_a_mood[]`)
- Calls `applyChanges` with `value: '{happy,ok}'` — the exact format `pg` returns for a custom enum array column
- Verifies the call resolves without a JSON error

https://claude.ai/code/session_01PeAMAcEkuNZuWN6zGT6nNZ
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_01PeAMAcEkuNZuWN6zGT6nNZ)_


not-night-but:
@rathboma Are we okay with this? Feels like a bit of a bandaid, and I think it's possible that a malformed array may be then returned as a string and possibly mutate the value

rathboma:
@not-night-but Fair concern — updated in dc8989d5 to use `pg`'s own registered text-array parser (OID 1009) instead of passing the raw string through:

```typescript
if (value.startsWith('{') && value.endsWith('}')) {
  const parseTextArray = pg.types.getTypeParser(1009, 'text')
  return parseTextArray(value)
}
```

This decodes `{sad,happy}` into `["sad","happy"]` (a proper JS array) the same way pg handles built-in array types. The JS array then goes through the normal knex path, which already works correctly for text arrays (there are existing tests for it). No raw string ever reaches the DB.

---

Rovalx:
@rathboma, this is something I tried to fix back in April here:
https://github.com/beekeeper-studio/beekeeper-studio/pull/4139

I saw your comments but didn’t have time to jump into it. As long as your fix is simpler, will it also work with JSON Viewer? I have also added support for a “contains” filter for arrays. If you think my PR is no longer needed, I will close it.

not-night-but:
@Rovalx I'm so sorry I totally missed this. I think I'm going to merge this change as it handles it at the root of the problem, but I like your contains filter! Can you maybe open a separate PR with just the contains filter in it?

not-night-but (approved):

PR #4244, opened 2026-05-12
Fix two import bugs: generateColumnTypesFromFile TypeError and CSV error message

rathboma:
## Summary

Two unrelated import bugs, each with a regression test.

### Bug 1 — `generateColumnTypesFromFile` crashes on JSON/XLSX with heterogeneous rows

**File:** `apps/studio/src/lib/import/index.ts`

When a JSON or XLSX file contains rows where later objects have fields that were absent from the first row, `meta.fields` (derived from `Object.keys(data[0])`) will not include those extra keys. Inside `generateColumnTypesFromFile`, `columns.indexOf(key)` then returns `-1` for any such extra key.

The guard was:
```typescript
if (arrInd == null) return   // -1 == null is false in JS!
```
`-1 == null` evaluates to `false` (JavaScript loose-equality with `null` only matches `null`/`undefined`), so the guard never fires. Execution falls through to `dataAnalysis[-1].add(...)` — `dataAnalysis[-1]` is `undefined` — and throws:
```
TypeError: Cannot read properties of undefined (reading 'add')
```
This crashes the column-mapper step of the "Import from file" flow for any JSON export where optional fields differ between rows (a very common real-world shape from APIs).

**Fix:** `if (arrInd < 0) return`

---

### Bug 2 — CSV parse errors always showed `[object Object]` to the user

**File:** `apps/studio/src-commercial/backend/lib/import/formats/csv.ts`

The PapaParse `chunk` callback built a correct human-readable error string inside an `if (Array.isArray(errors))` block, but then fell through (no `return`) to a second assignment:
```typescript
this.error = errors.join(', ');  // always runs; PapaParse error objects have no .toString()
```
This overwrote the properly-formatted message with `[object Object]` every time a CSV parse error occurred (wrong delimiter, malformed quoting, etc.).

**Fix:** Collapse to a single code path that always builds the formatted message.

---

## Test plan

- [ ] `yarn test:unit` — both new test cases pass
- [ ] Manual: import a JSON file with optional fields (e.g. row 1 has `{id, name}`, row 2 has `{id, name, email}`) — import mapper should open without error
- [ ] Manual: import a CSV file with a deliberately wrong delimiter — error dialog should show a readable message, not `[object Object]`

https://claude.ai/code/session_01DcMDWdH9cTTgiN6oBNzttx

---
_Generated by [Claude Code](https://claude.ai/code/session_01DcMDWdH9cTTgiN6oBNzttx)_


apps/studio/src-commercial/backend/lib/import/formats/csv.ts
```diff
@@ -22,16 +22,11 @@ export default class extends Import {
...
-            parser.abort();
-          }
-          // I don't think this should be able to happen, we shall see
-          this.error = errors.join(', ');
+          const errorSet = new Set()
+          errors.forEach(e => {
+            errorSet.add(e?.message || String(e))
+          })
+          this.error = Array.from(errorSet).join(', ');
           parser.abort();
```
not-night-but, line 35: Just based on the comment, I'm wondering if there was maybe supposed to be a return here? Like within the `Array.isArray` block, after the `parser.abort`. Either way, does it make sense to continue after this if there are errors?

not-night-but (approved):

PR #4248, opened 2026-05-13
Fix #4223: JSONB string scalars now display with outer quotes in table cells

### LINKED ISSUE #4223: BUG: jsonb cells render the same regardless of whether they're a JSON string scalar or a JSON object
> Split out from #3510 (which I'd like to close in favor of #4222 — that PR fixes the insert-path double-encoding). This issue tracks the remaining display bug.

**Describe the bug**

The table view's `jsonb` cell renderer collapses two distinct Postgres values — a `jsonb` *object* and a `jsonb` *string scalar* whose text happens to look like JSON — into the same visual representation. They have different `jsonb_typeof` results, different query semantics, and different wire representations, but Beekeeper shows them identically. There is no way to tell from the table view that a cell is actually a string containing escaped JSON rather than a real JSON object.

This is what made #3510 hard to notice: the insert was being double-encoded, but the cell looked correct.

**To Reproduce**

```sql
CREATE TABLE temp (id int, j jsonb);

-- Row A: a proper JSONB object
INSERT INTO temp VALUES (1, '{"hello":"world"}'::jsonb);

-- Row B: a JSONB STRING scalar whose textual value happens to be JSON
INSERT INTO temp VALUES (2, to_jsonb('{"hello":"world"}'::text));

SELECT id, jsonb_typeof(j) AS type, j FROM temp;
```

Open the `temp` table in Beekeeper's table view.

**Expected behavior**

Row A and Row B should look different, the way they do in psql / DBeaver / DataGrip:

```
 id |  type  |           j
----+--------+-----------------------------
  1 | object | {"hello": "world"}
  2 | string | "{\"hello\":\"world\"}"
```

Row B should display the surrounding quotes and the backslash escapes — i.e. the raw `jsonb` text representation Postgres returns.

**Actual behavior**

Both rows render as `{"hello":"world"}` in the table cell, with no visual distinction.

**Additional context**

- Originally reported (alongside the insert bug) in #3510.
- Repro and writeup credit: @srjchauhan (https://github.com/beekeeper-studio/beekeeper-studio/issues/3510#issuecomment-4253719294).
- The companion insert-path bug is fixed by #4222.

rathboma:
$(cat <<'EOF'
## The Bug

Fixes #4223 — JSONB cells render the same regardless of whether they're a JSON string scalar or a JSON object.

PostgreSQL stores two fundamentally different JSONB values the same way in the table view:

```sql
-- Row A: a proper JSONB object
INSERT INTO temp VALUES (1, '{"hello":"world"}'::jsonb);

-- Row B: a JSONB STRING scalar whose text value looks like JSON
INSERT INTO temp VALUES (2, to_jsonb('{"hello":"world"}'::text));
```

Before this fix, both rows displayed as `{"hello":"world"}` — indistinguishable. `psql`, DBeaver, and DataGrip all display row B as `"{"hello":"world"}"` (with outer quotes and escaped inner characters).

## Root Cause

`node-postgres` calls `JSON.parse()` on every JSONB column value before handing the result to the application:

| JSONB value | PostgreSQL wire text | After `JSON.parse` (JS value) |
|---|---|---|
| object `{"hello":"world"}` | `{"hello":"world"}` | JS object `{hello:"world"}` |
| string scalar `"{"hello":"world"}"` | `"{"hello":"world"}"` | JS string `'{"hello":"world"}'` |

So by the time `jsonMutator` in `tools.ts` sees the data, the distinction is entirely preserved in the JS type — objects for JSONB objects, strings for JSONB string scalars.

`jsonMutator` previously short-circuited with `if (!_.isObject(value)) return value`, returning the string as-is. The Tabulator cell then rendered it as `{"hello":"world"}` — visually identical to the object.

## The Fix

Wrap JS string values in `JSON.stringify()` before returning from `jsonMutator`. This renders the cell with the JSON text representation of a string — outer double-quotes and escaped inner characters — so the two types are visually distinct:

| Type | JS value received | Cell displays (before) | Cell displays (after) |
|---|---|---|---|
| JSONB object | `{hello:"world"}` | `{"hello":"world"}` | `{"hello":"world"}` ✓ (unchanged) |
| JSONB string scalar | `'hello'` | `hello` ❌ | `"hello"` ✓ |
| JSONB string scalar (JSON-like) | `'{"hello":"world"}'` | `{"hello":"world"}` ❌ | `"{\"hello\":\"world\"}"` ✓ |

## Tests

**Unit** (`tests/unit/lib/data/mutators.spec.js`):
- `jsonMutator(null)` → `null`
- `jsonMutator("hello")` → `'"hello"'` (outer quotes added)
- `jsonMutator('{"hello":"world"}')` → differs from raw string (the bug scenario)
- `jsonMutator({hello:"world"})` → object with custom `toString()` producing `{"hello":"world"}`
- `jsonMutator(42)` / `jsonMutator(true)` → unchanged

**Integration** (`tests/integration/lib/db/clients/postgres/jsonb.spec.ts`):
- Confirms `node-postgres` returns JSONB string scalars as JS strings (not objects)
- Verifies `jsonMutator` produces distinct output for JSONB object vs. string scalar vs. JSON-like string scalar

https://claude.ai/code/session_014AbZd9uxuGVVf9WV2ExB8K
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_014AbZd9uxuGVVf9WV2ExB8K)_

### DIFF
--- a/apps/studio/src/lib/data/tools.ts
+++ b/apps/studio/src/lib/data/tools.ts
@@ -27,7 +27,7 @@ export const Mutators = {
       return this.bitMutator.bind(this, dialect)
     }
     if (dataType && dataType.startsWith('json') && mutateJSON) {
-      return this.jsonMutator.bind(this)
+      return this.jsonMutator.bind(this, dialect)
     }
     return this.genericMutator.bind(this)
   },
@@ -105,15 +105,28 @@ export const Mutators = {
   },
 
   /** Stringify json data for display in table cells.
-   * node-postgres returns JSONB scalar strings as JS strings (already parsed),
-   * so wrapping them in JSON.stringify preserves the JSON representation with outer quotes,
-   * making string scalars visually distinct from JSON objects.
+   *
+   * Dialect matters here because drivers differ in how they hand back JSON values:
+   *
+   * - PostgreSQL (node-postgres): calls JSON.parse() on the wire value before the app
+   *   sees it, so a JSONB *string scalar* arrives as a plain JS string while a JSONB
+   *   *object* arrives as a JS object. Wrapping the JS string in JSON.stringify()
+   *   preserves the distinction (cell shows `"hello"` not `hello`, and
+   *   `"{\"k\":\"v\"}"` not `{"k":"v"}`).
+   *
+   * - MySQL and other dialects: the driver returns JSON columns as raw unparsed strings,
+   *   so returning the string as-is lets the cell show the JSON text directly (correct).
    */
-  jsonMutator(value: any): JsonFriendly {
+  jsonMutator(dialect: Dialect | null | undefined, value: any): JsonFriendly {
     if (_.isNull(value)) return value
-    // A JS string means the driver already parsed the JSONB value and got a string scalar.
-    // Wrap it so the cell shows the JSON text representation (e.g. "hello", not hello).
-    if (_.isString(value)) return JSON.stringify(value)
+    if (_.isString(value)) {
+      // node-postgres pre-parses JSONB, so a string here is already a scalar value.
+      if (dialect === 'postgresql' || dialect === 'greengage' || dialect === 'redshift') {
+        return JSON.stringify(value)
+      }
+      // Other drivers return JSON as raw strings; display the text content directly.
+      return value
+    }
     if (!_.isObject(value)) return value
     try {
       return friendlyJsonObject(value)
--- a/apps/studio/src/lib/data/tools.ts
+++ b/apps/studio/src/lib/data/tools.ts
@@ -104,9 +104,17 @@ export const Mutators = {
 
   },
 
-  /** Stringify json data for MySQL column */
+  /** Stringify json data for display in table cells.
+   * node-postgres returns JSONB scalar strings as JS strings (already parsed),
+   * so wrapping them in JSON.stringify preserves the JSON representation with outer quotes,
+   * making string scalars visually distinct from JSON objects.
+   */
   jsonMutator(value: any): JsonFriendly {
-    if(!_.isObject(value)) return value;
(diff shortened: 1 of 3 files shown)


apps/studio/src/lib/data/tools.ts
```diff
@@ -104,9 +104,17 @@ export const Mutators = {
 
   },
 
-  /** Stringify json data for MySQL column */
+  /** Stringify json data for display in table cells.
+   * node-postgres returns JSONB scalar strings as JS strings (already parsed),
+   * so wrapping them in JSON.stringify preserves the JSON representation with outer quotes,
+   * making string scalars visually distinct from JSON objects.
+   */
   jsonMutator(value: any): JsonFriendly {
```
rathboma: Does this screw up optimistic parsing of TEXT as JSON (eg for SQLite)?

rathboma: Good catch — yes, it did. I've pushed a fix in the latest commit.

The problem is that `mysql2` returns JSON column values as raw unparsed strings (e.g. `'{"hello":"world"}'` for a JSON object), while `node-postgres` pre-parses JSONB values so a JS string means the value *is* a string scalar. Applying `JSON.stringify()` unconditionally broke MySQL by adding spurious outer quotes to every JSON object cell.

The fix follows the same pattern as `bitMutator` — the dialect is now threaded through `resolveDataMutator` via `.bind(this, dialect)`, and `jsonMutator` only applies the `JSON.stringify()` wrapping for PostgreSQL-family dialects (`postgresql`, `greengage`, `redshift`). All other dialects return the string as-is, preserving the previous behaviour for MySQL, SQLite, SQL Server, etc.

Unit tests now cover both paths explicitly.

---

PR #4250, opened 2026-05-13
fix(ux): inline rename for queries, connections, and folders

azmy60:
Screenshot 2026-05-26 at 11 13 34


apps/studio/src/components/sidebar/core/favorite_list/FavoriteListItem.vue
```diff
@@ -92,14 +103,23 @@ export default Vue.extend({
...
+      event.stopPropagation();
+
       const options = [
         {
           name: "Open",
           handler: ({ item }) => this.$emit('open', item)
         },
         {
           name: "Rename",
-          handler: ({ item }) => this.$emit('rename', item)
```
not-night-but, line 102: If we're removing the emitter for this, we should probably remove the handler in FavoriteList.vue

not-night-but (changes requested):
I think there's just some dead code in `FavoriteList.vue` but other than that, I like this!

apps/studio/src/components/sidebar/core/FavoriteList.vue
```diff
@@ -82,10 +82,14 @@
             v-for="({ folder, queries, subfolders }) in foldersWithQueries"
             :key="`${folder.id}-${queries.length}`"
             :title="`${folder.name} (${queries.length})`"
+            :editable-label="folder.name"
```
not-night-but: I don't see a matching prop on the component. Is this used? Is it supposed to be name?

not-night-but (changes requested):

apps/studio/src/components/sidebar/core/FavoriteList.vue
azmy60: Yeah I missed this one

not-night-but:
@azmy60 This seems to have added horizontal scrollbars to the connection lists? I don't know if that was intended, but if it was I can't interact with them.
image

azmy60:
@not-night-but fixed!

not-night-but (approved):

PR #4263, opened 2026-05-15
fix(mongo): add aws4 dep to fix MONGODB-AWS auth

jim-fn:
## Summary

Adds `aws4` as a direct dependency in `apps/studio/package.json`. Without it, MongoDB connections using `authMechanism=MONGODB-AWS` fail at connection time with:

```
Optional module `aws4` not found. Please install it to enable AWS authentication
```

## Repro

1. Connect to a MongoDB cluster (e.g. Atlas) with a connection string that uses AWS IAM auth:
   `mongodb+srv://<host>/?authSource=$external&authMechanism=MONGODB-AWS&...`
2. Click Test / Connect.
3. Error shown above appears; connection fails.

Affects every release since **v5.7.0** (currently shipping in v5.7.2).

## Root cause

Commit ab732cc8 ("remove trino-client axios issues", 2026-04-09) bumped two major versions in passing:

\`\`\`diff
-    \"@mongosh/browser-runtime-electron\": \"^3.29.1\",
-    \"@mongosh/service-provider-node-driver\": \"^3.18.1\",
+    \"@mongosh/browser-runtime-electron\": \"^5.3.4\",
+    \"@mongosh/service-provider-node-driver\": \"^5.0.8\",
\`\`\`

The `@mongosh@5.x` packages pull in a newer `mongodb` driver (nested at `node_modules/@mongosh/service-provider-node-driver/node_modules/mongodb`) whose AWS auth provider (`lib/cmap/auth/mongodb_aws.js`) requires `aws4` as an **optional peer dependency** — the consumer must install it explicitly. The top-level `mongodb@^6.12.0` direct dep has the same expectation.

In earlier mongosh v3.x, this requirement was either bundled differently or not surfaced for the same connection path, so MONGODB-AWS auth worked out of the box before v5.7.0.

`aws4` is pure JS (no native deps), ~25 KB, MIT-licensed, and the canonical implementation used by both AWS SDK v2 and the mongodb driver. Adding it directly is the minimal fix.

## Change

\`\`\`diff
   \"dependencies\": {
     ...
+    \"aws4\": \"^1.13.2\",
     \"axios\": \"^1.15.0\",
     \"axios-retry\": \"^3.2.4\",
     ...
   }
\`\`\`

## Test plan

- [x] \`yarn install\` resolves cleanly
- [x] \`node_modules/aws4/package.json\` exists at the workspace root after install
- [x] Build a local Beekeeper Studio (\`yarn electron:build\`)
- [x] Connect to a MongoDB cluster using \`authMechanism=MONGODB-AWS\` — connection succeeds, no \`aws4 not found\` error
- [x] Smoke test a regular \`mongodb://user:pass@host/db\` connection still works (no regression)
- [x] CI green

## Notes

- Verified `aws4` is **not** currently in the produced asar bundle (`asar list app.asar | grep aws4` is empty), confirming it isn't being pulled in transitively despite `mock-aws-s3` being in `dependencies`.
- No issue filed for this yet — search for `aws4` and `MONGODB-AWS` in repo issues returns 0 hits, so likely few users hit AWS IAM auth.
- `yarn.lock` not regenerated in this PR — happy to update if maintainers prefer the lockfile change bundled.

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -62,6 +62,7 @@
     "@types/semver": "^7.7.0",
     "@uiw/codemirror-theme-monokai": "^4.23.10",
     "ansi-to-html": "^0.7.2",
+    "aws4": "^1.13.2",
     "axios": "^1.15.0",
     "axios-retry": "^3.2.4",
     "base64-url": "^2.3.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -5669,6 +5669,11 @@ aws-ssl-profiles@^1.1.1:
   resolved "https://registry.yarnpkg.com/aws-ssl-profiles/-/aws-ssl-profiles-1.1.2.tgz#157dd77e9f19b1d123678e93f120e6f193022641"
   integrity sha512-NZKeq9AfyQvEeNlN0zSYAaWrmBffJh3IELMZfRpJVWgrpEbtEpnjvzqBPf+mxoI287JohRDoa+/nsfqqiZmF6g==
 
+aws4@^1.13.2:
+  version "1.13.2"
+  resolved "https://registry.yarnpkg.com/aws4/-/aws4-1.13.2.tgz#0aa167216965ac9474ccfa83892cfb6b3e1e52ef"
+  integrity sha512-lHe62zvbTB5eEABUVi/AwVh0ZKY9rMMDhmm+eeyuuUQbQ3+J+fONVQOZyj+DdrvD4BY33uYniyRJ4UJIaSKAfw==
+
 axios-retry@^3.2.4:
   version "3.9.1"
   resolved "https://registry.yarnpkg.com/axios-retry/-/axios-retry-3.9.1.tgz#c8924a8781c8e0a2c5244abf773deb7566b3830d"


jim-fn:
## Verification

Ran an A/B test on this branch with `yarn install` against `node@24` (better-sqlite3 12.x rejects node ≥ 26):

### Before fix — `upstream/master` `apps/studio/package.json`

```
$ ls node_modules/aws4
ls: cannot access 'node_modules/aws4': No such file or directory
```

`require.resolve('aws4')` from each `mongodb_aws.js` location in the dep tree:

```
FAIL node_modules/mongodb/lib/cmap/auth/mongodb_aws.js
     -> Cannot find module 'aws4'
FAIL node_modules/@mongosh/service-provider-node-driver/node_modules/mongodb/lib/cmap/auth/mongodb_aws.js
     -> Cannot find module 'aws4'
```

This reproduces the user-visible error (`Optional module \`aws4\` not found...`).

### After fix — this branch

```
$ ls node_modules/aws4
LICENSE  README.md  aws4.js  lru.js  package.json
$ cat node_modules/aws4/package.json | jq -r .version
1.13.2
```

`require.resolve('aws4')` from every `mongodb_aws.js` in the tree:

```
OK   node_modules/mongodb/lib/cmap/auth/mongodb_aws.js
     -> /.../node_modules/aws4/aws4.js
OK   node_modules/mongodb-schema/node_modules/mongodb/lib/cmap/auth/mongodb_aws.js
     -> /.../node_modules/aws4/aws4.js
OK   node_modules/@mongodb-js/devtools-connect/node_modules/mongodb/lib/cmap/auth/mongodb_aws.js
     -> /.../node_modules/aws4/aws4.js
OK   node_modules/@mongosh/service-provider-core/node_modules/mongodb/lib/cmap/auth/mongodb_aws.js
     -> /.../node_modules/aws4/aws4.js
OK   node_modules/@mongosh/service-provider-node-driver/node_modules/mongodb/lib/cmap/auth/mongodb_aws.js
     -> /.../node_modules/aws4/aws4.js

5/5 mongodb driver locations can resolve aws4
```

The test uses `Module.createRequire(absPath).resolve('aws4')` — the exact same Node module-resolution algorithm the mongodb driver uses internally at `aws_temporary_credentials.js` / `mongodb_aws.js`.

### Ownership check (`yarn why aws4`)

```
info => Found "aws4@1.13.2"
info Reasons this module exists
   - "_project_#beekeeper-studio" depends on it
   - Hoisted from "_project_#beekeeper-studio#aws4"
info Number of shared dependencies: 0
```

No other dep in the tree provides `aws4` transitively — it's hoisted solely because of the direct declaration in this PR. Removing the line removes the package entirely (confirmed above).

### Lockfile

`yarn.lock` regenerated (separate commit, +5 lines for the `aws4@^1.13.2` entry).

### Caveats

- Node-resolution test, not a live MongoDB connection test, since AWS IAM connection requires creds. If maintainers want, I can also build the Electron app locally and capture a successful connection screenshot — happy to do that on request.

jim-fn:
### Live verification

Also confirmed end-to-end against a real MongoDB Atlas cluster using `authMechanism=MONGODB-AWS` with STS temporary credentials:

- Built this branch from source with `yarn lib:build && yarn bks:dev` (Electron dev mode, node@24 since `better-sqlite3@12.x` rejects node ≥ 26)
- New MongoDB connection → pasted the `mongodb+srv://...?authMechanism=MONGODB-AWS&...` URI → **Test**
- Result: **"Connection looks good!"** — no `Optional module \`aws4\` not found` error, connection succeeds

Same URI reproduces the failure on stock 5.7.2 install.

rathboma:
How do you auth with AWS? Does it use a pre-authenticated aws CLI or something like that?

I guess the question is what else do we need to do to fix this?

jim-fn:
Good question — in my setup I don't rely on a pre-authenticated AWS CLI at all. The AWS credentials are passed **inline in the connection URI**:

```
mongodb+srv://<AWS_ACCESS_KEY_ID>:<URL_ENCODED_SECRET>@cluster.example.mongodb.net/
  ?authMechanism=MONGODB-AWS
  &authSource=$external
  &authMechanismProperties=AWS_SESSION_TOKEN:<session-token>
```

(In my case these are short-lived STS credentials from `aws sts assume-role-with-web-identity`, which is why the access key starts with `ASIA` and an `AWS_SESSION_TOKEN` is required.)

The thing is — **regardless of where the credentials come from** (URI like above, `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`/`AWS_SESSION_TOKEN` env vars, EC2/ECS instance role, or Web Identity), the MongoDB Node driver always needs `aws4` to sign the auth request to the server with SigV4. That's the only thing missing from the stock install — the driver lists `aws4` as an *optional* peer dep and silently fails the auth handshake when it's not resolvable.

So nothing else needs fixing on Beekeeper's side. Adding `aws4` to `apps/studio/package.json` makes every `MONGODB-AWS` flow work (URI-inline, env, instance-role) for both the embedded driver and `mongosh`.

not-night-but:
@jim-fn Yeah so this is definitely needed. I think eventually I would like to rebuild the mongodb connect for because putting all that information into the URI makes me feel icky, but I wonder if for now you'd be willing to update the mongodb docs we have with how to connect with AWS?

rathboma (approved):

PR #4264, opened 2026-05-15
Fix: named query params clobbered when the same parameter appears multiple times

### LINKED ISSUE #4259: BUG: Saved queries with input variables unreliable on SQLite (name collision + inconsistent results)
**Describe the bug**
Saved queries with input variables behave unreliably on SQLite in two ways:

1. Using a variable name that matches a column name breaks the query, suggesting the variable substitution logic conflicts with SQL parsing when names collide.
2. The same query with the same input values sometimes produces different results on re-runs, pointing to a possible state issue or race condition in how variables are resolved at execution time.

**To Reproduce**
1. Create a saved query with input variables in SQLite
2. Use a variable name that matches a column name in the query
3. Run the query — it breaks
4. Even without name collision, re-running the same query with identical inputs can produce inconsistent results

**Expected behavior**
Variable substitution should be isolated from column names, and results should be deterministic for the same query and inputs.

**Screenshots / Gifs**
N/A

 - OS: [not provided]
 - App Version: [not provided]
 - Database type and version: SQLite (Cloudflare D1 local via wrangler)

**Additional context**
Reported by a user using Cloudflare D1 locally via wrangler, which uses a local .sqlite file.

rathboma:
## The Bug

Fixes #4259 — saved queries with named input variables produce wrong or broken results when the same parameter name appears more than once in the query.

### Root cause

`sql-query-identifier` returns one entry per **occurrence** of a named parameter in `placeholders`. So for:

```sql
SELECT * FROM users WHERE first_name = :name OR last_name = :name AND age > :age
```

`placeholders = [':name', ':name', ':age']`

The UI, however, deduplicates via `_.uniq` before prompting the user, so the user only enters one value for `:name`. `queryParameterValues = { ':name': "'Alice'", ':age': '25' }` → `values = ["'Alice'", '25']`.

`convertParamsForReplacement` then paired each placeholder by **index**:

```
index 0  :name  → obj.name = "'Alice'"   ✓
index 1  :name  → obj.name = '25'        ✗ overwrites!
index 2  :age   → obj.age  = undefined   ✗ nothing at values[2]
```

Result: `{ name: '25', age: undefined }` — `:name` gets the wrong value, `:age` gets nothing. The query silently breaks.

This explains the "inconsistent results on re-runs" symptom too: anything that affected insertion order of `queryParameterValues` (reactivity re-renders, typing order) could change which value landed at which index.

### The Fix

Deduplicate `placeholders` inside `convertParamsForReplacement` before building the named-params object. The deduplicated array aligns index-for-index with the unique-ordered `values` the UI collected:

```typescript
// Before
return placeholders.reduce((obj, val, index) => {
  obj[val.slice(1)] = values[index];
  return obj;
}, {});

// After
const uniquePlaceholders = _.uniq(placeholders);
return uniquePlaceholders.reduce((obj, val, index) => {
  obj[val.slice(1)] = values[index];
  return obj;
}, {});
```

Positional (`?`) params are unaffected — they go through a separate branch.

## Tests

Added 4 unit tests in `sql_tool.spec.js`:

- Positional params still work
- Normal unique named params still work
- Duplicate named params no longer overwrite the correct value
- End-to-end: a query with `:name` repeated twice substitutes `'Alice'` into both occurrences correctly

https://claude.ai/code/session_01M4ApjwLzSm4TYeYP4exCbx

---
_Generated by [Claude Code](https://claude.ai/code/session_01M4ApjwLzSm4TYeYP4exCbx)_

### DIFF
--- a/apps/studio/src/lib/db/sql_tools.ts
+++ b/apps/studio/src/lib/db/sql_tools.ts
@@ -19,21 +19,19 @@ export function canDeparameterize(params: string[]) {
   return !(params.includes('?') && params.some((val) => val != '?'));
 }
 
-export function convertParamsForReplacement(placeholders: string[], values: string[]): ParamItems | string[] {
+export function convertParamsForReplacement(placeholders: string[], values: string[] | Record<string, string>): ParamItems | string[] {
   if (placeholders.includes('?')) {
-    return values;
+    // Positional params: values is an ordered array, return as-is for sql-formatter.
+    return values as string[];
   } else {
-    // The same named parameter (e.g. :name) can appear multiple times in a query, so
-    // sql-query-identifier returns it once per occurrence in `placeholders`. The UI, however,
-    // deduplicates via _.uniq and only asks the user for one value per unique name.
-    // We must deduplicate here so that index i in `values` maps to the i-th *unique* placeholder,
-    // not the i-th occurrence — otherwise duplicate occurrences shift all subsequent indices and
-    // clobber or discard user-entered values.
-    const uniquePlaceholders = _.uniq(placeholders);
-    return uniquePlaceholders.reduce((obj, val, index) => {
-      obj[val.slice(1)] = values[index];
-      return obj;
-    }, {});
+    // Named params: values is a record keyed by the full placeholder (e.g. { ':name': "'Alice'" }).
+    // Strip the prefix character so sql-formatter gets { name: "'Alice'" }.
+    // Using a record rather than a positional array means duplicates in `placeholders` never
+    // shift indices or clobber values — every lookup goes through the key, not position.
+    const record = values as Record<string, string>;
+    return Object.fromEntries(
+      Object.entries(record).map(([k, v]) => [k.slice(1), v])
+    );
   }
 }
 
--- a/apps/studio/src/lib/db/sql_tools.ts
+++ b/apps/studio/src/lib/db/sql_tools.ts
@@ -23,9 +23,14 @@ export function convertParamsForReplacement(placeholders: string[], values: stri
   if (placeholders.includes('?')) {
     return values;
   } else {
-    // TODO (@day): this might not work with quoted params
-    // this will need to be more complex if we allow truly custom params
-    return placeholders.reduce((obj, val, index) => {
+    // The same named parameter (e.g. :name) can appear multiple times in a query, so
+    // sql-query-identifier returns it once per occurrence in `placeholders`. The UI, however,
+    // deduplicates via _.uniq and only asks the user for one value per unique name.
+    // We must deduplicate here so that index i in `values` maps to the i-th *unique* placeholder,
+    // not the i-th occurrence — otherwise duplicate occurrences shift all subsequent indices and
+    // clobber or discard user-entered values.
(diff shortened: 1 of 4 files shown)


apps/studio/tests/unit/lib/db/sql_tool.spec.js
```diff
@@ -128,6 +128,38 @@ describe("Text Selection", () => {
...
+  it("should build a named params object from unique placeholders", () => {
+    const result = convertParamsForReplacement([':name', ':age'], ["'Alice'", '25']);
+    expect(result).toEqual({ name: "'Alice'", age: '25' });
+  });
+
+  it("should not overwrite a named param value when the same placeholder appears multiple times", () => {
+    // When :name appears twice in the SQL, sql-query-identifier returns it twice in placeholders,
+    // but the UI only asks for one value (deduplicating via _.uniq). The second occurrence must not
+    // clobber the value entered for the first.
+    const result = convertParamsForReplacement([':name', ':name', ':age'], ["'Alice'", '25']);
```
rathboma: I don't think this fixes it. Test with name,age,name,age,name.

Really if params are named it should accept key/value for the values, or have some sort of stable sorting that guarantees correct results.

rathboma: Good catch — the index-based approach was still fragile. Updated in the latest commit.

`convertParamsForReplacement` now accepts `string[] | Record<string, string>` for `values`. For named params the caller passes `queryParameterValues` as-is (e.g. `{ ':name': "'Alice'", ':age': '25' }`), and the function strips the prefix from each key to produce `{ name: "'Alice'", age: '25' }`. Because lookup is by key rather than by index, duplicates in `placeholders` are irrelevant — the `name,age,name,age,name` pattern works correctly. Positional (`?`) path is unchanged.

---

not-night-but:
@rathboma No this is intentional. If you're naming parameters, every occurrence of that name will be replaced with that value. If you want each param to be separate then you need to use positional params, or just give it another name. Unless I'm misunderstanding what this is meant to fix?

apps/studio/src/lib/db/sql_tools.ts
```diff
@@ -19,14 +19,24 @@ export function canDeparameterize(params: string[]) {
...
   } else {
-    // TODO (@day): this might not work with quoted params
-    // this will need to be more complex if we allow truly custom params
-    return placeholders.reduce((obj, val, index) => {
-      obj[val.slice(1)] = values[index];
+    // Named params with legacy array input: deduplicate placeholders first so that
+    // index i maps to the i-th *unique* placeholder. Without dedup, a second occurrence
+    // of :name would overwrite the correct value and shift every subsequent index.
+    const uniquePlaceholders = _.uniq(placeholders);
+    return uniquePlaceholders.reduce((obj, val, index) => {
```
not-night-but: We only actually input named params as an array in the tests now, don't we? This "fix" means that the existing tests are now testing a scenario that doesn't happen

not-night-but (changes requested):

not-night-but (approved):

PR #4270, opened 2026-05-15
upgrade modal style redo

rathboma:


apps/studio/src/components/export/ExportModal.vue
```diff
@@ -326,7 +326,7 @@ export default {
     },
     upgradeModal() {
       this.closeModal()
-      this.$root.$emit(AppEvent.upgradeModal)
+      this.$root.$emit(AppEvent.upgradeModal, 'Multi-table Export')
     },
```
not-night-but, line 330: This doesn't seem to be called anywhere, we just call `importExportTables` above and that opens the tab with the old upsell

not-night-but (commented):
just a note about the multi table export not actually showing the modal, but other than that looks good! Not sure if that blocks this though, up to you

PR #4272, opened 2026-05-15
Add redaction utility to prevent sensitive data in logs

rathboma:
## Summary
Adds a new `redact()` utility function to safely sanitize sensitive data before logging. This prevents credentials, tokens, passwords, and other sensitive information from being exposed in application logs.

## Key Changes
- **New module**: `src/lib/log/redact.ts` with `redact()` and `isSensitiveKey()` functions
  - Recursively walks plain objects and arrays, replacing values under sensitive keys with `[REDACTED]`
  - Handles edge cases: circular references, Buffers, class instances, special objects (Error, Date, RegExp, etc.)
  - Comprehensive sensitive key detection via regex pattern (passwords, tokens, API keys, credentials, etc.)
  - Includes depth limiting to prevent infinite recursion

- **Comprehensive test coverage**: `tests/unit/lib/log/redact.spec.ts` with 204 lines of tests covering:
  - Sensitive vs. non-sensitive key classification
  - Nested object and array redaction
  - Buffer handling
  - Circular reference detection
  - Preservation of special object types
  - Edge cases (null/undefined values, class instances)

- **Applied redaction to existing logs**:
  - `mainBksConfig.ts`: Redact config objects before logging
  - `CloudClient.ts`: Removed full request/response logging (contained auth headers); now logs only method/URL/status
  - `OfflineLicense.ts`: Log only license metadata instead of full payload
  - `UtilityUsedConnectionModule.ts`: Redact connection objects before logging
  - `appDbHandlers.ts`: Log only object ID instead of full object
  - `SettingsPlugin.ts`: Changed from logging full result to logging only key and found status
  - `firebird.ts`, `bigquery.ts`, `postgresql.ts`: Redact connection configs before logging
  - `node-ssh-forward/index.ts`: Redact SSH connection options before logging
  - `LicenseKeyController.ts`: Redact sensitive request parameters
  - Removed unnecessary `console.log()` statements from `PinModule.ts` and `LicenseKey.ts`

## Implementation Details
- Uses `WeakSet` to track visited objects and detect circular references
- Distinguishes between plain objects (which are cloned and traversed) and class instances (returned as-is)
- Buffers are always redacted as they typically contain key material
- Sensitive key matching is case-insensitive and handles both camelCase and snake_case variants
- Preserves `null` and `undefined` values for sensitive keys rather than replacing them

https://claude.ai/code/session_01Huooy49LfZ4JDDVy5VtivQ

### DIFF
--- a/apps/studio/src-commercial/backend/lib/db/clients/firebird.ts
+++ b/apps/studio/src-commercial/backend/lib/db/clients/firebird.ts
@@ -1,4 +1,5 @@
 import electronLog from "@bksLogger";
+import { redact } from "@/lib/log/redact";
 import knexlib, { Knex } from "knex";
 import Client_Firebird from "@shared/lib/knex-firebird";
 import Firebird from "node-firebird";
@@ -278,7 +279,7 @@ export class FirebirdClient extends BasicDatabaseClient<FirebirdResult, Firebird
 
     this.firebirdOptions = config;
 
-    log.debug("create driver client for firebird with config %j", config);
+    log.debug("create driver client for firebird with config %j", redact(config));
 
     this.pool =  new Pool(BksConfig.db.firebird.maxConnections, config);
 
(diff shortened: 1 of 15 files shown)


not-night-but:
I feel like there has to be a way for us to do this in the utilityLogger and mainLogger. Like maybe through the transforms array on the transport?

PR #4282, opened 2026-05-20
Use icon buttons in query editor toolbar to save space

azmy60:
Screenshot 2026-05-20 at 20 05 59

Frees up horizontal space in the query editor toolbar by leaning on icons instead of text labels.

Why? Reduce toolbar width so more of it stays usable in narrower windows.


azmy60:
Not planned. Might be revisited later.

PR #4283, opened 2026-05-20
Fix query stream bug

not-night-but:
Currently these are just the failing tests to confirm the issue :)

The issue is: We run the entire query before we open the cursor so we can get the columns and the number of rows. This is BAD. Currently I'm thinking each cursor should implement a way to retrieve this information, and it will be retrieved along with the first chunk to write the header if necessary, along with storing the information for mutating rows. The total number of rows isn't going to be possible on most clients, so we'll need to make sure that we still show some sort of progress in the frontend (a marquee progress bar or a spinner)


apps/studio/src-commercial/backend/lib/db/clients/mongodb.ts
rathboma: should this be == or ===?

apps/studio/src-commercial/backend/lib/db/clients/surrealdb.ts
rathboma: same q

apps/studio/src/lib/db/clients/postgresql/PsqlCursor.ts
```diff
@@ -45,10 +54,14 @@ export class PsqlCursor extends BeeCursor {
       if (!this.client || !this.cursor) {
         reject("You need to call start first")
       } else {
-        this.cursor.read(this.chunkSize, (err, rows) => {
+        this.cursor.read(this.chunkSize, (err, rows, result) => {
+          log.info("RESULT: ", result)
```
rathboma: let's not log.info every result! Plz kill

apps/studio/src-commercial/backend/lib/db/clients/surrealdb/SurrealDBPool.ts
```diff
@@ -67,7 +67,6 @@ export class SurrealPool {
       log.info('CONFIG: ', this.config)
```
rathboma: we probably shouldn't log the config here

rathboma (approved):
couple of minor things to remove, otherwise, ship it!

PR #4294, opened 2026-05-22
Update search module to support fuzzy matching with character position

json-nan:
### Old 
before

### New

May-22-2026 10-19-13

## Summary
- improve quick search matching in `SearchModule.searchItems`
- normalize item titles and search input by removing separators such as underscores and spaces
- preserve fuzzy matching so partial/consecutive input like `onlineitems` and typo-like input such as `comters` can still find the expected item

## Changes
- add search-term normalization for indexed titles and user input
- map fuzzy-match highlight ranges back to the original title for correct UI highlighting
- keep the existing result shape and search limit behavior unchanged

## Problem
Quick search required separator-aware exact input in some cases. For example, an item named `online_items` was harder to find unless the user typed `online_items` exactly.

## Solution
Search now runs against a normalized version of each title that strips non-alphanumeric separators and lowercases the text before applying fuzzy matching. This makes matches more tolerant to formatting differences and minor missing-character input.

## Examples
- `onlineitems` → matches `online_items`
- `comters` → matches `computers`
- `userprofile` → matches `user_profile`

## Testing
- not run


apps/studio/src/store/modules/SearchModule.ts
```diff
@@ -22,15 +22,68 @@ const uf = new uFuzzy({
   intraIns: Infinity,
 });
 
+function buildSearchTarget(value: string) {
+  const positions: number[] = [];
+  const normalized = Array.from(value).reduce((result, char, index) => {
+    if (/^[\p{L}\p{N}]$/u.test(char)) {
```
azmy60, line 28: Could we make a variable for this or add a comment explaining what this does?

azmy60 (changes requested):

azmy60 (changes requested):
Can we also add a unit test for `searchItems`?

PR #4297, opened 2026-05-22
Move component initialization logic from data to created lifecycle hook

rathboma:
## Summary
Refactored Vue component initialization patterns to move instance property initialization from the `data()` function to the `created()` lifecycle hook. This improves code organization by separating reactive data properties from non-reactive instance initialization.

## Key Changes

- **TableTable.vue**: Moved `timeAgo` and `interval` initialization to `created()` hook; removed `tabulator: null` from data as it's set during initialization
- **TextEditor.vue**: Replaced `editor: null` with `editorReady: false` boolean flag; set `editorReady = true` when editor is initialized and `false` when destroyed
- **ImportFile.vue**: Replaced `tabulator: null` with `previewReady: false` boolean flag; set flag when preview is ready
- **TabQueryEditor.vue**: Moved `queryMagic` initialization to `created()` hook
- **SchemaBuilder.vue**: Removed unused `tabulator: Tabulator` type and `tabulator: null` from data
- **ImportMapper.vue**: Removed `tabulator: null` from data initialization
- **TableIndexes.vue**: Removed `tabulator: Tabulator` type and `tabulator: null` from data
- **ConnectionInterface.vue, CoreInterface.vue, GroupedTables.vue, TabPluginShell.vue, TabShell.vue, QueryEditHistory.vue, ResultTable.vue, ConnectionSidebar.vue, TableList.vue, TablePartitions.vue, TableRelations.vue, TableSchema.vue, DataEditor.vue, text-editor mixin**: Removed unused `split: null` and `tabulator: null` properties from data

## Implementation Details

- Boolean flags (`editorReady`, `previewReady`) are more semantically correct than storing object references for state tracking
- Non-reactive initialization (like `TimeAgo` instances and intervals) now properly belongs in `created()` rather than `data()`
- Removed unused property declarations that were never assigned or used
- Added null checks in computed properties where needed (e.g., `columnsWithFilterAndOrder` checks `initialized` flag)

https://claude.ai/code/session_01QVc8n5Jrc8jzieannwDzp4


apps/studio/src/components/common/texteditor/TextEditor.vue
```diff
@@ -84,7 +84,7 @@ export default {
   ],
```
azmy60, line 84: I think we should've got rid of this component a while ago in favor of UI Kit, but this isn't the PR for that ofc. Just want to mention it.

azmy60 (approved):

PR #4306, opened 2026-05-26
Integration test for ident update

not-night-but:
Also actually provide dialect to identifier :upside_down_face: 
Fix some issues with sql server and duckdb


rathboma (approved):
after making defaultSchema just return _defaultSchema, it's gud

PR #4314, opened 2026-05-27
Rename openTab table/structure `table` prop to `name`

### LINKED ISSUE #4108: FEAT: Add plugin API for opening table data view & structure

This will support ER Diagram plugin so users can open the table view / structure of the selected table.

The API looks like:

```ts
import { openTab } from "@beekeeperstudio/plugin";

openTab("tableData", { schema: "public", table: "customer" });
openTab("tableStructure", { schema: "public", table: "customer" });
```

azmy60:
## What

Rename the `table` option to `name` for the plugin `openTab` API:

```ts
openTab("tableTable", { name: "customer", schema: "public" });
openTab("tableStructure", { name: "customer", schema: "public" });
```

## Why

So the options match the `Table` type signature, which uses `{ name: string; schema: string }`. Previously these used `table` while everything else referring to a table uses `name`, which was inconsistent.

## Breaking change

This is a breaking change to the `openTab("tableTable" | "tableStructure", ...)` options, but it's fine — no plugins currently use these APIs.

## Scope / follow-up

This PR covers the in-repo side only:
- `PluginStoreService.openTab` reads `options.name` (and the param type now points at `RequestMap["openTab"]["args"]`).
- Plugin API reference docs updated.

The `@beekeeperstudio/plugin` package (separate repo) still needs the matching `table` → `name` rename in `OpenTableTableTabOptions` / `OpenTableStructureTabOptions`, then a version bump in `apps/studio/package.json`, for this to work end-to-end and type-check cleanly.

### DIFF
--- a/apps/studio/src/services/plugin/web/PluginStoreService.ts
+++ b/apps/studio/src/services/plugin/web/PluginStoreService.ts
@@ -324,13 +324,13 @@ export default class PluginStoreService {
     }
 
     if (options.type === "tableStructure") {
-      const table = this.findTableOrThrow(options.table, options.schema);
+      const table = this.findTableOrThrow(options.name, options.schema);
       this.appEventBus.emit(AppEvent.openTableProperties, { table });
       return;
     }
 
     if (options.type === "tableTable") {
-      const table = this.findTableOrThrow(options.table, options.schema);
+      const table = this.findTableOrThrow(options.name, options.schema);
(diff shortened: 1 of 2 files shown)


rathboma (approved):
Just make sure there haven't been any other changes since you wrote this that added new `OpenTab` calls

azmy60:
@rathboma We only use `openTab("query")` in AI Shell which is totally unrelated to this changes.

PR #4323, opened 2026-05-28
Feat/snowflake

not-night-but:


apps/studio/src-commercial/backend/lib/db/clients/snowflake/SnowflakeCursor.ts
```diff
@@ -0,0 +1,78 @@
...
+export class SnowflakeCursor extends BeeCursor {
+  private readonly options: CursorOptions;
+  private stmt: RowStatement;
+  private error?: SnowflakeError;
+  private cursorPos: number = 0;
+  private rowBuffer: any[][] = [];
+  private conn: Connection;
+
+  constructor(options: CursorOptions) {
+    // TODO (@day): chunksize :)
```
not-night-but: Remove

apps/studio/src-commercial/backend/lib/db/clients/snowflake/SnowflakeCursor.ts
```diff
@@ -0,0 +1,78 @@
...
+    super(options.chunkSize);
+    this.options = options;
+  }
+
+  async start() {
+    this.conn = await this.options.pool.acquire();
+    return await new Promise<void>((resolve, reject) => {
+      this.stmt = this.conn.execute({
```
not-night-but, line 36: I don't think we need to be setting this, looks like from a previous experiment ?

apps/studio/src-commercial/backend/lib/db/clients/snowflake/SnowflakeCursor.ts
```diff
@@ -0,0 +1,78 @@
...
+  private error?: SnowflakeError;
+  private cursorPos: number = 0;
+  private rowBuffer: any[][] = [];
+  private conn: Connection;
+
+  constructor(options: CursorOptions) {
+    // TODO (@day): chunksize :)
+    super(options.chunkSize);
+    this.options = options;
+  }
```
not-night-but, line 22: OH! Need to do the update from the queryStream fix

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1059 @@
...
+    const data = await this.driverExecuteSingle(sql);
+
+    return data.rows.map((row) => ({
+      name: row.NAME,
+      schema: row.SCHEMA
+    } as TableOrView));
+  }
+
+  async listViews(filter?: FilterOptions): Promise<TableOrView[]> {
+    // TODO (@day): do we need a custom wrapIdent
```
not-night-but: We have one now, use it 🙃

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1059 @@
...
+    this.version = await this.getVersion();
+    this._defaultSchema = await this.getSchema();
+  }
+
+  async disconnect(): Promise<void> {
+    await super.disconnect();
+    await this.pool.clear();
+  }
+
+  // TODO (@day): need to be async? prob not
```
not-night-but: remove

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1059 @@
...
+
+  // afaik this doesn't exist in snowflake, unless we want to shove tasks and streams into here
+  async listTableTriggers(_table: string, _schema?: string): Promise<TableTrigger[]> {
+    return [];
+  }
+
+  // Only available on hybrid tables, which I can't create on a trial account :(
+  async listTableIndexes(_table: string, _schema?: string): Promise<TableIndex[]> {
+    return [];
+  }
```
not-night-but: Oh maybe we can look into this now

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1059 @@
...
+  async query(queryText: string, tabId: number, _options?: any): Promise<CancelableQuery> {
+    let stmt: snowflake.RowStatement | snowflake.FileAndStageBindStatement = null;
+    const hasReserved = this.reservedConnections.has(tabId);
+    const conn = hasReserved ? this.peekConnection(tabId) : await this.pool.acquire();
+    const cancelable = createCancelablePromise(errors.CANCELED_BY_USER);
+
+    return {
+      execute: async(): Promise<NgQueryResult[]> => {
+        // TODO (@day): use the snowflake dialect
+        log.info('RUNNING: ', queryText);
```
not-night-but: Still waiting for this to be merged/released

azmy60 (changes requested):
## MFA Passcode modal

1. I think there's a good chance people will perceive "this is not saved" even without the explicit "(this is not saved)" text, because the passcode prompt doesn't appear inside the connection form, but rather in a popup modal. So I'm fine with not being explicit here, and I'd just remove the text.

But if we want to keep it, which is fine too, it should be a helper text below the input (see CONTRIBUTING.md for how to make one), and probably add asterisk (*) to the input label so it's like `MFA Passcode*` and then the helper is like`*Passcode isn't saved` or something like that. But honestly, I prefer the first solution cause it's simpler.

Image

2. The close button looks broken when hovered. You can use the [BaseModal](https://github.com/beekeeper-studio/beekeeper-studio/pull/4131), and forget about the styles, class names, etc.

Image

### Optional: further steps to improve the UI

3. Is the MFA passcode limited to 6 digits? If so, I don't think the modal needs to be that wide, but that can be a separate issue cause I know that's the default style(?).

4. There are many examples of One-time password UI designs that you can "steal". I would not be too fancy about it, just try this fundamental stuff is good enough; increase the input's font size 2-10x, increase the letter-spacing, font-weight, and padding. No other changes, just style that input alone.

azmy60 (changes requested):
## Bugs after connected 

1. Missing schema in the schema builder. No way for me to specify which schema I want the new table in. This works in postgres btw.

Image

2. Missing feedback when applying changes even for a small change. It feels like it's "lagging". At the very least, the apply button should be disabled, which is already a form of feedback. Maybe setting `tab.isRunning` to true can be good too.

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1058 @@
...
+        fields: r.columns.map((c) => ({
+          name: c.name,
+          id: c.name,
+          dataType: c.type
+        }))
+      }
+    });
+  }
+
+  async executeApplyChanges(changes: TableChanges, tabId?: number): Promise<TableUpdateResult[]> {
```
azmy60, line 544: Has this become repetitive now? If so, in the future we'd want to put this in the BasicDatabaseClient and remove it from each database class.

docs/user_guide/connecting/snowflake.md
```diff
@@ -0,0 +1,44 @@
...
+If you select "Multi-Factor Authentication with Duo", you will receive a notification from Duo on your device to authenticate when you attempt to connect.
+
+## Authentication Token Caching
+
+The Snowflake driver can cache Authentication Tokens locally for a limited amount of time to reduce the number of prompts required for authentication. This requires your Account to allow client token caching:
+
+### Caching For MFA
+Run this query:
+```snowflake
+ALTER ACCOUNT SET ALLOW_CLIENT_MFA_CACHING = TRUE;
```
azmy60, line 36: @not-night-but Is it possible if the users can check a "Enable token cache" box in the MFA modal instead of doing this manually? I guess it's not that hard, right? thoughts? @rathboma 

We can make a custom handler in the utility and then call a custom method from the snowflake class to run this query.

azmy60 (commented):
In terms of code, I'd replace the modal with the new <BaseModal>, the example is in the link, that way we don't need to think too much on the styles and html structure.

I tested the branch for running ERD plugin, worked perfectly.

Overall, I don't see any major things. Good stuff.

docs/user_guide/connecting/snowflake.md
not-night-but: Just for posterity: As mentioned in slack, this query is something that affects the ENTIRE account and thus should not be something that we just run silently. I think most people using BKS wouldn't be connecting with an account that would have these perms anyways, and would have to contact their system admin

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
not-night-but: Oh yes for sure, this is kind of what I've been saying in that we need to do an audit of our db clients to 1. make sure that we have a standard for how everything is done (we have an unspoken one at the moment that is not always followed) and then 2. I think we can further abstract a lot of what we're doing up into BasicDatabaseClient. I don't think it's something for us to do in this pr though :)

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1088 @@
...
+
+  async listDatabases(_filter?: DatabaseFilterOptions): Promise<string[]> {
+    // doing it this way should ensure we don't need a database selected
+    const sql = `
+      SHOW DATABASES;
+    `;
+
+    const data = await this.driverExecuteSingle(sql);
+
+    return data.rows.map((row) => row.name);
```
azmy60, line 202: Are we sure this is not `row.NAME`?

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1088 @@
...
+
+    try {
+      const data = await this.driverExecuteSingle(sql);
+
+      return data.rows
+        .map((row) => ({
+          name: row.name,
+          schema: row.schema_name
+        } as TableOrView));
+    } catch {
```
azmy60: Should we log the error regardless? Maybe with `log.debug`?

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1088 @@
...
+    let stmt: snowflake.RowStatement | snowflake.FileAndStageBindStatement = null;
+    const hasReserved = this.reservedConnections.has(tabId);
+    const conn = hasReserved ? this.peekConnection(tabId) : await this.pool.acquire();
+    const cancelable = createCancelablePromise(errors.CANCELED_BY_USER);
+
+    return {
+      execute: async(): Promise<NgQueryResult[]> => {
+        // TODO (@day): use the snowflake dialect
+        log.info('RUNNING: ', queryText);
+        const commands = identify(queryText, { strict: false, dialect: undefined });
```
azmy60: Can we use `this.identifyCommands`?

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1088 @@
...
+        cancelable.cancel();
+      }
+    }
+  }
+
+
+  async executeQuery(queryText: string, options?: any): Promise<NgQueryResult[]> {
+    const data = await this.driverExecuteMultiple(queryText, options);
+
+    const commands = identify(queryText, { strict: false });
```
azmy60: Same

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
azmy60: or log.warn

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1088 @@
...
+    const hasReserved = this.reservedConnections.has(options?.tabId);
+    let conn: Connection = null
+
+    if (hasReserved) {
+      conn = this.peekConnection(options?.tabId);
+    } else if (options?.connection) {
+      conn = options?.connection;
+    } else {
+      conn = await this.pool.acquire();
+    }
```
azmy60, line 956: Not for this PR, but I think we probably want to abstract this in the BasicDatabaseClient in the future.

azmy60: Something like a new class would be nice too. Similar to the transcoder pattern:

```ts
class SnowflakeConnectionManager extends BasicDatabaseConnection {
  // return connection
  async get() {}
}

class SnowflakeClient extends BasicDatabaseClient {
  connectionManager = SnowflakeConnectionManager;

  async rawExecuteQuery(query: string, options: ExecuteQueryOptions) {
    options.connection // always defined
  }
}

class BasicDatabaseClient {
  connectionManager: BasicDatabaseConnection;
  async driverExecuteSingle(query, options) {
    if (!options.connection) {
      return this.rawExecuteQuery(query, {...options, connection: await this.connectionManager.get()})
    }
  }
}
```

azmy60: Maybe just move rawExecuteQuery to the BasicDatabaseConnection??

azmy60: We don't want to call rawExecuteQuery directly right? always call driverExecuteSingle or driverExecuteMultiple?

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
```diff
@@ -0,0 +1,1088 @@
...
+          arrayMode: options.arrayMode,
+          rowCount: stmt.getNumRows(),
+          affectedCount: stmt.getNumUpdatedRows()
+        };
+        return result;
+      })
+    }))
+
+    if (!hasReserved && !options?.connection) {
+      this.pool.release(conn);
```
azmy60, line 990: Another case why we probably want to make a new class for connection manager. It's easy to miss this.

apps/studio/src/components/common/modals/BaseModal.vue
```diff
@@ -1,6 +1,12 @@
 <template>
   <portal to="modals">
-    <modal :name="name" class="base-modal-root">
+    <modal
+      :name="name"
+      @before-open="$emit('before-open', $event)"
+      @opened="$emit('opened')"
+      @closed="$emit('closed')"
+      class="base-modal-root beekeeper-modal"
```
azmy60: Hmm, I'm afraid adding ".beekeeper-modal" can add unwanted styles, and I'd avoid it for that.

apps/studio/src/components/common/modals/InputEphemeralModal.vue
```diff
@@ -0,0 +1,117 @@
+<template>
+  <portal to="modals">
+    <base-modal
+      :name="name"
+      @before-open="beforeOpen"
+      @opened="opened"
+      @closed="closed"
+    >
+      <template #title>
+        <div class="dialog-c-title has-icon">
```
azmy60: Okay, I see why you added beekeeper-modal there so it's fine. I'll clean it up later in a different PR.

azmy60: Actually I can do it now

azmy60 (changes requested):

azmy60 (approved):
I think this is good overall. Pretty straightforward. Good for merge once the feedback is addressed!

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
azmy60: this doesn't need to be addressed here

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
not-night-but: Yeah. Annoyingly all `SHOW` commands from snowflake return lowercase columns, but if you try to do a `SELECT col` it will automatically return that as `COL`

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
not-night-but: Had to merge my other PR first, now we can!

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
not-night-but: Yes we don't want to call rawExecuteQuery. The clients have to implement it, we call through the inherited interface. I do think we need to further abstract the connection management, but to do that I think we also need to further standardize how we query in clients, as some have multiple ways of doing it. Definitely not for this PR though

apps/studio/src-commercial/backend/lib/db/clients/snowflake.ts
not-night-but: Oh 100%. The only reason I know to do this is because I built it lol

PR #4330, opened 2026-05-30
fixes: saved query no longer gets removed from its folder when saved from the query editor tab.

### LINKED ISSUE #4329: BUG: Saved query is removed from its folder after saving edits in the query tab
**Describe the bug**
Saved queries are silently moved out of their folder whenever the query is saved from the query editor tab. If a saved query lives inside a folder and the
  user edits its text and hits Save (Ctrl/Cmd+S or the Save button), the query's folder assignment is cleared — the query reappears at the root of the Saved
  Queries sidebar. The user never asked to move the query; the save action itself destroys the folder link.

**To Reproduce**
 1. Open Beekeeper Studio and connect to any database.
  2. Open the Saved Queries sidebar (left panel).
  3. Create a new folder (e.g. "Reports").
  4. Create or open an existing saved query and drag it into the "Reports" folder. Confirm it now appears under "Reports" in the sidebar.
  5. Double-click the query to open it in a new query editor tab.
  6. Make any trivial edit to the SQL text (add a space, a comment — anything that flips the unsaved-changes state).
  7. Hit Save (Ctrl/Cmd+S, or the Save button in the toolbar).
  8. Observe the Saved Queries sidebar: the query has jumped out of "Reports" and is now at the root.
  9. Reload / restart the app to confirm the change is persisted (it is — not just a UI glitch).

**Expected behavior**
Saving the contents of a saved query should only update the query's text/title/excerpt. The query's folder assignment and its sidebar ordering should be
  untouched by a normal save — they should only change when the user explicitly drags the query to a different folder, picks a folder in the save dialog, or
  reorders items.

**Screenshots / Gifs**

https://github.com/user-attachments/assets/d0ff22b0-804b-4c36-b351-6eef8d7bec57

  - OS: macOS (reproduced on macOS); likely platform-independent.
  - App Version: current master branch at time of report. Present on any build that includes the saved-query folders feature.
  - Database type and version: not database-specific — the affected data is stored in Beekeeper Studio's internal app database, not in the user's connected
  database. Reproduces regardless of the connection type.

**Additional context**
-

Dojeto:
The tab keeps a local snapshot of the saved query that never resyncs with the `data/queries` store. After moving the query into a folder via the
  sidebar, the snapshot's `queryFolderId` (and `position`) are stale, so the next save dispatches the stale values and the backend merge overwrites
   the real ones.

  ## Fixes #4329

  In `TabQueryEditor.saveQuery`, override the payload's `queryFolderId` and `position` from the live `savedQueries` store before dispatching the
  save.

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -1357,6 +1357,13 @@ import { KeybindingPath } from '@/common/bksConfig/BksConfigProvider'
             const payload = _.clone(this.query)
             payload.text = this.unsavedText
             payload.excerpt = payload.text.substr(0, 250)
+            if (payload.id) {
+              const latest = this.savedQueries.find(q => q.id === payload.id)
+              if (latest) {
+                payload.queryFolderId = latest.queryFolderId
+                payload.position = latest.position
+              }
+            }
             this.$modal.hide(`save-modal-${this.tab.id}`)
             const id = await this.$store.dispatch('data/queries/save', payload)
             this.tab.queryId = id


not-night-but (approved):

not-night-but:
@Dojeto Thanks! I cherry picked this onto the branch `rc-58` so we can push it out with the next release :)

PR #4343, opened 2026-06-01
feat: Enable custom SSL certificate support in ClickHouse client

mastercactapus:
This pull request adds support for custom SSL certificates when connecting to ClickHouse databases, and updates the frontend to reflect this new capability in the connection form.

**Backend: Enhanced SSL/TLS Support for ClickHouse Connections**
- Added support for custom SSL certificates in the `ClickHouseClient` by conditionally setting the `tls` configuration based on the presence of certificate files (`sslCaFile`, `sslCertFile`, `sslKeyFile`). If certificates are not provided, or if `sslRejectUnauthorized` is not set, the client will not verify the server certificate, to be consistent with behavior elsewhere.

**Frontend: Updated Connection Form**
- Enabled support for complex SSL options in the ClickHouse connection form by setting `support-complex-s-s-l` to `true` in `ClickHouseForm.vue`, allowing users to provide custom certificates via the UI.

### DIFF
--- a/apps/studio/src-commercial/backend/lib/db/clients/clickhouse.ts
+++ b/apps/studio/src-commercial/backend/lib/db/clients/clickhouse.ts
@@ -63,6 +63,9 @@ import { errors } from "@/lib/errors";
 import { IDbConnectionServer } from "@/lib/db/backendTypes";
 import { ChangeBuilderBase } from "@shared/lib/sql/change_builder/ChangeBuilderBase";
 import { ClickHouseCursor } from "./clickhouse/ClickHouseCursor";
+import { readFileSync } from 'fs';
+import { NodeClickHouseClientConfigOptions } from "@clickhouse/client/dist/config";
+import https from 'https'
 
 interface JSONResult {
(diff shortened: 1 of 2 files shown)


not-night-but:
@mastercactapus Thanks for the contribution! Have you been able to verify that this works? Is there any way for us to add some integration tests for this functionality?

mastercactapus:
@not-night-but Yes, I tested it on my personal machine last night with a custom CA; and this AM I've been using this branch at work (original motivation for fixing it) with the default "don't reject unauthorized" behavior without issue.

As for integration tests, always a good idea, but I didn't see anything setup for clickhouse yet (looks like only postgres is covered in e2e) so I tried to keep this change as minimal as possible.

not-night-but:
@mastercactapus Yeah our e2e tests are more for just ensuring that electron is still behaving. We do have integration tests for clickhouse, but I agree it might take a bit more work to actually test this case.

Thanks again for the contribution, and please feel free to file more if you feel so inclined :)

not-night-but (approved):

PR #4352, opened 2026-06-02
feat: Add tables overview dashboard

### LINKED ISSUE #4055: FEAT: Full table overview with statistics (record count, size, free space, index size)

Currently, there's no built-in overview to quickly see table statistics across all tables in a database. Users who want to monitor table growth need to write custom queries instead of having this information readily available in the UI.

Add a full table overview/dashboard that displays key statistics for all tables:
- Total number of records per table
- Total size per table
- Free space per table
- Index size per table

This would be visible when connecting to a database, allowing users to immediately see how table growth is progressing without writing queries.

- Writing custom queries for each database - works but defeats the purpose of a GUI tool
- Using database-specific admin tools - but users want this in Beekeeper Studio
- Third-party monitoring tools - overkill for quick development checks

**Additional context**

User feedback: "When i log in i want to see how table growth is progressing. I can get that with a query but its one of the reasons i stick with querious for now..."

This feature exists in tools like Querious and is valuable for:
- Monitoring database growth trends
- Identifying tables that need optimization
- Quick health checks during development
- Space management and planning

Should work across all major databases (PostgreSQL, MySQL, etc.) - each has system tables/views for this metadata

---
*Created from user email feedback*

Francisco-R-Mendonca:
Add a new Tables Overview tab to inspect table-level metrics across a database, including row count, total size, table size, index size, and fragmentation indicators when supported.

The feature introduces a shared overview model and database client contract, with engine-specific implementations for PostgreSQL and SQLite. Other clients expose the same contract but return unsupported or empty behaviour until proper engine-specific support is added.

Add unit tests for the overview component and database client behaviour.

Closes #4055

Demonstration video:
https://github.com/user-attachments/assets/25bf1bd9-79b2-4310-b900-3bf65364787d

### DIFF
--- a/apps/studio/src-commercial/backend/handlers/connHandlers.ts
+++ b/apps/studio/src-commercial/backend/handlers/connHandlers.ts
@@ -1,6 +1,6 @@
 import { UserSetting } from "@/common/appdb/models/user_setting";
 import { IConnection } from "@/common/interfaces/IConnection";
-import { DatabaseFilterOptions, ExtendedTableColumn, FieldDescriptor, FieldEditData, FilterOptions, NgQueryResult, OrderBy, PrimaryKeyColumn, Routine, SchemaFilterOptions, StreamResults, SupportedFeatures, TableChanges, TableColumn, TableFilter, TableIndex, TableInsert, TableOrView, TablePartition, TableProperties, TableResult, TableTrigger, TableUpdateResult } from "@/lib/db/models";
+import { DatabaseFilterOptions, ExtendedTableColumn, FieldDescriptor, FieldEditData, FilterOptions, NgQueryResult, OrderBy, PrimaryKeyColumn, Routine, SchemaFilterOptions, StreamResults, SupportedFeatures, TableChanges, TableColumn, TableFilter, TableIndex, TableInsert, TableOrView, TablePartition, TableProperties, TableResult, TableTrigger, TableUpdateResult, TableOverview, TablesOverview } from "@/lib/db/models";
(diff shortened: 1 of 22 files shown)


Francisco-R-Mendonca:
@not-night-but hi! Ana and I worked on a new feature. Could you take a look when you have a chance?

not-night-but:
@Francisco-R-Mendonca Hey! Sorry for the delay. I think this is an interesting feature, but I think it might be better suited to being a plugin perhaps? I think regardless the style needs to be worked on, as it doesn't look very cohesive with the rest of the app

Francisco-R-Mendonca:
@not-night-but Thanks for the feedback! We’re happy to make changes. Could you share a bit more detail on what you’d like us to adjust here, especially around the plugin direction / UI fit with the rest of the app? That’d help us make sure we’re taking it in the right direction.

not-night-but:
@Francisco-R-Mendonca In terms of UI direction, I think this probably needs to be a Tabulator table, maybe similar to the table in `TableSchema.vue`. In terms of splitting it out into a plugin, the documentation for that is [here](https://docs.beekeeperstudio.io/plugin_development/). I think for the most part the functions that are added to the client can just be queries built within the plugin and then queried there. You can see an example of a plugin in the [ai-shell](https://github.com/beekeeper-studio/bks-ai-shell) and the [er-diagram](https://github.com/beekeeper-studio/bks-er-diagram)

anabdsantos:
@not-night-but Thanks, this is really helpful! We'll work on it and come back with an updated PR.

PR #4359, opened 2026-06-03
Fix/query selection identify error

### LINKED ISSUE #4350: BUG: annot paste SQL starting with MySQL DEFINER using backticked wildcard host (`user`@`%`) into empty editor on macOS
**Describe the bug**
When the SQL editor is completely empty, pasting a MySQL/MariaDB procedure starting with a `CREATE DEFINER` clause does not work if the definer contains a backticked wildcard host.

Example:

```sql
CREATE DEFINER=`example_user`@`%` PROCEDURE test_proc()
BEGIN
    SELECT 1;
END;
```

Nothing gets pasted into the editor.

I narrowed it down to this part:

```sql
`example_user`@`%`
```

If I remove the `@`%` / wildcard host part, the paste works. Also, if I type any random character into the editor first and then paste the same SQL, the paste works.

**To Reproduce**
1. Open Beekeeper Studio on macOS.
2. Open a MySQL/MariaDB connection.
3. Open a new empty SQL editor/query tab.
4. Copy this SQL text to the clipboard:

```sql
CREATE DEFINER=`example_user`@`%` PROCEDURE test_proc()
BEGIN
    SELECT 1;
END;
```

5. Try to paste it into the empty SQL editor.
6. Nothing is pasted.
7. Type any random character into the empty editor.
8. Paste the same SQL text again.
9. The paste works.

**Expected behavior**
The SQL text should be pasted normally into the SQL editor, even when the editor is empty and the pasted content starts with:

```sql
CREATE DEFINER=`example_user`@`%`
```

**Screenshots / Gifs**

 - OS: macOS Tahoe 26.5
 - App Version: 5.8.0
 - Database type and version: MySQL / MariaDB

**Additional context**
This looks like an editor/paste edge case rather than a SQL syntax issue, because the exact same SQL can be pasted once the editor already contains at least one character.

The problematic sequence seems to be related to the MySQL definer user/host expression:

```sql
`example_user`@`%`
```

Removing the wildcard host part makes the paste work.

azmy60:
fix #4350 
Screenshot 2026-06-03 at 22 23 38


azmy60:
need re-branching to rc-58

Issue #4355, opened 2026-06-03
BUG: Beekeeper does not follow symlinks (e.g. for az-cli)


philipbrembeck:
**Describe the bug**

Beekeeper Studio does not auto-discover tools installed via Homebrew on macOS, and does not follow symlinks when resolving configured tool paths. Homebrew exposes a stable symlink at `/opt/homebrew/bin/` (e.g. `/opt/homebrew/bin/az`, `/opt/homebrew/bin/pg_dump`) that always points to the currently installed version. Beekeeper neither picks these up automatically nor resolves them correctly when set manually. 

As you can see in the screenshot below, it always resolves the underlying path of the symlink (here to 2.85.0) instead of just using the symlink itself. 

After a `brew upgrade`, the underlying versioned path changes and Beekeeper silently loses the reference.

**To Reproduce**

1. Install `azure-cli` or `postgresql` via Homebrew on macOS (`brew install azure-cli`)
2. Open Beekeeper and go to the tool path settings for `az` or `pg_dump`
3. Observe that no path is auto-detected
4. Manually set the path to `/opt/homebrew/bin/az`
5. Run `brew upgrade azure-cli` (when a new version is available) 
6. Return to Beekeeper Studio and attempt to use the Azure login or pg_dump feature
7. Tool is no longer found

**Expected behavior**

Beekeeper should either auto-discover tools from `/opt/homebrew/bin/` on macOS (it is on `$PATH` in any normal shell session), or resolve symlinks at runtime rather than at save-time, so that `brew upgrade` does not silently break tool integrations.

**Screenshots / Gifs**

<img width="631" height="782" alt="Image" src="https://github.com/user-attachments/assets/d1d8d8a1-2a57-4da3-86c3-d9ef0c75f68b" />

**Version Information**

- OS: macOS
- App Version:  5.7.3
- Database type and version: unrelated

**Additional context**

`/opt/homebrew/bin/` is the standard Homebrew prefix on Apple Silicon Macs. `/usr/local/bin/` is the equivalent on Intel Macs. Both should be checked.

rathboma:
Thanks for the bug report! Yes agreed we should check homebrew paths

not-night-but:
The fix has been merged into `rc-58` and so should be fixed in the next release. Thanks for the report!

PR #4360, opened 2026-06-03
Fix symlink resolution for Homebrew CLI tools (issue #4355)

rathboma:
## Summary

Fixes a critical issue where Beekeeper Studio could not discover or reliably use Homebrew-installed CLI tools (az, aws, pg_dump, mysqldump, etc.) because:
1. macOS GUI apps don't inherit the shell's PATH, so Homebrew bin directories were missing from discovery
2. The native file picker resolved symlinks to version-pinned Cellar targets, which broke after `brew upgrade`

### New Components & Handlers
- **`CliPathPicker.vue`** - Reusable CLI binary path picker component with auto-discovery via "Find" button
  - Uses `noResolveAliases` option on macOS to preserve symlinks instead of resolving to versioned Cellar paths
  - Emits `found` event when discovery succeeds, allowing parent components to react
  
- **`cliHandlers.ts`** - New backend handler for CLI discovery
  - `cli/which` handler that runs `which`/`where` with extended PATH
  - Prepends Homebrew bin directories (`/opt/homebrew/bin`, `/usr/local/bin`) on macOS so stable symlinks are discovered
  - Returns the symlink path, not the version-pinned target

### Updated Components
- **`CommonIam.vue`** (AWS) - Replaced inline CLI path picker with `CliPathPicker` component
  - Removed manual `tryFindAWSCli()` method
  - Wired profile discovery to the component's `found` event
  
- **`CommonEntraId.vue`** (Azure) - Replaced inline CLI path picker with `CliPathPicker` component
  - Removed manual `tryFindAzCli()` method
  - Simplified to use component's auto-discovery on mount

- **`BaseCommandClient.ts`** - Added `noResolveAliases` to file picker options for backup tools
  - Ensures backup tool paths also preserve symlinks

### Removed
- `backup/whichDumpTool` handler from `backupHandlers.ts` - replaced by unified `cli/which` handler

### Testing
- Comprehensive integration test (`tool-path-symlink.spec.ts`) covering:
  - Auto-discovery returns stable symlinks, not version-pinned Cellar paths
  - Running via symlink continues to work across `brew upgrade` scenarios
  - Demonstrates why the old behavior (resolved paths) breaks after upgrades

## Implementation Details

The fix addresses both parts of the issue:
- **Part A (File Picker)**: `noResolveAliases` option tells macOS's native open panel to return the symlink as-is
- **Part B (Discovery)**: Extended PATH with Homebrew bin dirs allows `which` to find tools via their stable symlinks

The test suite validates against the real filesystem with no mocks, ensuring the solution works with actual Homebrew installations and symlink semantics.

https://claude.ai/code/session_01NKC5tfog9LVbpkBB39cWEa

### DIFF
--- a/apps/studio/src/components/common/form/CliPathPicker.vue
+++ b/apps/studio/src/components/common/form/CliPathPicker.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="form-group">
+  <div class="form-group cli-path-picker">
     <label :for="inputId">
       {{ label }}
       <i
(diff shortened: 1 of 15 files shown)


apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -223,8 +184,13 @@ export default {
...
   async mounted() {
-    const cliPath = await this.tryFindAWSCli();
-    await this.tryFindAWSProfiles(cliPath);
+    // CliPathPicker auto-discovers on mount and emits `found` with the
+    // discovered path; AWS-specific profile fetching is wired to that event.
+    // Still run a profile fetch here for the case where cliPath was already
+    // persisted from a prior session (no discovery needed, no `found` fires).
+    if (this.config.iamAuthOptions?.cliPath) {
+      await this.tryFindAWSProfiles(this.config.iamAuthOptions.cliPath);
+    }
```
not-night-but: There's a race condition here. the cli path picker runs which, and this depends on which, but is only run here. We probably want to run this on a watcher for the cliPath?

apps/studio/src/components/common/form/CliPathPicker.vue
```diff
@@ -0,0 +1,117 @@
...
+      class="alert alert-danger"
+      v-show="!cliFound"
+    >
+      <i class="material-icons-outlined">warning</i>
+      <div>
+        NO CLI FOUND, Please refer to our
+        <a :href="docsHref">Beekeeper Docs</a>
+        for more information
+      </div>
+    </div>
```
not-night-but, line 52: Looks like there's a style issue here?:
Image

This is what it looks like in the backup picker: 

Image

not-night-but (changes requested):
Just two things. We probably also want this on `rc-58`

not-night-but (approved):

PR #4365, opened 2026-06-04
[DEMO] Linux packaging and distribution via omnipackage

olegantonyan:
Hello
I made a software that might be useful for Beekeeper Studio - https://omnipackage.org

It allows indie developers to re-use existing Linux packaging infrastructure for distribution, and gives users native auto-updatable packages specifically for their distro. It's essentially a wrapper on top of native Linux tools to make RPM&DEB + S3 client to manage repositories.

For users the entrypoint could look like this - https://repositories-test.omnipackage.org/beekeeper-studio/install.html (static page generated after build and put in S3 along with repo files).
Choose distro, copy-paste commands and install the software. From now on they'll pick up updates from you automatically through distro's package manager (`dnf update`, `apt upgrade`, etc) as if beekeeper-studio was in distro's standard repos.

It runs entirely in Github Actions (but not limited to - it's just a CLI that needs Linux with containers). You bring own S3-compatible bucket and GPG key. The demo linked above was built in my fork (https://github.com/olegantonyan/beekeeper-studio/actions/runs/26930574479) and hosted in my Cloudlflare R2 bucket (which is nice for indie repos with free egress and 10G storage, but requires custom domain). 

This PR is just a demo and not supposed to be merged. If you decide to pull the trigger on it we'd need to make a few decisions and improvements, like stable vs dev builds, containers caching, distro support, etc. Feel free to close this PR, or give any or none feedback in general.


not-night-but:
@olegantonyan Cool software! I think our packaging is in a good place at the moment. If we were to change how we do things I think we would want to discuss it a lot more thoroughly, but thanks for the demo!

PR #4367, opened 2026-06-05
Add StarRocks

### LINKED ISSUE #3169: FEAT: StarRocks support
StarRocks is an open source analytics database:
https://www.starrocks.io/

Similar to Trino and ClickHouse

### LINKED ISSUE #4356: BUG: Cannot browse table data on StarRocks due to information_schema query parsing error
**Describe the bug**
Cannot browse table data or expand table tree when connecting to StarRocks via MySQL protocol. Manual `SELECT` queries work fine, but automatic metadata queries fail with a column resolution error.

**To Reproduce**
1. Connect to StarRocks database using MySQL protocol (default port 9030)
2. In the left sidebar, expand a database and double-click any table
3. See error message

**Expected behavior**
Table data should load and display normally, similar to how Beekeeper works with MySQL.

**Screenshots / Gifs**

Image

 - OS: macOS 26.5
 - App Version 5.8.1
 - Database type and version StarRocks
 
**Additional context**
Root cause appears to be that StarRocks cannot parse backtick-quoted alias-qualified columns like `` `r`.`data_type` `` in `information_schema` queries.

Test results:
- ✅ `SELECT data_type FROM information_schema.columns LIMIT 1;` works
- ❌ `` SELECT `r`.`data_type` FROM information_schema.columns `r` LIMIT 1; `` fails with same error

azmy60:
resolves #3169 , resolves #4356


rathboma:
@azmy60 broken tests

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -298,6 +298,11 @@ function filterDatabase(
   return true;
 }
 
+export type ExecuteApplyChangesOptions = {
+  /** Apply the changes in autocommit instead of wrapping them in a transaction. */
+  autocommit?: boolean;
+};
+
```
not-night-but: This feels like too much work for a bit of a hack imo? It can just be a flag in executeApplyChanges (or we just store whether we are a starrocks or a mysql client), but I also wonder about the other uses of runWithTransaction may cause issues? We could also just override that.

apps/studio/tests/integration/lib/db/clients/starrocks.spec.js
```diff
@@ -0,0 +1,198 @@
...
+    // StarRocks ALTER is limited and asynchronous, and it rejects DDL inside a
+    // transaction. Disable the alter/rename common tests.
+    alter: {
+      everything: true,
+      renameSchema: true,
+      renameTable: true,
+      renameView: true,
+    },
+  },
+}
```
not-night-but: This should be its own dialect file that we use for starrocks, dialectFor shouldbe returning this

apps/studio/src/shared/lib/dialects/models.ts
```diff
@@ -32,6 +32,7 @@ export function dialectFor(s: string): Dialect | null {
       return 'greengage'
     case 'mariadb':
     case 'tidb':
+    case 'starrocks':
```
not-night-but: In this case we may want our own dialect for starrocks, as it looks like we have to disable a LOT in the tests

apps/studio/src/shared/lib/knex-starrocks/index.js
```diff
@@ -0,0 +1,134 @@
...
+// stock mysql2 dialect. Its DDL, however, diverges from MySQL:
+//   - no UNSIGNED type
+//   - AUTO_INCREMENT only on BIGINT
+//   - PRIMARY KEY must be declared at table level, AFTER the column list,
+//     e.g. `CREATE TABLE t (...) PRIMARY KEY(id)` (inline `primary key` and
+//     a `, primary key(...)` clause inside the parens both fail to parse)
+//   - no FOREIGN KEY constraints
+//   - no `engine = ...` / charset / collate clauses
+// Only the schema/table/column compilers are overridden here; everything else
+// is inherited from mysql2.
```
not-night-but, line 13: I feel weird about specifically building something for tests and also not using it?

But beyond that, if this is true, then we probably need to take a look at the ChangeBuilder

not-night-but (changes requested):
A couple issues to look at. We need a dialect file for starrocks, we should not be constructing it specifically for tests imo. We also may need a change builder.

not-night-but:
Just need to fix the tests and then I think this is probably good to go

not-night-but (approved):

Issue #4366, opened 2026-06-04
BUG: SSH doesn't skip bad config entries


not-night-but:
Typical ssh behaviour is to attempt matching entries in order until something authenticates. In our case, if the first Identity file is bad somehow (eg. doesn't exist) it can break the whole system. I'm not sure how this integrates with the "filtering agent" so that will need to be looked at, but at the very least we should not error out if there is a bad entry in the ssh config, we should simply ignore it

PR #4368, opened 2026-06-05
Fix SSH tunnel to skip missing IdentityFile entries (#4366)

rathboma:
## Summary

Fixes SSH tunnel connections failing when `~/.ssh/config` lists multiple `IdentityFile` entries and the first one doesn't exist on disk. The tunnel now mirrors OpenSSH's behavior of skipping unreadable identity files and trying the next available one, instead of aborting the connection.

## Changes

- **`tunnel.ts`**: Refactored key resolution logic to handle missing/unreadable identity files gracefully
  - Added `tryReadKeyFile()` helper that returns `undefined` for missing files instead of throwing
  - Added `resolveKeyForAuth()` function that iterates through identity files in order, skipping any that can't be read
  - Updated both main SSH and bastion key resolution to use the new logic
  - Falls back to OpenSSH default identity files when no explicit keys are available

- **`tunnel.spec.ts`**: Added unit tests for the new behavior
  - Test that skips a missing `IdentityFile` and reads the next existing one
  - Test that doesn't throw when all `IdentityFile` entries are missing (falls back to agent)

- **`ssh-skip-bad-identity.spec.js`**: Added integration test with real SSH agent and server
  - Sets up a real `ssh-agent` with a valid ed25519 key
  - Configures `~/.ssh/config` with a non-existent `IdentityFile` first, then the valid key second
  - Verifies the tunnel successfully authenticates by skipping the bad entry

## Implementation Details

The fix distinguishes between two modes:
- **Automatic (agent) mode**: Tries identity files in order from `~/.ssh/config`, skipping any that can't be read, then falls back to OpenSSH defaults. Bad entries never abort the connection.
- **Explicit key mode**: Honors the user's explicit choice exactly—if they picked a key that can't be read, that's a real error and surfaces as expected.

This matches OpenSSH's behavior where `ssh(1)` skips unusable `IdentityFile` entries and continues until one authenticates.

https://claude.ai/code/session_01YNRPwCo2A6X4vcZ6BvE5xm


apps/studio/src/store/index.ts
```diff
@@ -55,6 +55,14 @@ function shouldPromptCockroachJwt(config: Nullable<IConnection>) {
     !config?.password;
 }
 
+// Surface non-fatal ~/.ssh/config issues (untrusted/invalid config, missing
+// IdentityFile) as warning toasts after a connect or test.
+function notifySshConfigWarnings(warnings?: string[]) {
+  for (const warning of warnings || []) {
+    Vue.prototype.$noty.warning(warning, { timeout: 8000 });
```
rathboma: I think a single noty with all warnings would be better. Be sure it's formatted well

apps/studio/tests/integration/lib/db/clients/ssh-skip-bad-identity.spec.js
```diff
@@ -0,0 +1,169 @@
...
+// throws ENOENT and the whole connection fails — even though the SSH agent
+// holds a perfectly good key that OpenSSH would have used.
+//
+// This test sets up the exact OpenSSH-faithful scenario:
+//   - a valid ed25519 key is loaded into a real ssh-agent and authorized on
+//     the server,
+//   - ~/.ssh/config lists a non-existent IdentityFile FIRST, then the valid
+//     key second, with IdentitiesOnly yes.
+// OpenSSH would skip the missing first entry and authenticate with the second.
+// Beekeeper currently errors out on the bad first entry, so this test FAILS
```
rathboma: update this comment, this is fixed

apps/studio/tests/integration/lib/db/clients/ssh-skip-bad-identity.spec.js
```diff
@@ -0,0 +1,169 @@
+// Regression test for https://github.com/beekeeper-studio/beekeeper-studio/issues/4366
+//
+// OpenSSH behaviour: when ~/.ssh/config lists several IdentityFile entries,
+// ssh(1) tries them in order and simply SKIPS any entry it cannot use (e.g. a
+// file that doesn't exist on disk), continuing until one identity
+// authenticates. A bad entry never aborts the connection.
+//
+// Beekeeper must mimic this exactly. Today it does not: connection-provider
```
rathboma: Remove this comment, as we've fixed this problem

apps/studio/src/lib/ssh/sshConfigReader.ts
```diff
@@ -7,26 +7,86 @@ import { resolveHomePathToAbsolute } from "@/handlers/utils";
...
+  }
+
   try {
     const raw = fs.readFileSync(configPath, "utf-8");
     const config = SSHConfig.parse(raw);
-    const result = config.compute(host, { ignoreCase: true });
+    // Pass the connection username so `Match user`/`Match localuser` rules are
+    // evaluated against it. Without it, compute() falls back to the OS user and
+    // those Match blocks never fire.
+    const result = config.compute(
```
rathboma, line 91: Add an option to config.default.ini in the security section to allow users to disable match processing for ssh config-file reading. Read that value here. Also add a test to check it. Also list this option in the security section of the docs

apps/studio/src/store/index.ts
```diff
@@ -55,6 +56,17 @@ function shouldPromptCockroachJwt(config: Nullable<IConnection>) {
     !config?.password;
 }
 
+// Surface non-fatal ~/.ssh/config issues (untrusted/invalid config, missing
+// IdentityFile) as a single warning toast after a connect or test.
+function notifySshConfigWarnings(warnings?: string[]) {
```
rathboma: I think the notify code should live in the connection component, not in the store.

apps/studio/src/lib/ssh/sshConfigReader.ts
```diff
@@ -4,29 +4,101 @@ import os from "os";
...
+    return endResult;
+  }
+
   try {
     const raw = fs.readFileSync(configPath, "utf-8");
     const config = SSHConfig.parse(raw);
-    const result = config.compute(host, { ignoreCase: true });
+    // `Match` blocks apply directives via dynamic criteria and `Match exec` runs
+    // arbitrary commands. Operators can disable Match processing entirely via
+    // [security] allowSshConfigMatch; drop the blocks so compute() never sees them.
```
rathboma: This is not how to do this, the library supports this natively:

```md
`Match exec` criteria are evaluated by default to match OpenSSH behavior. If
the config text may be untrusted, pass `{ matchExec: false }` to skip
`Match exec` sections while still computing `Host` sections and non-exec
`Match` criteria:

config.compute('walden', { matchExec: false })
```

apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -548,6 +560,7 @@ export default Vue.extend({
         this.connectionError = null
         const connected = await this.$store.dispatch('test', this.config)
         if (!connected) return false
+        this.notifySshConfigWarnings(connected)
```
rathboma: now test fails if there are no warnings.... Just make the warnings part of the store, and have them map to the component's computed values. Then use a `watch` to display them. Clear any warnings on test/connect. This is more idiomatic Vue

PR #4391, opened 2026-06-11
fix for numerics and error handling

### LINKED ISSUE #4388: Bigquery chokes on NUMERIC
**Describe the bug**
Bigquery is choking on results when the column contains a decimal.

**To Reproduce**

1. Run a (simple) query that contains a NUMERIC column.
2. The query will hang:
3. Check the utility.log
4. See some results like this:

```
[2026-06-11 13:56:00.361] [error] (UtilityProcess)            ERROR SENDING MESSAGE:  {
  id: '4f477ac8-2900-4b26-9d85-f76a6ff91d47',
  type: 'reply',
  data: [
    {
      command: 'SELECT',
      rows: [ { priceRetail: '19.95' } ],
      fields: [ { name: 'priceRetail', id: 'priceRetail' } ],
      rowCount: 1,
      affectedRows: 0,
      arrayMode: false,
      totalRowCount: 1
    }
  ]
} 
 ERROR:  {}
```

**More info**

I tested on multiple tables/columns in my data and found if I cast from NUMERIC the problem does not occur.
```
Fail:    SELECT `total_price` FROM raw_shopify.orders LIMIT 1
Succeed: SELECT CAST(`total_price` AS int) FROM raw_shopify.orders LIMIT 1
Succeed: SELECT CAST(`total_price` AS string) FROM raw_shopify.orders LIMIT 1
```

**Expected behavior**
Should not fail.

**Screenshots / Gifs**

Image

 - OS: Mac M3
 - App Version: 5.8.1
 - Database type: BigQuery

**Additional context**

not-night-but:
Fix #4388 

Numerics in bigquery are returned as a Big.js object, which isn't caught by our existing row data parser. Add a branch that actually converts that data to something that can be sent across port bounds. 

Also improved error handling for cases like this so that the request doesn't just hang, an error created and sent back across the port.


apps/studio/src/lib/db/clients/bigquery.ts
```diff
@@ -636,8 +636,13 @@ export class BigQueryClient extends BasicDatabaseClient<BigQueryResult> {
       const parsedRow = {}
       Object.keys(row).forEach((key) => {
         let strValue = row[key]
-        if (strValue != null && (Object.prototype.hasOwnProperty.call(strValue, 'value'))) {
-          strValue = row[key].value
+        if (strValue !== null && typeof strValue === 'object') {
```
azmy60: Can we have a big query unit tests for this? Just for avoiding regression. We might want to extract this into a testable function.

azmy60 (dismissed):

PR #4393, opened 2026-06-12
Fix path traversal in plugin manifest id (GHSA-3wfm-5rhc-mg5c)

rathboma:
## Summary

Fixes a path-traversal vulnerability (GHSA-3wfm-5rhc-mg5c) in the plugin system. `manifest.id` is read verbatim from a plugin's own `manifest.json` and used as a path segment by `PluginFileManager.getDirectoryOf()`, which backs `remove()` / `update()` / `getManifest()`. A plugin shipping an `id` containing `../` causes recursive deletion of an attacker-chosen directory the moment the user clicks **Uninstall** (`fs.rmSync(..., { recursive: true, force: true })`).

This extends the earlier containment fix in `df17b5c32` (which only covered `getPath()`) to the remaining, more dangerous sinks, and closes the underlying bug class.

Note: the plugin system is disabled by default, so default-configuration users are not exposed; exploitation requires the plugin system to be manually enabled and a malicious plugin installed, so it requires physical access to be dangerous.

## Changes (defense in depth)

- **`getDirectoryOf()`** — containment check; refuses to resolve outside the plugins directory (also rejects the empty/`.` id, which would resolve to the plugins root and let `remove()` delete every installed plugin).
- **`isValidPluginId()`** — single-path-segment whitelist (`^[a-zA-Z0-9][a-zA-Z0-9._-]*$`, length-capped) enforced wherever a manifest is loaded from disk (`scanPlugins`, `getManifest`).
- **Identity consistency** — `scanPlugins` requires `manifest.id` to match its own directory; `getManifest` and `installPlugin` require it to match the expected/registry id, so the on-disk directory name and the manifest id can never diverge.

## Testing

- New `tests/unit/security/pluginFileManagerUninstall.exploit.spec.ts` confirms the exploit (`remove("../../victim-userdata")`) now throws and the victim directory survives, plus the scan/consistency cases.
- All plugin unit tests pass (4 suites, 11 tests); ESLint clean on changed files.

### DIFF
--- a/apps/studio/src/services/plugin/PluginFileManager.ts
+++ b/apps/studio/src/services/plugin/PluginFileManager.ts
@@ -7,6 +7,7 @@ import {
   DownloaderReport,
 } from "nodejs-file-downloader";
 import { Manifest, PluginView, Release } from "./types";
+import { InvalidPluginManifestError } from "./errors";
 import extract from "extract-zip";
 import { tmpdir } from "os";
 
@@ -279,23 +280,7 @@ export default class PluginFileManager {
 
       try {
         const manifest = JSON.parse(manifestContent);
-        // The id is later used as a path segment (uninstall/update) and as an
-        // identity key. Trust the directory name we found it in, not the
-        // self-declared id: reject malformed ids and any manifest whose id does
-        // not match its own directory, so a plugin cannot redirect filesystem
-        // operations at another location.
-        if (!isValidPluginId(manifest.id)) {
-          log.warn(
-            `Plugin in "${dir}" has a missing or invalid id. Skipping.`
-          );
-          continue;
-        }
-        if (manifest.id !== dir) {
-          log.warn(
-            `Plugin in "${dir}" declares mismatched id "${manifest.id}". Skipping.`
-          );
-          continue;
-        }
+        this.validateManifest(manifest, dir);
         manifests.push(manifest);
       } catch (e) {
         log.error(`Failed to parse manifest for plugin "${dir}":`, e);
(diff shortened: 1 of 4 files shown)


apps/studio/src/services/plugin/PluginFileManager.ts
```diff
@@ -257,7 +278,25 @@ export default class PluginFileManager {
...
       try {
-        manifests.push(JSON.parse(manifestContent));
+        const manifest = JSON.parse(manifestContent);
+        // The id is later used as a path segment (uninstall/update) and as an
+        // identity key. Trust the directory name we found it in, not the
+        // self-declared id: reject malformed ids and any manifest whose id does
+        // not match its own directory, so a plugin cannot redirect filesystem
+        // operations at another location.
+        if (!isValidPluginId(manifest.id)) {
+          log.warn(
```
azmy60: Should just throw an error and let the `catch` log the error.

apps/studio/src/services/plugin/PluginFileManager.ts
```diff
@@ -257,7 +278,25 @@ export default class PluginFileManager {
...
+            `Plugin in "${dir}" has a missing or invalid id. Skipping.`
+          );
+          continue;
+        }
+        if (manifest.id !== dir) {
+          log.warn(
```
azmy60: Same.

azmy60 (approved):
Requested some minor changes. But overall, this is good!

azmy60 (approved):

PR #4394, opened 2026-06-12
Upgrade to Vite 8

rathboma:
## Summary

Upgrades both workspaces (`apps/studio` and `apps/ui-kit`) from Vite 5.4 to **Vite 8**, and switches the Vue 2 plugin to the Vite 8–compatible fork [`vite-ng-plugin-vue2`](https://github.com/beekeeper-studio/vite-plugin-vue2).

Vite 8 dropped esbuild for Oxc/Rolldown and bumped its minimum Node to `^20.19 || >=22.12` (the repo `.nvmrc` is already `v22.22.0`, so no Node change is needed). The official `@vitejs/plugin-vue2` does not support Vite 8, so this points at our fork instead.

## Changes

- **Plugin**: replace `@vitejs/plugin-vue2` (studio + ui-kit) and the leftover `vite-plugin-vue2` (ui-kit) with `vite-ng-plugin-vue2`, and update the `vite.config` imports. The fork is installed directly from a git branch that has the built `dist/` committed, so no build step runs on install.
- **Vite**: bump `vite` to `^8.0.8` in both workspaces (resolves to 8.0.16).
- **ui-kit pure-CSS entry**: `lib/style.scss` (which `@use`s every component stylesheet) was declared as a standalone Rollup input. In library mode `cssCodeSplit` defaults to `false`, and a pure-CSS *entry* chunk hits a broken code path in Vite 8's internal `css-post` plugin (it reads the per-entry CSS map that is only populated when `cssCodeSplit` is `true`), crashing the build. Importing `style.scss` from `lib/index.ts` instead moves it into the JS graph and removes the standalone entry. The merged output is **byte-for-byte identical** to the previous `dist/style.css`.
- **ui-kit CSS filename**: set `build.lib.cssFileName: "style"` so the merged library CSS keeps the `dist/style.css` name that the package `exports` map and consumers (`@import "@beekeeperstudio/ui-kit/style.css"`) depend on — Vite 8 otherwise names it after the package (`ui-kit.css`).

## Verification

- `yarn lib:build` — ui-kit library builds; `dist/style.css` is byte-for-byte identical to the Vite 5 output (2,599,250 bytes).
- `yarn workspace beekeeper-studio build` — both the esbuild (main) and Vite 8 (renderer) builds succeed.
- `yarn workspace @beekeeperstudio/ui-kit test` — 72/72 vitest tests pass (vitest 4.1.8 on Vite 8).
- The `vite-ng-plugin-vue2` fork's own test suite passes (18/18).

## Companion change

Depends on the plugin fork branch `claude/beekeeper-vite-8-upgrade-thnhtf` in `beekeeper-studio/vite-plugin-vue2`, which carries the committed `dist/` artifacts so it can be consumed as a git dependency.

https://claude.ai/code/session_01Nqem4WCC8SSgmP5y8sSpv3

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nqem4WCC8SSgmP5y8sSpv3)_


apps/ui-kit/package.json
```diff
@@ -123,8 +122,8 @@
     "tabulator-tables": "beekeeper-studio/tabulator#f9d3c0cdf0933a9a1bc5056b773c02a68a309fa1",
     "tinyduration": "^3.2.4",
     "typescript": "~5.8.3",
-    "vite": "~5.4.19",
-    "vite-plugin-vue2": "^2.0.3",
+    "vite": "^8.0.8",
+    "vite-ng-plugin-vue2": "github:beekeeper-studio/vite-plugin-vue2#claude/beekeeper-vite-8-upgrade-thnhtf",
```
not-night-but: Use the npm release

not-night-but (approved):
Need to change the uikit package.json but other than that looks good

PR #4396, opened 2026-06-15
Query, connection, and folder sharing

azmy60:
How to share / set permissions of a file?

1. Click a connection > find a Share button
2. Or, right click a connection/query/folder > Share

Screenshot 2026-06-22 at 16 03 14

Screenshot 2026-06-22 at 16 03 29


not-night-but:
If a folder is set to `Restricted` and then you attempt to set a child team perm to `View`, we get an error in the console but we don't give any feedback in the UI: 
image

azmy60:
I also changed the connection list to become like this:

image

The "SSH" should show in the list when the SSH is set to enabled.

azmy60:
Latest commit (4b3bd3433005a1c91f001ebc6f22da519f1b6475) depends on https://github.com/beekeeper-studio/cloud/pull/145

rathboma:
Some UX feedback.

### Moving a shared connection to personal folder
1. It should let me do this, but give me a confirmation dialog saying "all team access will be removed"
2. The error message makes no sense (doesn't matter if you add the dialog). It should say "You can't move this to your personal folder because it is shared with other workspace members"
image

### Visual indicator for shared connections
Which of these items is shared with specific people? Which of these items is read only? If I have 1000 connections how do I visualize that without having to click share on each and every one?
Maybe this is some icon in the tree view, maybe it's a separate modal, maybe there's a 'manage sharing' view on the cloud app, but there should be some way to do it.
image

### Personal items should have a different share dialog
image
This is in my personal folder, why does my team have edit access?


### Something is funky with 'recent' section - have you merged in master lately?
image
What is this demo db, I've never connected to that in the workspace.

### Import connections has no scroll
image
Again you might just be on a really old fork

## No perms for folders?
image

rathboma:
@azmy60 if day says the code is good - I'm happy for you to merge it and work out the UX issues in separate PRs

apps/studio/src/common/interfaces/IQueryFolder.ts
```diff
@@ -1,6 +1,6 @@
+import { IShareable } from "./IShareable"
 
-
-export interface IFolder {
+export interface IFolder extends IShareable {
```
not-night-but, line 4: I know this change wasn't made in this PR, but this should definitely also extend Transport, I believe the lsp is complaining about this in appDbHandlers

apps/studio/src/common/interfaces/ISavedQuery.ts
```diff
@@ -1,6 +1,6 @@
+import { IShareable } from "./IShareable"
 
-
-export default interface ISavedQuery {
+export default interface ISavedQuery extends IShareable {
```
not-night-but, line 3: Did this just accidentally take over from TransportFavoriteQuery? Should they be merged? And should we stick to the naming convention (actually asking here lol)

apps/studio/src/components/common/form/MultiSelect.vue
```diff
@@ -0,0 +1,359 @@
...
+    filteredSuggestions() {
+      const query = this.value.trim().toLowerCase();
+      const suggestions = this.suggestions.filter(
+        (option) => !this.selectedOptions.includes(option)
+      );
+      if (!query) {
+        return suggestions;
+      }
+      return suggestions.filter((option) =>
+        this.filterText(option).toLowerCase().includes(query)
```
not-night-but: We do have a fuzzy finding library in the app, I think it might make more sense to use that

apps/studio/src/components/common/modals/ShareModal.vue
```diff
@@ -0,0 +1,722 @@
...
+            v-for="grant of accessGrants"
+            :key="grant.id"
+            :class="{
+              highlight: highlightedMembers.includes(grant.membershipId),
+            }"
+          >
+            <div class="icon">{{ getInitials(grant.membership.name) }}</div>
+            <div class="label">
+              {{ grant.membership.name }}
+              <span v-if="isItYou(grant.membership)"> (You)</span>
```
not-night-but: `isItYou` is expecting a user ID, not a membership

apps/studio/src/components/common/form/MultiSelect.vue
```diff
@@ -0,0 +1,359 @@
...
+    },
+    listboxId() {
+      return `multi-select-${this._uid}-listbox`;
+    },
+    activeOption() {
+      return this.filteredSuggestions[this.activeIndex];
+    },
+    filteredSuggestions() {
+      const query = this.value.trim().toLowerCase();
+      const suggestions = this.suggestions.filter(
```
not-night-but, line 151: This isn't always filtering out already selected suggestions:

Image

apps/studio/src/components/common/form/MultiSelect.vue
```diff
@@ -0,0 +1,359 @@
...
+  },
+  beforeDestroy() {
+    document.removeEventListener("mousedown", this.maybeClose);
+    window.removeEventListener("scroll", this.updatePosition, true);
+    window.removeEventListener("resize", this.updatePosition);
+  },
+});
+</script>
+
+<style scoped>
```
not-night-but: Need to set the lang to scss

apps/studio/src/components/common/modals/ShareModal.vue
```diff
@@ -0,0 +1,722 @@
...
+      if (!this.isCloud) {
+        return;
+      }
+      this.resetState();
+      this.subjectId = options.subjectId;
+      this.subjectType = options.subjectType;
+      this.$modal.show("share-modal");
+    },
+    async handleOpened() {
+      this.$refs.search.$el.querySelector("input").focus();
```
not-night-but: This ref doesn't always exist, lets make sure that doesn't cause issues

apps/studio/src/components/common/modals/ShareModal.vue
```diff
@@ -0,0 +1,722 @@
...
+      } finally {
+        this.initiallyLoadingGrants = false;
+      }
+    },
+    async loadMembershipsOnce() {
+      if (!this.membershipsLoaded) {
+        await this.loadMemberships();
+        this.membershipsLoaded = true;
+      }
+    },
```
not-night-but, line 324: Is there a reason not to just load this when we load the access grants on open? We don't even have to await it, we can just set a bool and show a spinner when they search, but we fire this whenever the search changes, which I think means we could end up spamming a bunch of requests until one of them actually finishes.

apps/studio/src/components/common/modals/ShareModal.vue
```diff
@@ -0,0 +1,722 @@
...
+        canWrite = false;
+      } else if (this.permission === "view") {
+        canRead = true;
+        canWrite = false;
+      } else if (this.permission === "edit") {
+        canRead = true;
+        canWrite = true;
+      } else {
+        throw new Error(`Invalid permission. Got "${this.permission}"`);
+      }
```
not-night-but: Shouldn't we just use `permissionToGrant` here?

apps/studio/src/components/common/modals/ShareModal.vue
```diff
@@ -0,0 +1,722 @@
...
+  mounted() {
+    this.registerHandlers(this.rootBindings);
+  },
+  beforeDestroy() {
+    this.unregisterHandlers(this.rootBindings);
+  },
+});
+</script>
+
+<style scoped>
```
not-night-but: Make sure to set the language
```suggestion
<style scoped lang="scss">
```

apps/studio/src/components/connection/AddDatabaseForm.vue
```diff
@@ -96,7 +99,12 @@
   import { mapState } from 'vuex';
 
   export default {
-    props: [],
+    props: {
```
not-night-but: It looks like we don't use this, and even if we did - we don't show the button that shows this modal if the database doesn't support adding a database or the user is in read-only mode

apps/studio/src/components/connection/CommonEntraId.vue
```diff
@@ -59,7 +61,7 @@
             class="btn btn-flat btn-icon"
             type="button"
             @click.prevent="signOut"
-            :disabled="signingOut"
+            :disabled="disabled || signingOut"
```
not-night-but: Would we not expect them to sign in as their own account? Or would we want them to copy an entra connection to their own workspace?

apps/studio/src/components/connection/CommonIam.vue
```diff
@@ -74,37 +77,38 @@
...
           v-model="config.iamAuthOptions.awsProfile"
+          :disabled="disabled"
         >
       </div>
 
       <div v-show="isKeyAuth">
         <div class="form-group">
           <label for="Access Key ID">Access Key ID</label>
-          <masked-input :value="config.iamAuthOptions.accessKeyId" @input="val => config.iamAuthOptions.accessKeyId = val" :type="'password'" />
+          <masked-input :value="config.iamAuthOptions.accessKeyId" @input="val => config.iamAuthOptions.accessKeyId = val" :type="'password'" :disabled="disabled" />
```
not-night-but: Same question as with entra. I would assume that they would want to use their own IAM identity

apps/studio/src/components/connection/SqliteForm.vue
```diff
@@ -12,6 +12,7 @@
             input-id="Database"
             editable
             show-create-button
+            :disabled="disabled"
```
not-night-but: This means that the shared connection requires the same filepath on every machine. I mean I doubt people are sharing sqlite connections, but this seems to be a common theme with different connection types. We have no way to mark something as per-user within a connection

apps/studio/src/components/connection/SqlServerForm.vue
```diff
@@ -130,7 +132,7 @@
         </div>
       </div>
     </common-server-inputs>
-    <common-entra-id v-show="azureAuthEnabled" :auth-type="authType" :config="config" />
+    <common-entra-id v-show="azureAuthEnabled" :auth-type="authType" :config="config" :disabled="disabled" />
     <!-- Kerberos requires the real FQDN/SPN; an SSH tunnel routes through localhost and
          breaks SPN matching, so the tunnel is not offered for integrated auth. -->
     <common-advanced v-show="!azureAuthEnabled && !windowsAuthEnabled" :config="config" />
```
not-night-but: This should also be disabled, no?

apps/studio/src/components/editor/QueryEditHistory.vue
```diff
@@ -96,6 +96,7 @@
             Close
           </button>
           <button
+            v-if="query.canWrite"
```
not-night-but: This can be undefined

apps/studio/src/components/sidebar/connection/ConnectionListItem.vue
```diff
@@ -21,27 +21,15 @@
             @cancel="rename = false"
           />
         </div>
```
not-night-but, line 23: I think we should revert the style changes here. I think they're unnecessary and they make it so we can display less information

apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -689,13 +691,19 @@ import { KeybindingPath } from '@/common/bksConfig/BksConfigProvider'
...
+        return this.savedQueries.find((q) => q.id === this.tab.queryId);
+      },
       readOnly() {
         if (this.tab.isLoading) {
           return true;
         }
         if (this.remoteDeleted) {
           return true;
         }
+        if (this.isCloud && !this.savedQuery.canWrite) {
```
not-night-but: savedQuery can be undefined (ie it is an unsaved query tab) and cause a whole bunch of fun things: 

Image

apps/studio/src/lib/cloud/controllers/AccessGrantsController.ts
```diff
@@ -0,0 +1,143 @@
...
+    options: {
+      subjectType: SubjectType;
+      subjectId: number;
+    }
+  ): Promise<IAccessGrant> {
+    if (q.id)
+      throw new CloudError(
+        400,
+        `Cannot create ${this.name} - it already has an ID`
+      );
```
not-night-but: Style wise we should put curly braces here

apps/studio/src/lib/cloud/controllers/AccessGrantsController.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    const subjectTypeUrl = map[options.subjectType];
+    if (!subjectTypeUrl) {
+      throw new Error(`Invalid subjectType: ${options.subjectType}`);
+    }
```
not-night-but: We do this in almost every method, probably should just add a helper

apps/studio/src/lib/cloud/controllers/AccessGrantsController.ts
```diff
@@ -0,0 +1,143 @@
...
+    );
+    return res(response, this.plural);
+  }
+
+  async upsert(
+    q: IAccessGrant,
+    options: { subjectId: number; subjectType: SubjectType }
+  ): Promise<IAccessGrant> {
+    const method = q.id ? this.update.bind(this) : this.create.bind(this);
+    return await method(q, options);
```
not-night-but: I think it's cleaner and clearer to do this in an if statement, rather than binding and all that

apps/studio/src/store/modules/data/access_grant/CloudAccessGrantModule.ts
```diff
@@ -0,0 +1,93 @@
...
+    },
+    async list(
+      context,
+      subjectType: SubjectType,
+      subjectId: number
+    ): Promise<IAccessGrant[]> {
+      return await havingCli(context, (cli) =>
+        cli.accessGrants.list(undefined, { subjectId, subjectType })
+      );
+    },
```
not-night-but: Call me crazy, but I don't see where we use this?

not-night-but (changes requested):
Looks good! There are some bugs and some typing issues that I think should be addressed, and I think we should revert the change tot he `ConnectionListItem` as I don't think it really makes sense to change that? Anyways, good job as always

apps/studio/src/components/sidebar/connection/ConnectionListItem.vue
azmy60: Hmm not sure why it's here to begin with. Probably I was experimenting stuff here lol

azmy60: Nvm, I remember. But I agree though.

apps/studio/src/common/interfaces/ISavedQuery.ts
azmy60: I don't think this can be merged to TransportFavoriteQuery since "shareable" (or perms) is a cloud feature, not for local workspace.

I do like following the naming convention, though.

apps/studio/src/components/common/form/MultiSelect.vue
azmy60: It doesn't need scss though?

apps/studio/src/components/common/modals/ShareModal.vue
azmy60: This is intentional because plain css works perfectly fine. But if we do want scss for a reason I may not know yet, sure I can change that.

apps/studio/src/components/common/form/MultiSelect.vue
azmy60: Any clues to replicate?

apps/studio/src/components/common/modals/ShareModal.vue
azmy60: Nice catch!

apps/studio/src/components/common/modals/ShareModal.vue
not-night-but: The vue lang server just gets mad at me for the nesting: 
image
(it's okay with it if we switch to scss so idk)

apps/studio/src/components/common/form/MultiSelect.vue
not-night-but: Same as the share modal, the vue lang server doesn't like this without it being scss for some reason:
image

apps/studio/src/components/common/form/MultiSelect.vue
not-night-but: I think the fuzzy change might've fixed it? I'm not sure, I suspect it had something to do with comparing object references here but idk lol. When I had first done it I had added bob, pressed escape, and then typed Martinez and he popped up again

not-night-but (approved):

Issue #3327, opened 2025-06-27
FEAT: Unix timestamp tooltips in table view


Sinu5oid:
**Is your feature request related to a problem? Please describe.**
My feature request isn't related to a problem, but an improvement can be made with relatively small efforts. Currently, there are tooltips for integer-typed cells in the table view, which are the result of [this mutation](https://github.com/beekeeper-studio/beekeeper-studio/blob/b10fd29355b0c7ec992884961999fba04ecc9fc8/apps/studio/src/mixins/data_mutators.ts#L64). However, since converted value is placed in Date constructor, it is [required](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#value) to be in milliseconds. If there's regular unix timestamp (in second precision) stored in the column, the result is wrong (e.g. `1970-01-21T06:22:37.464Z` instead of `2025-06-26T17:04:24.000Z`).

**Describe the solution you'd like**
It would be nice to support not only JavaScript timestamps (in ms) but UNIX timestamps (in seconds) as well. Shouldn't require much engineering, would be enough to perform a simple boundary / distance check for a value (e.g. subtract both `Number(cellValue)` and `Number(cellValue) * 1_000` from `Date.now()` and pick one that's smallest, as a quick workaround)

**Describe alternatives you've considered**
I didn't consider any other alternatives other than external converters, as I started using Beekeeper Studio only recently.

**Additional context**

![Image](https://github.com/user-attachments/assets/7801ac7a-e040-4af6-aac5-22b689344054)

**Database type, if appropriate**
Doesn't relate to any specific database type.

Sinu5oid:
Optionally, the same logic may be applied to nanosecond and other precision timestamps, e.g. go applications may use one of those internally:
* https://pkg.go.dev/time#Time.Unix (s)
* https://pkg.go.dev/time#Time.UnixMicro (μs)
* https://pkg.go.dev/time#Time.UnixMilli (ms)
* https://pkg.go.dev/time#Time.UnixNano (ns)

not-night-but:
Yeah I think this would be an easy improvement to the usability of the app. There are a couple improvements we want to make to the experience with date/timestamp data. This should be on the list.

junkystu:
I store a lot of unix timestamps in my database and this would be very helpful!

debesyla:
Is there any manual way, setting to force it somehow manually?

PR #4401, opened 2026-06-16
feat: unixtime query magic

Squidysquid1:
image

This pr is similar to the accepted issue: https://github.com/beekeeper-studio/beekeeper-studio/issues/3327 

Some design decisions:
- Multiple flags can be present last one wins
- defaults to US locale if not present (I found the same pattern in the money formatter)
- I truncate anything below ms since Date only supports ms
- I am using, dateStyle: 'short', timeStyle: 'short' . It might be applicable to more people to have timeStyle: 'medium' the more I think about it, but not sure.

### DIFF
--- a/apps/studio/src/lib/magic/magics/format_magics/UnixTimeMagic.ts
+++ b/apps/studio/src/lib/magic/magics/format_magics/UnixTimeMagic.ts
@@ -1,0 +1,116 @@
+import { CellComponent } from 'tabulator-tables';
+import { Magic } from '../../Magic';
+import { MagicColumn } from '../../MagicColumn';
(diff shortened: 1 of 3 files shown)


not-night-but:
@Squidysquid1 Thanks for the contribution, this looks great! Any chance you could add a section to the docs for this? Just add it to the `docs/user_guide/query-magics.md` file

Squidysquid1:
For sure, will get to this tonight

not-night-but:
@Squidysquid1 Great! Thanks so much for the contribution!

not-night-but (approved):

PR #4407, opened 2026-06-17
fix: omit DEFAULT columns from INSERT when user leaves cell empty

### LINKED ISSUE #4362: BUG: Table view INSERT explicitly sets NULL on columns with DEFAULT values instead of omitting them
**Describe the bug**
When adding a new row via the table view and leaving columns empty, Beekeeper explicitly inserts NULL into every column — including columns that have a DEFAULT value and are NOT NULL. This causes the insert to fail because NULL violates the NOT NULL constraint, and the default value is never applied.

Manually omitting those columns from the INSERT statement works correctly and lets the database apply the default.

**To Reproduce**
1. Open a SQL Server table that has a NOT NULL datetime column with a default of `(getdate())`
2. Add a new row via the table view, leaving the datetime column empty
3. Save changes
4. See error — Beekeeper is inserting NULL explicitly into that column

**Expected behavior**
Columns left empty in the table view that have a DEFAULT value defined should be omitted from the INSERT statement, allowing the database to apply the default.

**Screenshots / Gifs**
N/A

- OS: Not specified
- App Version: Not specified
- Database type and version: SQL Server (likely affects other databases too)

**Additional context**
Reported for datetime columns with `(getdate())` as default, but likely affects any column type with a DEFAULT value.

rathboma:
## Bug

Fixes #4362

When a user adds a new row in the table view and leaves a column blank, Beekeeper was building an INSERT that explicitly set every unfilled column to `NULL` — including columns that have a database `DEFAULT` value.

For a `NOT NULL` column with a `DEFAULT` (e.g. `created_at TIMESTAMP NOT NULL DEFAULT GETDATE()`), explicitly inserting `NULL` violates the constraint and the insert fails. Omitting the column entirely is the correct behaviour: the database then applies its own default.

**Example (SQL Server):**
```sql
-- Table definition
CREATE TABLE orders (
  id INT PRIMARY KEY,
  name VARCHAR(100),
  created_at DATETIME NOT NULL DEFAULT GETDATE()
);

-- What Beekeeper was generating (FAILS):
INSERT INTO orders (id, name, created_at) VALUES (1, 'Widget', NULL);
-- Error: Cannot insert the value NULL into column 'created_at'

-- What it should generate (WORKS):
INSERT INTO orders (id, name) VALUES (1, 'Widget');
-- Database applies DEFAULT GETDATE() automatically
```

## Fix

`buildInsertQuery()` in `apps/studio/src/lib/db/clients/utils.ts` already receives full column metadata (including `hasDefault`) — it just wasn't using it. The fix adds a check: if a column is marked `hasDefault: true` and the user's value is `null`/`undefined`, delete that key from the data object before passing it to Knex.

**What is unchanged:**
- Columns without DEFAULT metadata: null values are passed through as before
- DEFAULT columns where the user supplied an explicit value: included normally
- Bit conversion and `?`-in-identifier escape hacks: unaffected

## Tests

Six new unit tests in `tests/unit/lib/db/clients/buildInsertQuery.spec.ts` cover:
- Omit DEFAULT column when value is `null`
- Omit DEFAULT column when value is `undefined`
- Keep DEFAULT column when user supplies a real value
- Keep null on a non-DEFAULT column (explicit NULL insert is valid)
- Omit multiple DEFAULT columns simultaneously
- No regression when no column metadata is passed

All 6 pass; all 115 existing db-related unit tests continue to pass.

https://claude.ai/code/session_01D1K8rvWkZ6hdGwLS3cs1ry

---
_Generated by [Claude Code](https://claude.ai/code/session_01D1K8rvWkZ6hdGwLS3cs1ry)_


azmy60 (commented):
I'm not sure if a unit test is sufficient for this. Shouldn't we add integration tests too? Are we sure all databases react the same way?

rathboma:
Yuck, sorry the response got generated by Claude. Yeah we probably do need an integration test

PR #4416, opened 2026-06-18
Add SQL Server integrated (Windows/Kerberos) authentication support

rathboma:
## Summary

Adds support for SQL Server integrated authentication (SSPI/Kerberos/NTLM) via the `msnodesqlv8` native ODBC driver. This enables passwordless connections using the OS user's identity on Windows, Linux, and macOS.

## Key Changes

- **SQL Server client (`sqlserver.ts`)**: 
  - Added `connectWindowsAuth()` method to establish connections via `msnodesqlv8` instead of the default `tedious` driver
  - Added `findWindowsAuthDriver()` to probe for installed ODBC drivers (prefers modern drivers with TLS 1.2 support)
  - Added `withDeadline()` utility to wrap promises with JS-level timeouts, since native ODBC connection timeouts don't reliably cancel stalled SSPI/Kerberos handshakes
  - Updated `configDatabase()` to configure trusted connection mode when `windowsAuthEnabled` is set
  - Routes connections through `msnodesqlv8` when integrated auth is selected

- **UI Components**:
  - Updated `SqlServerForm.vue` to add "Windows / Kerberos (Integrated)" authentication option
  - Modified `CommonServerInputs.vue` to hide username/password fields when integrated auth is selected
  - Added feature gating (Enterprise Authentication tier required)

- **Database Models & Types**:
  - Added `windowsAuthEnabled` boolean field to connection models (`saved_connection`, `used_connection`)
  - Updated `IDbConnectionServerConfig` and `ISimpleConnection` interfaces
  - Added database migration to add the new column

- **Build & Dependencies**:
  - Added `msnodesqlv8` as an optional dependency in `package.json`
  - Updated `electron-builder-config.js` to unpack native `.node` binaries for `msnodesqlv8`
  - Updated `esbuild.mjs` to mark `msnodesqlv8` as external

- **Testing & Documentation**:
  - Added Windows-only integration test (`sqlserver-winauth.spec.ts`) covering both localhost and hostname connections
  - Added GitHub Actions workflow (`windows-login-tests.yaml`) to run tests on Windows with SQL Server Express
  - Added comprehensive user guide (`sql-server.md`) covering prerequisites, Kerberos vs NTLM, and troubleshooting for Windows/Linux/macOS

## Implementation Details

- **Protocol negotiation**: The same code path handles both Kerberos (domain-joined, matching SPN) and NTLM (fallback) — the OS/ODBC layer determines which is used
- **Timeout handling**: Custom `withDeadline()` wrapper ensures stalled SSPI handshakes don't hang the app indefinitely; native ODBC timeouts only cover TCP connect, not the post-connect TDS prelogin
- **Driver probing**: Attempts modern ODBC drivers first (18, 17) for TLS 1.2 support, falls back to legacy `{SQL Server}` driver on Windows only
- **Error messages**: Detailed diagnostics guide users through setup (missing drivers, SPN issues, Kerberos ticket problems)
- **Feature gating**: Integrated auth is part of the paid Enterprise Authentication tier

https://claude.ai/code/session_01BBtbdbH7wGU7JVhPCmLKYu

### DIFF
--- a/apps/studio/src/components/connection/SqlServerForm.vue
+++ b/apps/studio/src/components/connection/SqlServerForm.vue
@@ -21,7 +21,7 @@
           SQL Server Options
         </h4>
         <div class="advanced-body">
-          <div class="form-group">
+          <div class="form-group" v-show="!windowsAuthEnabled">
             <label for="domain">
               Domain
               <i
--- a/apps/studio/src/components/connection/SqlServerForm.vue
+++ b/apps/studio/src/components/connection/SqlServerForm.vue
@@ -15,6 +15,15 @@
         </option>
       </select>
     </div>
+    <div v-show="windowsAuthEnabled" class="alert alert-info">
+      <i class="material-icons-outlined">info</i>
+      <div>
+        Integrated authentication uses the current OS login &mdash; no username or
+        password. On Linux and macOS it also requires unixODBC, the Microsoft ODBC
+        Driver 18 for SQL Server, and a valid Kerberos ticket (kinit).
+        <a href="https://docs.beekeeperstudio.io/user_guide/connecting/sql-server/">Setup guide</a>
+      </div>
+    </div>
     <common-server-inputs v-show="!azureAuthEnabled" :config="config" :hide-credentials="windowsAuthEnabled">
       <div class="advanced-connection-settings">
         <h4 class="advanced-heading">
--- a/apps/studio/src/components/connection/SqlServerForm.vue
+++ b/apps/studio/src/components/connection/SqlServerForm.vue
@@ -7,12 +7,15 @@
         <option value="default">
           Username / Password
         </option>
+        <option value="windows">
+          Windows / Kerberos (Integrated)
+        </option>
         <option :key="`${t.value}-${t.name}`" v-for="t in authTypes" :value="t.value">
           {{ t.name }}
         </option>
       </select>
     </div>
-    <common-server-inputs v-show="!azureAuthEnabled" :config="config">
+    <common-server-inputs v-show="!azureAuthEnabled" :config="config" :hide-credentials="windowsAuthEnabled">
       <div class="advanced-connection-settings">
         <h4 class="advanced-heading">
           SQL Server Options
@@ -71,12 +74,18 @@
     components: {CommonEntraId, CommonServerInputs, CommonAdvanced },
     props: ['config'],
     mounted() {
-      this.authType = this.config?.azureAuthOptions?.azureAuthType || 'default'
-      this.azureAuthEnabled = this.config?.azureAuthOptions?.azureAuthEnabled || false
+      if (this.config?.windowsAuthEnabled) {
+        this.authType = 'windows'
+        this.windowsAuthEnabled = true
+      } else {
+        this.authType = this.config?.azureAuthOptions?.azureAuthType || 'default'
+        this.azureAuthEnabled = this.config?.azureAuthOptions?.azureAuthEnabled || false
+      }
     },
     data() {
       return {
         azureAuthEnabled: false,
+        windowsAuthEnabled: false,
         authType: 'default',
         authTypes: AzureAuthTypes,
       }
@@ -85,25 +94,47 @@
       async authType() {
         if (this.authType === 'default') {
           this.azureAuthEnabled = false
+          this.windowsAuthEnabled = false
+          this.config.windowsAuthEnabled = false
+          this.config.azureAuthOptions.azureAuthType = undefined
+          return
+        }
+
+        // Integrated (Windows/Kerberos) and Azure auth are both paid features.
+        if (this.$store.getters.isCommunity) {
+          this.$root.$emit(AppEvent.upgradeModal, "Enterprise Authentication");
+          this.authType = 'default'
+          return
+        }
+
+        if (this.authType === 'windows') {
+          this.azureAuthEnabled = false
+          this.windowsAuthEnabled = true
+          this.config.windowsAuthEnabled = true
           this.config.azureAuthOptions.azureAuthType = undefined
         } else {
-          if (this.$store.getters.isCommunity) {
-            this.$root.$emit(AppEvent.upgradeModal, "Enterprise Authentication");
-            this.authType = 'default'
-          } else {
-            this.azureAuthEnabled = true
-            this.config.azureAuthOptions.azureAuthType = this.authType
-          }
+          this.azureAuthEnabled = true
+          this.windowsAuthEnabled = false
+          this.config.windowsAuthEnabled = false
+          this.config.azureAuthOptions.azureAuthType = this.authType
         }
       },
       azureAuthEnabled() {
         this.config.azureAuthOptions.azureAuthEnabled = this.azureAuthEnabled
       },
       config() {
-        if (this.config.azureAuthOptions.azureAuthEnabled) {
-          this.authType = this.config.azureAuthOptions.azureAuthType;
+        if (this.config.windowsAuthEnabled) {
+          this.authType = 'windows'
+          this.windowsAuthEnabled = true
+          this.azureAuthEnabled = false
+        } else if (this.config.azureAuthOptions?.azureAuthEnabled) {
+          this.authType = this.config.azureAuthOptions.azureAuthType
+          this.windowsAuthEnabled = false
+          this.azureAuthEnabled = true
         } else {
-          this.authType = 'default';
+          this.authType = 'default'
+          this.windowsAuthEnabled = false
+          this.azureAuthEnabled = false
         }
       },
     },
--- a/apps/studio/src/lib/db/clients/sqlserver.ts
+++ b/apps/studio/src/lib/db/clients/sqlserver.ts
@@ -1278,112 +1278,103 @@ export class SQLServerClient extends BasicDatabaseClient<SQLServerResult, Transa
       )
     }
 
-    const { driver, legacy } = await this.findWindowsAuthDriver()
-    if (legacy) {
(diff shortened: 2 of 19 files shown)


apps/studio/src/lib/db/clients/sqlserver.ts
```diff
@@ -1237,6 +1259,133 @@ export class SQLServerClient extends BasicDatabaseClient<SQLServerResult, Transa
...
+    } catch {
+      throw new Error(
+        (process.platform === 'win32'
+          ? 'Integrated authentication is unavailable: the msnodesqlv8 native module could not be loaded. Try reinstalling Beekeeper Studio.'
+          : 'Integrated authentication is unavailable: the msnodesqlv8 native module could not be loaded. Install unixODBC and the Microsoft ODBC Driver 18 for SQL Server, then reinstall Beekeeper Studio.') +
+        ` See ${WIN_AUTH_DOCS_URL} for setup.`
+      )
+    }
+
+    const { driver, legacy } = await this.findWindowsAuthDriver()
```
rathboma: **Every connect opens two authenticated connections.** `findWindowsAuthDriver()` performs a full `rawSql.open(...)` — a complete SSPI/Kerberos (or NTLM) handshake against the server — purely to discover the driver name, then immediately closes it. `connectWindowsAuth()` then opens a *second* real connection via `mssql/msnodesqlv8`. So each `connect()` authenticates twice; against a domain controller that's two Kerberos exchanges (and two SQL Server logins, visible in audit logs) for one user action.

Consider folding probing into the actual connect: loop the candidate drivers and attempt the real `mssql` pool connect with each, advancing only on the driver-missing errors (`IM002` / "Data source name not found") and keeping the first that succeeds. That removes the throwaway handshake and the two code paths that build connection strings differently (see the related comment on the probe string vs. `beforeConnect`).

---

apps/studio/src/lib/db/clients/sqlserver.ts
```diff
@@ -1237,6 +1259,133 @@ export class SQLServerClient extends BasicDatabaseClient<SQLServerResult, Transa
...
+    if (process.platform === 'win32') {
+      candidates.push({ driver: 'SQL Server', legacy: true })
+    }
+
+    const PROBE_TIMEOUT_S = 10
+
+    for (const candidate of candidates) {
+      const connStr =
+        `Driver={${candidate.driver}};Server=${server},${port};Database=${database};` +
+        `Trusted_Connection=yes;TrustServerCertificate=${trustCert};`
```
rathboma: The probe validates a connection string that is **not** the one the real connection uses. Here the probe hand-builds `Driver=...;Server=...;Database=...;Trusted_Connection=yes;TrustServerCertificate=...`, while `connectWindowsAuth()` starts from mssql's *generated* `conn_str` and patches only `Driver=` and `Trusted_Connection=`. Those two strings can differ in load-bearing ways — most notably `Encrypt` (ODBC Driver 18 defaults to `Encrypt=yes`; mssql's generated string may set it explicitly to something else). A probe that succeeds therefore doesn't guarantee the real connect succeeds, which undercuts the stated rationale ("surface real connection errors"). If the probe stays, build both strings through one helper so they can't drift.

---

apps/studio/src/lib/db/clients/sqlserver.ts
```diff
@@ -1237,6 +1259,133 @@ export class SQLServerClient extends BasicDatabaseClient<SQLServerResult, Transa
...
+    const CONNECT_TIMEOUT_S = 15
+
+    // Patch mssql's generated connection string before msnodesqlv8 uses it: mssql
+    // ignores the driver name, emits Trusted_Connection as a boolean (must be
+    // yes/no), and omits TrustServerCertificate.
+    const connecting = new sqlWindows.ConnectionPool({
+      ...this.dbConfig,
+      connectionTimeout: CONNECT_TIMEOUT_S * 1000,
+      beforeConnect: (cfg: any) => {
+        cfg.conn_str = cfg.conn_str
```
rathboma: These regex patches silently no-op if the token isn't present. `.replace(/Driver=[^;]*/i, ...)` only injects the discovered driver if mssql's generated `conn_str` already contains `Driver=`; if it doesn't, no driver is set and ODBC falls back to whatever default it picks (or errors), with no indication why. Same for `Trusted_Connection=`. Since the whole feature hinges on the right driver being used, this should append the clause when the match is absent rather than assume mssql always emits it — or assert that the replacement actually happened.

---

apps/studio/src/components/connection/SqlServerForm.vue
```diff
@@ -7,12 +7,15 @@
...
+        <option value="windows">
+          Windows / Kerberos (Integrated)
+        </option>
         <option :key="`${t.value}-${t.name}`" v-for="t in authTypes" :value="t.value">
           {{ t.name }}
         </option>
       </select>
     </div>
-    <common-server-inputs v-show="!azureAuthEnabled" :config="config">
+    <common-server-inputs v-show="!azureAuthEnabled" :config="config" :hide-credentials="windowsAuthEnabled">
```
rathboma, line 27: `hide-credentials` hides only the user/password row, but the **Domain** field in the advanced section (just below) stays visible when integrated auth is selected. The new docs explicitly say Domain (password-based NTLM) is "distinct from integrated authentication", and the integrated `configDatabase()` branch ignores `config.domain` entirely. An editable-but-ignored field tends to generate confusion/support tickets — consider hiding Domain too when `windowsAuthEnabled` is true.

---

rathboma (commented):
Reviewed for correctness, legibility, and whether the feature does what the description claims. Overall this is solid, well-scoped work: the schema/migration/interface plumbing is consistent with the existing `trustServerCertificate`/Azure patterns, the UI gating mirrors the Azure flow correctly (no regression to Azure auth), and the docs + CI workflow are genuinely thorough. The error messages are clear and actionable.

The one design point worth addressing before merge is the **double authenticated connection** on every connect — the driver probe does a full SSPI/Kerberos handshake just to discover the driver name, then a second real connection is opened. Inline comments cover that plus a few smaller items. Nothing here is a hard blocker, but the probe/connect split is worth a second look.

**Main items (inline):**
1. `connectWindowsAuth` → `findWindowsAuthDriver` opens two full authenticated connections per `connect()`. Folding driver discovery into the real connect loop would halve the handshakes and collapse the duplicated connection-string logic.
2. The probe validates a different connection string than the real connect uses (notably `Encrypt` defaults), so a passing probe doesn't guarantee the real connect succeeds.
3. The `Driver=`/`Trusted_Connection=` regex patches silently no-op if the token is absent from mssql's generated `conn_str` — should append rather than assume.
4. The **Domain** field stays visible under integrated auth even though it's ignored and the docs call it a distinct mechanism.

**Smaller things (not inline):**
- `withDeadline()` is the stated timeout strategy, but `findWindowsAuthDriver` re-implements its own inline `setTimeout` race instead of reusing it. The per-candidate resolve-not-reject semantics differ, so it's defensible, but the two timeout mechanisms (15s deadline vs. 10s probe + 2s fallback vs. native `conn_timeout`) are worth a brief note on why they differ.
- Feature gating is UI-only (`isCommunity` check in the form). That matches the existing Azure auth gating, so it's consistent, but flagging that a hand-edited saved connection would bypass it — same caveat as Azure.
- `msnodesqlv8` is an `optionalDependency` with a prebuilt native binary; worth confirming the packaged Linux/macOS builds actually include and load the `.node` (the `asarUnpack` glob looks right, but it's the kind of thing that only shows up at runtime on a real install).

I could not exercise the runtime path (no Windows/SQL Server + domain here), so the SSPI behaviour itself is reviewed by reading only — the CI workflow covers the NTLM-over-localhost case, and the spec's comment is honest that real Kerberos needs manual validation.

---

rathboma:
Thanks for the thorough review. Addressed the four inline items in c41cbae3:

1. **Double handshake (1281) + probe-string drift (1345)** — folded driver discovery into the real connect. `findWindowsAuthDriver()` is gone; `connectWindowsAuth()` now loops the candidate ODBC drivers attempting the actual `mssql/msnodesqlv8` pool connect and keeps the first that opens. A missing driver fails fast at the driver-manager level (IM002, before any handshake), so only the present driver authenticates — one handshake, and the string that's validated *is* the string used. No more throwaway connection, no second code path building strings differently.
2. **Silent regex no-op (1299)** — replaced the `.replace()` patches with a `setClause()` helper that appends `Driver=`/`Trusted_Connection=`/`TrustServerCertificate=` when mssql's generated `conn_str` omits them, so the discovered driver can't be silently dropped.
3. **Domain field (form:18)** — now hidden when `windowsAuthEnabled` is true, matching the docs (NTLM Domain is a distinct mechanism, ignored by the integrated `configDatabase()` branch).

The inline probe-timeout race is also gone, so `withDeadline` is now the single timeout mechanism on the live connect.

**One thing worth a sanity-check on real hardware:** the driver-missing-vs-real-error classification now reads mssql's *wrapped* error (flattening `originalError`/`precedingErrors`) instead of the raw msnodesqlv8 error the old probe used. I matched it against `IM002`/`IM003`/"data source name not found"/"can't open lib"/etc., but I can't exercise the "Driver 18 absent, 17 present" fallback here (no Windows/ODBC). If mssql ever masks IM002 into a fully opaque string, that fallback would surface as a hard error instead of advancing to 17 — worth confirming on a box with only Driver 17 installed.

On the smaller notes: the `isCommunity` gating is UI-only by design (same as Azure auth — a hand-edited saved connection bypasses both), and the `msnodesqlv8` `asarUnpack` `.node` load is the one thing only a real packaged Linux/macOS install will confirm. Happy to adjust either if you'd prefer a different approach.

---

rathboma:
Approved (I can't officially approve it myself, but this is good enough)

PR #4421, opened 2026-06-20
Fix ResultTable menu errors and add SQL escaping

Squidysquid1:
**Fix flatMap error: pass ranges array instead of single range**
https://github.com/beekeeper-studio/beekeeper-studio/pull/4372 forgot to pass in ranges in the results table which resulted in a user not being able to right click in the result table once a query has been made.

**Fix toLowerCase error: handle undefined dataType in asIn copy**
The column metadata lookup via conn/listTableColumns returns undefined for dataType during a query result. This caused normalizeDataType() to fail when calling .toLowerCase() on undefined.

Added a null check. This probably could be handled better in the future. If you have any ideas, please throw them my way.
 
**Escaping copied sql string**
When pressing the "Copy for IN statement" button after a query has been made it will not properly escape the string.

- **Before:** ('A soft, creamy, surface-ripened cow's milk cheese.')
- **After:** ('A soft, creamy, surface-ripened cow''s milk cheese.')


apps/studio/src/common/utils.ts
```diff
@@ -377,3 +377,9 @@ export function isNumericDataType (dataType) {
 
   return numericStarts.some(t => base.startsWith(t))
 }
+
+/** Escape a string value for use in SQL by doubling single quotes */
+export function escapeSqlString(value: any): string | null {
```
rathboma: One level up (ResultTable) you should have access to `dialectData` which provides `escapeString` functions for the specific database in use. If you can update the PR to use that instead it would be helpful.

apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -404,7 +404,7 @@ import { stringToTypedArray } from '@/common/utils'
 
           return [
             this.openEditorMenu(cell),
-            this.setAsNullMenuItem(range),
+            this.setAsNullMenuItem(ranges),
```
rathboma, line 409: You should be able to pass in `this.dialectData` here as an extra argument

rathboma (dismissed):
Thank you so much for the bug fix PR!

Only one change - let's use the available `dialectData`.

Also - if you can add a test that fails before the fix, and passes after the fix, that would be excellent too. We usually require that we cover bug fixes with tests whenever possible.

Squidysquid1:
Thank you, I will get to this tonight when I have some time.

Squidysquid1:
I have implemented using dialectData. Thanks for pointing this out, I did see there was escaping elsewhere in the repo but didn't know how to plug it in.

Also note: the tests were primarily generated by llm, I do not know the testing framework well enough to properly validate, but should be super good enough.

not-night-but (approved):

not-night-but:
@Squidysquid1 Thanks so much for the contribution!

PR #4422, opened 2026-06-20
Feature / Edit table AUTO_INCREMENT value for MySQL/MariaDB

psmithson:
### What & why
Adds the ability to view and edit a table's next `AUTO_INCREMENT` value from the
**Structure → Columns** tab for **MySQL** and **MariaDB**. Previously this value wasn't
visible or editable in the UI — changing it meant raw SQL. The auto-increment column's
**Extra** cell now shows the current value (e.g. `auto_increment 1042`) and is editable;
applying runs `ALTER TABLE … AUTO_INCREMENT = n`, reusing the existing column-edit/Apply flow.

_(Per CONTRIBUTING I know an issue-first is preferred — happy to discuss the design or
close this if it doesn't fit your roadmap.)_

### Implementation
- Modeled as a **table-level** change: new `AlterTableSpec.autoIncrement` +
  `MySqlChangeBuilder.endSql()` emitting `ALTER TABLE … AUTO_INCREMENT = n` (composes
  cleanly with renames/adds/drops).
- `listTableColumns` reads `information_schema.TABLES.AUTO_INCREMENT` (only when a specific
  table is requested) and attaches it to the auto-increment column.
- UI: the auto-increment column's Extra cell is editable + pre-loaded; collected as a
  table-level change, not a per-column alteration. Display labels it `auto_increment <value>`;
  the editor opens with just the number.

### Scope
- ✅ MySQL, MariaDB. ⛔ TiDB excluded (cached per-node ID ranges / different AUTO_INCREMENT
  semantics). Other engines unaffected.

### Safety / edge cases
- `^auto_increment$` exact match — never conflicts with other EXTRA flags
  (`on update CURRENT_TIMESTAMP`, generated columns).
- Validates a positive integer; reports clamping after apply (the engine raises the value to
  max+1 if a lower one is requested).
- Column-reorder DDL preserves AUTO_INCREMENT; the cell is editable only when the value is readable.

### Testing
- 5 new unit tests for the SQL generation (`mysql.spec.js`); full unit suite passes.
- Lint clean. No new dependencies.
- Manually verified end-to-end against a remote MariaDB: edit + Apply, clamp message,
  validation, and tooltip.

### Environment tested
- OS: Windows 10 (built from source). DB: MariaDB (remote).

### Screenshots
image
image
image
image


apps/studio/src/components/tableinfo/TableSchema.vue
```diff
@@ -512,10 +583,80 @@ export default Vue.extend({
...
+      return this.cellFormatter(cell, formatterParams, onRendered)
+    },
+    extraCellTooltip(_e: any, cell: CellComponent, _onRendered: any): string {
+      if (this.isAutoIncrementRow(cell.getRow())) {
+        const value = escapeHtml(`${cell.getValue() ?? ''}`)
+        const ago = this.autoIncrementLoadedAt
+          ? ` (loaded ${this.formatLoadedAgo(Date.now() - this.autoIncrementLoadedAt)} ago)`
+          : ''
+        return `Next AUTO_INCREMENT value: ${value}${ago}. ` +
+          `The next inserted row uses this value. It can't be set below the largest existing value.`
```
not-night-but: I think this might be a bit verbose for a tooltip (also if you're using auto_increment you should know this). Can probably just be `Next AUTO_INCREMENT value: ${value}${ago}.`

apps/studio/src/components/tableinfo/TableSchema.vue
```diff
@@ -413,6 +472,18 @@ export default Vue.extend({
...
         this.$noty.success(`${this.table.name} Updated`)
+
+        // MySQL/MariaDB silently clamp AUTO_INCREMENT up to (highest existing value + 1).
+        // Re-read the real value and let the user know when their request was adjusted.
+        if (requestedAutoIncrement !== undefined) {
+          const actual = this.readCurrentAutoIncrement()
+          if (actual != null && actual !== requestedAutoIncrement) {
+            this.$noty.info(
+              `Auto-increment is now ${actual}. ${this.dialect === 'mysql' ? 'MySQL/MariaDB' : 'The database'} ` +
+              `won't set it below the largest existing value, so the requested ${requestedAutoIncrement} was adjusted.`
```
not-night-but: I think this wording could be a bit more brief. `The database clamped the auto-increment value to ${value}` or something like that. Maybe add a read more link to the relevant mysql docs

apps/studio/src/components/tableinfo/TableSchema.vue
```diff
@@ -512,10 +583,80 @@ export default Vue.extend({
...
+    // instead of the numeric value we display in the cell.
+    sanitizeReorderData(rows: any[]): any[] {
+      return rows.map((r) => {
+        if (r?.autoIncrementColumn && r.extraKeyword != null) {
+          return { ...r, extra: r.extraKeyword }
+        }
+        return r
+      })
+    },
+    // Read the current next AUTO_INCREMENT value from the freshly loaded columns.
```
not-night-but: A nitpick, but my pet peeve with LLM generated code is the overly verbose comments. If the code is doing something that requires explaining, I get it, but stuff like this where it's rewording the function name just clutters things up imo. If you're up for it I'd appreciate if you could trim them down :)

not-night-but (changes requested):
Looks good. I had some nitpicks on wording for notifications and such, as well as the sheer amount of comments (that is more personal preference than anything though). If you're able to make the changes that'd be great! If not, let me know and I can myself

apps/studio/src/components/tableinfo/TableSchema.vue
psmithson: Hello! I lived in Alberta myself a long time ago, nice to meet you. 
I agree with your points, I'll make these corrections and send them soon, thanks!

apps/studio/src/components/tableinfo/TableSchema.vue
psmithson: Agreed, and done :)

apps/studio/src/components/tableinfo/TableSchema.vue
psmithson: Good idea, I also added a single Read More button, AND made 2 different links to MariaDB or MySQL documentation respectively, depending on the brand of DB you're using.

psmithson:
All comments written again in the same concise style as the rest of the code.

Addressed all three points and merged in the latest master to clear the conflict — ready for another look, thanks!

apps/studio/src/components/tableinfo/TableSchema.vue
psmithson: Here is how the Read More button looks:
image

psmithson:
I also fixed my previous "Read More" button that was rendering as raw HTML. This latest version uses the standard Noty.button to render correctly. 
image

not-night-but:
@psmithson Looking good, just two things:
image
We probably want to update this from the response from the database.

And could we add some integration tests for this? We already have the stack set up to spin up a docker container and such, you should just need to add some cases to `mysql.spec.js`

(I also can't actually get the database to *not* clamp the value, but I'm not sure if I just have something misconfigured)

psmithson:
Sorry for the late reply.

Thanks for catching that. The table was actually showing the correct database value — the stale number was in the notification. `submitApply` read the auto-increment before the refreshed columns had landed, so it compared your 10 against the old 3 and reported a clamp that never happened. `updateTableColumns` now returns the columns it fetched and `submitApply` reads from those.

That also explains the clamp you couldn't get rid of — nothing was misconfigured on your end. Any change at all looked like a mismatch, so the notification fired on every apply. A real clamp only happens when you set a value below the current maximum, where MySQL raises it to `max(id) + 1`.

About the integration tests, I added six cases in `mysql.spec.js`: reading the value, tables with no auto-increment column, setting it, clamping below the maximum, applying alongside other alterations, and ignoring a non-positive value. Verified locally against MariaDB 12.3; CI covers 5.1, 5.7 and 8.

psmithson:
@not-night-but Bumping this one — I pushed the fixes on Aug 2: the AUTO_INCREMENT value now reads back from the database response, and I added six integration tests in mysql.spec.js covering read/set/clamping/edge cases, verified against MariaDB 12.3.

PR #4424, opened 2026-06-22
Replace Azure signing with Google Cloud KMS for Windows code signing

rathboma:
## Summary
Migrates Windows code signing infrastructure from Azure Key Vault (using AzureSignTool) to Google Cloud KMS (using jsign). This change updates the CI/CD workflow and signing implementation to use GCP's key management service instead of Azure.

## Key Changes

- **GitHub Actions Workflow** (`studio-publish.yml`):
  - Added `id-token: write` permission for OIDC authentication with GCP
  - Removed Azure credential extraction step and related job outputs
  - Replaced Azure Login action with Google Cloud authentication using workload identity federation
  - Added gcloud setup and Java installation (required for jsign)
  - Added jsign installation with SHA-256 checksum verification
  - Added Windows code signing certificate decoding from base64-encoded secret
  - Updated build environment variables from Azure (`KEYVAULT_URL`, `KV_WIN_CERTIFICATE`) to GCP (`GCP_KMS_KEYRING`, `GCP_KMS_KEY`)

- **Code Signing Script** (`apps/studio/build/win/sign.js`):
  - Replaced AzureSignTool command with jsign (Java-based signing tool)
  - Changed from Azure Key Vault authentication to Google Cloud KMS authentication
  - Updated to use GCP OAuth token minting at sign time
  - Changed timestamp server from DigiCert to Sectigo (RFC3161 compatible)
  - Updated signing algorithm configuration for jsign compatibility
  - Improved code comments to document GCP KMS integration details

## Implementation Details

- Uses Google Cloud workload identity federation for keyless authentication in CI
- jsign is downloaded and verified via SHA-256 checksum before use
- GCP OAuth token is freshly minted at signing time to prevent expiration during builds
- Windows code signing certificate (public chain) is stored as a base64-encoded secret and decoded at runtime
- Private key remains in Google Cloud KMS and never leaves the service

https://claude.ai/code/session_01UyseejFjJW4grrWxorCwGD

### DIFF
--- a/.github/workflows/studio-publish.yml
+++ b/.github/workflows/studio-publish.yml
@@ -100,20 +100,9 @@ jobs:
         if: matrix.os.setup_flatpak
         run: bash ./.github/scripts/install-build-deps.sh
 
-      - name: Authenticate to Google Cloud
-        if: matrix.os.type == 'windows'
-        uses: google-github-actions/auth@v2
-        with:
-          workload_identity_provider: ${{ secrets.gcp_workload_identity_provider }}
-          service_account: ${{ secrets.gcp_service_account }}
-
-      - name: Set up gcloud
-        if: matrix.os.type == 'windows'
-        uses: google-github-actions/setup-gcloud@v2
-
       - name: Set up Java (for jsign)
         if: matrix.os.type == 'windows'
-        uses: actions/setup-java@v4
+        uses: actions/setup-java@v5
         with:
           distribution: temurin
           java-version: '21'
@@ -295,6 +284,18 @@ jobs:
         run: |
           echo "$DATA" | base64 --decode > ~/mac-certificate.p12
 
+      # Authenticate just before signing so the access token (1h lifetime) is
+      # fresh. token_format: access_token outputs the token jsign uses as its
+      # KMS store password, so no gcloud token minting is needed in sign.js.
+      - name: Authenticate to Google Cloud
+        id: auth
+        if: matrix.os.type == 'windows'
+        uses: google-github-actions/auth@v3
+        with:
+          token_format: access_token
+          workload_identity_provider: ${{ secrets.gcp_workload_identity_provider }}
+          service_account: ${{ secrets.gcp_service_account }}
+
         # Oh hello future Matthew, why do we split out windows build?
         # well CSC_LINK gets picked up EVEN THOUGH YOU HAVE
         # A CUSTOM sign.js. Obviously a bug, but this is a workaround
@@ -304,6 +305,7 @@ jobs:
         env:
           GCP_KMS_KEYRING: "${{ secrets.gcp_kms_keyring }}"
           GCP_KMS_KEY: "${{ secrets.gcp_kms_key }}"
+          GCP_ACCESS_TOKEN: "${{ steps.auth.outputs.access_token }}"
           # WIN_CERT_FILE and JSIGN_JAR are exported to $GITHUB_ENV by earlier steps
           PYTHON_PATH: "${{'$PYTHON_PATH' }}"
           PYTHONPATH: "${{ '$PYTHONPATH' }}"
--- a/.github/workflows/studio-publish.yml
+++ b/.github/workflows/studio-publish.yml
@@ -3,6 +3,7 @@ name: Studio - Build & Publish
 permissions:
   contents: read
   actions: write
+  id-token: write
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/studio-publish.yml
```diff
@@ -114,15 +100,50 @@ jobs:
         if: matrix.os.setup_flatpak
         run: bash ./.github/scripts/install-build-deps.sh
 
-      - name: Install azuresigntool
-        run: 'dotnet tool install --global AzureSignTool --version 7.0.1'
+      - name: Authenticate to Google Cloud
         if: matrix.os.type == 'windows'
+        uses: google-github-actions/auth@v2
```
rathboma: Should use auth v3

.github/workflows/studio-publish.yml
```diff
@@ -114,15 +100,50 @@ jobs:
...
+        with:
+          workload_identity_provider: ${{ secrets.gcp_workload_identity_provider }}
+          service_account: ${{ secrets.gcp_service_account }}
 
-      - name: Azure Login
+      - name: Set up gcloud
         if: matrix.os.type == 'windows'
-        uses: azure/login@v2
+        uses: google-github-actions/setup-gcloud@v2
```
rathboma: also should use v3

.github/workflows/studio-publish.yml
```diff
@@ -114,15 +100,50 @@ jobs:
...
+
+      - name: Set up Java (for jsign)
+        if: matrix.os.type == 'windows'
+        uses: actions/setup-java@v4
```
rathboma: should use v5

.github/workflows/studio-publish.yml
```diff
@@ -114,15 +100,50 @@ jobs:
...
-          creds: '{"clientId":"${{ needs.create_draft_release.outputs.azure_client_id }}","clientSecret":"${{ secrets.keyvault_auth_secret }}","subscriptionId":"${{ secrets.azure_subscription_id }}","tenantId":"${{ needs.create_draft_release.outputs.azure_tenant_id }}"}'
+          distribution: temurin
+          java-version: '21'
+
+      - name: Install jsign
+        if: matrix.os.type == 'windows'
+        shell: powershell
+        env:
+          JSIGN_VERSION: '7.4'
+          JSIGN_SHA256: '2abf2ade9ea322acc2d60c24794eadc465ff9380938fca4c932d09e0b25f1c28'
```
rathboma, line 115: double check these against the github releases page for jsign

https://github.com/ebourg/jsign/releases

.github/workflows/studio-publish.yml
rathboma: Also looks like we can have this create the credentials file that we can use later, so no need to gen a new token in the sign.js. The token is valid for an hour, is that right?

apps/studio/build/win/sign.js
```diff
@@ -6,29 +6,43 @@ function isEmpty(value) {
...
-  const timeserver = "http://timestamp.digicert.com"
+  // Sectigo's RFC3161 timestamp server with SHA-256, per their code signing docs.
+  const timeserver = "http://timestamp.sectigo.com"
 
-  // Updated to use Azure managed identity authentication via azure/login action
-  // This uses the token obtained from the azure/login step in the GitHub workflow
-  // The -kvm flag enables managed identity authentication (uses Azure CLI credentials)
+  // Mint a fresh GCP OAuth access token at sign time so it can't expire
+  // mid-build. gcloud is provisioned by the setup-gcloud step in CI.
+  const token = cp.execSync('gcloud auth print-access-token', {
```
rathboma: I think we can just use the credentials generated by the github auth. I believe they last an hour? This build takes ~10 mins. You could move the google auth after dependency installation if you want it to be faster.

rathboma: Done. sign.js now reads the token from `GCP_ACCESS_TOKEN` (the auth step's `token_format: access_token` output) instead of shelling out to `gcloud`. The access token defaults to a 1-hour lifetime, so it comfortably covers the ~10-min build. I also moved the auth step to immediately before the Windows build so the token is freshly issued at sign time — that keeps it safe even if the dependency-install retries run long.

---

.github/workflows/studio-publish.yml
rathboma: Yes — 1-hour default lifetime. Switched the auth step to `token_format: access_token` and pass that token to jsign via env, so no token is minted in sign.js anymore (jsign's `GOOGLECLOUD` store password needs an actual access-token string rather than an ADC credentials file, and this output gives us exactly that).

---

.github/workflows/studio-publish.yml
rathboma: Removed this step entirely — with `token_format: access_token` the auth action emits the token directly, so `gcloud` is no longer needed for signing. Let me know if you'd rather keep it around for other reasons.

---

.github/workflows/studio-publish.yml
rathboma: Verified against the releases page: **7.4** is the latest (7.5+ return 404), and the pinned SHA-256 (`2abf2ade…1c28`) matches the actual `jsign-7.4.jar` — I downloaded it and confirmed its manifest reports `Implementation-Version: 7.4`, `Main-Class: net.jsign.JsignCLI`.

---

PR #4430, opened 2026-06-23
Add SQL Server integrated (Windows/Kerberos) authentication support

rathboma:
## Summary

Adds passwordless integrated authentication (SSPI/Kerberos/NTLM) for SQL Server connections, allowing users to authenticate using their OS identity without supplying credentials. This is an Enterprise Authentication feature.

## Key Changes

- **SQL Server client integration**: Added `connectWindowsAuth()` method that routes through the native `msnodesqlv8` driver instead of the pure-JS `tedious` driver, enabling SSPI support. Includes a `withDeadline()` wrapper to prevent indefinite hangs during stalled Kerberos/NTLM handshakes.

- **ODBC driver probing**: Implemented `findWindowsAuthDriver()` to discover installed ODBC drivers (preferring modern versions 17/18 for TLS 1.2 support, with fallback to legacy `{SQL Server}` on Windows). Probes directly via `msnodesqlv8` to surface real connection errors.

- **UI updates**: 
  - Added "Windows / Kerberos (Integrated)" authentication option to `SqlServerForm.vue`
  - Hidden username/password fields when integrated auth is selected via `hideCredentials` prop on `CommonServerInputs.vue`
  - Gated feature behind Enterprise Authentication paywall

- **Database schema**: Added `windowsAuthEnabled` boolean column to `saved_connection` and `used_connection` tables via migration.

- **Configuration**: Added `windowsAuthEnabled` field to `IDbConnectionServerConfig` interface and connection models.

- **Documentation**: Created comprehensive setup guide (`docs/user_guide/connecting/sql-server.md`) covering prerequisites by OS (Windows, Linux, macOS), Kerberos vs NTLM negotiation, and troubleshooting.

- **CI/Testing**:
  - Windows integration test (`.github/workflows/windows-login-tests.yaml`): Provisions SQL Server Express on `windows-2022`, validates NTLM negotiation via `localhost` and hostname
  - Kerberos integration test (`.github/workflows/sqlserver-kerberos-tests.yaml`): Full Docker stack (Samba AD DC + SQL Server on Linux + Jest runner) validating real Kerberos negotiation
  - Jest specs: `sqlserver-winauth.spec.ts` (Windows NTLM) and `sqlserver-kerberos.spec.ts` (Kerberos)
  - Docker assets: Samba provisioning, SQL Server domain join, Kerberos client setup

- **Build config**: Updated `electron-builder-config.js` to unpack `msnodesqlv8` native addon from the ASAR archive so it remains loadable at runtime.

- **Optional dependency**: Added `msnodesqlv8@^5.1.5` as an optional dependency (gracefully fails with a helpful error if not available).

## Implementation Details

- The `withDeadline()` helper wraps promises with a JS-level timeout because `msnodesqlv8`'s native `conn_timeout` does not reliably cancel stalled SSPI/Kerberos handshakes—only the TCP connect phase. This guarantees the app stays responsive even if the handshake hangs.

- Connection string patching in `beforeConnect` corrects `mssql`'s generated ODBC string: fixes driver name format, converts `Trusted_Connection` boolean to `yes/no`, and injects `TrustServerCertificate` when needed.

- The feature self-gates: Windows tests skip on non-Windows platforms; Kerberos tests skip unless `SQLSERVER_KERBEROS_TEST=1` (set by the dedicated workflow).

- All user-facing error messages follow the neutral statement style (no "we" language) per project conventions.

https://claude.ai/code/session_01AiMpG5DKL1CLDmNkpk4zuT


apps/studio/src/lib/db/clients/sqlserver.ts
```diff
@@ -1279,6 +1279,8 @@ export class SQLServerClient extends BasicDatabaseClient<SQLServerResult, Transa
     }
 
     const trustCert = this.dbConfig.options?.trustServerCertificate
+    const encrypt = this.dbConfig.options?.encrypt
```
rathboma: Is this correct?

rathboma: Yes. `this.dbConfig.options.encrypt` is set in `configDatabase()`'s integrated-auth branch from the form's SSL toggle (`encrypt: server.config.ssl`) — the same `ssl → encrypt` mapping the standard `tedious` path uses just below it. This line just reads it back, mirroring the `trustCert = this.dbConfig.options?.trustServerCertificate` line directly above; `connectWindowsAuth` then turns it into the ODBC `Encrypt=yes/no` clause.

The integration test confirms the round-trip: with SSL enabled the connection reports `encrypt_option = TRUE`, and `FALSE` when off.

Happy to read it straight off `this.server.config.ssl` instead if you'd prefer not to round-trip through `options` — but it's consistent with the adjacent `trustCert`/`serverSpn` reads as written.

---

apps/studio/src/lib/db/clients/sqlserver.ts
```diff
@@ -1412,9 +1410,14 @@ export class SQLServerClient extends BasicDatabaseClient<SQLServerResult, Transa
       }
 
       // trustedConnection delegates auth to the OS (SSPI -> Kerberos/NTLM) via msnodesqlv8.
+      // The integrated-auth encryption/cert/SPN settings live in sqlServerOptions;
+      // connectWindowsAuth() translates encryptionMode into the ODBC Encrypt/strict clauses.
+      const sqlServerOptions = server.config.sqlServerOptions || {};
       config.options = {
         trustedConnection: true,
-        trustServerCertificate: server.config.trustServerCertificate,
```
rathboma, line 1417: if not connecting with Kerberos or ODBC, this needs to be set from the config as before

rathboma: SQL Server still uses `tedious` for everything except integrated auth — the driver selection is unchanged from the original feature:

```ts
// connect(), sqlserver.ts
if (this.server.config.windowsAuthEnabled) {
  this.pool = await this.connectWindowsAuth()        // msnodesqlv8 / ODBC
} else {
  this.pool = await new ConnectionPool(this.dbConfig).connect()  // tedious
}
```

So Username/Password and Azure both go through tedious; only **Kerberos / Windows (integrated)** uses ODBC (it's the only path that can do SSPI). This `config.options` block is inside `if (server.config.windowsAuthEnabled)`, so its `encryptionMode`/`serverCertificate`/`serverSpn` keys are only ever consumed by `connectWindowsAuth()` — they never reach tedious.

The non-integrated branches still set `encrypt`/`trustServerCertificate` from config as before: the username/password branch builds `config.options` from `server.config.ssl` + the cert files (unchanged), and the Azure branch sets `encrypt: true`. My change only swapped what the integrated branch reads (flat `ssl`/`kerberosSpn` → `sqlServerOptions`); it didn't touch driver selection or the other branches.

So nothing regresses for non-ODBC connections. Let me know if you'd prefer I make the "integrated ⇒ ODBC only" intent more explicit here (e.g. a comment or guard) — happy to.

---

rathboma (commented):
LGTM

PR #4437, opened 2026-06-26
Use mssql dialect for sql server, and update quoting logic

not-night-but:
Update some of the quoting logic in the vendored complete from upstream, and move to using the mssql dialect in codemirror (the note about it being broken is from before we upgraded codemirror and no longer applies).


rathboma (approved):
LGTM, but test if possible

PR #4450, opened 2026-06-30
Paste as new rows + minor UX improvements

rathboma:
- Now it's possible to paste copied data as new rows in a table vs overwriting existing rows.

This is good if you have a lot of data to port over and you don't want to have to use the import system.

Also:
- backspace now deletes selected range data (as expected)
- new keybinding for past as new rows


rathboma (commented):
LGTM

PR #4457, opened 2026-07-01
Replace Pluralize

not-night-but:
This is a very stripped down of the pluralize lib. Idk if we're ever going to need the irregular words or the uncountable rules, I removed them but it would be very easy to add in. I just kept the general regex rules and I've confirmed that it works properly for every use in the app. Also confirmed that this does fix the issues with edit data and the workspace button.


rathboma (approved):
lgtm

PR #4459, opened 2026-07-01
Add support for ssh include directive

Squidysquid1:
I rely on the Include directive to separate my ssh configs by groups of machines, this pr gives support for said directive. 

This is not a full implementation, but it should have the features that most people need. For example one thing not supported is an include directive within a host block:
```
# Supported:
Include ~/.ssh/conf.d/*.conf

# Not supported:
Host my-server
    HostName 192.168.1.50
    User dev
    Include ~/.ssh/extras/my-server-settings.conf`
```

There is a world where include directory resolving lives in: https://github.com/beekeeper-studio/ssh-config 
But that kind of moves away from the purpose of that library in its current state.

### DIFF
--- a/apps/studio/src/lib/ssh/sshConfigReader.ts
+++ b/apps/studio/src/lib/ssh/sshConfigReader.ts
@@ -25,6 +25,50 @@ export interface SshConfigResult {
   warnings?: SshConfigWarning[];
 }
 
+type ReadConfigResult = {
+  content: string;
+  warning?: SshConfigWarning;
+};
+
+function readConfigFile(configPath: string): ReadConfigResult {
+  const trustIssue = configTrustIssue(configPath);
+
+  if (trustIssue) {
+    log.warn(`Ignoring ${configPath}: ${trustIssue}`);
+    return {
+      content: "",
+      warning: {
+        code: "untrusted",
+        message: `${tildify(configPath)} was ignored because ${trustIssue}.`,
+      },
+    };
+  }
+
+  try {
+    return {
+      content: fs.readFileSync(configPath, "utf8"),
+    };
+  } catch (err) {
+    const message = err instanceof Error ? err.message : String(err);
+
+    log.warn(`Ignoring ${configPath}: ${message}`);
+    return {
+      content: "",
+      warning: {
+        code: "invalid",
+        message: `${tildify(configPath)} could not be read and was ignored.`,
+      },
+    };
+  }
+}
+
+// ssh(1) resolves a relative Include against the directory of the config that
+// declared it, not the working directory.
+function resolveIncludePattern(pattern: string, includeRoot: string): string {
+  const expanded = resolveHomePathToAbsolute(pattern);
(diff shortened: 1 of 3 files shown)


Squidysquid1:
I changed my approach to be more limited after looking at similar approaches taken by other node-based projects.

apps/studio/src/lib/ssh/sshConfigReader.ts
```diff
@@ -57,30 +101,65 @@ function configTrustIssue(configPath: string): string | null {
...
+  const rootConfig = readConfigFile(rootConfigPath);
+  if (rootConfig.warning) {
+    endResult.warnings = [rootConfig.warning];
     return endResult;
   }
 
+  const includeRoot = path.dirname(rootConfigPath);
+
+  // Expanded in place so ordering holds. Nested Includes are not expanded.
+  const includeRegex = /^Include[ \t]+(.+)$/gim;
```
not-night-but, line 124: This means that it can't be indented, so this would only apply for root level Includes? Include can appear inside a Match or Host block according to the man page.

not-night-but (commented):
Great work! I think this would be super cool to have, but I am worried about having to essentially reimplement openssh's functionality. Currently this implementation is missing the ability to specify multiple pathnames in the Include clause, environment variable expansion, and the ability to specify Include blocks within Match/Host blocks. I guess going off of the docs you provided these limitations are expected. Maybe we allow this functionality to be toggled on/off? I'm not sure. Thoughts @rathboma ?

Squidysquid1:
Thanks for reviewing! Yeah, that’s what Ive come across, I tried to limit the implementation and keep it small and simple for the most basic case. There are lots of projects that don’t even bother implementing all of this complexity from my brief exploring of other node projects. I could update to a more feature filled implementation, with the cost of complexity, if you think that’s best for the project.

The actual spec has a good amount of cases to name a few (along with up to 16 recursions of globbed files):

Include=filename
Include = filename "file 2”
Include “filename”
Include file1 file2

And the same rules inside a block:
Host production-server
	Include prod_keys_config

One approach that I ruled out early on was using the ssh cli and calling `ssh -G <hostname>` because then you would need to package the binary and have all sorts of issues (probably), but it would be the “easiest” way of getting the correct config. 

Writing from my phone hopefully everything was typed out correctly since more annoying to fact check everything.

not-night-but:
@Squidysquid1 Yeah from what I've seen I'm not even sure if other projects even bother trying to parse the ssh config. I think in general something is better than nothing, and I think I'm okay with a more stripped back simple implementation at first as long as it is well documented. I think if someone needs Including in Host blocks or recursion or whatever in the future, we can add it then.

As long as Matthew is okay with this implementation then I think we're good, I might just spend some time fooling around with it to be sure.

not-night-but:
@Squidysquid1 Alrighty I messed around with this a bit and it works well! Thank you for the contribution! Hopefully we'll be able to expand on this in the future.

not-night-but (approved):

PR #4475, opened 2026-07-08
Inline folder creation

azmy60:
https://github.com/user-attachments/assets/22a43ef7-813b-4055-9aa3-7503ea7378d9

NOTE
- In cloud workspace, clicking the "create folder" icon will create a personal folder by default.
- The default name is "Untitled folder", "Untitled folder 2", "Untitled folder 3", etc..


rathboma:
@azmy60 what happens if I press `esc`? Does that cancel the folder creation? It probably should I think?

rathboma (approved):

azmy60:
This is conflicting with #4502 . I will include this feature in that PR instead.

PR #4478, opened 2026-07-08
Fix parameter detection on initial query load

rathboma:
## Summary
Fixed a timing issue where SQL parameters were not detected on the first callback when loading a saved query with `paramTypes` supplied at construction. Previously, `paramTypes` defaulted to `null` and was only corrected one transaction later, causing the initial query split and `onQuerySelectionChange` callback to run without parameter type information.

## Changes
- Moved `dialectState` and `paramTypeState` field definitions inside the `querySelection()` extension factory function so they can be seeded with the values supplied at construction time
- Updated `dialectState.create()` to use the `dialect` parameter (defaulting to `"generic"`)
- Updated `paramTypeState.create()` to use the `paramTypes` parameter (defaulting to `null`)
- Added test case verifying that parameters are correctly detected in the first callback when a query is loaded with `paramTypes` supplied at construction

## Implementation Details
The state fields are now scoped to the extension factory function, allowing them to capture the construction-time parameters. This ensures the very first query identification uses the correct dialect and parameter types, rather than waiting for effects to be dispatched in a subsequent transaction. This is particularly important for the initial `onQuerySelectionChange` callback fired when the editor is first interacted with.

https://claude.ai/code/session_01GiGADkxQEx7E3dqEoAKfEj

### DIFF
--- a/apps/ui-kit/lib/components/sql-text-editor/extensions/querySelection.ts
+++ b/apps/ui-kit/lib/components/sql-text-editor/extensions/querySelection.ts
@@ -43,31 +43,6 @@ interface QuerySelectionState {
 // Effects for updating the state
 const setDialectEffect = StateEffect.define<Options["dialect"]>();
 const setParamTypesEffect = StateEffect.define<Options["paramTypes"]>();
-// State for storing dialect
-const dialectState = StateField.define<Options["dialect"]>({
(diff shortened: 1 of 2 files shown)


azmy60 (approved):
LGTM!

PR #4480, opened 2026-07-10
Add SAP HANA database support

rathboma:
## Summary
Adds comprehensive support for SAP HANA as a new database client, enabling users to connect to and interact with HANA instances through Beekeeper Studio.

### Database Client Implementation
- **HanaClient** (`src-commercial/backend/lib/db/clients/hana.ts`): Full-featured database client implementing the standard client interface with support for:
  - Schema and database listing
  - Table, view, column, trigger, and index introspection
  - Routine (procedure/function) discovery with parameter details
  - Primary key and foreign key relationship mapping
  - Query execution with result formatting
  - Table data operations (select, insert, update, delete)
  - Filtering and pagination support

### Connection & Authentication
- **HanaPool** (`src-commercial/backend/lib/db/clients/hana/HanaPool.ts`): Connection pooling implementation with:
  - Configurable pool size (default 8 connections)
  - Connection reuse and lifecycle management
  - Query cancellation support via `ALTER SYSTEM CANCEL WORK`
  
- **HanaForm.vue**: UI component for HANA-specific connection configuration with support for multiple authentication methods:
  - Password authentication
  - JWT token authentication
  - SAML assertion authentication
  - X.509 certificate-based authentication with optional key password
  - SSL/TLS configuration options

### Dialect & Query Building
- **HanaData** (`src/shared/lib/dialects/hana.ts`): Dialect definition with:
  - HANA-specific data types (ALPHANUM, SHORTTEXT, VARBINARY, etc.)
  - Identifier wrapping and escaping rules
  - Constraint action support (including RESTRICT)
  - Feature flags for read-mostly stage 1 support

- **HanaChangeBuilder** (`src/shared/lib/sql/change_builder/HanaChangeBuilder.ts`): DDL statement generation for:
  - Column addition/alteration with HANA's parenthesized syntax
  - Column renaming and type changes
  - Default value management

### Data Access
- **HanaCursor** (`src-commercial/backend/lib/db/clients/hana/HanaCursor.ts`): Cursor implementation for paginated table data retrieval with filtering and ordering support

### Configuration & Types
- **HanaAuthType enum** and **HanaOptions interface** in `src/lib/db/types.ts` for type-safe authentication configuration
- **HanaOptionsEncryptTransformer** in `src/common/appdb/transformers/Transformers.ts` for secure credential storage
- Database migration (`src/migration/20260709_add_hana_options.js`) to add HANA options column to saved connections

### Testing
- **Integration test suite** (`tests/integration/lib/db/clients/hana.spec.ts`) with 30+ test cases covering:
  - Version reporting and schema discovery
  - Table/view/column introspection
  - Primary and composite key detection
  - Foreign key relationship mapping
  - Index and trigger listing
  - Routine discovery with parameters
  - Data querying and filtering
  - CRUD operations
  - Transaction handling

- **Test container setup** (`tests/integration/lib/db/clients/hana/container.ts`) using SAP HANA Express Edition for isolated testing

### Integration Points
- Updated client registry in `src/lib/db/clients/index.ts` with HANA configuration
- Added HANA to connection type enums and UI routing
- Integrated with existing connection provider and encryption infrastructure
- Added `@sap/hana-client` npm dependency
- Updated build configuration to handle HANA client native dependencies

## Notable Implementation Details

- Uses `@sap/hana-client` native driver for direct HANA connectivity
- Implements system schema filtering to hide HANA's hundreds of `_SYS_*` and `SYS` objects from user listings
- Supports both single-tenant and multi-tenant HANA deployments with optional database name configuration
- Handles SSH tunneling and mapped ports correctly by deferring database name specification
- Implements safe pagination with LIMIT/OFFSET using HANA's maximum limit constant
- Provides

https://claude.ai/code/session_01WQKckf2rXSK7e5px967Pyb


apps/studio/src-commercial/backend/lib/db/clients/hana.ts
```diff
@@ -0,0 +1,992 @@
...
+      customRoutines: false,
+      comments: true,
+      properties: true,
+      partitions: false,
+      editPartitions: false,
+      backups: false,
+      backDirFormat: false,
+      restore: false,
+      indexNullsNotDistinct: false,
+      transactions: false,
```
rathboma, line 185: why don't we support the transactions feature?

Need: add a comment to the PR with everything that is not currently supported by this implementation, but that technically can be supported by the driver given the time and effort.

rathboma: `transactions: false` is a scope decision, not a driver limitation. The driver fully supports transactions — `HanaConn` already implements `setAutoCommit`/`commit`/`rollback` — but the only app features that consume the transactions/manualCommit flags are data editing and manual-commit mode, and stage 1 stubs out all write paths (`executeApplyChanges` throws). Advertising `transactions: true` would enable UI affordances that dead-end. It should flip to `true` together with data editing.

Full inventory of what's disabled in this stage but supportable by `@sap/hana-client` given time and effort:

**Data operations**
- Table data editing (insert/update/delete + result editing) — plain DML; the shared `buildInsertQuery`/`buildUpdateQueries` utils apply, and commit/rollback already exist on `HanaConn`
- Transactions / manual-commit mode — as above
- Import from file — the driver has `execBatch` (batched prepared inserts), which would make imports fast
- Bind parameters — values are currently escaped and inlined (with operator/count validation); the driver supports prepared statements with parameters everywhere

**Schema operations** — `HanaChangeBuilder` already generates most of the ALTER grammar; the missing piece is tracking existing column types, since HANA's `ALTER (col ...)` syntax restates the type
- Create/drop/duplicate/truncate tables
- Add/drop/rename/retype columns, defaults, constraints
- Create/drop indexes, rename table/view
- Table and column comments (`COMMENT ON` works; the builder's `setComment` is implemented, just disabled in the UI)
- Generated columns (`GENERATED ALWAYS AS`)
- `CREATE SCHEMA` as the "add database" affordance (creating an actual tenant DB is a SYSTEMDB admin operation, so schemas are the practical unit)

**Advanced**
- Running stored procedures from the sidebar (`customRoutines`) — driver supports `CALL` with parameter metadata via `getParameterInfo`
- Partition listing — HANA has table partitioning; `SYS.PARTITIONED_TABLES` / `M_TABLE_PARTITIONS` could feed `listTablePartitions`
- Kerberos authentication — driver supports it; deferred from v1 (password/LDAP, JWT, SAML, X.509 shipped)
- LOB streaming — the driver can stream LOB columns; values currently load whole
- Driver-native connection pool — the driver ships `createPool`; the hand-rolled pool was kept because cancellation needs per-connection session ids for `ALTER SYSTEM CANCEL WORK IN SESSION`

**Not driver-solvable (listed for completeness)**
- Backups/restore — server-side `BACKUP DATA` to the server's file system with admin privileges, not a client concern
- Charsets/collations — HANA is Unicode-only; nothing to list
- Materialized views — no classic materialized views in HANA
- Shell — no embedded hdbsql equivalent

---

PR #4497, opened 2026-07-16
Add SQL: Select * to table context menus

### LINKED ISSUE #3095: FEAT: Right-click “Select *” Shortcut from Table Context Menu
From a customer:

When exploring tables, it can be slightly tedious to manually open a new query tab and write out a SELECT * FROM [tablename] just to inspect the data quickly. This slows down quick exploration and testing.

"I'd love to be able to right-click on a table name in the sidebar and see a context menu option like "SQL: Select *". Clicking it would automatically open a new SQL tab and populate it with SELECT * FROM [tablename]."

Typing the SQL manually each time, or relying on autocomplete — but neither are as fast as a simple right-click option.

**Additional context**

This is a small quality-of-life feature, but very helpful for anyone reviewing or testing database contents regularly.

All supported relational databases (e.g., PostgreSQL, MySQL, SQL Server, etc.)

coyaSONG:
## What changed

- add a `SQL: Select *` action to the shared table sidebar context menu
- load every table, view, or materialized-view column and generate an editable column list
- cap the generated query at 1,000 rows and open it in a formatted SQL tab
- quote explicit PostgreSQL column identifiers, including reserved words and names with spaces or quotes

## Why

This makes quick table exploration available directly from the sidebar while following the accepted direction in #3095: the generated query names each column so users can remove unwanted fields easily.

## User impact

Right-clicking a table-like entity now creates a dialect-aware query without requiring users to type the table or column names manually. Existing read-only behavior is unchanged because this action only creates a query tab.

## Validation

- `yarn all:lint` (0 errors; 365 existing warnings)
- `yarn workspace beekeeper-studio test:unit --runInBand --no-watchman` (82 suites, 3,108 tests passed)
- `yarn workspace beekeeper-studio build`
- `yarn lib:build`
- `yarn run electron:build --linux AppImage` completed the UI/app build and Linux packaging preparation; final AppImage assembly could not execute the downloaded Linux helper on macOS arm64 (`spawn Unknown system error -86`)

## AI assistance

Codex assisted with implementation and tests. I reviewed the resulting diff and ran the validation above.

Fixes #3095

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -1496,7 +1496,7 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
       })
     }
 
-    const selectSQL = `SELECT ${selects.join(', ')}`
+    const selectSQL = `SELECT ${selects.map(wrapIdentifier).join(', ')}`
     const baseSQL = `
       FROM ${wrapIdentifier(options.schema)}.${wrapIdentifier(options.table)}
       ${filterString}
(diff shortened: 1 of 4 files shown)


not-night-but:
This should be fixed with #4524

PR #4498, opened 2026-07-16
Fix Ctrl+Z undoing multiple steps at once in the query editor

### LINKED ISSUE #4491: BUG: Ctrl+Z undoes 2-3 actions at a time in the query editor
## Description

Pressing Ctrl+Z (Windows/Linux) or Cmd+Z (macOS) in the query editor undoes 2–3 history steps at once instead of one. Using Edit → Undo from the menu works correctly (one step at a time).

## Root Cause

In `src/common/menus/MenuItems.ts`, the undo menu item has both `role: 'undo'` and an explicit `accelerator: "CommandOrControl+Z"` with a `click` handler:

```ts
undo: {
  id: 'undo',
  label: 'Undo',
  accelerator: 'CommandOrControl+Z',
  click: actionHandler.undo,
  role: 'undo',
},
```

In Electron, `role: 'undo'` already binds Ctrl+Z/Cmd+Z natively. The explicit `accelerator` + `click` fires an additional `webContents.undo()` call on the same keypress. Combined with CodeMirror's own Ctrl+Z binding in the renderer, this can trigger undo 2–3 times per keypress.

Edit → Undo works fine because it only triggers the click path once without the role firing again.

## Fix

Either remove `role: 'undo'` and rely on the `click` + `accelerator` alone, or remove `accelerator` and `click` and rely solely on `role: 'undo'`.

## Impact

High — dangerous when editing live data (updates/deletes). User reported losing more changes than intended while amending queries.

muhammadusamahoyrr:
## Problem

Pressing Ctrl+Z (Cmd+Z) in the query editor undoes 2-3 history steps at once instead of one. Using Edit > Undo from the menu works correctly. Fixes #4491.

## Root cause

On Windows/Linux the app uses the custom in-window menu bar (`NewAppMenu.vue`), which registers every menu accelerator that has a `click` handler through `v-hotkey`. For undo/redo, one Ctrl+Z keypress therefore fires twice:

1. the menu hotkey -> `ClientMenuActionHandler.undo()` -> IPC -> `webContents.undo()`
2. CodeMirror's own Ctrl+Z keymap in the editor

`NewAppMenu.shortcut()` already skips items marked `registerAccelerator: false` - that's how `cut`/`copy`/`paste` avoid this exact double-execution - but `undo`/`redo` were missing the flag.

## Fix

Mark `undo` and `redo` as `registerAccelerator: false` in `MenuItems.ts`, matching the existing cut/copy/paste idiom. The accelerator is still displayed in the menu, Edit > Undo/Redo clicks still work via the `click` path, and the macOS native `role` behavior is unchanged. The shortcut itself is now handled solely by the focused editor.

## Testing

Added `tests/unit/common/menus/menuItems.spec.ts` asserting that all editor-scoped edit items (`undo`, `redo`, `cut`, `copy`, `paste`) are display-only and remain clickable with their roles intact. `yarn test:unit` passes locally apart from pre-existing environment-dependent suites (unbuilt ui-kit lib / DB-connection suites), which are unrelated to this change.

### DIFF
--- a/apps/studio/src/common/menus/MenuItems.ts
+++ b/apps/studio/src/common/menus/MenuItems.ts
@@ -22,8 +22,12 @@ export function menuItems(actionHandler: IMenuActionHandler, settings: IGroupedU
     undo: {
       id: 'undo',
       label: "Undo",
+      // Displayed only — the focused editor (CodeMirror, text inputs)
+      // handles the shortcut itself. Registering it would fire an extra
+      // webContents.undo() per keypress, undoing 2-3 steps at once.
       accelerator: "CommandOrControl+Z",
       click: actionHandler.undo,
+      registerAccelerator: false,
       role: 'undo',
     },
     redo: {
@@ -31,6 +35,7 @@ export function menuItems(actionHandler: IMenuActionHandler, settings: IGroupedU
       label: "Redo",
       accelerator: platformInfo.isWindows ? 'Ctrl+Y' : 'Shift+CommandOrControl+Z',
       click: actionHandler.redo,
+      registerAccelerator: false,
       role: 'redo',
     },
     cut: {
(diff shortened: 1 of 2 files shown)


not-night-but (approved):

not-night-but:
Thanks for the contribution!

PR #4499, opened 2026-07-16
Fix Trino query errors being swallowed as empty results

### LINKED ISSUE #4489: BUG: Trino query errors silently swallowed — syntax errors show as 0-row success
## Description

When a Trino query fails (syntax error, schema error, etc.), the UI shows a successful 0-row result instead of an error message. The user has no indication the query failed.

## Root Cause

In `src-commercial/backend/lib/db/clients/trino.ts`, `driverExecuteSingle()` iterates over the Trino QueryIterator but never checks `r.error` on the yielded results:

```typescript
for await (const r of result) {
  const { data: resultData, columns: resultColumns } = r
  columns = resultColumns
  if (resultData) rows.push(...resultData)
  // r.error is never checked!
}
```

When Trino returns a query error, the response has an `error` field and no `nextUri`. The `trino-client` `QueryIterator` does not throw — it returns the error response as a normal value with `done: true`. Since there's no `data`, `rows` stays empty and Beekeeper returns a successful 0-row result.

## Fix

Check `r.error` inside the loop and throw if present:

```typescript
for await (const r of result) {
  if (r.error) throw new Error(r.error.message ?? JSON.stringify(r.error))
  const { data: resultData, columns: resultColumns } = r
  columns = resultColumns
  if (resultData) rows.push(...resultData)
}
```

## Reported By

User on Linux, Beekeeper 5.8.1, connecting to a Trino instance.

muhammadusamahoyrr:
## Problem

When a Trino query fails (syntax error, unknown column, resource limits, ...), the UI shows a successful 0-row result instead of an error. The user has no indication the query failed. Fixes #4489.

## Root cause

The `trino-client` `QueryIterator` doesn't throw on query failure - when Trino returns an error, the response is yielded as a normal iterator value carrying an `error` field and no `data`. `rawExecuteQuery` in `trino.ts` only collected `data`/`columns` from each yielded result, so a failed query fell through to the `rows.length === 0` branch and was returned as a clean empty result.

## Fix

Check `r.error` on each yielded result and throw with the Trino error name and message (e.g. `SYNTAX_ERROR: line 1:1: mismatched input 'SELCT'`), so the failure propagates through `executeQuery` and surfaces in the UI like other drivers' errors.

## Testing

Extended `tests/unit/lib/db/clients/trino.spec.ts` (the mocked trino-client now supports queueing per-test results):
- throws with the Trino error name/message when the iterator yields an error result
- throws even when the error arrives after partial data pages
- successful queries still return rows (control)

All 11 tests in the suite pass, including the 8 pre-existing ones.

### DIFF
--- a/apps/studio/src-commercial/backend/lib/db/clients/trino.ts
+++ b/apps/studio/src-commercial/backend/lib/db/clients/trino.ts
@@ -441,6 +441,14 @@ export class TrinoClient extends BasicDatabaseClient<TrinoResult> {
       const rows: any[] = []
 
       for await (const r of result) {
+        // The trino-client iterator doesn't throw on query failure - it
+        // yields the error response as a normal value, so without this
+        // check a failed query looks like a successful 0-row result.
+        if (r.error) {
+          const { errorName, message } = r.error
+          throw new Error(errorName ? `${errorName}: ${message}` : message)
+        }
+
         const { data: resultData, columns: resultColumns } = r
         columns = resultColumns
 
--- a/apps/studio/tests/integration/lib/db/clients/trino.spec.ts
+++ b/apps/studio/tests/integration/lib/db/clients/trino.spec.ts
@@ -113,7 +113,7 @@ describe('Trino integration tests', () => {
         expect(typeof firstRow).toBe('object')
       }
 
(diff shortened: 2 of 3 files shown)


not-night-but:
Thanks for the fix!

not-night-but (approved):

PR #4503, opened 2026-07-17
ui: add context menu to table empty space for new row

mazzoccantelorenzo:
Adding a new row using the bottom button feels slow and confusing. The plus button is far from the data and its icon is too generic. Users might not realize it means add row. This creates friction during data entry tasks.

https://github.com/user-attachments/assets/f9dfcb56-fe4c-4421-8daa-437bb792f3ee

This is a simple improvement, but i think effective because using other db clients it's much more easy to add rows, instead of going down to that + button to add a row and then go up with your eyes, some friction that can be removed easily.

We add a right click context menu to the empty space below the table. The menu provides a clear text option to add a row. We attach an event listener to the table wrapper and prevent the default browser menu. We skip the custom menu if the user clicks inside existing data rows.

This makes the user experience completely frictionless and intuitive.


apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -760,6 +761,35 @@ export default Vue.extend({
     this.registerHandlers(this.rootBindings)
   },
   methods: {
+    onTableWrapperContextMenu(e: MouseEvent) {
+      const target = e.target as HTMLElement;
+      // Do not show custom menu if click is on a row or header.
+      // Tabulator handles its own context menus for those elements.
+      if (target.closest('.tabulator-row') || target.closest('.tabulator-header')) {
```
azmy60: What happens if we click other elements like the row header and the top left corner?

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -760,6 +761,35 @@ export default Vue.extend({
...
+      
+      // Prevent default browser context menu on the empty space.
+      e.preventDefault();
+      
+      // Define context menu options for the empty space.
+      // We add an option to insert a new row to improve user experience.
+      const menuOptions = [
+        {
+          name: "Add row",
+          shortcut: String(this.$bksConfig.keybindings.general.addRow),
```
azmy60: This shows the raw value from the configuration file. Instead, you should use `this.$bksConfig.getKeybindings("context-menu", "general.addRow")`.

azmy60 (changes requested):
Hey @mazzoccantelorenzo, I think this is great! But there's an issue that needs to be addressed.

apps/studio/src/components/tableview/TableTable.vue
mazzoccantelorenzo: https://github.com/user-attachments/assets/ef778df9-bfef-4e88-b465-e16d3509ed4d

This is what happens now, after doing some polishment. I've stressed it a lot 

Registrazione schermo 2026-07-23 alle 18.15.34

apps/studio/src/components/tableview/TableTable.vue
mazzoccantelorenzo: Refined, take a look now! Any edit can be done, no problem.

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -760,6 +761,40 @@ export default Vue.extend({
     this.registerHandlers(this.rootBindings)
   },
   methods: {
+    onTableWrapperContextMenu(e: MouseEvent) {
+      const target = e.target as HTMLElement;
+      // Do not show custom menu if click is on a row, header, row header, or corner.
+      // Tabulator handles its own context menus for those elements.
+      if (
+        target.closest('.tabulator-row') ||
```
azmy60: Is it possible to combine the selectors like this instead?

```
target.closest('.tabulator-row, .tabulator-header, .tabulator-row-header, .tabulator-corner')
```

azmy60 (changes requested):

apps/studio/src/components/tableview/TableTable.vue
mazzoccantelorenzo: > Is it possible to combine the selectors like this instead?
> 
> ```
> target.closest('.tabulator-row, .tabulator-header, .tabulator-row-header, .tabulator-corner')
> ```

Changed as you said 👍

azmy60 (approved):
Looks good, thank you!

PR #4504, opened 2026-07-17
ui: change table edit trigger from double to single click

mazzoccantelorenzo:
Users had to double click every cell to enter edit mode. Editing a long row required too many clicks and felt very slow. The double click requirement adds unnecessary friction to simple data entry tasks.
BEFORE:

https://github.com/user-attachments/assets/18b984ad-89f5-4694-b13c-134bf8deb7d0

AFTER:

https://github.com/user-attachments/assets/8b166964-8fe5-4216-948b-37a18744a1a6

We reduce 50% of friction by doing this little improvement.

We change the global tabulator configuration to trigger edit mode on a single click. This cuts the required clicks in half. Data entry is now fast and intuitive.

This improves overall user experience significantly.

### DIFF
--- a/apps/studio/default.config.ini
+++ b/apps/studio/default.config.ini
@@ -53,6 +53,8 @@ maxColumnWidth = 1000
 minColumnWidth = 100
 maxInitialWidth = 500
 defaultColumnWidth = 125
+; The event that triggers cell edit mode ("click" or "doubleclick")
+editTrigger = doubleclick
 
 [ui.tableTriggers]
 maxColumnWidth = 300
--- a/apps/studio/src/common/tabulator.ts
+++ b/apps/studio/src/common/tabulator.ts
@@ -63,7 +63,9 @@ export function tabulatorForTableData(
     resizableColumnGuide: true,
     movableColumns: true,
     height: "100%",
-    editTriggerEvent: "dblclick",
+    editTriggerEvent: window.bksConfig.ui.tableTable.editTrigger === "click" 
+      ? "click" 
+      : "dblclick",
     debugInvalidComponentFuncs: false,
     history: true,
     keybindings: {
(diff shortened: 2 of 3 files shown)


azmy60 (changes requested):
Hey! Thanks for your contribution. I think you have a valid point, but I don't think it's wise to make this change because this is a core feature that many people are already used to, especially in spreadsheet apps this is a common behavior. So I would agree to have this merged if you could add this as an option to the configuration file. 

The steps are:
1. Add the new option to `apps/studio/default.config.ini`
2. Use the new config by calling `window.bksConfig`, for example, `window.bksConfig.ui.table.editTrigger`.

mazzoccantelorenzo:
> Hey! Thanks for your contribution. I think you have a valid point, but I don't think it's wise to make this change because this is a core feature that many people are already used to, especially in spreadsheet apps this is a common behavior. So I would agree to have this merged if you could add this as an option to the configuration file.
> 
> The steps are:
> 
> 1. Add the new option to `apps/studio/default.config.ini`
> 2. Use the new config by calling `window.bksConfig`, for example, `window.bksConfig.ui.table.editTrigger`.

I did some polishment and refining, take a look and we can iterate, if you dont like something we can change it, no problem :D

apps/studio/src/config/typesGenerator.ts
```diff
@@ -31,6 +31,7 @@ export function generateConfigTypes(): void {
     dtsGen
       .generateIdentifierDeclarationFile("IBksConfig", processedConfig)
       .replace("declare const IBksConfig:", "declare interface IBksConfig")
+      .replace(/editTrigger: string;/g, "editTrigger: 'click' | 'doubleclick';")
```
azmy60: Interesting. I don't think we want to modify the types generator this way. It's a good idea, but it'll be a bit dishonest because `editTrigger` can be an arbitrary string, because we don't have a validation process. I think it's best to not strictly type this for now.

azmy60 (changes requested):

apps/studio/src/config/typesGenerator.ts
mazzoccantelorenzo: No problem, it has been removed. Let me know what to do next 👍

azmy60 (approved):
This looks good, thanks!

PR #4507, opened 2026-07-18
Position database dropdown icons absolutely

### LINKED ISSUE #4483: FEAT: Allow resizing the database location dropdown
The database location dropdown has a fixed width. When database names or paths are long, they are truncated, and I have to horizontally scroll to read the full value. This becomes annoying when switching between database locations frequently.

Allow the database location dropdown to be resized horizontally so that longer database names or paths can be viewed without requiring horizontal scrolling. Ideally, the dropdown would remember the width I set for future use.

The current workaround is to use the horizontal scrollbar, but this is slow and inconvenient when changing database locations often.

manos02:
- Wrapped icons in a div
- Positioned `.db-actions` absolutely at right
- Gave `.dropdown-search width: 100%` plus right padding to reserve space for the icon strip

Closes: https://github.com/beekeeper-studio/beekeeper-studio/issues/4483

### DIFF
--- a/apps/studio/src/components/sidebar/core/DatabaseDropdown.vue
+++ b/apps/studio/src/components/sidebar/core/DatabaseDropdown.vue
@@ -17,22 +17,24 @@
         placeholder="Select a database..."
         class="dropdown-search"
       />
-      <a
-        v-if="supportsMultipleDatabases"
-        class="refresh"
-        @click.prevent="refreshDatabases"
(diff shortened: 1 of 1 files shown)


not-night-but:
@manos02 Oh okay! I was misunderstanding what you were requesting. Thanks so much for the fix, this definitely works better

not-night-but (approved):

PR #4508, opened 2026-07-18
fix: show loading spinner and detailed error on Test connection

### LINKED ISSUE #4456: BUG: Test connection button shows no feedback on timeout
## Description

When clicking the "Test" button on the connection screen and the connection times out, the UI shows no feedback — no loading state, no error message, nothing happens visually.

The error only appears in the JavaScript console (Help → Show Developer Tools → Console), and only after switching to another connection and back.

## Steps to Reproduce

1. Configure a connection that will time out (e.g. wrong host or SSL mismatch)
2. Click the "Test" button
3. Observe: no spinner, no error, no response in the UI
4. Switch to a different connection, then switch back
5. Observe: the JS console now shows the timeout error

## Expected Behavior

The Test button should show a loading indicator while connecting and display the error message in the UI when it fails.

## Environment

- macOS (reported by user)
- Connection type: MySQL

## Additional Context

Reported by a user comparing Beekeeper to Sequel Ace. The workaround is checking the JS console via Help → Developer Tools.

deeferentleeg:
Closes #4456

## What

When clicking the **Test** button on the connection screen and the connection times out or fails, the UI showed no loading indicator and the toast notification used a generic "Error establishing a connection" message without the actual error detail.

## Changes

In `apps/studio/src/components/ConnectionInterface.vue`:

- **Loading spinner on the Test button**: added the existing `LoadingSpinner` component (already used in `LoadingSSOModal.vue`), shown via `v-if="testing"`. Gives immediate feedback that a test is in progress, even when the connection takes a long time to time out.
- **Detailed error in the toast**: changed the `testConnection()` catch block to include the actual error message (`Error establishing a connection: <error>`) instead of the generic hardcoded string. The `ErrorAlert` below the button already shows the full error; the toast is now useful too.

The Connect path is unchanged.

## Testing

1. Configure a connection that will time out (wrong host / SSL mismatch)
2. Click Test
3. Spinner appears on the Test button while testing
4. On failure, the toast includes the actual error (e.g. `Error establishing a connection: connect ETIMEDOUT 10.0.0.1:3306`)
5. The ErrorAlert below the button still shows the full error with helpful links

### DIFF
--- a/apps/studio/src/components/ConnectionInterface.vue
+++ b/apps/studio/src/components/ConnectionInterface.vue
@@ -213,7 +213,8 @@
                       type="button"
                       @click.prevent="testConnection"
                     >
-                      Test
+                      <loading-spinner v-if="testing" :size="12" />
+                      <template v-else>Test</template>
                     </button>
                     <button
                       :disabled="testing || connecting"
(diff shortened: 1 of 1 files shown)


not-night-but:
@deeferentleeg Looks like there are some style issues that need to be ironed out: 
image

Issue #4492, opened 2026-07-14
BUG: Autocomplete freezes and crashes app on large schemas


JeanPaiva42:
## Description

On large database schemas, the SQL autocomplete/typeahead freezes the app and eventually crashes it. Reported on version 5.9.1.

## Expected Behavior

Autocomplete should handle large schemas gracefully — either by limiting the completion set, lazy-loading, or providing a way to disable it.

## Workaround Requested

The user wants to disable autocomplete entirely as a workaround, but no such option exists in the UI or config file.

## Notes

The internal `disableSchemaCompletion` and `disableKeywordCompletion` flags exist in the SQL text editor extension config (`apps/ui-kit/lib/components/sql-text-editor/extensions/customSql.ts`) but are not exposed to users through any UI setting or `user.config.ini` key.

## Related

Exposing these flags (or adding a simpler `[ui.queryEditor] disableAutocomplete = true` config key) would give users an immediate workaround while the performance issue is investigated.

rathboma:
How large? We need more details!

JeanPaiva42:
Reproduction details from a user:

- PostgreSQL with ~20,000 tables in the public schema (~half generated by pg_partman)
- All 20k tables are passed to the CodeMirror completion engine with no limit or lazy-loading (see `entities()` computed in `TabQueryEditor.vue`)
- Symptom: hangs for ~20 seconds, then the renderer window goes blank. DevTools shows 'page disconnected' — meaning the renderer process crashes (likely OOM)
- macOS, Beekeeper 5.9.1

The completion engine pre-loads all schema entities eagerly. At this scale it exhausts renderer memory. Fix needs either: entity count cap before passing to completion, lazy/on-demand schema loading, or exposing `disableSchemaCompletion` flag to users.

PR #4509, opened 2026-07-18
Add settings toggle to disable SQL autocomplete as workaround for #4492

deeferentleeg:
## Summary

This PR exposes the existing `disableSchemaCompletion` and `disableKeywordCompletion` flags — already defined in the SQL text editor config type (`apps/ui-kit/lib/components/sql-text-editor/extensions/customSql.ts`) — as a user-facing settings toggle. This gives users an immediate workaround for the autocomplete freeze/crash on large schemas while the deeper performance issue is investigated.

**This is a workaround, not a root-cause fix for the underlying performance problem.** As @rathboma noted in the issue, more details are needed about the schema sizes involved. This PR simply gives affected users a way to keep working.

Addresses #4492

## What changed

The `disableSchemaCompletion` and `disableKeywordCompletion` flags already existed in the `SQLConfig` type extension in `customSql.ts`, but were **not actually consumed** by the vendored `sql()` function. This PR wires them through end-to-end:

### ui-kit (`@beekeeperstudio/ui-kit`)
1. **`vendor/@codemirror/lang-sql/src/sql.ts`** — The vendored `sql()` function now checks `disableSchemaCompletion` and `disableKeywordCompletion`. When either is `true`, the corresponding completion source (schema-based or keyword-based) is omitted entirely from the `LanguageSupport` extensions.
2. **`sql-text-editor/props.ts`** — Added a `disableAutocomplete` boolean prop (default `false`).
3. **`sql-text-editor/SqlTextEditor.vue`** — Passes `disableAutocomplete` through to the `SqlTextEditor` constructor as both `disableSchemaCompletion` and `disableKeywordCompletion`.

### studio (desktop app)
4. **`TabQueryEditor.vue`** — Added a `disableSqlAutocomplete` computed property that reads the `disableSqlAutocomplete` user setting from the Vuex settings store, and passes it as `:disable-autocomplete` to the editor component.
5. **`QueryEditorStatusBar.vue`** — Added a "Disable SQL Autocomplete" toggle in the editor settings dropdown (alongside "Editor keymap" and "Wrap Text"). Toggling it saves the setting via `settings/save`.
6. **Migration** — Added `20260714_add_disable_sql_autocomplete_setting.js` to seed the `disableSqlAutocomplete` boolean user setting (default `false`). Registered in `migration/index.js`.

## How to use

1. Open a query tab
2. Click the **⚙️ settings** button in the query editor status bar
3. Toggle **"Disable SQL Autocomplete"**
4. Autocomplete (both schema-based and keyword-based) is now fully disabled — no freeze, no crash

The setting persists across sessions via the existing user settings store.

## Notes

- The flags were already defined in the type system (`SQLConfig` in `customSql.ts`) and already used in the test suite (`completions.spec.ts`), so this PR completes the plumbing rather than introducing new concepts.
- The vendored `sql()` uses `(config as any).disableSchemaCompletion` because the upstream `@codemirror/lang-sql` `SQLConfig` type doesn't include these custom fields — they're added by Beekeeper's type extension in `customSql.ts`. Using `as any` avoids importing the custom type into the vendored file.
- This disables **all** autocomplete (schema + keywords) via a single toggle. If more granularity is desired later (e.g. disable schema completion but keep keyword completion), the underlying flags already support that — only the UI would need to change.

---

🤖 This PR was created with AI assistance. The implementation was generated programmatically and follows existing patterns in the codebase (settings store, migration system, editor props).

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -72,6 +72,7 @@
         :columns-getter="columnsGetter"
         :default-schema="defaultSchema"
         :language-id="languageIdForDialect"
+        :disable-autocomplete="disableSqlAutocomplete"
         :clipboard="$native.clipboard"
         :replace-extensions="replaceExtensions"
         :context-menu-items="editorContextMenu"
@@ -718,6 +719,9 @@ import { KeybindingPath } from '@/common/bksConfig/BksConfigProvider'
       },
       editorComponent() {
         return this.connectionType === 'surrealdb' ? SurrealTextEditor : SqlTextEditor;
+      },
+      disableSqlAutocomplete() {
+        return this.settings?.disableSqlAutocomplete?.value ?? false;
       },
       enabled() {
         return !this.dialectData?.disabledFeatures?.queryEditor;
--- a/apps/studio/src/components/editor/QueryEditorStatusBar.vue
+++ b/apps/studio/src/components/editor/QueryEditorStatusBar.vue
@@ -223,6 +223,15 @@
             Wrap Text
           </x-label>
         </x-menuitem>
+        <x-menuitem
+          togglable
+          :toggled="disableSqlAutocomplete"
+          @click.prevent="toggleDisableSqlAutocomplete"
+        >
+          <x-label class="flex-between">
+            Disable SQL Autocomplete
+          </x-label>
+        </x-menuitem>
       </x-menu>
     </x-button>
   </statusbar>
@@ -307,6 +316,9 @@ export default {
         this.$store.dispatch('settings/save', { key: 'hideResultsDropdown', value })
       }
     },
+    disableSqlAutocomplete() {
+      return this.settings?.disableSqlAutocomplete?.value ?? false
+    },
     rowCount() {
       return this.result && this.result.rows ? this.result.rows.length : 0
     },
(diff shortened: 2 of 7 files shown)


not-night-but:
This feels like it might be better suited as a key in the ini file rather than a user setting. @azmy60 Thoughts?

azmy60 (changes requested):
Overall, I think the code looks correct! But as @not-night-but said, this should be a new config key in the `default.config.ini` file.

And from me, I have two requests:

1. I don't know if you actually need to disable **all** completions. If the issue was caused by a large schema, then we might just want to disable the `schemaCompletion`? I don't think we need to disable the `keywordCompletionSource` too, right? Regardless, It probably makes sense to be able to configure it this way (from the config file), for example:
```ini
; Controls which sources are used for query editor autocomplete.
; Possible values: all, schema-only, keyword-only, none
[ui.queryEditor]
autocomplete = all
```
2. This is optional, but it'd be awesome if the `disableSchemaCompletion` is a state, and not just a static flag. This allows CodeMirror to update the completion in real time. Right now, it seems like we'll need to rebuild the whole text editor manually in order to see the effect of switching the flag from `false` to `true` or vice versa. We've done this in many parts, e.g. toggling vim mode.

Thanks!

not-night-but:
Superseded by #4688

PR #4511, opened 2026-07-19
Fix MongoDB Decimal128 rendering as raw bytes in table/JSON view

### LINKED ISSUE #4258: BUG: MongoDB Decimal128 view
Describe the bug
When viewing MongoDB documents in JSON view, fields of type ‎⁠Decimal⁠ are rendered as raw bytes instead of a human‑readable value.

Example:"price": {
  "bytes": {
    "0": 176,
    "1": 4,
    "2": 0,
    "3": 0,
    "4": 0,
    "5": 0,
    "6": 0,
    "7": 0,
    "8": 0,
    "9": 0,
    "10": 0,
    "11": 0,
    "12": 0,
    "13": 0,
    "14": 60,
    "15": 48
  }
}

This looks similar to how Mongo ObjectID is shown as bits in issue #3024, but for ‎⁠Decimal⁠ values.

To Reproduce

1. Connect to a MongoDB database.
2. Open a collection that has a ‎⁠Decimal⁠ field (e.g. ‎⁠price⁠).
3. Open JSON view (right‑click on a record → “See Details” or use the JSON viewer panel).
4. Observe that the ‎⁠Decimal⁠ field is rendered as a ‎⁠bytes⁠ object instead of a readable decimal value.

Expected behavior
‎⁠Decimal⁠ (e.g. ‎⁠Decimal128⁠) fields should be displayed as a human‑readable decimal number (string/number), not as a bytes object.

Screenshots / Gifs
Add a screenshot of the JSON viewer showing the ‎⁠Decimal⁠ field rendered as bytes.

Version Information (please complete the following information):

- OS: MacOS 26.3.1
- App Version: Version 5.7.2 (5.7.2)
- Database type and version: MongoDB

Additional context
Behavior is analogous to the Mongo ObjectID JSON view issue described in [#3024](https://github.com/beekeeper-studio/beekeeper-studio/issues/3024), but affects ‎⁠Decimal⁠ fields such as ‎⁠price⁠.

gaurav0107:
## Problem

When viewing MongoDB documents that contain a `Decimal128` field, the value is rendered as a raw `{ bytes: { 0: 176, 1: 4, ... } }` object instead of a human-readable decimal number, both in the table view and the JSON detail view. This mirrors the earlier ObjectId issue (#3024), but for `Decimal128`.

Fixes #4258.

## Root cause

The MongoDB client serializes row values through a per-field transcoder pipeline (`BasicDatabaseClient.serializeQueryResult`), which selects a transcoder by each field's `bksType`. `MongoDBClient.parseQueryResultColumns` only tags a column as `OBJECTID` when the value is an `ObjectId`; every other BSON type — including `Decimal128` — falls through to `UNKNOWN`. With no matching transcoder, the raw `Decimal128` object (which carries an internal `bytes` buffer) is passed straight to the view and JSON-stringified as `{ bytes: {...} }`.

## Fix

Add a `Decimal128` transcoder alongside the existing ObjectId one, following the same pattern:

- `MongoDBDecimal128Transcoder` (in `src/lib/db/serialization/transcoders.ts`) serializes a `Decimal128` to its readable decimal string via `value.toString()` (e.g. `"12.34"`).
- A new `DECIMAL` member on `BksFieldType`.
- `MongoDBClient.parseQueryResultColumns` now detects `instanceof Decimal128` and tags the column `DECIMAL`, and the client registers the new transcoder.

`deserializeCheckByValue` deliberately returns `false`: `deserializeValue` picks a transcoder from the value alone with no field context, so claiming every string would hijack unrelated edited-string write-backs. The transcoder is therefore scoped to the read/display path, which is what the issue reports.

## Testing

Added `apps/studio/tests/unit/lib/db/serialization/transcoders.spec.ts` — pure unit tests for the new transcoder (no live database):

- serializes `Decimal128` to its readable decimal string, preserving precision for both high-precision and large-magnitude values
- returns a non-`Decimal128` value unchanged
- `serializeCheckByField` matches only `DECIMAL` fields
- `deserialize` round-trips a decimal string back to an equal `Decimal128`
- `deserializeCheckByValue` never claims a value (documents the display-only write-back scope)

### DIFF
--- a/apps/studio/src-commercial/backend/lib/db/clients/mongodb.ts
+++ b/apps/studio/src-commercial/backend/lib/db/clients/mongodb.ts
@@ -1,12 +1,12 @@
 import { IDbConnectionServer } from "@/lib/db/backendTypes";
 import { BasicDatabaseClient, ExecutionContext, QueryLogOptions } from "@/lib/db/clients/BasicDatabaseClient";
 import { DatabaseElement, IDbConnectionDatabase } from "@/lib/db/types";
-import { AggregationCursor, Collection, Db, Document, MongoClient, ObjectId } from 'mongodb';
+import { AggregationCursor, Collection, Db, Decimal128, Document, MongoClient, ObjectId } from 'mongodb';
 import rawLog from '@bksLogger';
 import { BksField, BksFieldType, CancelableQuery, ExtendedTableColumn, NgQueryResult, OrderBy, PrimaryKeyColumn, Routine, SchemaFilterOptions, StreamResults, SupportedFeatures, TableChanges, TableColumn, TableDelete, TableFilter, TableIndex, TableInsert, TableOrView, TableProperties, TableResult, TableTrigger, TableUpdate, TableUpdateResult } from "@/lib/db/models";
 import { CreateTableSpec, IndexAlterations, TableKey } from "@/shared/lib/dialects/models";
 import _ from 'lodash';
-import { MongoDBObjectIdTranscoder } from "@/lib/db/serialization/transcoders";
+import { MongoDBDecimal128Transcoder, MongoDBObjectIdTranscoder } from "@/lib/db/serialization/transcoders";
 import { ElectronRuntime as MongoRuntime } from '@mongosh/browser-runtime-electron';
 import { NodeDriverServiceProvider } from '@mongosh/service-provider-node-driver';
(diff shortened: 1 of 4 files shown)


not-night-but:
@gaurav0107 Thanks for the contribution! Any chance we could add a test to the `mongodb.spec.ts` integration tests? For confirmation and regression purposes

PR #4524, opened 2026-07-22
Add "select top 100 rows:

### LINKED ISSUE #4515: FEAT: Add context menu shortcut to generate basic SELECT query

Every time I want to query a table, I have to manually type out the boilerplate:
```sql
select *
from ...
limit ...
```

Add a "Select 100 rows" option to the right-click context menu on a table (see screenshot below). Selecting it would automatically generate:
```sql
select *
from ...
limit ...
```

This would save time on writing the basic query for quick data exploration. ClickHouse offers this feature, and I've found it really handy, I imagine other users would too.

An alternative could be a settings option to customize the behavior of double clicking a table. Currently, double clicking opens the "view data" screen. it would be nice to optionally have it generate the query above instead.

**Additional context**

Image

manos02:
Closes: https://github.com/beekeeper-studio/beekeeper-studio/issues/4515

- Added select top 100 rows option when right clicking on a table

image


not-night-but:
Oh that is a satisfying change. I was worried you were going to have to add a new method for all of the clients. The only thing I would say is I think we should probably move it to be below the SQL: Create entry in the context menu just to keep things organized, and maybe have it be called `SQL: Select Top`

manos02:
> Oh that is a satisfying change. I was worried you were going to have to add a new method for all of the clients. The only thing I would say is I think we should probably move it to be below the SQL: Create entry in the context menu just to keep things organized, and maybe have it be called `SQL: Select Top`

@not-night-but  Done!

rathboma (approved):

rathboma:
Awesome, thanks!

PR #4532, opened 2026-07-23
fix: hide selected dropdown text while searching

### LINKED ISSUE #4520: BUG: Selected database text overlaps with new input in dropdown search
**Describe the bug**
When a database is selected in the dropdown and the user tries to type the name of another database, the previously selected text remains visible behind the input, causing it to overlap with the text being typed.

**To Reproduce**
1. Select a database using the dropdown.
2. Click the dropdown and start typing another database name.
3. Notice that the new text overlaps with the previously selected text.

**Expected behavior**
While typing, only the text being entered should be visible. The previously selected value should not remain displayed behind it.

**Screenshots / Gifs**
Image

AnandaBintang:
Closes #4520

## Summary

Fixes the database dropdown text overlap when a selected database exists and the user starts typing a new search value.

The selected vue-select label was still rendered behind the search input while typing. This hides the selected label only during vue-select's `vs--searching` state, so the typed search text is the only visible text.

## Changes

- Hide `.vs__selected` while `.dropdown-search` is searching.
- Apply the same fix to the shared UI kit dropdown styles.

https://github.com/user-attachments/assets/5e276b5b-24a2-40b3-a124-28aa9acf9a79


azmy60 (changes requested):
Hey @AnandaBintang , thanks for the PR! This works really well, but could we remove the test please? This is fine without test, and I think it'll just create noise.

AnandaBintang:
> Hey @AnandaBintang , thanks for the PR! This works really well, but could we remove the test please? This is fine without test, and I think it'll just create noise.

Sure, thanks for the feedback!

azmy60 (approved):
Awesome, thanks!

PR #4536, opened 2026-07-24
feat: add data type colors for TableTable and ResultTable

### LINKED ISSUE #3334: FEAT: Add colours based of data type (like Heidi)
Would like a feature similar to Heidi where different datatypes are displayed with different colours. For example

Image

Functionality to either choose your own colours based off data type or default colours would do as well.

AnandaBintang:
Closes #3334

## Summary

Adds customizable result-grid colors by data type.

- Adds default light and dark palettes for text, numbers, dates, booleans, binary values, and other types
- Adds the **Data type colors** dialog under **View**
- Supports enabling/disabling datatype coloring and editing colors per palette
- Selects the matching palette for Light, Dark, Solarized, Solarized Dark, and System themes

Screenshot 2026-07-24 at 21 10 35

https://github.com/user-attachments/assets/7c843032-fc52-4f54-b62a-94fe716dde23


apps/studio/src/components/common/modals/DataTypeColorsModal.vue
```diff
@@ -0,0 +1,244 @@
...
+              <div v-for="family in families" :key="family" class="color-row">
+                <label :for="`data-type-color-${theme}-${family}`">{{ labels[family] }}</label>
+                <input
+                  :id="`data-type-color-${theme}-${family}`"
+                  class="color-swatch"
+                  type="color"
+                  :value="settings[theme][family]"
+                  :aria-label="`${labels[family]} color`"
+                  @input="setColor(family, $event.target.value)"
+                >
```
not-night-but, line 38: The styling for this doesn't seem to properly match  the active theme

Image

apps/studio/src/components/common/modals/DataTypeColorsModal.vue
```diff
@@ -0,0 +1,244 @@
...
+                  :aria-label="`${labels[family]} color hex value`"
+                  @change="setColor(family, $event.target.value)"
+                >
+              </div>
+            </div>
+          </fieldset>
+        </div>
+        <div class="vue-dialog-buttons">
+          <button type="button" class="btn btn-flat" @click="resetPalette">
+            Reset this palette
```
not-night-but, line 53: I'd just say `Reset Palette`

apps/studio/src/components/common/modals/DataTypeColorsModal.vue
```diff
@@ -0,0 +1,244 @@
...
+  normalizeDataTypeColorSettings,
+} from '@/common/dataTypeColors'
+
+const families: DataTypeColorFamily[] = ['text', 'number', 'dateTime', 'boolean', 'binary', 'other']
+
+export default Vue.extend({
+  data() {
+    return {
+      modalName: 'data-type-colors-modal',
+      settings: normalizeDataTypeColorSettings(undefined) as DataTypeColorSettings,
```
not-night-but, line 85: This just ends up as the default settings anyways, shouldn't we just use that?

apps/studio/src/components/common/modals/DataTypeColorsModal.vue
```diff
@@ -0,0 +1,244 @@
...
+      if (this.themeValue === 'system') {
+        this.systemTheme.addEventListener('change', this.handleSystemThemeChange)
+      }
+    },
+    stopListeningForSystemThemeChanges() {
+      this.systemTheme.removeEventListener('change', this.handleSystemThemeChange)
+    },
+    handleSystemThemeChange() {
+      if (this.isOpen && this.themeValue === 'system') this.theme = this.activePaletteTheme()
+    },
```
not-night-but, line 141: Shouldn't the theme change be handled by the media queries in `app.scss`? I think realistically we should only need to fetch the system theme when the modal opens, we don't need to listen for changes

apps/studio/src/components/editor/ResultTable.vue
```diff
@@ -436,6 +437,7 @@ import { stringToTypedArray } from '@/common/utils'
             }),
             { separator: true },
             ...commonColumnMenu,
+            this.dataTypeColorsMenuItem(),
```
not-night-but, line 440: Yeah I think this should be in the cog menu in the statusbar, it feels kinda out of place in the column header

Image

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -888,6 +888,7 @@ export default Vue.extend({
           }),
           { separator: true },
           ...commonColumnMenu,
+          this.dataTypeColorsMenuItem(),
```
not-night-but, line 891: Same as above, I think we should move this to the settings cog in the statusbar

apps/studio/src/App.vue
```diff
@@ -135,6 +138,38 @@ export default Vue.extend({
     }),
     editorFontSize() {
       return this.$store.state.settings?.settings?.editorFontSize?.value || 14
+    },
+    dataTypeColors() {
+      return normalizeDataTypeColorSettings(this.$store.state.settings?.settings?.dataTypeColors?.value)
```
not-night-but, line 143: Probably worth adding a getter to the settings store for this like we do for themeValue

not-night-but (changes requested):
@AnandaBintang Cool feature! I think a lot of people will get some great use out of this.

I'm actually not sure this really belongs in the View toolbar, I think we should instead add an entry to the menu in the status bar in the query editor and the table table view:

Image
Image

That's just a thought though. If we do keep this in the View menu, then we need a mapping in `ClientMenuActionHandler.ts` as well, as this currently will only work on Mac.

I do find the theme toggle in the modal might be mildly confusing, I think it would be fine to just display the palette for the current theme? Or a better way to communicate that the palette is for that theme, as right now it just seems like an option. I think maybe having it be a dropdown might be better. 

Image

Well done though! I think this will be fantastic to have in the app, just some minor things to iron out. I'll also ping @azmy60 as he is the more opinionated one when it comes to styles :)

apps/studio/src/mixins/data_mutators.ts
```diff
@@ -69,11 +79,20 @@ export default {
...
       let cellValue = cell.getValue()
+      const cellElement = cell.getElement()
+
+      if (params.dataTypeColors) {
+        cellElement.classList.remove(...Object.values(dataTypeColorClasses))
+      }
+
+      if (params.dataTypeColors && !_.isNil(cellValue) && !(_.isString(cellValue) && _.isEmpty(cellValue))) {
+        cellElement.classList.add(dataTypeColorClasses[dataTypeColorFamily(params.dataType, cellValue)])
+      }
```
azmy60, line 95: I'm not sure why we need to add logics for this. Why not just add the new class all the time?

```suggestion

     classNames.push(`data-type-${params.dataType}`)
```

azmy60 (changes requested):

azmy60 (changes requested):
@AnandaBintang I think this is an amazing PR for improving the UX overall! But the only thing that we don't want now is the ability to customize / override the colors.

I don't think it will work well with our plan with theming the app with custom css or some type of media in the future. Keeping it out for now will help us avoid conflicts later.

azmy60:
I just updated the PR by making it a lot more simpler so, no more customization, it uses colors from the text editor, and users are only allowed to enable/disable this feature via configuration file. 
image

apps/studio/src/common/dataTypeColors.ts
```diff
@@ -0,0 +1,23 @@
+import { isDateDataType, isNumericDataType, normalizeDataType } from '@/common/utils'
+
+export type DataTypeColorFamily = 'string' | 'number' | 'dateTime' | 'boolean' | 'binary' | 'other'
+
+export function getDataTypeColorFamily(dataType?: string, value?: unknown): DataTypeColorFamily {
```
rathboma: I don't like this function.

We should be using the `bksType` and expanding that to new types if we don't have them. This is brittle and feels like it will break easily, and is duplicating work that we also do elsewhere (formatters, editors, bksType, etc)

rathboma (changes requested):
Fine except for the type parsing which is unmaintainable and will certainly drift.

apps/studio/src/common/dataTypeColors.ts
azmy60: @rathboma `bksType` doesn't recognize basic primitives like string, number, etc.. Should we imlpement the basic support in a different PR?

azmy60:
@not-night-but @rathboma This is ready for another review. `bksType` only support `BINARY` type so this PR won't make that much of a difference. I also made a new PR #4693 for extending the type.

PR #4537, opened 2026-07-24
feat(sqlite): support SQLCipher-encrypted databases

### LINKED ISSUE #625: QUESTION: SQLite with SQLCipher encryption
**Details:**
I'd like to open an encrypted SQLite database. From the popup dialog when I select the SQLite connection type, I only see the database file input and the connection name input. Is there any workaround to specify the encryption settings and the password?

Operating System: macOS Catalina (10.15.7)
App Version [help -> about]: 1.10.2 (1.10.2)

N/A

guilhermeprokisch:
Closes #625.

Adds support for opening and working with encrypted SQLite databases (SQLCipher and compatible schemes), one of the longest-standing feature requests.

image

## Approach

Swaps `better-sqlite3` for the drop-in [`better-sqlite3-multiple-ciphers`](https://github.com/m4heshd/better-sqlite3-multiple-ciphers) build (same API, actively maintained, recommended in #625) and adds a password + cipher UI to the SQLite connection form.

## Changes

- **Driver**: alias `better-sqlite3` → `better-sqlite3-multiple-ciphers`. Every existing import (client, cursor, knex, TypeORM app DB) keeps working unchanged.
- **Client**: apply `PRAGMA cipher` / `legacy` / `key` when a connection password is set, in both `SqliteClient` and the streaming `SqliteCursor`. Databases created from an encrypted connection are encrypted with the same key.
- **SQLCipher compatibility**: the multiple-ciphers `sqlcipher` scheme defaults to `legacy=0`, which is **not** binary-compatible with databases produced by the real SQLCipher library. Default to `legacy=4` so genuine SQLCipher files open, and expose a compatibility selector (v4–v1) for older databases.
- **UI**: new "Encryption" section on the SQLite form — cipher dropdown (SQLCipher, ChaCha20, AES-256/128-CBC, RC4), password field, and a SQLCipher compatibility selector. The password uses the existing encrypted `password` column and respects "Save Passwords".
- **Persistence**: new `sqliteOptions.{cipher,cipherCompatibility}` on saved/used connections via migration, plumbed through the connection provider.
- **Errors**: clear messages for encrypted-without-password and wrong-key/cipher connects instead of the cryptic "file is not a database".

## Testing

Integration tests cover a genuine SQLCipher v4 file (connect, read, stream), a SQLCipher v3 file requiring the compatibility selector, and the failure paths (no password, wrong password). Existing SQLite integration suite passes against the new driver (unencrypted behaviour unchanged). Both esbuild and vite production builds pass.


Waldenesque:
I don't have an opinion either way on supporting encrypted databases, however if the decision is made to accept this change, it's probably better to wait a little longer before moving ahead.

Recently `better-sqlite3` was refactored to be a Node-API addon rather than directly interfacing with Node internals. This is a fantastic change over the long term, since it will make ongoing maintenance simpler and easier, and it eliminates the need for lifecycle scripts which are disabled by default starting with npm 12. So this is great news, however in the short term it is a significant transition.

The author of `better-sqlite3-multiple-ciphers` @m4heshd is aware of all this, however there has not yet been a new version published since the Node-API refactor. It would probably make sense to let `better-sqlite3-multiple-ciphers` update to the latest `better-sqlite3`, and then wait a modest period of time to ensure any potential bugs or edge cases can be handled before making this switch in Beekeeper Studio.

Thank you.

m4heshd:
@Waldenesque Just released [`v13.0.0-beta.0`](https://github.com/m4heshd/better-sqlite3-multiple-ciphers/releases/tag/v13.0.0-beta.0). It would be very helpful if the community, including you, could test it. And yes, now that it relies on a stable runtime API, it would be so easy to use BS3MC as a replacement.

m4heshd:
FYI, as no issues have been reported with the beta, stable release [`v13.0.3`](https://github.com/m4heshd/better-sqlite3-multiple-ciphers/releases/tag/v13.0.3) of BS3MC has been published. I support this PR.

Waldenesque:
This is great to hear. I should clarify, I'd checked this repo to see if folks were aware of the new `better-sqlite3`, and noticed this PR. I'm not familiar with typical practices here, and also I didn't realize `better-sqlite3-multiple-ciphers` was so close to shipping an update, so I worded my earlier comment in a generic overview style.

The two primary contributors here know best when things are good to move forward. Thanks to both of you for doing this work!

guilhermeprokisch:
Bumped the dependency to `better-sqlite3-multiple-ciphers@^13.0.3` (the stable Node-API build) in the latest commit — thanks @m4heshd for turning that around so quickly, and @Waldenesque for flagging the Node-API transition.

The full SQLite integration suite passes against v13, including the SQLCipher v4 and v3 encryption cases (connect / read / stream) and the failure paths (no password, wrong password, wrong compatibility revision). Both the esbuild and vite production builds are green, and existing unencrypted SQLite behaviour is unchanged.

Ready for review whenever a maintainer has a chance.

PR #4538, opened 2026-07-24
fix: fix inability to discover or launch az cli on windows

jasonnutter:
Today on Windows (Beekeeper v5.9.2), it appears that discovering and using the `az` CLI for Azure CLI Authentication are both broken. This is due to a couple of issues:

1. On Windows, environment variables are case-insensitive, and as a result, the `PATH` variable is at `process.env.Path`, not `process.env.PATH` (which the code currently looks for).
2. The `az` cli is a `cmd` file, and since Node patched [CVE-2024-27980](https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2#command-injection-via-args-parameter-of-child_processspawn-without-shell-option-enabled-on-windows-cve-2024-27980---high), using `child_process.spawn` with `{ shell: false}` for `.cmd` files results in `EINVAL` (as described in the aforementioned security release, and which you see in the app if you manually find the `az.cmd` file and run test/connect, as shown in the screenshot below).

This PR does the following:

1. Use `process.env.Path` instead of `process.env.PATH`, if set (i.e. on Windows).
2. Use `child_process.exec` instead of `process.spawn` on Windows (as [recommended](https://nodejs.org/api/child_process.html#spawning-bat-and-cmd-files-on-windows) in the Node docs) for both discovering and executing the `az` CLI.

I've manually verified these fixes on Windows and also verified working functionality on Mac is not broken (haven't testing Linux).

## Screenshots

1. `az` CLI not found on Windows even though installed in default location and in PATH:
Screenshot 2026-07-24 135930

2. `EINVAL` error when manually finding `az.cmd` file and testing the connection:
Screenshot 2026-07-23 091619

3. After fixes in this PR, app is able to discover and run `az` CLI on Windows for Azure CLI Authentication.
image

Question: I'm not sure if you'd like to see tests for this? Let me know.

Thanks!

### DIFF
--- a/apps/studio/src-commercial/backend/handlers/cliHandlers.ts
+++ b/apps/studio/src-commercial/backend/handlers/cliHandlers.ts
@@ -51,7 +51,7 @@ export const CliHandlers: ICliHandlers = {
 
       proc.on('close', (code) => {
         if (code === 0) {
-          const result = stdout.trim().split('\n')[0]; // pick first match
+          const result = stdout.trim().split(/\r?\n/)[0]; // pick first match
           resolve(result);
         } else {
           reject(`cli/which failed (code ${code})\nSTDERR: ${stderr}\nSTDOUT: ${stdout}`);
--- a/apps/studio/src-commercial/backend/handlers/cliHandlers.ts
+++ b/apps/studio/src-commercial/backend/handlers/cliHandlers.ts
@@ -1,4 +1,4 @@
-import { exec, spawn } from "child_process";
+import { spawn } from "child_process";
 import path from "path";
 import platformInfo from "@/common/platform_info";
 
@@ -25,17 +25,14 @@ export const CliHandlers: ICliHandlers = {
 
     // Pass the environment explicitly so we can prepend Homebrew bin dirs.
     const env = { ...process.env };
-    const newPath = [...extraToolSearchDirs, env.Path].filter(Boolean).join(path.delimiter);
-    if (env.Path) {
-      // Windows is case-insensitive, so PATH is usually capitalized as Path
-      env.Path = newPath;
+    if (env.Path && platformInfo.isWindows) {
+      env.Path = [...extraToolSearchDirs, env.Path].filter(Boolean).join(path.delimiter);
     } else {
-      env.PATH = newPath;
+      env.PATH = [...extraToolSearchDirs, env.PATH].filter(Boolean).join(path.delimiter);
     }
 
     return new Promise((resolve, reject) => {
-      // Use exec on windows as spawn won't launch cmd files (e.g. azure cli) on windows with shell: false, see: CVE-2024-27980
-      const proc = platformInfo.isWindows ? exec(`${command} ${toolName}`, { env }) : spawn(command, [toolName], { shell: false, env });
+      const proc = spawn(command, [toolName], { shell: false, env });
 
       let stdout = '';
       let stderr = '';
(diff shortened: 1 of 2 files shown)


apps/studio/src-commercial/backend/handlers/cliHandlers.ts
```diff
@@ -25,19 +25,26 @@ export const CliHandlers: ICliHandlers = {
 
     // Pass the environment explicitly so we can prepend Homebrew bin dirs.
     const env = { ...process.env };
-    env.PATH = [...extraToolSearchDirs, env.PATH].filter(Boolean).join(path.delimiter);
+    const newPath = [...extraToolSearchDirs, env.Path].filter(Boolean).join(path.delimiter);
+    if (env.Path) {
```
rathboma: There's a chance that there is both a `Path` and a `PATH` - Mac and Linux are case sensitive, so someone could also have a `Path` set.

I think we want to also do a platform check:
```
if (env.Path && platformInfo.isWindows) {
  // set Path
}

// always set PATH
```

apps/studio/src-commercial/backend/handlers/cliHandlers.ts
```diff
@@ -25,19 +25,26 @@ export const CliHandlers: ICliHandlers = {
...
+      // Windows is case-insensitive, so PATH is usually capitalized as Path
+      env.Path = newPath;
+    } else {
+      env.PATH = newPath;
+    }
 
     return new Promise((resolve, reject) => {
-      const proc = spawn(command, [toolName], { shell: false, env });
+      // Use exec on windows as spawn won't launch cmd files (e.g. azure cli) on windows with shell: false, see: CVE-2024-27980
+      const proc = platformInfo.isWindows ? exec(`${command} ${toolName}`, { env }) : spawn(command, [toolName], { shell: false, env });
```
rathboma: this exec command creation is unescaped and so would allow a potential RCE depending on the payload of the toolname.

I think we should use `execFile` instead because it supports passing args as an array, similar to spawn
--> https://nodejs.org/api/child_process.html#child_processexecfilefile-args-options-callback

rathboma: We would also have to capture both stdout and stderr, this should work the same as `spawn` I believe.

apps/studio/src/lib/db/authentication/azure.ts
```diff
@@ -178,23 +179,25 @@ export class AzureAuthService {
...
         'account',
         'get-access-token',
         '--resource',
         BksConfig.azure.azSQLLoginScope,
         '--output',
         'json'
-      ], { shell: false });
+      ];
+      // Use exec on windows as spawn won't launch cmd files (e.g. azure cli) on windows with shell: false, see: CVE-2024-27980
+      const proc = platformInfo.isWindows ? exec(`"${options.cliPath}" ${commandArgs.join(' ')}`) : spawn(options.cliPath, commandArgs, { shell: false });
```
rathboma: Same here, can we use `execFile` rather than manually joining args and cli strings?

rathboma (dismissed):
AMAZING find. Thank you so much for fixing this for us. Great work and a knowledge of Windows that I don't have.

I've added a couple of comments about using `execFile` instead of exec because it doesn't require us to manually build the execution string, but I think if we have that then we're good to go.

Also we should check for Windows with the `Path` vs `PATH` check.

Again - this is amazing, thanks so much.

azmy60 (approved):

PR #4539, opened 2026-07-24
Fix overlapping context menus in the data grid

### LINKED ISSUE #4517: BUG: Right-click in editing cell opens two overlapping context menus
### Describe the bug
When editing a cell in the data grid, right-clicking inside that cell opens two context menus at the same time (native edit menu + Beekeeper menu), and they overlap.

### To Reproduce
1. Open table data in editable mode.
2. Start editing a specific cell (cursor visible in cell editor).
3. Right-click inside the same cell.

### Expected behavior
Only one context menu should open (or menus should be handled in a non-overlapping way).

### Actual behavior
Two context menus open and overlap, making actions hard to use.

### Environment
- Beekeeper Studio: 5.9.2
- OS: Windows 11

### Screenshot

Image

manos02:
- Added `event.defaultPrevented` guard to the root listener. 
image

Closes: https://github.com/beekeeper-studio/beekeeper-studio/issues/4517

### DIFF
--- a/apps/studio/src/mixins/assignContextMenuToAllInputs.ts
+++ b/apps/studio/src/mixins/assignContextMenuToAllInputs.ts
@@ -21,6 +21,10 @@ export const assignContextMenuToAllInputs: ComponentOptions<Vue> = {
 
   methods: {
     ctxMenu_showContextMenu(event: MouseEvent) {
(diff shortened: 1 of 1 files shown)


rathboma (approved):

rathboma:
thanks!

PR #4542, opened 2026-07-26
Fix macOS cold-start open-file/open-url ignoring the URL

### LINKED ISSUE #4214: BUG: Cold start the app by double clicking sqlite file or via url schema stays in New Connection
**Describe the bug**
BUG: Cold start the app by double clicking sqlite file or via url schema stays in New Connection, do it a second time while the app is running then it works as intended.

**To Reproduce**
1. Kill the app
2. Double clicking sqlite file

**Expected behavior**
Launching the app with a url schema or sqlite file works immediately

 - OS: Mac
 - App Version 5.7.2 (5.7.2)

scriptease:
Fixes #4214

## Problem

On macOS, cold-starting the app by double-clicking a SQLite file or via a URL scheme (`postgresql://...`) opened a blank New Connection screen — the file/URL was ignored. Doing the same thing again while the app was already running worked fine.

## Root cause

On a macOS cold start, `open-file` / `open-url` fire **before** `app.ready`. Both handlers called `initBasics()`, which starts with `ProtocolBuilder.createAppProtocol()` → `protocol.registerBufferProtocol()`. That API touches the default session, which throws before ready. The handler's promise rejected silently, the URL was dropped, and the `ready` handler then built a plain empty window.

This regressed when protocol registration was moved to the top of `initBasics()` (4f192ec43).

## Fix (`apps/studio/src-commercial/entrypoints/main.ts`)

- Register the `app://` and `plugin://` protocols in the `ready` handler instead of `initBasics()` — registered exactly once, and only when legal (previously they were also re-registered on every `open-url`).
- `open-file` / `open-url` arriving before `ready` are queued in `pendingOpens`; the `ready` handler drains the queue alongside positional CLI args and opens one window per URL.
- `initBasics()` now memoizes its promise. The old `if (initialized) return settings` guard returned immediately while init was still running, handing a concurrent caller `undefined` settings.
- `initializeSecurity` / `initializeFileHelpers` / `createUtilityProcess` now run unconditionally in `ready`; previously they were skipped entirely when the app was launched with a positional file argument.
- Guarded `activate` so it can't race the startup path before the initial window exists.

## Testing

Verified on macOS (arm64) with unsigned local builds via LaunchServices (`open -a`), app fully quit before each test:

| Scenario | Before | After |
|---|---|---|
| Cold start, double-click/`open` a `.db` file | blank New Connection | connects to the file |
| Cold start, `mysql://root@localhost:3306` | blank New Connection | connects |
| Cold start, `postgresql://...` | blank New Connection | reaches connect (fails only if no server) |
| Warm app, same opens | works | still works |

Verified via the `used_connection` table in app.db as well as visually.

https://claude.ai/code/session_01CJKHfV26Pct7fyYVjvmNwe

### DIFF
--- a/apps/studio/src-commercial/entrypoints/main.ts
+++ b/apps/studio/src-commercial/entrypoints/main.ts
@@ -45,6 +45,8 @@ function initUserDirectory(d: string) {
 
 let utilityProcess: Electron.UtilityProcess
 let newWindows: number[] = [];
+let appReady = false
+let pendingOpens: string[] = []
 
 async function createUtilityProcess() {
   if (utilityProcess) {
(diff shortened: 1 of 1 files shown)


apps/studio/src-commercial/entrypoints/main.ts
```diff
@@ -220,22 +225,26 @@ app.on('ready', async () => {
...
+  await createUtilityProcess()
+
+  // positional arguments, plus anything macOS handed us via open-file /
+  // open-url before we were ready
+  const urls = [...platformInfo.parsedArgs._, ...pendingOpens]
+  pendingOpens = []
+  appReady = true
+
+  if (urls.length > 0) {
+    urls.forEach((url: string) => buildWindow(settings, { url }))
```
azmy60, line 245: Do we need `Promise.all` here?

azmy60 (changes requested):

PR #4546, opened 2026-07-31
ui: add multi-select deletion for saved queries

ylyra:
I'm currently dealing with an issue where this would help me a lot. I have so many items in my saved queries that it's practically impossible to delete them one by one because of how long it would take.

Example:
image

----

# Ai slop text bellow

## Summary
- Add multi-select deletion for Saved Queries via per-item checkboxes, Cmd/Ctrl+click toggle, and Shift+click range selection
- Show a bottom Remove/Cancel toolbar when one or more queries are selected
- Bulk delete uses a single confirmation dialog instead of one per item
- Hide the query code icon while bulk selection mode is active

## Test plan
- [ ] Select multiple saved queries with checkboxes, Cmd/Ctrl+click, and Shift+click
- [ ] Remove toolbar appears; Cancel clears selection without deleting
- [ ] Remove shows one confirm dialog and deletes all selected queries
- [ ] Context menu Delete on a single item still works with per-item confirm
- [ ] Drag-and-drop reordering still works
- [ ] `yarn test:unit tests/unit/components/sidebar/FavoriteListBulkDelete.spec.ts` passes (7 tests)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/studio/src/assets/styles/app/sidebar/favorite-list.scss
+++ b/apps/studio/src/assets/styles/app/sidebar/favorite-list.scss
@@ -66,6 +66,11 @@
     height: initial;
     line-height: initial;
     padding: $gutter-h ($gutter-w * 0.75);
+    &.bulk-selection-active {
+      .item-icon {
+        visibility: hidden;
+      }
+    }
     &:hover {
       transition: display 0.15s ease-in-out;
       .badge {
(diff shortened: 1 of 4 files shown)


not-night-but:
@ylyra This is definitely something that we need, but the entire structure of the rendering for saved queries has changed in #4502 so this feature would have to be built on top of that PR, otherwise this is just going to get overwritten in the merge

ylyra:
@not-night-but thanks for letting me know. I'll update it based on that PR.

not-night-but:
@ylyra no problem, thanks for contributing! Ping me when you're ready for a review

ylyra:
I went ahead and created a new PR based on #4502 since it was quicker and easier than updating the existing one.

PR #4547, opened 2026-07-31
fix: show favorite title in delete confirmation

ylyra:
## Summary
- Use `favorite.title` (with `name` fallback) in the saved-query delete confirmation dialog

## Test plan
- [ ] Open a saved query in the sidebar favorites list
- [ ] Delete it via context menu — dialog shows the query title
- [ ] Multi-select delete (`removeCheckedFavorites`) — each dialog shows the correct title

Before:
image

After:
image


not-night-but (approved):
Oh good catch!

PR #4549, opened 2026-07-31
ui: add multi-select deletion for saved queries

ylyra:
I'm currently dealing with an issue where this would help me a lot. I have so many items in my saved queries that it's practically impossible to delete them one by one because of how long it would take.

This PR uses #4502 as a starting point since entire structure of the rendering for saved queries has changed

Example:
image

----

# Ai slop text bellow

## Summary

- Add multi-select deletion for Saved Queries via per-item checkboxes, Cmd/Ctrl+click toggle, and Shift+click range selection
- Show a bottom Remove/Cancel toolbar when one or more queries are selected
- Bulk delete uses a single confirmation dialog instead of one per item
- Works in both the tree sidebar and cloud workspace search flat results

## Test plan

- [ ] Select multiple saved queries with checkboxes, Cmd/Ctrl+click, and Shift+click
- [ ] Remove toolbar appears; Cancel clears selection without deleting
- [ ] Remove shows one confirm dialog and deletes all selected queries
- [ ] Context menu Delete on a single item still works with per-item confirm
- [ ] Drag-and-drop reordering still works in the tree sidebar
- [ ] `yarn test:unit --testPathPattern=FavoriteListBulkDelete` passes (7 tests)


ylyra:
hi, @not-night-but, ready for review. 😅😅

not-night-but:
@ylyra I changed the base to the pr's branch just so the diff isn't polluted, looks like there are some conflicts

ylyra:
@not-night-but fixed 😄

apps/studio/src/common/utils/folderTree.ts
```diff
@@ -75,6 +75,45 @@ export function buildItemNodes<T extends HasId & { position?: number }>(
 }
 
 /** Transform {@link TreeNodeMoveEvent} into a consumable payload for the reorder action. */
```
azmy60: The comment is detached from the `parseReorderTarget`. Looks like a merge conflict issue?

azmy60 (commented):
I just want to comment; the base of this branch is currently being reviewed but I don't think there would be a significant change, only bug fixes.

Right now, this PR only applies to a query list, which is fine for now, but it'd be much better if it's done the same to the connection list (ConnectionSidebar.vue) as well.

I would redo this implementation inside the tree component, so in the app, I can just pass `selectedIds` like this:

```vue
<tree
  :folders="folderNodes"
  :items="itemNodes"
  :expandedIds="expandedIds"
   ...
  :selectedIds="selectedIds"
  @update:selectedIds="setSelectedIds"
>
```

*Notice that you still own the `selectedIds`, meaning you still can control which node can be selected externally, how many can be selected, etc..

apps/studio/src/common/utils/folderTree.ts
ylyra: thanks, fixed

ylyra:
@azmy60 done, sorry for the delay

ylyra:
Since PR #4502 has been merged, I switched the base back to master.

apps/studio/tests/unit/components/sidebar/ConnectionSidebarRemove.spec.ts
```diff
@@ -0,0 +1,180 @@
+import ConnectionSidebar from "@/components/sidebar/ConnectionSidebar.vue";
```
azmy60: I'm not really convinced with this UI test. It'll just create noise to the codebase. Please remove it.

apps/studio/tests/unit/components/sidebar/FavoriteListBulkDelete.spec.ts
```diff
@@ -0,0 +1,216 @@
+import FavoriteList from "@/components/sidebar/core/FavoriteList.vue";
```
azmy60: Same here.

apps/studio/src/components/sidebar/core/FavoriteList.vue
```diff
@@ -203,10 +207,13 @@
               <favorite-list-item
                 :item="node.ref"
                 :active="isActive(node.ref)"
-                :selected="selected === node.ref"
+                :selected="isChecked(node.ref) || selected === node.ref"
```
azmy60: I don't think we should keep `selected === node.ref` if we introduce `selectedIds`.

Also, can we not use `isChecked` and just call `selectedIds.include(node.id)`?

apps/studio/src/components/sidebar/core/FavoriteList.vue
```diff
@@ -203,10 +207,13 @@
...
+                :checked="isChecked(node.ref)"
```
azmy60: What's the difference between `checked` and `selected`?

apps/studio/src/components/sidebar/core/FavoriteList.vue
```diff
@@ -203,10 +207,13 @@
...
+                :bulk-selection-active="bulkSelectionActive"
                 :class="{ 'drag-pending': (pendingSaveIds || []).includes(node.ref.id) }"
                 @remove="remove"
                 @select="select"
+                @toggle-check="toggleChecked"
```
azmy60: Why do we make a new `toggle-check` event when we have a `select` event?

azmy60 (changes requested):

apps/studio/tests/unit/components/sidebar/ConnectionSidebarRemove.spec.ts
ylyra: done

apps/studio/tests/unit/components/sidebar/FavoriteListBulkDelete.spec.ts
ylyra: done

apps/studio/src/components/sidebar/core/FavoriteList.vue
ylyra: `checked` is the bulk-delete checkbox (selectedIds) and `selected` is the row highlight, the last clicked item, and any checked items so they stay visually marked.

apps/studio/src/components/sidebar/core/FavoriteList.vue
ylyra: They do different things. 

`select` is the row click highlight the item, shift-range, or cmd/ctrl-toggle. A normal click doesn’t actually check it.

`toggle-check` is a plain checkbox click, which always has to add/remove from bulk selection. 

If we sent that through `select`, it would just highlight the row and the checkbox wouldn’t toggle. Shift/cmd/ctrl on the checkbox still go through `select`. Same split as the connection list.

apps/studio/src/components/sidebar/core/FavoriteList.vue
ylyra: done

apps/studio/src/components/sidebar/core/FavoriteList.vue
ylyra: done

apps/studio/src/components/sidebar/core/FavoriteList.vue
ylyra: folded it into `select`

not-night-but:
@ylyra Looks like something is broken in this branch? It's preventing the frontend from compiling from what I can tell

ylyra:
hi @not-night-but, fixed it. It was caused by a missing comma when merging master into it.

apps/studio/src/common/utils/folderTree.ts
```diff
@@ -79,6 +117,22 @@ export function buildItemNode<T extends HasId>(
   };
 }
 
+/**
+ * Collect item refs in the same order the ui-kit Tree renders them (expanded
+ * folders only; subfolders before sibling items within a folder).
+ */
+export function collectVisibleItemRefs<T extends HasId>(
```
azmy60: I don't see this used anymore. Please remove it.

apps/studio/src/common/utils/folderTree.ts
```diff
@@ -1,10 +1,48 @@
...
 import type {
   FolderNode,
   ItemNode,
   TreeNodeMoveEvent,
 } from "@beekeeperstudio/ui-kit";
-import { HasId } from "@/common/interfaces/IGeneric";
-import { IFolder } from "@/common/interfaces/IQueryFolder";
+import { collectVisibleItemIds } from "@beekeeperstudio/ui-kit/tree/helpers";
+
+export { collectVisibleItemIds };
```
azmy60: We shouldn't re-export this.

apps/studio/src/common/utils/folderTree.ts
```diff
@@ -1,10 +1,48 @@
...
+
+export function toggleSelectedId(
```
azmy60: Looks like this is a duplicate of the UI Kit's function? If so, please export this function from UI Kit instead.

apps/studio/src/common/utils/folderTree.ts
```diff
@@ -1,10 +1,48 @@
...
+  id: ItemNode["id"]
+): ItemNode["id"][] {
+  const index = selectedIds.indexOf(id);
+  if (index === -1) {
+    return [...selectedIds, id];
+  }
+  return selectedIds.toSpliced(index, 1);
+}
+
+export function rangeSelectVisibleIds(
```
azmy60: Same as `toggleSelectedId`, I think we should also export this from UI Kit.

apps/studio/src/components/sidebar/core/favorite_list/FavoriteListItem.vue
```diff
@@ -5,11 +5,24 @@
   >
     <a
       class="list-item-btn"
+      v-tooltip.bottom.delay="{
```
azmy60: I think we don't need to use v-tooltip here because we already use `:title`. This was probably came through from a merge conflict?

azmy60 (changes requested):
@ylyra This is close! We still have a couple of minor changes in the code and more here for UX:

1. Pressing esc should cancel the selection.

2. Let's not show the checkbox when we hover it. It's OK to show this when the bulk selection is active.
Image

3. If I activate the bulk selection, somehow the item height is squeezed a little bit, and the checkbox also gets larger. (This happens in connection sidebar. Query sidebar is good.)
4. If I select multiple items, and then right click, I should only see the "Delete" option.

Image

5. Also, the delete option (from the context menu) does not work.

6. I'm not sure if these buttons are effective so I wouldn't add them for now because;
  a. It was not obvious for me at first that I should look down to find that there is a delete button. I was on my laptop, and if I could barely find this button on a small screen, how about a larger screen? I feel like this is a pattern that's more common in mobile apps than desktop apps.
  b. Are we sure that "delete" is the action that most people will look for? Is this the right place for it?

Anyway, no need to answer those. I'm just telling you my thought process. But I would remove these buttons for now.

Image

After those are deleted, the only way to delete is by using the context menu. We might want to find a way to cancel the selection other than pressing esc but that could be done in another time.

7. The checkbox is too close to the name.

Image

ylyra:
hi @azmy60 

5. sorry about that. in my mind only the "remove" at the bottom was enough.
6. I just reused them since they were already hidden. But if you think it’s better to remove them, I can do that.

Items: 1, 2, 3, 4, 5 and 7 fixed.

apps/studio/src/components/sidebar/core/favorite_list/FavoriteListItem.vue
ylyra: removed, it came through from a merge conflict

apps/studio/src/common/utils/folderTree.ts
ylyra: done

apps/studio/src/common/utils/folderTree.ts
ylyra: done

apps/studio/src/common/utils/folderTree.ts
ylyra: done

apps/studio/src/common/utils/folderTree.ts
ylyra: done

azmy60:
@ylyra Yes please :)

ylyra:
@azmy60 done

azmy60 (changes requested):
Love it! I found two more bugs:

1. If I click an item, and then ctrl+click the second item, the first item should be selected as well.

https://github.com/user-attachments/assets/3e3d262d-0b5a-453a-aa7d-f6e8742619eb

2. If I have multiple selection, and then I click an item without pressing `ctrl` or `shift`, it should deselect everything except the item that I click.

I think these are the last UX bugs that will make it feel more native. I'm ready to merge this when the they are addressed. Overall, this is a fantastic job!

ylyra:
done

PR #4551, opened 2026-08-01
fix(postgres): use selected overload for SQL: Create on overloaded functions

### LINKED ISSUE #4513: BUG: SQL: Create opens top overload instead of selected PostgreSQL function
### Describe the bug
For overloaded PostgreSQL functions (same name, different parameter lists), `SQL: Create` from the context menu does not use the selected function from the tree.

It always generates the script for the top entry in the function group.

### To Reproduce
1. Create multiple functions with the same name in one schema, e.g.:
   - `my_func(a int)`
   - `my_func(a int, b int)`
   - `my_func(a int, b int, c int)` (optional)
2. In Beekeeper Studio, open the function tree and select a lower overload (not the first/top one).
3. Right-click -> `SQL: Create`.

### Expected behavior
`SQL: Create` should open the definition for the exact selected overload (signature-aware).

### Actual behavior
It always opens the definition of the top overload in the tree (typically the one with fewer parameters), regardless of which overload is selected.

### Additional notes
The tree appears to order overloads by parameter count or lexicographic parameter signature, but regardless of ordering, SQL generation always resolves to the top entry.

### Environment
- Beekeeper Studio: 5.9.2
- OS: Windows 11
- Database: PostgreSQL

AdityaBhosale22:
## Summary

Fixes #4513.

When using **SQL: Create** on overloaded PostgreSQL functions, Beekeeper Studio always generated the definition for the first overload instead of the selected one.

This change propagates the routine identifier through the existing flow and uses it when generating the CREATE script so the selected overload is resolved correctly.

## What changed

- Added an optional routine identifier parameter to `getRoutineCreateScript`.
- Passed the selected routine's identifier from the routine tree through the IPC layer.
- Updated the PostgreSQL implementation to filter by routine OID when the identifier is available.
- Added a unit test covering overloaded PostgreSQL functions.

## Behavior

Before:

- `SQL: Create` always returned the first overload matching the function name.

After:

- `SQL: Create` returns the definition for the exact overload selected in the routine tree.

When no routine identifier is available, the existing lookup behavior is preserved.

## Testing

- Added a unit test covering overloaded PostgreSQL functions.
- Verified existing PostgreSQL tests continue to pass.
- Confirmed no behavior changes for other database clients.

### DIFF
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -286,8 +286,11 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
         r.routine_schema as routine_schema,
         r.routine_name as name,
         r.routine_type as routine_type,
-        r.data_type as data_type
+        r.data_type as data_type,
+        p.oid::text as oid
       FROM INFORMATION_SCHEMA.ROUTINES r
+      LEFT JOIN pg_catalog.pg_proc p
+        ON r.specific_name::text = p.proname || '_' || p.oid::text
       where r.routine_schema not in ('sys', 'information_schema',
                                      'pg_catalog', 'performance_schema')
         ${schemaFilter ? `AND ${schemaFilter}` : ''}
@@ -329,6 +332,7 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
         returnType: row.data_type,
         entityType: 'routine',
         id: row.id,
+        oid: row.oid,
         routineParams: params.map((p, i) => {
           return {
             name: p.parameter_name || `arg${i + 1}`,
--- a/apps/studio/src/lib/db/clients/postgresql.ts
+++ b/apps/studio/src/lib/db/clients/postgresql.ts
@@ -1030,18 +1030,20 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
     return data.rows.map((row) => `${createViewSql}\n${row.pg_get_viewdef}`);
   }
 
-  async getRoutineCreateScript(routine: string, _type: string, schema: string = this._defaultSchema): Promise<string[]> {
+  async getRoutineCreateScript(routine: string, _type: string, schema: string = this._defaultSchema, id?: string): Promise<string[]> {
     const sql = `
       SELECT pg_get_functiondef(p.oid)
       FROM pg_proc p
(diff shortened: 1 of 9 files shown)


azmy60 (approved):
LGTM

apps/studio/src/lib/db/clients/postgresql.ts
```diff
@@ -1030,18 +1030,20 @@ export class PostgresClient extends BasicDatabaseClient<QueryResult, PoolClient>
...
 
-  async getRoutineCreateScript(routine: string, _type: string, schema: string = this._defaultSchema): Promise<string[]> {
+  async getRoutineCreateScript(routine: string, _type: string, schema: string = this._defaultSchema, id?: string): Promise<string[]> {
     const sql = `
       SELECT pg_get_functiondef(p.oid)
       FROM pg_proc p
              LEFT JOIN pg_catalog.pg_namespace n ON n.oid = p.pronamespace
       WHERE proname = $1
         AND n.nspname = $2
+        ${id ? 'AND p.oid = $3::oid' : ''}
```
not-night-but, line 1044: It looks like id isn't actually an oid. We grab it from `INFORMATION_SCHEMA.ROUTINES.specific_name` which is of the type `name`, so this throws an error:

Image

not-night-but (changes requested):
We don't actually get the `oid` of the routine. the `id` field on `Routine` is of the pg type `name`. We may need to add another field to `Routine` and modify the script to get the `oid`.

This also needs a regression test in the postgres integration test suite.

AdityaBhosale22:
> We don't actually get the `oid` of the routine. the `id` field on `Routine` is of the pg type `name`. We may need to add another field to `Routine` and modify the script to get the `oid`.
> 
> This also needs a regression test in the postgres integration test suite.

Thanks for catching that! You're right I assumed Routine.id was the PostgreSQL OID, but it's actually information_schema.routines.specific_name. I'll update the routine listing to expose the real pg_proc.oid separately, thread that through the SQL Create path, and add a regression test to the PostgreSQL integration suite.

AdityaBhosale22:
Thanks for catching that @not-night-but! I updated the implementation to expose the actual pg_proc.oid as a separate optional field on Routine while keeping Routine.id unchanged. The SQL Create flow now uses the real OID, and I added a regression test to the PostgreSQL integration test suite covering overloaded functions.

not-night-but:
@AdityaBhosale22 Nice! Thanks for the patch !

not-night-but (approved):

PR #4552, opened 2026-08-01
feat(editor): add configurable SQL keyword casing for autocomplete

### LINKED ISSUE #4501: FEAT: Autocomplete should insert SQL keywords in UPPERCASE
## Feature request

When using SQL autocomplete, typing:

```sql
sel
```

and pressing Enter inserts:

```sql
select
```

I would like an option to insert SQL keywords in uppercase instead:

```sql
SELECT
```

This should apply to all SQL keywords, for example:

- SELECT
- FROM
- WHERE
- GROUP BY
- ORDER BY
- INSERT
- UPDATE
- DELETE

### Suggested solution

Add a setting such as:

- Autocomplete keyword case:
  - Preserve (current behavior)
  - UPPERCASE
  - lowercase

Many SQL developers use uppercase keywords as their coding style. This option would make Beekeeper Studio consistent with IDEs such as DataGrip and DBeaver.

**Beekeeper Studio version:** 5.9.2

AdityaBhosale22:
## Summary

Fixes #4501.

Adds a configurable SQL keyword casing option for autocomplete.

Users can now choose how SQL keywords are inserted when accepting autocomplete suggestions:

- Preserve (default)
- UPPERCASE
- lowercase

## What changed

- Added a configurable keyword case setting.
- Applied keyword casing through the SQL autocomplete pipeline.
- Exposed the setting in the query editor settings menu.
- Added a migration to register the new setting.
- Added unit tests covering the new behavior.

## Behavior

Before:

- Autocomplete always inserted SQL keywords using the default casing.

After:

- SQL keyword insertion respects the configured keyword case.
- Only SQL keywords are transformed.
- Table names, column names, schemas, and other identifiers are unchanged.
- Existing behavior is preserved by default.

## Testing

- Added unit tests for:
  - Preserve (default)
  - UPPERCASE
  - lowercase
  - Multiple SQL keywords
  - Identifiers remaining unchanged
  - Runtime setting changes

- Verified existing tests continue to pass.
- Verified linting and TypeScript checks produce no new errors.


apps/ui-kit/tests/unit/sql-text-editor/completions.spec.ts
```diff
@@ -3,9 +3,10 @@
```
azmy60, line 3: This is a built in feature from CodeMirror and most likely it's already tested so please don't add tests for this :)

apps/studio/tests/unit/migration/20260801_add_keyword_case_setting.spec.ts
```diff
@@ -0,0 +1,18 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
azmy60: Please remove this too

azmy60 (changes requested):
Thank you for the contribution! I have some corrections before we can merge this:

1. Instead of adding a new key to a `user_setting` table, this should be a new key in `default.config.ini` file.
  - To access ini configuration, you can use `this.$bksConfig` inside Vue components, or `window.bksConfig` if it's outside of Vue.
2. I don't see `preserve` to be implemented? If so, please remove it. We should be good to have two possible values for the new config; `lowercase (as default)` and `uppercase`.
3. I also don't think this is a feature that's worth to test. Please remove the tests :)

AdityaBhosale22:
> Thank you for the contribution! I have some corrections before we can merge this:
> 
> 1. Instead of adding a new key to a `user_setting` table, this should be a new key in `default.config.ini` file.
> 
> * To access ini configuration, you can use `this.$bksConfig` inside Vue components, or `window.bksConfig` if it's outside of Vue.
> 
> 2. I don't see `preserve` to be implemented? If so, please remove it. We should be good to have two possible values for the new config; `lowercase (as default)` and `uppercase`.
> 3. I also don't think this is a feature that's worth to test. Please remove the tests :)

Thanks for the review! I'll update the implementation to use default.config.ini, remove the migration and tests, and simplify the option to lowercase/uppercase as suggested.

apps/studio/tests/unit/migration/20260801_add_keyword_case_setting.spec.ts
AdityaBhosale22: > Please remove this too

Removed the migration and migration test

PR #4556, opened 2026-08-02
fix: restore table row delete shortcuts via config on macOS

ylyra:
## Summary
- Fix table row delete shortcuts on macOS where `deleteSelection` collided with `nullSelection` after platform keybinding conversion
- Route `deleteSelection` through config-driven `keydown` capture so `Delete`, `fn+Delete`, and `Cmd+Backspace` all delete selected rows reliably
- Add `deleteSelection[] = ctrlOrCmd+backspace` to default config and keep `Backspace` for setting cells to NULL
- Show correct Command-based shortcut labels in the table context menu on Mac

## Test plan
- [ ] Open a table with a primary key in browse view
- [ ] Select a row and press Delete (external keyboard / fn+Delete) — row is marked as deleted with pending changes
- [ ] Select a row and press Cmd+Backspace — row is marked as deleted
- [ ] Select a single cell and press Backspace — cell value becomes NULL
- [ ] Right-click a row and confirm the Delete menu shows the configured shortcut
- [ ] Save with Cmd+S and verify the row is removed from the database
- [ ] Run `yarn test:unit tests/unit/components/tableview/TableTable.keybindings.spec.ts tests/unit/config.spec.js`

Made with [Cursor](https://cursor.com)


apps/studio/src/common/bksConfig/BksConfigProvider.ts
```diff
@@ -140,12 +140,12 @@ const uiModifierMap: ModifierMap = {
...
-  CMDORCTRL: "Control",
-  COMMAND: "Control",
-  CONTROLORCOMMAND: "Control",
-  COMMANDORCONTROL: "Control",
+  CMD: (isMac) => (isMac ? "Command" : "Control"),
+  CTRLORCMD: (isMac) => (isMac ? "Command" : "Control"),
+  CMDORCTRL: (isMac) => (isMac ? "Command" : "Control"),
+  COMMAND: (isMac) => (isMac ? "Command" : "Control"),
+  CONTROLORCOMMAND: (isMac) => (isMac ? "Command" : "Control"),
+  COMMANDORCONTROL: (isMac) => (isMac ? "Command" : "Control"),
```
not-night-but: No we can't do this, `xel` expects Control and will map it to either `Ctrl` or the Command symbol depending on the device. These changes break the display of mac keybinds
image

apps/studio/src/common/bksConfig/BksConfigProvider.ts
```diff
@@ -218,10 +218,6 @@ export function convertKeybinding(
       if (mod === "ctrlorcmd") {
         mod = platform === "mac" ? "meta" : "ctrl";
       }
-
-      if (mod === "delete" && platform === "mac") {
-        mod = "backspace";
-      }
```
not-night-but, line 224: I'm not sure flat out removing this is the right call? The issue was that MacOS maps delete to a different keycode than Linux and Windows. Maybe there is a better way to ensure that v-hotkey actually uses delete?
https://github.com/beekeeper-studio/beekeeper-studio/commit/7f0c390f95ef07614088f2fe27ab7b46190fd5e7

apps/studio/src/components/tableview/TableTable.vue
```diff
@@ -1105,6 +1106,20 @@ export default Vue.extend({
...
+      const bindingList = Array.isArray(bindings) ? bindings : [bindings]
+
+      for (const binding of bindingList) {
+        if (matchesVHotkeyBinding(e, binding)) {
+          e.preventDefault()
+          this.deleteTableSelection(e)
+          return
+        }
+      }
+    },
```
not-night-but: I think we should just continue to use v-hotkey. This seems to be over complicating things

apps/studio/src/plugins/ConfigPlugin.ts
```diff
@@ -4,6 +4,60 @@ import { BksConfigProvider, KeybindingPath } from "@/common/bksConfig/BksConfigP
...
+
+  return (
+    event.keyCode === keyCode &&
+    event.ctrlKey === ctrl &&
+    event.altKey === alt &&
+    event.shiftKey === shift &&
+    event.metaKey === meta
+  );
+}
+
```
not-night-but: Yeah no this is just going to be annoying to maintain in the future, and I'm not totally sure it's necessary?

not-night-but (changes requested):
I can see how some if this may be a problem, but I think introducing another handler for binding keyboard shortcuts is just not a good idea. I think maybe we just need to look at binding the proper keycode for delete rather than mapping delete to backspace? Or change the default keybind for one of those actions.

apps/studio/src/common/bksConfig/BksConfigProvider.ts
ylyra: oh, my bad. did not see that breaking change. fixed it

apps/studio/src/common/bksConfig/BksConfigProvider.ts
ylyra: reverted the lines.

apps/studio/src/components/tableview/TableTable.vue
ylyra: you right, removed this. thanks 😄

not-night-but (approved):
Looks good. I'm not sure this totally fixes the issue you were seeing, but I think to fix that we might have to add some extra functionality to the config system (like a `noneOrCmd` or way to override bindings for a specific platform. But I think this is good for now. Thanks!

PR #4558, opened 2026-08-02
fix: detect direct AppImage execution for auto-updater

### LINKED ISSUE #4540: BUG: AppImage auto-update only works with AppImageLauncher, not when running AppImage directly
## Description

The auto-updater is only enabled when `process.env.DESKTOPINTEGRATION === 'AppImageLauncher'` (see `mainPlatformInfo.ts` line 87). This means users who run the AppImage directly — the most common case — get no update functionality at all. Nothing appears in the log and Help > Check for Software Updates does nothing.

## Root Cause

```ts
```

The AppImage runtime sets `$APPIMAGE` for all AppImage executions, but Beekeeper only checks for AppImageLauncher's specific env variable.

## Fix

Also check for `process.env.APPIMAGE` to detect AppImage execution regardless of launcher:

```ts
isAppImage: p.env.DESKTOPINTEGRATION === 'AppImageLauncher' || !!p.env.APPIMAGE,
```

Note: the existing `dealWithAppImage()` hack that rewrites `$APPIMAGE` to `$ARGV0` is specifically for AppImageLauncher and should only run in that case.

## Impact

Docs state AppImage provides automatic updates, but this only works with AppImageLauncher installed. Users running AppImage directly have no update path other than manual download.

cynicalmindset:
## Summary

Fix AppImage detection so the auto-updater is available when Beekeeper Studio is run directly as an AppImage.

### Changes

- Detect AppImage execution using the `APPIMAGE` environment variable in addition to `DESKTOPINTEGRATION=AppImageLauncher`.
- Restrict the AppImageLauncher-specific `APPIMAGE` rewrite workaround to AppImageLauncher only.

### Why

Previously, `isAppImage` was only true when Beekeeper Studio was launched through AppImageLauncher. As a result, users running the AppImage directly did not have access to the built-in auto-updater, despite automatic updates being documented for AppImage installations.

Using the `APPIMAGE` environment variable allows Beekeeper Studio to correctly detect direct AppImage execution while preserving the existing AppImageLauncher-specific workaround.

closes: #4540

### DIFF
--- a/apps/studio/src/background/update_manager.ts
+++ b/apps/studio/src/background/update_manager.ts
@@ -13,7 +13,7 @@ autoUpdater.logger = log
 
 // HACK(mc, 2019-09-10): work around https://github.com/electron-userland/electron-builder/issues/4046
 function dealWithAppImage() {
-  if (platformInfo.isAppImage) {
+  if (process.env.DESKTOPINTEGRATION === 'AppImageLauncher') {
     // remap temporary running AppImage to actual source
     // THIS IS PROBABLY SUPER BRITTLE AND MAKES ME WANT TO STOP USING APPIMAGE
     // eslint-disable-next-line
--- a/apps/studio/src/common/platform_info/mainPlatformInfo.ts
+++ b/apps/studio/src/common/platform_info/mainPlatformInfo.ts
@@ -93,7 +93,7 @@ export function mainPlatformInfo(): IPlatformInfo {
     isFlatpak: !!p.env.FLATPAK_ID || existsSync('/.flatpak-info'),
     isPortable: isWindows && p.env.PORTABLE_EXECUTABLE_DIR,
     isDevelopment: isDevEnv,
-    isAppImage: p.env.DESKTOPINTEGRATION === 'AppImageLauncher',
+    isAppImage: p.env.DESKTOPINTEGRATION === 'AppImageLauncher' || !!p.env.APPIMAGE,
     sshAuthSock: p.env.SSH_AUTH_SOCK,
     sshConfigExists: existsSync(join(homeDirectory, '.ssh', 'config')),
     defaultSshIdentityFile: ['id_ed25519', 'id_ecdsa', 'id_rsa', 'id_dsa']


apps/studio/src/common/platform_info/mainPlatformInfo.ts
```diff
@@ -93,7 +93,7 @@ export function mainPlatformInfo(): IPlatformInfo {
     isFlatpak: !!p.env.FLATPAK_ID || existsSync('/.flatpak-info'),
     isPortable: isWindows && p.env.PORTABLE_EXECUTABLE_DIR,
     isDevelopment: isDevEnv,
-    isAppImage: p.env.DESKTOPINTEGRATION === 'AppImageLauncher',
+    isAppImage: p.env.DESKTOPINTEGRATION === 'AppImageLauncher' || !!p.env.APPIMAGE,
```
not-night-but, line 96: I actually think we might want to split these up? `APPIMAGE` is set regardless of if it is running inside AppImageLauncher, so we can just have `isAppImage: !!p.env.APPIMAGE` and then also have an `isAppImageLauncher` variable to be used in places like `dealWithAppImage`

apps/studio/src/background/update_manager.ts
```diff
@@ -13,7 +13,7 @@ autoUpdater.logger = log
 
 // HACK(mc, 2019-09-10): work around https://github.com/electron-userland/electron-builder/issues/4046
 function dealWithAppImage() {
-  if (platformInfo.isAppImage) {
+  if (process.env.DESKTOPINTEGRATION === 'AppImageLauncher') {
```
not-night-but, line 16: Probably not for this PR, but I just looked into this issue, and it looks like it is no longer an issue as of `AppImageLauncher` v2.2.0, which was released in 2020. Might be something worth looking into if that is now standard version in stable repos.

not-night-but (changes requested):
Thanks for the patch! I think it might be useful to be able to differentiate between raw AppImage and AppImageLauncher in other places in the future, so I think we should put that on the platforminfo object as well.

apps/studio/src/common/platform_info/mainPlatformInfo.ts
cynicalmindset: Um yeah make sense , I'll split isAppImage and isAppImageLauncher into separate fields on platformInfo and update dealWithAppImage() to use the new property

PR #4559, opened 2026-08-02
fix: reposition AND/OR operators in table filters

### LINKED ISSUE #4433: FEAT: Reposition the AND/OR in MongoDB's sql thing
It's more of a visual inconsistency. When you have more than one statement the AND/OR is put before the first line even though it applies to the second line. Currently it is:
AND [line 1]
AND [line 2]
[line 3]

when it should be:
[line 1]
AND [line 2]
AND [line 3]

The AND/OR button should precede the argument it applies to

N/A

**Additional context**

I experience this in MongoDB. I don't know if this editor is used in other databases or not.

Image

bnquon:
added a hidden spacer when raw filters are unavailable, keeping the and/or operators aligned with the filters they apply to. the spacer uses `visibility: hidden` because the empty spacer inherited an unintended button hover effect.

screenshot shown for MongoDB and also separately tested in Trino
image

closes #4433

### DIFF
--- a/apps/studio/src/assets/styles/app/core-tabs.scss
+++ b/apps/studio/src/assets/styles/app/core-tabs.scss
@@ -449,6 +449,9 @@
       padding-left: 0.45rem;
       padding-right: 0.25rem;
     }
+    .filter-mode-spacer {
+      visibility: hidden;
+    }
     .multiple-filter  {
       flex-grow: 1;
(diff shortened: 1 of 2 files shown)


not-night-but:
@bnquon Thanks for the patch!

not-night-but (approved):

PR #4560, opened 2026-08-02
fix: preserve table column order in SQL->SQL exports

### LINKED ISSUE #4266: FEAT: Exported row content should maintain the same column order as the original table.
If the table column order is:
id,
name,
age

The export data file SQL content is
insert into `test` (
`age`, 
`id`, 
`name`
) values (20, 1, 'a');

Keep the column order is friendly.

I can confirm that this issue still exists in version 5.6.2.

Is there any plan to fix this in a future release?

Thank you.

bnquon:
- works around Knex sorting object keys before generating insert queries, that's why export had different order than in the UI
- uses Knex’s raw query helper to build the insert body directly from ordered column and value arrays instead of passing a row object to Knex (to avoid the object sorting mentioned above)
  - investigated a way to use the original `this.rowToObject()` function, looping over the returned keys and values but JS/TS doesn't preserve the order of integer-like string keys such as `"10"` or `"2"` during object enumeration
- preserves existing bit, json, fallback-column, and Knex behavior
- adds a defensive SQLite check for undefined values although the ui should never pass these, and this matches Knex’s existing SQLite restriction
- passes existing unit tests and added some to cover the ordering behavior

SS showing sql outputs after exporting, **left** is _BEFORE_ and **right** is _AFTER_:

image

closes #4266

### DIFF
--- a/apps/studio/src/lib/export/formats/sql.ts
+++ b/apps/studio/src/lib/export/formats/sql.ts
@@ -66,6 +66,17 @@ export class SqlExporter extends Export {
     return ""
   }
 
+  private getColumnNames(rowLength: number, columns: TableColumn[]): string[] {
+    const storedColumns = this.dedupedColumns
+    const hasColumnNames = (candidate: TableColumn[]) =>
+      candidate.length === rowLength && candidate.every(column => column.columnName != null)
+
+    if (hasColumnNames(storedColumns)) return storedColumns.map(column => column.columnName)
+    if (hasColumnNames(columns)) return columns.map(column => column.columnName)
+
+    return Array.from({ length: rowLength }, (_value, index) => `col_${index + 1}`)
+  }
+
   formatRow(rowArray: any[], columns: TableColumn[] = []): string {
     const sanitized = rowArray.map((val, idx) => {
       // Handle bit columns - convert booleans and buffers to 0/1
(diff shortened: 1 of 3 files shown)


apps/studio/src/lib/export/formats/sql.ts
```diff
@@ -82,15 +93,37 @@ export class SqlExporter extends Export {
...
+      }
+    
+      if (_.isPlainObject(value)) {
+        return JSON.stringify(value)
+      }
+
+      return value
+    })
+
+    const insertBody = this.knex.raw(
```
azmy60, line 127: I'm assuming the reason why the column order is wrong because knex just follows the order from the argument.

I don't know if this is true, but for example, if I run `knex.insert({ b, c, a }).toQuery()`, then the output would be `INSERT INTO ... (b, c, a)`. If I rearrange the keys before passing it to `knex` like `knee.insert({ a, b, c }).toQuery()`, then the output might be `INSERT INTO ... (a, b, c)`?

If this is true then you don't need to construct the raw SQL, but just simply rearrange the argument keys. There's a way to rearrange object keys like that. I hope that makes sense?

azmy60 (dismissed):

apps/studio/src/lib/export/formats/sql.ts
bnquon: hi I dug into this a bit and its actually bc Knex sort's the keys when passing through an object, it does `columns = Object.keys(data[i]).sort()` here in [line 1289](https://github.com/knex/knex/blob/2.5.1/lib/query/querycompiler.js#L1274-L1297)

here is what your example scenario code outputted btw:
input: 
```
node -e 'const knex=require("knex")({client:"pg"}); const rows=[{b:2,c:3,a:1},{a:1,b:2,c:3}]; rows.forEach(row=>console.log("keys:",Object.keys(row),"sql:",knex("example").insert(row).toQuery()))'
```

output:
```
keys: [ 'b', 'c', 'a' ] sql: insert into "example" ("a", "b", "c") values (1, 2, 3)
keys: [ 'a', 'b', 'c' ] sql: insert into "example" ("a", "b", "c") values (1, 2, 3)
```

azmy60: Awesome, thanks for looking into it!

azmy60:
I feel like this PR is a feature rather than a bug fix, because `INSERT INTO ... (a,b,c) VALUES ("a","b","c")` is essentially the same as `INSERT INTO ... (b,c,a) VALUES ("b","c","a")` so it doesn't really matter what the order is, unless I'm missing something.

I'd suggest that we keep the default behavior (let knex does its job like previously), and add a checkbox in the UI so users can decide whether they want to apply this feature or not. Thoughts @not-night-but ?

not-night-but:
Yeah I think that's a good middle ground

bnquon:
added a “Preserve Column Order” checkbox under SQL Advanced Options. it defaults to off, lmk if this looks good!

image

not-night-but (approved):

not-night-but:
@bnquon Thanks again!

not-night-but (approved):

PR #4562, opened 2026-08-03
Fix/appimage launcher detection

### LINKED ISSUE #4540: BUG: AppImage auto-update only works with AppImageLauncher, not when running AppImage directly

cynicalmindset:
closes: #4540 
continuation of PR #4558

### DIFF
--- a/apps/studio/src/background/update_manager.ts
+++ b/apps/studio/src/background/update_manager.ts
@@ -13,7 +13,7 @@ autoUpdater.logger = log
 
 // HACK(mc, 2019-09-10): work around https://github.com/electron-userland/electron-builder/issues/4046
 function dealWithAppImage() {
-  if (platformInfo.isAppImage) {
+  if (platformInfo.isAppImageLauncher) {
     // remap temporary running AppImage to actual source
     // THIS IS PROBABLY SUPER BRITTLE AND MAKES ME WANT TO STOP USING APPIMAGE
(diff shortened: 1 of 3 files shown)


not-night-but:
@cynicalmindset. Thanks! In the future, no need to open a new PR after making changes. You can just push them up to the same branch and the PR will update

not-night-but (approved):

PR #4563, opened 2026-08-03
fix(postgres): use selected overload for SQL: Create on overloaded functions

### LINKED ISSUE #4513: BUG: SQL: Create opens top overload instead of selected PostgreSQL function

drannonymousxx:
## Summary

Fixes #4513.

### Problem

For overloaded PostgreSQL functions, `SQL: Create` always generated the definition for the first overload instead of the selected overload.

### Solution

- Preserve the selected routine identifier through the existing call chain.
- Pass the routine identifier to `getRoutineCreateScript`.
- Safely extract the PostgreSQL OID from PostgreSQL's `specific_name`.
- Query the selected overload by OID when available.
- Preserve the existing lookup path for backward compatibility.

### Testing

- Added unit test coverage.
- Verified backward compatibility when no routine identifier is supplied.

### DIFF
--- a/apps/studio/src-commercial/backend/handlers/connHandlers.ts
+++ b/apps/studio/src-commercial/backend/handlers/connHandlers.ts
@@ -70,7 +70,7 @@ export interface IConnectionHandlers {
   'conn/getTableCreateScript': ({ table, schema, sId }: { table: string, schema?: string, sId: string }) => Promise<string>,
   'conn/getViewCreateScript': ({ view, schema, sId }: { view: string, schema?: string, sId: string }) => Promise<string[]>,
   'conn/getMaterializedViewCreateScript': ({ view, schema, sId }: { view: string, schema?: string, sId: string }) => Promise<string[]>,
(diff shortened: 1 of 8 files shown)


not-night-but:
This was fixed in #4551, thanks though!

PR #4564, opened 2026-08-04
fix: display formatted keybindings in tooltips

### LINKED ISSUE #4525: BUG: Some shortcuts aren't displayed nicely
It looks like we are showing raw shortcuts in tooltips (unparsed).

Image

We might need a vue helper like `this.$keybindingLabel('path.to.keybinding')` or something like that.

bnquon:
added the helper suggested in the issue to display formatted keybindings in tooltips instead of the unparsed versions.

updated `DeepKeyOf` in `global.d.ts` so array-valued keybindings, like `general.refresh`, are recognized as valid type-safe paths when passed to the helper.

added unit test

demo showing keybindings in tooltips:

https://github.com/user-attachments/assets/103ef428-d113-40f2-8ca8-5d28c0b9fe4b

also replaced these ones in the `structure` view of tables that were showing 'Meta+R' for example before
image

closes #4525

### DIFF
--- a/apps/studio/jest.config.js
+++ b/apps/studio/jest.config.js
@@ -32,6 +32,7 @@ module.exports = {
     '^@commercial(.*)$': '<rootDir>/src-commercial/$1',
     '^@bksLogger$': '<rootDir>/src/lib/log/mainLogger.ts',
     '^@tests(.*)$': '<rootDir>/tests/$1',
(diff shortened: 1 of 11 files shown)


not-night-but:
@bnquon Nice! This has been bugging me for a while lol so appreciate the fix.

not-night-but (approved):

PR #4571, opened 2026-08-05
Directory import

not-night-but:


apps/studio/src/components/common/modals/MoveItemModal.vue
```diff
@@ -14,52 +14,15 @@
...
-              v-if="parentId === props.node.ref.id"
-              class="current-location"
-            >
-              (current location)
-            </span>
-          </button>
-        </template>
-      </tree>
+      <folder-tree-picker
+        :key="`${folderPath}-${target.value.id}`"
```
azmy60, line 18: Not sure what `:key` is for?

apps/studio/src/components/connection/SaveConnectionForm.vue
```diff
@@ -17,14 +17,12 @@
...
       <label>Folder <i v-if="!isUltimate && !isCloud" class="material-icons menu-icon">stars</i></label>
-      <select v-model="config.connectionFolderId" :disabled="disabled || (!isUltimate && !isCloud)">
-        <option :value="null">
-          No folder
-        </option>
-        <option v-for="f in folders" :key="f.id" :value="f.id">
-          {{ f.name }}
-        </option>
-      </select>
+      <folder-tree-picker
```
azmy60: To be frank, I don't feel like this works at all. Progressive loading for folders is kinda dead anyway if this is how we want to use the `<tree>` component.

Image

It's much better if it's a copy of `<file-picker>` but for workspace folders:

Image

apps/studio/src/lib/cloud/controllers/GenericController.ts
```diff
@@ -82,6 +82,8 @@ export abstract class GenericController<T extends HasId> {
   async import(qs: T[]): Promise<T[]> {
     const response = await this.axios.post(url(this.path, "import"), {
       [this.plural]: qs,
+    }, {
+      timeout: 20000
```
azmy60, line 86: Worth adding an ini config for this.

azmy60 (approved):

apps/studio/src/components/common/modals/MoveItemModal.vue
not-night-but: It forces the component to re-init when the key changes

apps/studio/src/components/connection/SaveConnectionForm.vue
not-night-but: Yeah I really don't like it here. I think we definitely at least need the modal here

PR #4573, opened 2026-08-05
Add command-line database connections

### LINKED ISSUE #361: QUESTION: Can I connect to a database server from the command line?
**Details:**
Is it possible to pass all necessary parameters on the command line. e.g. db type, hostname/ip, port, user, password, database? I could not find any command line examples in the documentation.

eliasthecactus:
## Summary

- accept connection URLs and database files as positional arguments, `--url`, or `-u`
- serialize Electron startup initialization and queue early macOS open events
- preserve complete MongoDB URIs and their options
- support `mssql://` plus SQL Server authentication selection through the `auth` query parameter
- document command-line usage and SQL Server authentication parameters

## SQL Server authentication values

- `sql-password` (default), optionally with `domain` for NTLM
- `windows` or `kerberos`, with ODBC encryption/certificate/SPN options
- `azure-sso`
- `azure-service-principal`
- `azure-cli`

## Verification

- 22 targeted unit tests passed
- 2 Electron launch E2E tests passed
- Electron main and renderer builds passed
- live Docker-backed CLI connections verified for PostgreSQL, MySQL, MariaDB, SQL Server, MongoDB, Redis, and ClickHouse
- integrated SQL Server auth was confirmed to select the ODBC path; this macOS host lacks the required unixODBC/msnodesqlv8 native dependency, so a real Kerberos login was not available locally

Closes #361


apps/studio/src-commercial/entrypoints/main.ts
```diff
@@ -112,55 +113,77 @@ log.debug("####################################")
...
+    menuHandler = new MenuHandler(electron, settings, bksConfig)
+    menuHandler.initialize()
+    log.debug("Building the window")
+    log.debug("managing updates")
+    manageUpdates(settings.useBeta.valueAsBool)
+    ipcMain.on(AppEvent.openExternally, (_e: electron.IpcMainEvent, args: unknown[]) => {
+      safeOpenExternal(args?.[0])
+    })
+    return settings
+  })()
```
not-night-but: This seems like a very unnecessary refactor? We were handling this the exact same way except with an async function and the initialized boolean before, and I think that is much easier to read than this

apps/studio/src-commercial/entrypoints/main.ts
```diff
@@ -112,55 +113,77 @@ log.debug("####################################")
...
+  if (runtimePromise) return runtimePromise
+
+  runtimePromise = (async () => {
+    initializeSecurity()
+    initializeFileHelpers()
+    await createUtilityProcess()
+  })()
+
+  return runtimePromise
+}
```
not-night-but: I'd rather this just be an actual async function than returning a promise like this. 
```suggestion
let runtimeInitialized = false;
async function initRuntime(): Promise<void> {
  if (runtimeInitialized) return;
  runtimeInitialized = true;

  initializeSecurity();
  initializeFileHelpers();
  await createUtilityProcess();
}
```

PR #4576, opened 2026-08-05
feat: track query origins in query history

### LINKED ISSUE #3802: BUG: Queries executed via AI Shell are not recorded in Query History
Image

**TODO**

1. Plugins use `runQuery` API to run queries, so make sure that we record that when the API call is completed.
2. After that, we'll need to be able to tell if a query is originated from the plugin or from the app.
3. It'd be nice to have a dropdown that filters the origin too.

**Question**: What will happen if we double click a query that's originated from a plugin?

All can be done in one PR or separate PRs.

bnquon:
- plugin calls to `runQuery` are now saved to query history with their origin and plugin ID.
  - added migration to support this
- history entries record whether they originated from the app or a plugin.
- added origin icons and a dropdown for filtering local history.
- existing local history entries default to the app origin.
- double-clicking a plugin history item opens it normally in the query editor.
  - rerunning a plugin history item from the normal editor records it as an app query (shown in demo).

### cloud workspace behavior

- plugin queries still go through the existing history save path for cloud workspaces and appear in cloud history.
- the cloud backend currently does not preserve or return origin metadata.
  - cloud history items therefore use the generic code icon in the app.
  - the origin dropdown is hidden for cloud workspaces because filtering is not currently supported there.

the cloud-specific fallback is documented in the relevant history files and can be removed once the cloud api stores+returns origin metadata.

for plugins, history saving runs in the background so it doesn’t slow down AI/plugin query responses and failures are logged.

tests were added for plugin history saving, origin filtering, origin icons, and local origin persistence.

demo: tracking **local** app and plugin queries in history with dropdown interactions

https://github.com/user-attachments/assets/96f30a28-b8de-488d-af6a-188d60ffeb4b

closes #3802


apps/studio/src/components/sidebar/core/HistoryList.vue
```diff
@@ -151,6 +187,12 @@ import SidebarLoading from '@/components/common/SidebarLoading.vue'
           return text
         }
       },
+      originInfo(item) {
+        return this.originOptions.find(origin => origin.value === item.origin) || {
+          label: 'Unknown',
+          icon: 'code'
+        }
+      },
```
not-night-but, line 216: I think it might be nice to also be able to match on plugin id here? Specifically for the case of the AI Shell. Then use the same icon that we use for the AI Shell tab if that's the origin

apps/studio/src/components/sidebar/core/HistoryList.vue
```diff
@@ -27,6 +27,32 @@
           <input type="checkbox" id="show-all-history-checkbox" v-model="showAllHistory">
           <label for="show-all-history-checkbox" class="show-all-history-text">Show all</label>
         </div>
+        <!-- Cloud history does not include query origin metadata because the cloud backend does not support it. -->
+        <div
```
not-night-but, line 31: I'm not sure I love how this looks with the show all button above, we might want to make some changes to show all. Thoughts @azmy60

apps/studio/src/common/interfaces/QueryOrigin.ts
```diff
@@ -0,0 +1,10 @@
+export type QueryOrigin = 'app' | 'plugin'
+
+export const QUERY_ORIGIN_OPTIONS: ReadonlyArray<{
+  value: QueryOrigin
+  label: string
+  icon: string
+}> = [
+  { value: 'app', label: 'App', icon: 'code' },
+  { value: 'plugin', label: 'Plugin', icon: 'extension' }
+]
```
not-night-but, line 41: Yeah I think add the AI Shell and maybe also the ERD as separate origins from Plugin? I think maybe we treat the plugin origin as a fallback.

apps/studio/src/common/interfaces/IUsedQuery.ts
```diff
@@ -7,4 +8,7 @@ export default interface IUsedQuery {
   updatedAt: Date | number | null
   connectionId: number | null
   queryId: number | null
+  // The cloud backend does not support query origins, so cloud history may omit these fields.
```
not-night-but, line 11: Once we get this merged I can add this to cloud as well

not-night-but (changes requested):
Okay first of all, so sorry for missing this originally, appreciate you keeping it updated.

Second, I love this. I think this will work well going forward with some of the changes we have planned (working on doing a proper launch of the plugin system soon).

I left a couple suggestions. Mainly I think we should have the ability to match origin on specific plugin IDs. Right now that would just be for bks-ai-shell and bks-er-diagram (don't quote me on those ids lol). In the future that would allow us to add entries for any other bundled or official plugins. 

Also a note: CI is failing because it looks like 3 new test files were added. We're in the middle of moving all our tests from jest to vitest, so any new jest files will trigger an error. You can look in `apps/studio/tests/VITEST_MIGRATION.md` for more details, but for the most part all you need to do is move the unit tests into `apps/studio/tests/vitest/unit/` and make sure to use the vitest equivalents of `describe`, `test`, and so on.

Again, thanks so much for the contribution!!

apps/studio/src/common/interfaces/QueryOrigin.ts
bnquon: hi thanks a lot for the feedback! just to clarify what you mean by treating Plugin as a fallback, should the filters work like:
1. Plugin only shows other/unrecognized plugin queries, while AI Shell and ER Diagram have their own separate filters 

or

2. Plugin shows all plugin queries, including AI Shell and ER Diagram, while AI Shell and ER Diagram can also be filtered individually?

apps/studio/src/components/sidebar/core/HistoryList.vue
bnquon: yep, done — AI Shell and ER Diagram are now matched by plugin ID and use their own icons.

apps/studio/src/components/sidebar/core/HistoryList.vue
bnquon: yeah, i can see that. i left this part as-is for now while we wait for @azmy60's thoughts.

apps/studio/src/common/interfaces/QueryOrigin.ts
bnquon: i went ahead with the first option for now: Plugin only handles unrecognized plugin IDs, while AI Shell and ER Diagram have their own filters. happy to adjust if you meant the second one.

not-night-but: @bnquon Yes sorry I did mean the first one

apps/studio/src/components/sidebar/core/HistoryList.vue
azmy60: Agreed. Honestly, I forgot what "show all" is for? Maybe we could merge it with the new dropdown and call it "Filter"?

not-night-but: Show All is to show used queries from all connections, not just the current one. I mean if the dropdown could be a multi select then that might work? They are two different things though. Ones the source of the query and the other is the destination so idk

PR #4586, opened 2026-08-07
only poll cloud modules on the screen that shows them

azmy60:
- poll connections when not connected
- poll queries when connected

### DIFF
--- a/apps/studio/src/store/modules/data/query/CloudQueryModule.ts
+++ b/apps/studio/src/store/modules/data/query/CloudQueryModule.ts
@@ -39,6 +39,10 @@ export const CloudQueryModule: DataStore<ISavedQuery, State> = {
       // noop
     },
     async poll(context) {
+      if (!context.rootState.connected) {
+        return
+      }
+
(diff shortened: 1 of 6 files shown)


apps/studio/src/store/modules/data/query/CloudQueryModule.ts
```diff
@@ -39,8 +39,10 @@ export const CloudQueryModule: DataStore<ISavedQuery, State> = {
       // noop
     },
     async poll(context) {
-      const expandedFolderIds = context.rootState.sidebar.queries.expandedIds
-      await context.dispatch('loadByParentIds', expandedFolderIds)
+      if (context.rootState.connected) {
+        const expandedFolderIds = context.rootState.sidebar.queries.expandedIds
+        await context.dispatch('loadByParentIds', expandedFolderIds)
+      }
```
not-night-but: Ideally this should only poll when the query sidebar panel is open

not-night-but (approved):

PR #4588, opened 2026-08-07
fix saving query payload and poll saved query

not-night-but:
Slim down the payload for the save request so we don't write anything inadvertently.

When tab is active, poll for changes. If no unsaved changes and there are remote changes, update the editor value and send a noty. If there are both remote and local changes, merge editor pops up. If remote is deleted, local state gets wiped and a notice is shown.

### DIFF
--- a/apps/studio/src/components/TabQueryEditor.vue
+++ b/apps/studio/src/components/TabQueryEditor.vue
@@ -672,7 +672,9 @@ import { KeybindingPath } from '@/common/bksConfig/BksConfigProvider'
         },
         editingResult: false,
         resultsEditData: [],
-        resultEditableMap: []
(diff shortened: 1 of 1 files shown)


apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -1982,6 +1992,53 @@ import { KeybindingPath } from '@/common/bksConfig/BksConfigProvider'
             },
           }
         };
+      },
+      maybePollOriginalText() {
+        if (this.tab.queryId && this.isCloud) {
```
azmy60, line 1997: I'd check if `this.pollInterval` is already defined here.

azmy60 (approved):

PR #4598, opened 2026-08-09
Improve TableLength record count fetching

### LINKED ISSUE #311: BUG: Slow render time when opening table in tab due to counting rows
**Describe the bug**

When previewing the content of the table, the tab is very slow to load (on the order of 10+ seconds) if the table is large (has millions of rows). This is because BeeKeeper is issuing a `SELECT COUNT(*) from table;` statement to display the number of rows at the bottom left corner of the tab, and this query takes several seconds to run. A simple select against the table runs fast (`SELECT * FROM table LIMIT 1000;`).

I am connecting to a MySQL 5.7.22 database and counting rows on an InnoDB table.

**To Reproduce**
1. Navigate to a table with millions of rows and click the "Open in a new tab" button

**Expected behavior**
I would expect the table preview to load quickly, and then the table row count to load in the lower left corner after a delay once all the rows have been counted. Perhaps there could be an option to not produce table row counts, since `SELECT COUNT(*) from table` runs slowly and generates work for the database.

 - OS: Mac
 - App Version: 1.7.4

**Additional context**
I am connecting to a MySQL 5.7.22 database and counting rows on an InnoDB table.

### LINKED ISSUE #1179: BUG: Row count is wrong when table filter is on
**Describe the bug**
The total number of rows is not getting updated when filtering the results

**To Reproduce**
1. Go to a table with more than 1 record
2. Filter so you'll have 1 record match
3. See the total number of rows in the left bottom
4. See the total number is representing the total rows count instead of the total filtered rows count

**Expected behavior**
Get the total filtered rows count

 - OS: Mac
 - App Version 3.4.1 (3.4.1)
 - Database type: Postgres

### LINKED ISSUE #1684: FEAT: get total of filtered records
When I filter the records using the panel I don't see how many records are shown. I do get the total records in the status bar....but I want to know how many has been filtered

### LINKED ISSUE #2825: FEAT: Auto fetch number of records
Automatically display the total number of records without requiring an extra click or make it an optional setting.

ylyra:
## Summary

- Auto-fetch record count on mount and when table/filters change
- In-memory cache keyed by table + filters to avoid duplicate COUNT(*) on remount
- Stale-response guard when fetches overlap
- Manual click still refreshes (bypasses cache)
- Unit tests for fetch, cache, filter/table changes, and race conditions

## Related issues

Resolves #2825
Resolves #1684
Resolves #1179

## Performance / trade-offs

Historically, auto-fetch was removed in #739 (fix #311) because `COUNT(*)` can be expensive on large tables. This PR:

- Does **not** block tab loading (fetch is async in the status bar)
- Reuses cached counts across remounts for the same table/filters
- Keeps click-to-refresh as an explicit opt-in to re-run the query

Should auto-fetch on mount be behind a per-connection setting (as suggested in #2825 comments), with default off?

## Test plan

- [ ] Open a table tab — count loads automatically
- [ ] Apply/remove filters — count updates for filtered total
- [ ] Close and reopen same table tab — count comes from cache (no extra query until click)
- [ ] Click the count badge — forces refresh
- [ ] Large table — tab opens quickly; count may load later
- [ ] `yarn test:unit -- TableLength.spec.ts` passes

[asciicast](https://github.com/user-attachments/assets/5b741ebd-60de-4900-a572-3c8d58e246a5)


rathboma:
@ylyra I'm open to making this change, but I think it requires a little investigation into which databases this is expensive for and which it is cheap for. We should enable for cheap, disable for expensive.

By cheap I mean ~O(1), expensive I mean >= O(log n).

ylyra:
hi @rathboma, i’ll look into which databases this integration might be more expensive for and which ones could be cheaper. Thanks for the feedback!

ylyra:
@rathboma Investigated getTableLength() cost per client.

Now auto-fetch is on only where the count is ~O(1): Postgres/Greengage (reltuples) and MySQL family (SHOW TABLE STATUS). Everything else stays off.

New flag: autoFetchRecordCount (default false). 

Click-to-refresh still works on all databases.

rathboma:
When someone is filtering a table - does it still stay on? In that situation it's probably not O(1) right?

ylyra:
only if the user enables the new config `autoFetchFilteredRecordCount`, then it's not O(1) anymore

apps/studio/src/components/common/tableLengthCache.ts
```diff
@@ -0,0 +1,28 @@
+export type RecordCountCacheEntry = {
+  totalRecords: number | string | null,
+  error: Error | null,
+}
+
+const recordCountCache = new Map<string, RecordCountCacheEntry>()
+
+export function buildRecordCountCacheKey(
```
rathboma, line 8: So you can actually use TableTable's `tableId` which is basically the same thing - a unique key that we use to save configuration related to the specific table.

You need to account for workspace/connection/schema/table -- and sometimes workspace/connection can be null because it's a "quick connect" - we don't remember anything for a quick connect.

apps/studio/src/components/common/TableLength.vue
```diff
@@ -15,47 +15,130 @@
...
+        if (!newKey || newKey === oldKey) return
+        this.scheduleFetch()
+      },
+      immediate: true,
+    },
+  },
   methods: {
-    async fetchTotalRecords() {
+    applyCachedResult(cacheKey: string) {
+      const cached = getCachedRecordCount(cacheKey)
```
rathboma, line 78: So the cache is basically just a Map, right?

One issue is - this cache lives forever (so long as the connection is open), we probably don't want it to live forever.

Do we even need the cache? We're only enabling length fetching for fast databases anyway, feels like we don't need the cache, or if we do it should only be short, like 10s or something

rathboma (changes requested):
Sorry for the back-and-forth, big question about the cache, both the key and the value of it.

apps/studio/src/components/common/TableLength.vue
ylyra: Yes. 

If that's okay, I can add a cache of 30 seconds or 1 minute. 

The cache is being used as a safety measure to prevent multiple requests when switching between tabs too quickly.

apps/studio/src/components/common/tableLengthCache.ts
ylyra: Changed to use tableId

PR #4599, opened 2026-08-09
fix(table): use filtered count when jumping to last page

ylyra:
## Summary
- `jumpToLastPage()` now uses `getFilteredDataCount` when filters are active
- Matches existing logic in `TableLength.vue`
- `Math.max(1, ...)` avoids page 0 when count is empty

## Test plan
- [ ] Large table, no filters → End goes to correct last page
- [ ] Filter reduces rows → End goes to last **filtered** page
- [ ] Raw SQL filter string → same behavior
- [ ] Filter with zero rows → stays on page 1

# Before

https://github.com/user-attachments/assets/569da52b-bc33-4d4f-acbc-6cfe946f6c51

# After

https://github.com/user-attachments/assets/23cdf145-76db-4989-86bf-972fe8fd25db


not-night-but:
@ylyra Thanks for the fix!

not-night-but (approved):

PR #4608, opened 2026-08-10
Fix for being unable to close tabs

rathboma:
- getting a tab close confirmation then dismissing with esc didn't clear the 'closing' flag
- This meant that no tab could be closed
- We now properly track confirmation modal state

### DIFF
--- a/apps/studio/src/components/common/modals/ConfirmationModal.vue
+++ b/apps/studio/src/components/common/modals/ConfirmationModal.vue
@@ -1,5 +1,9 @@
 <template>
-  <base-modal :name="id" @submit="confirm">
+  <base-modal
+    :name="id"
+    @submit="confirm"
+    @closed="reportClose(false)"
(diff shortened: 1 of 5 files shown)


not-night-but (approved):

apps/studio/src/components/common/modals/ConfirmationModal.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <base-modal :name="id" @submit="confirm">
+  <base-modal
+    :name="id"
+    @submit="confirm"
+    @closed="reportClose(false)"
```
azmy60, line 5: Not sure if it'll cause issues; this OK if we close the modal by clicking outside of the modal, but `reportClose()` is called twice if we click the "cancel" or "confirm" button.

azmy60 (approved):

PR #4612, opened 2026-08-11
fix(editor): show hours field at exactly one hour in query elapsed timer

YuriNachos:
## Summary
- The query elapsed-time formatter (`apps/studio/src/lib/time/formatSeconds.ts`) used a strict `seconds > 3600` check to decide whether to render the `HH:` prefix.
- At exactly 3600 seconds (1 hour) the prefix was dropped, so the timer on the query status bar briefly rendered `00:00` before flipping to `01:00:01` one tick later.
- Changed the bound to `>=` so the one-hour boundary renders as `01:00:00`. The elapsed counter in `TabQueryEditor.vue` increments by whole seconds, so this only ever received integers; the off-by-one was purely the inclusive-vs-exclusive bound.

## Test plan
- [x] Added `apps/studio/tests/unit/lib/time/formatSeconds.spec.ts` covering sub-minute, sub-hour, and the one-hour boundary.
- [x] Confirmed the new test fails against the old code (`formatSeconds(3600)` returned `00:00`) and passes with the fix.
- [x] `npx eslint src/lib/time/formatSeconds.ts tests/unit/lib/time/formatSeconds.spec.ts` — clean.
- [x] `yarn workspace beekeeper-studio test:unit -- tests/unit/lib/time/formatSeconds.spec.ts` — 2 passed.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/apps/studio/src/lib/time/formatSeconds.ts
+++ b/apps/studio/src/lib/time/formatSeconds.ts
@@ -2,5 +2,5 @@ export default (seconds: number) => {
   const h = Math.floor(seconds / 3600).toString().padStart(2, '0');
   const m = Math.floor((seconds % 3600) / 60).toString().padStart(2, '0');
   const s = (seconds % 60).toString().padStart(2, '0');
-  return `${seconds > 3600 ? `${h}:` : ''}${m}:${s}`;
+  return `${seconds >= 3600 ? `${h}:` : ''}${m}:${s}`;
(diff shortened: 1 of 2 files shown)


not-night-but:
@YuriNachos Thanks for the fix!

not-night-but (approved):

PR #4629, opened 2026-08-13
Decouple release publishing from build workflow

rathboma:
## Summary

Moves release publishing tasks (deb/rpm repository uploads, snap store publishing, and downloads mirroring) from the build workflow (`studio-publish.yml`) to a new dedicated workflow (`release-published.yml`) that triggers when a draft release is manually published. This ensures no channel ships before the manual publish click and allows independent retry/recovery of publishing steps.

## Key Changes

- **New workflow**: `release-published.yml` handles all user-facing publishing:
  - Mirrors release assets to R2 downloads bucket and updates `latest.json` manifest
  - Publishes deb/rpm packages to apt/rpm repositories (stable channel only)
  - Uploads snaps to the Snapcraft store (edge for all releases, stable for latest only)
  - Supports `workflow_dispatch` for idempotent re-runs and backfilling

- **New script**: `mirror_release_downloads.sh` downloads GitHub release assets and mirrors them to R2, with intelligent manifest generation that respects GitHub's canonical "latest" release

- **Updated `studio-publish.yml`**:
  - Removed `publish_snapcraft` and `publish_repositories` jobs
  - Removed artifact upload step for deb/rpm/snap files (no longer needed for inter-job transfer)
  - Kept build and draft release creation; publishing now happens separately

- **Updated `extract_channel.sh`**: Now accepts version as a parameter (for release-published.yml usage) while maintaining backward compatibility with package.json fallback

## Implementation Details

- Serialized concurrency in `release-published.yml` ensures near-simultaneous publishes converge on GitHub's canonical latest release
- Downloads mirroring is safe to re-run via `workflow_dispatch` — uploads overwrite and manifest is written last
- Repository publishing (deb/rpm) only runs for the "latest" channel, determined by tag format (v1.2.3 → latest, v1.2.3-beta.1 → beta)
- Snap uploads go to edge for all releases; stable promotion only happens for latest channel

https://claude.ai/code/session_01RLJ3hsY7SBCCAnM5neCAY1


.github/scripts/mirror_release_downloads.js
```diff
@@ -0,0 +1,159 @@
...
+// (web repo: _assets/js/lib/github.js GithubAsset + controllers/download.js).
+function assetEntry(asset, urlBase) {
+  const name = asset.name
+  const extension = name.split('.').pop()
+
+  let arch = 'x86_64'
+  if (name.includes('arm64') || name.includes('aarch64')) {
+    arch = 'arm64'
+  } else if (name.includes('armhf')) {
+    arch = 'armhf'
```
rathboma: we can totally remove support for arm32bit, we don't build it anymore

.github/scripts/mirror_release_downloads.js
```diff
@@ -0,0 +1,159 @@
...
+      console.log(`${progress}: downloading`)
+      run('gh', ['release', 'download', tag, '--repo', REPO, '--dir', workdir, '--pattern', asset.name])
+      console.log(`${progress}: uploading`)
+      upload(
+        path.join(workdir, asset.name),
+        `s3://${R2_BUCKET}/${prefix}/${asset.name}`,
+        asset.contentType,
+        'public, max-age=31536000, immutable'
+      )
+      console.log(`${progress}: done`)
```
rathboma, line 123: should we delete the file after upload succeeds so as not to blow past actions storage?

.github/scripts/mirror_release_downloads.js
```diff
@@ -0,0 +1,159 @@
...
+    })
+
+    // Only the release GitHub reports as latest may write the manifest. This
+    // one rule covers prereleases, re-published old versions, and
+    // near-simultaneous publishes: drafts and prereleases are never "latest",
+    // and whatever order concurrent runs finish in, the manifest converges on
+    // the canonical latest.
+    const latestTag = capture('gh', ['api', `repos/${REPO}/releases/latest`, '--jq', '.tag_name'])
```
rathboma: This should be moved to a separate step in the workflow, so we fail early if we can't get this info. This script should only be mirroring the release files.

.github/workflows/release-published.yml
```diff
@@ -0,0 +1,197 @@
...
+      # and the mirror script itself only mirrors. latest.json is written only
+      # when the published tag matches this.
+      - name: Determine latest release
+        id: latest_release
+        run: |
+          latest=$(gh api repos/beekeeper-studio/beekeeper-studio/releases/latest --jq '.tag_name')
+          echo "Latest release: $latest"
+          echo "tag=$latest" >> "$GITHUB_OUTPUT"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
rathboma, line 84: Do you need this, or can you use the built-in GH workflow permission system to provision access to the api?

.github/workflows/release-published.yml
```diff
@@ -0,0 +1,197 @@
...
+          # 2025-01-15: AWS made changes to their clis that broke compatibility
+          # for third party services.
+          # This works around that.
+          # https://github.com/aws/aws-sdk-ruby/issues/3166
+          AWS_REQUEST_CHECKSUM_CALCULATION: WHEN_REQUIRED
+          AWS_RESPONSE_CHECKSUM_VALIDATION: WHEN_REQUIRED
+
+  publish_repositories:
+    needs: identify_channel
+    if: needs.identify_channel.outputs.channel == 'latest'
```
rathboma, line 106: can't we just do the same thing for the r2 mirror? Do we need to check if there's a later version available? If this is latest aren't we on the latest release guaranteed?

.github/workflows/release-published.yml
rathboma: It's needed — the `permissions: contents: read` block at the top is the built-in system and is already what scopes this token, but it only scopes the auto-generated `GITHUB_TOKEN`; it doesn't expose it to processes. The `gh` CLI specifically requires the token via a `GH_TOKEN` env var in Actions (it errors out with "set the GH_TOKEN environment variable" otherwise), so this line just hands the same built-in, permissions-scoped token to `gh` — no extra credential involved. Same pattern as the `GH_TOKEN` env on the build steps in studio-publish.yml.

---

.github/workflows/release-published.yml
rathboma: `channel == 'latest'` only says the tag is a stable version (not beta/alpha) — not that it's the *newest* release. They diverge when an older stable version is published (or re-published via workflow_dispatch) after a newer one exists, e.g. a v6.0.x hotfix after v6.1 ships: its channel is still "latest", but writing latest.json would point every website download at the old version.

deb/rpm don't need the guard because apt/yum repos are versioned pools — uploading an older package doesn't downgrade anyone. latest.json is a single last-write-wins pointer, so it gets the one extra API call. That check also honors the "Set as latest release" checkbox on publish (GitHub's `releases/latest` is exactly what that checkbox controls), and makes two stable releases published back-to-back converge on the right manifest regardless of run order.

---

PR #4647, opened 2026-08-14
fix(sidebar): drag-and-drop reordering is discarded on the next render

andreasarf:
Dragging a connection or saved query in the sidebar looks like it does nothing — the item lifts, you drop it, and it snaps back.

The drop actually saves fine; `reorder` writes the new `position` to the database. The trouble is that `sortedItemNodes` builds the rendered list from `sort.field` instead:

```js
result = _.orderBy(this.itemNodes, `ref.${this.sort.field}`)
```

`position` never gets a say, so the order you just saved is thrown away on the next render. Cloud workspaces already sorted by `position`, so this only shows up locally.

### Why it broke

Sorting here has always been a one-shot action rather than a live view sort — `reorderBySort` rewrites `position` for every connection and offers an Undo. The order on screen was `position`.

Before the tree rewrite, the old `sortedConnections` computed had a single consumer, and that consumer re-sorted by `position` anyway, so its ordering was never actually used. When `67d8043a1` wired it into the tree as `sortedItemNodes`, that unused ordering quietly became the real one.

### The fix

Render by `position` in both sidebars.

Sorting by Colour still works — the colour map it needs lives in `applySortOrder`, which is what `reorderBySort` calls. The copy removed here was a duplicate. Cloud behaviour is unchanged, since it already sorted by `position`.

### One thing to watch

With sort set to Name, the list now shows saved `position` order instead of re-alphabetizing on every render. That's how it worked before 6.0, but if someone's stored positions are stale, the list may not look alphabetical until they pick a sort again.

### Testing

Added `SidebarRenderOrder.spec.ts` covering position vs name/title, position vs colour rank, missing positions, and ties. I checked these actually catch the bug — putting the old code back fails all four connection cases. Full suite passes (88 suites, 3158 tests).

### DIFF
--- a/apps/studio/src/components/sidebar/ConnectionSidebar.vue
+++ b/apps/studio/src/components/sidebar/ConnectionSidebar.vue
@@ -469,29 +469,12 @@ export default {
       return this.connectionsPollError || this.foldersPollError || null
     },
     sortedItemNodes() {
-      // Cloud has no sort buttons — drag and drop is the only way to reorder,
-      // and it lands in `position`.
-      if (this.isCloud) {
-        return _.sortBy(this.itemNodes, 'ref.position')
-      }
-      let result = []
-      if (this.sort.field === 'labelColor') {
-        const mappings = {
-          default: -1,
-          red: 0,
-          orange: 1,
-          yellow: 2,
-          green: 3,
-          blue: 4,
-          purple: 5,
-          pink: 6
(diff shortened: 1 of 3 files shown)


not-night-but:
Oh yeah good catch! Thanks for the patch

not-night-but (approved):

PR #4648, opened 2026-08-14
Autocomplete defaults overhaul - restoring upper case

rathboma:
- match-case now the default for keyword casing
- upper/lower case can be forced in config
- Fixed identifier quoting for MySQL and other DBs that don't need quoted mixed-case entity names
- Allow customization of the quoting character
- Slight tweak to query format context menu to make it cohesive
- Updated the docs for autocomplete


apps/studio/default.config.ini
```diff
@@ -45,6 +45,20 @@ defaultFormatter = bk-default
...
+; preserve = follow the typed prefix: SEL -> SELECT, sel -> select;
+;            completing with no prefix (e.g. Ctrl+Space) inserts uppercase
+; upper    = always uppercase
+; lower    = always lowercase
+keywordCasing = preserve
+; When completed table/column names get quoted.
+; auto   = only names the database can't reference unquoted: special
+;          characters always; MixedCase only where unquoted names are
+;          case-folded (e.g. PostgreSQL)
+; always = also quote any name that isn't all-lowercase, on every database
```
not-night-but: This is a bit confusing, because I would assume that always would *always* quote identifier autocompletions

apps/ui-kit/lib/components/sql-text-editor/SqlTextEditor.vue
```diff
@@ -83,22 +86,36 @@ export default Vue.extend({
...
+      // "Custom..." hands off to the embedding app (e.g. a formatter modal).
+      const customItems = this.allowPresets
+        ? [
+            {
+              label: "Custom...",
+              id: "format-custom",
+              handler: () => this.$emit("bks-open-custom-formatter"),
+            },
+          ]
+        : [];
```
not-night-but: This doesn't seem super necessary? Right below Format Query we have Open Query Formatter:

Image

Custom... and Open Query Formatter do the same thing

apps/studio/default.config.ini
rathboma: I see, yes probably right

apps/ui-kit/lib/components/sql-text-editor/SqlTextEditor.vue
rathboma: I meant to remove 'open query formatter', my bad

not-night-but: Yeah I think that makes the most sense

rathboma: I just reverted this change

not-night-but (approved):

PR #4649, opened 2026-08-15
fix: show empty state and disable 'more' in connection switch popover

### LINKED ISSUE #4150: BUG: The switch connection pop-out needs an empty state when there are no connections you can switch to

Hyhyhyyy:
## Summary
Fixes #4150.

The "Switch Connection" popover (the quick switcher opened from the connection button menu) had two rough edges:

- When there are no other connections to switch to, the list was just blank. This adds a friendly empty-state message instead.
- The "more" button (which toggles between recent and all saved connections) stayed clickable even when there was nothing more to reveal. It is now disabled in that case.

## Changes
- `ConnectionButton.vue`: render a `quick-switcher-empty` element when `displayedConnections` is empty, and bind `:disabled` on the "more" button to a new `hasMoreConnections` computed property.
- `sidebar.scss`: add styling for `.quick-switcher-empty` and the disabled state of `.quick-switcher-item`.

## How to test
1. Open the connection menu on a tab and choose **Switch Connection**.
2. With only the current connection available, the popover shows "No other connections to switch to" and the "more" button is disabled.
3. With additional saved connections, the list shows them and "more" toggles to the full saved list as before.

### DIFF
--- a/apps/studio/src/assets/styles/app/sidebar/sidebar.scss
+++ b/apps/studio/src/assets/styles/app/sidebar/sidebar.scss
@@ -701,3 +701,19 @@
   flex-grow: 1;
   max-height: calc(50vh - 70px);
 }
+
+.quick-switcher-empty {
+  padding: 12px 6px;
+  color: $text-lighter;
+  text-align: center;
+  font-size: 0.85rem;
+}
+
+.quick-switcher-item:disabled {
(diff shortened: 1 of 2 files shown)


azmy60:
Closing this as a duplicate of #4155

PR #4654, opened 2026-08-16
fix(sidebar): duplicate saved queries with full SQL text

ylyra:
## Summary
- Duplicate a saved query now loads the full record (including `text`) before cloning, matching Export.
- Save errors show a toast instead of failing silently.
## Test plan
- [ ] Right-click a saved query in the sidebar → Duplicate → a "Copy of …" query appears with the same SQL
- [ ] Duplicate a query inside a folder → the copy stays in that folder
- [ ] If save fails, an error toast appears


apps/studio/src/components/sidebar/core/FavoriteList.vue
```diff
@@ -616,10 +616,21 @@ export default {
...
-      cloned.title = 'Copy of ' + cloned.title
-      await this.$store.dispatch('data/queries/save', cloned)
-      this.$noty.success('Query duplicated')
+      try {
+        const fullQuery = await this.$store.dispatch('data/queries/findOne', query.id)
+        const cloned = await this.$store.dispatch('data/queries/clone', query)
+        cloned.title = 'Copy of ' + cloned.title
+        cloned.text = fullQuery.text
+        cloned.excerpt = fullQuery.excerpt ?? cloned.excerpt
+        const payload = _.omit(cloned, [
```
rathboma: I think this is fine overall, I'd prefer a `_.pick` to whitelist the fields you want - title, text, parent_id, position, vs an omit, as we'll end up adding more fields so this will drift.

rathboma (changes requested):
Let's move to opting in columns for duplication rather than opting out, then we're gtg

apps/studio/src/components/sidebar/core/FavoriteList.vue
ylyra: updated to use _.pick instead, thansk. 😁

PR #4660, opened 2026-08-18
add updated by status to query editor

azmy60:
https://github.com/user-attachments/assets/e64ae932-584b-4e24-affd-19e746abc430


rathboma:
To save space - I think you can just show the material icon (no name), then on hover, it shows a v-tooltip with the last edit. I'd also make the icon less bold (font-weight).

rathboma (approved):
Looks good, think just this one small change would help

PR #4664, opened 2026-08-18
Fix typing transaction commands in manual commit

not-night-but:
A typed commit/rollback command in manual commit mode would always rollback the transaction first. 

Typed commit/rollback in the query editor would also release the connection before running the queries so a race condition could happen

Sql Server is annoying and requires functions to be called instead of the transaction statements being run on the connection. We now remove those commands from the query and manually run the transaction functions, as well as run the commands sequentially


apps/studio/src/components/TabQueryEditor.vue
```diff
@@ -1655,7 +1655,7 @@ import { KeybindingPath } from '@/common/bksConfig/BksConfigProvider'
               this.enteredTransactionFromIdent = true;
               this.hasActiveTransaction = true;
             } else if (this.isManualCommit && this.hasActiveTransaction && endTransaction > startTransaction) {
-              await this.toggleCommitMode();
+              shouldToggle = true;
```
azmy60, line 1658: Looks like this is ignored if we show `parameters-modal-...` at line 1669?

azmy60 (changes requested):
Close to being approved. I'd check if the `shouldToggle` is stable enough.

apps/studio/src/components/TabQueryEditor.vue
not-night-but: @azmy60 When we show the modal we return from this function, and then it is called again on submit of the modal, so we shouldn't miss this at all

rathboma (approved):

azmy60 (approved):

PR #4665, opened 2026-08-19
Create a query by right clicking a folder

azmy60:
https://github.com/user-attachments/assets/c72a9672-aa7e-4374-8b71-e23b19617ac1


rathboma:
When I create a new query, I think I'd rather have it open a query tab and just name the query 'untitled query'. Focus the query editor. Creating in the sidebar with no content doesn't make much sense to me. If I'm creating a query, it's because I want to edit it right now.


I think the easiest change is:
- It works exactly as you have it right now
- After naming the query, it opens the query in a new tab for editing

rathboma (approved):
Fine to merge, one minor change I'd like to see - pressing `esc` when naming the query (or renaming a query) should discard the changes, same as clicking away does.

PR #4666, opened 2026-08-19
Create a connection by right clicking a folder

azmy60:


apps/studio/src/components/sidebar/ConnectionSidebar.vue
```diff
@@ -226,7 +226,10 @@
                   <tree-folder
                     v-bind="props"
                     v-else
-                    :class="{ 'just-created': justCreatedFolderId === props.node.ref.id }"
+                    :class="{
+                      'commited': commitedType === 'folder'
```
rathboma: committed has two `t`s :-)

rathboma (approved):
looks good, minor typo running through the sidebar, but we can ignore for now

rathboma (dismissed):
Ok actually I changed my mind. Same problem as the query workflow.

Here's what I expect:
Right click -> new connection

It doesn't _need_ to add it to the sidebar at this point, but it could add it as 'untitled'. Really all this should do is call 'new connection', but pre-fill the folder.

If I don't save it, it should disappear.

I don't think saving a new sqlite connection by default is what I would expect it to do, also it doesn't work - there's an error (local workspace) because we're breaking connection field requirements

rathboma:
I think this is harder than connection because we don't have tab-headers with save state like we do for queries

apps/studio/src/components/sidebar/ConnectionSidebar.vue
azmy60: lol :D

azmy60: English is hard

azmy60:
I'll implement your solution. But be aware that this is a UX issue for first time users I think. If I click the `+ New connection` button, nothing happens on the screen. It's the same thing if we do your solution, which is just `+ New connection` and then prefill the folder. This is just my personal experience when I used beekeeper for the first time.

azmy60:
@rathboma to be clear, your workflow is valid. But there's a deeper UX problem, which is probably why implementing this is harder for connection, and not for query.

rathboma:
@azmy60 so - how do we fix the UX issue? That's the real linchpin that unlocks this feature and makes it intuitive.

One idea that doesn't change the UX too much:
1. Empty state screen is not the new connection dropdown, but something else, like 'welcome to beekeeper studio'
2. Clicking 'new connection' opens a modal or a popover of the main content area for you to select the database engine or import a url. It has a 'cancel' button and a `(x)` button, so you can exit the creation flow. This can also have name/folder in it.
3. Then it pops you onto the normal connection editing screen. to fill in the details

azmy60:
@rathboma I'll play with that

azmy60:
I added a "Select connection type" modal and icons. Changes are minimal.

https://github.com/user-attachments/assets/2193f0c1-ffb5-4755-85f7-ee0f4162ee9b

rathboma:
1. Needs to work for users who are on a free version, and they select a paid connection type
2. It broke all the integration tests -- those need fixing (psst: have Claude do it)

azmy60:
Added a new component `UpgradeAlert.vue`.

image
image

not-night-but (approved):

Issue #2990, opened 2025-04-10
BUG: Vim Mode Doesn't Auto-Focus on New Tab


JeanPaiva42:
**Describe the bug**

When using Vim mode in the SQL editor, opening a new tab (Cmd+T) does not automatically focus the editor. The user has to click manually before typing. Interestingly, if you open another new tab and then close it, the editor in the previous tab regains focus.

To Reproduce

1. Open a database connection.
2. Enable Vim mode (through the gear icon on the bottom right of the screen)
3. Press Cmd+T (or Ctrl+T) to open a new tab.
4. Try to start typing immediately — nothing happens until you manually click the editor.
5. Open another new tab, then close it — now the editor in the previous tab is focused automatically.

**Expected behavior:**

- The SQL editor should automatically receive focus when opening a new tab, even in Vim mode, so users can immediately start typing without extra clicks.

**Screenshots / Gifs**
- N/A — reproducible through the steps above.

Version Information (please complete the following information):

OS: Mac 
App Version: 5.1.5

Database type and version: Happens regardless of DB type

Additional context

* This issue is specific to Vim mode. Without Vim mode enabled, the editor auto-focuses correctly when a new tab is opened.

Issue #2953, opened 2025-03-31
BUG: Vim: binding to "*y or "*p causes error


not-night-but:
**Describe the bug**
When trying to map a key to the system register y/p, we end up getting an error related to the maximum call stack size. This might just be a codemirror thing that we can't fix, but we should at least look into it.

**To Reproduce**
1. Open an editor in vim mode
2. Type `:` in normal mode to start entering a command
3. enter `nmap y "*y`
4. in normal mode, press y
5. Observe the 4000ish errors that are emitted

**Expected behavior**
We can bind keys to the system register so that the default behaviour is using the system register.

**Screenshots / Gifs**
![Image](https://github.com/user-attachments/assets/30e81c2c-67d3-4d98-89e1-4a021b0fa47a)

Issue #1930, opened 2024-01-31
VIM: :w does not trigger save on the correct tab


austinwilcox:
When I try to save the current tab I am editing, and I have another opened tab that is unsaved, it does not save my current file I am editing, it only tries to save the new tab that is on the right of my current tab I am editing. This is in VIM mode only with ```:w```.
![bug-with-save-in-vim](https://github.com/beekeeper-studio/beekeeper-studio/assets/43357911/d7b0073f-d2d9-4afd-a070-866ff2d2cad7)

austinwilcox:
This is another issue that I am happy to fix.

PR #4669, opened 2026-08-19
Vim mode: fix focus, ex commands, and .beekeeper.vimrc

### LINKED ISSUE #3446: BUG: CTRL+P no longer responses when in vim normal mode
When I'm using VIM mode and in NORMAL mode: 

Version: 5.2.12

Quick search popup will show up 

Version: 5.3.4

Nothing happens unless I click outside of text buffer/textarea

I also noticed that in version 5.3.4 the cursor will get stuck in normal mode and navigation keys no longer work at times but I'm unable to replicate this 100% of the time.

rathboma:
Vim mode has collected a few issues over the years, and the CodeMirror 6 migration added some more. This fixes the open ones and fills in some gaps.

**This needs hands-on testing.** The vimrc parser has unit tests, but the app itself could not be run where this was written, so nothing below has been confirmed in a real session. If you use vim mode, please work through the list and say which of these actually hold.

## Things that were broken

- **The editor no longer goes dead after you run a query.** Run a query, then tap `Esc` a few times in normal mode. It used to stop accepting keys entirely — the cursor turned into a hollow outline and you had to click back into the editor to recover. (#3446)
- **A new tab is ready to type in immediately.** Press `Cmd/Ctrl+T` and start typing. Vim mode used to need a click first, and the known workaround was opening a second tab and closing it again. Worth checking on the very first tab after a cold start, which was the worst case. (#2990)
- **`:w` saves the tab you are actually in.** Open two tabs with unsaved text and `:w` in the first one. It used to save whichever tab was opened most recently. (#1930)
- **`Ctrl+P` opens quick search again from normal mode.** Vim binds `Ctrl+P` to "move up", which took the shortcut over during the CodeMirror 6 move. `k` still moves up. If you would rather have vim's binding back, `nnoremap <C-p> k` in your `.beekeeper.vimrc` reclaims it. (#3446)
- **`"*y` and `"*p` should work consistently**, including in the Mongo shell tab where they previously did nothing at all.

## New

- **The editor has a status line along the bottom.** It shows `--NORMAL--` / `--INSERT--` / `--VISUAL--`, any keys waiting on the rest of a command, and count prefixes. `:` commands and `/` searches are typed there now too. Please flag if it looks wrong in your theme.
- **`Esc` cancels a running query again.** Only from normal mode with nothing pending — `Esc` out of insert or visual mode just returns you to normal mode, as it should. `Ctrl-Esc` still cancels from any mode.

## `.beekeeper.vimrc`

- **`noremap`, `nnoremap`, `inoremap` and `vnoremap` now work.** `nnoremap y "*y` is the supported way to make a plain `y` use the system clipboard. Written as `nmap y "*y` it used to emit thousands of errors; that mistake is now reported as a single message instead. (#2953)
- **Comments and ordinary whitespace no longer break the file.** Lines starting with `"` are comments, and tabs or runs of spaces between fields are fine. Previously anything other than a single space was rejected, and a comment line was treated as a broken mapping.
- **`unmap`, `nunmap`, `iunmap`, `vunmap` and the `mapclear` family are supported.**
- **`set` is supported** — `set ignorecase`, `set nonumber`, `set tabstop=4`.
- **`<leader>` is supported**, with `let mapleader = ","` (or `g:mapleader`). Defaults to `\`.
- **A line that cannot be parsed now tells you so**, in a notification naming the line number. It used to fail silently unless you had the developer console open. The rest of the file still applies.
- **Your vimrc is read once at startup** rather than re-read for every tab you open.

## Ex commands

These already existed but were never written down: `:w`, `:q`, `:qa`, `:x`, `:wq`, `:tabnew [name]`. They are in the docs now.

## Please poke at

- Focus generally — clicking between the editor and the results grid, switching tabs, `Ctrl+L`. The focus fix is the largest behavioural change here.
- **Non-vim users too:** the focus fix applies to the default keymap as well, so it is worth a sanity check with vim mode off.
- Whether the status line is readable in both light and dark themes.
- Whether `Esc` ever cancels a query when you did not mean it to.
- Anything in your existing `.beekeeper.vimrc` that behaves differently.

Docs: `docs/user_guide/sql_editor/editor.md`.

Closes #3446, #2990, #1930, #2953

---
_Generated by [Claude Code](https://claude.ai/code/session_016tFqRXyPEpgUJYGMU83wjF)_

### DIFF
--- a/apps/studio/src/lib/editor/vim.ts
+++ b/apps/studio/src/lib/editor/vim.ts
@@ -1,186 +1,237 @@
-import Vue from "vue";
+export type VimMode = "normal" | "insert" | "visual";
 
+/**
+ * A directive parsed out of `.beekeeper.vimrc`. The shapes match the
+ * `vimKeymaps` prop of the ui-kit text editor, which applies them in order.
+ */
 export type IMapping = {
+  type?: "map";
   mappingMode: string;
   lhs: string;
   rhs: string;
-  mode: string;
+  mode?: VimMode;
+  /** `noremap` and friends: don't re-expand the rhs through other mappings. */
+  noremap?: boolean;
 };
 
-interface Config {
-  exCommands?: { name: string, prefix: string, handler: () => void }[];
+export type IUnmapping = {
+  type: "unmap";
+  lhs: string;
+  mode?: VimMode;
+};
+
+export type IMapClear = {
+  type: "mapclear";
+  mode?: VimMode;
+};
+
+export type IVimOption = {
+  type: "set";
+  name: string;
+  value: string | boolean;
+};
+
+export type VimDirective = IMapping | IUnmapping | IMapClear | IVimOption;
+
+export interface VimrcParseError {
+  /** 1-based, so it lines up with what an editor shows. */
+  line: number;
+  text: string;
+  reason: string;
 }
 
-export function applyConfig(codeMirrorVimInstance: any, config: Config) {
-  const { exCommands } = config;
-  if (exCommands) {
-    exCommands.forEach(({ name, prefix, handler }) => {
-      codeMirrorVimInstance.defineEx(name, prefix, handler);
-    });
+export interface VimrcParseResult {
+  directives: VimDirective[];
+  errors: VimrcParseError[];
+}
+
+const MAP_COMMANDS: Record<string, { mode?: VimMode; noremap: boolean }> = {
+  map: { noremap: false },
+  nmap: { mode: "normal", noremap: false },
+  imap: { mode: "insert", noremap: false },
+  vmap: { mode: "visual", noremap: false },
+  noremap: { noremap: true },
+  nnoremap: { mode: "normal", noremap: true },
+  inoremap: { mode: "insert", noremap: true },
+  vnoremap: { mode: "visual", noremap: true },
+};
+
+const UNMAP_COMMANDS: Record<string, VimMode | undefined> = {
+  unmap: undefined,
+  nunmap: "normal",
+  iunmap: "insert",
+  vunmap: "visual",
+};
+
+const MAPCLEAR_COMMANDS: Record<string, VimMode | undefined> = {
+  mapclear: undefined,
+  nmapclear: "normal",
+  imapclear: "insert",
+  vmapclear: "visual",
+};
+
+const DEFAULT_LEADER = "\\";
+const MAPLEADER = /^let\s+(?:g:)?mapleader\s*=\s*(.+)$/;
+const LEADER = /<leader>/gi;
+
+function unquote(value: string): string {
+  const trimmed = value.trim();
+  const quote = trimmed[0];
+  if ((quote === '"' || quote === "'") && trimmed.endsWith(quote) && trimmed.length > 1) {
+    return trimmed.slice(1, -1);
   }
+  return trimmed;
 }
 
-async function readVimrc(): Promise<string[]> {
-  const data: string | null = await Vue.prototype.$util.send('config/readVimrc');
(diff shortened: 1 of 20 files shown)


apps/studio/src/lib/editor/vim.ts
```diff
@@ -1,186 +1,237 @@
...
+    result.push(mapping);
+  });
 
+  return result;
+}
 
+/** @deprecated Use `parseVimrc`, which also reports what it could not parse. */
+export function createVimCommands(vimrcContents: string[]): VimDirective[] {
+  return parseVimrc(vimrcContents).directives;
+}
```
not-night-but: Why keep the old stuff around?

not-night-but (changes requested):
I think we've left some vestigial code that really doesn't need to be there, other than that, not totally sure how to review this other than express my growing hatred for comments 😆 

Looks like there's some issues with the status line in dark mode:

Image

And when typing a command:
Image

Also, redo doesn't seem to work? `u` (undo) works fine, but `<C-r>` does not redo as it should. It's bound in the default bindings in the vim package: https://github.com/replit/codemirror-vim/blob/8640966b6977f84d2197e6adfd521fb737184587/packages/codemirror-vim-core/vim.js#L234

Also if you open the quick search and then close it, focus does not return to the editor, but this isn't a vim mode thing.

rathboma:
Thanks — pushed 375caad for the parts I could act on.

**Vestigial code.** Removed `resetAppliedKeymaps` (nothing called it) and the `createVimCommands` shim, whose only remaining caller was a test asserting the shim still existed. The spec uses `parseVimrc` directly now.

**Comments.** Fair hit, there were ~190 added lines of them. Cut by about half, keeping only the ones recording something the code can't show: why removing a built-in vim mapping breaks `mapclear` and `noremap`, why the panel rules need the `&` qualifier to survive theme mount order, and why observed focus must not be written into the `is-focused` prop.

**Redo.** I traced this as far as I can without running the app, and I don't think I can confirm the cause from here — one question would settle it. When you press Ctrl-R in normal mode, does the find-and-replace panel open, does nothing happen, or does the window reload?

- Find panel opens → the ui-kit editor was claiming `Mod-r` for find-and-replace. I've stopped it doing that in vim mode (it duplicated `Mod-f`, which already does the same thing). Note this only affects editors created after the keymap is switched, so it needs a new tab or a restart to show up.
- Nothing happens → vim consumed the key but `redo` didn't run, which is a different bug and probably in the history wiring.
- Window reloads → something above the renderer is taking it. The Reload Window menu item is Ctrl+Shift+R, so that isn't it, but it would rule the renderer out.

For what it's worth, vim's plugin registers ahead of that keymap and calls `preventDefault`/`stopPropagation` when it handles a key, so precedence *should* already favour vim — which is why I'd rather ask than assume the `Mod-r` change fixed it.

**Status line in dark mode.** I can't see either screenshot — GitHub's attachment CDN returns 403 through this session's proxy. Could you describe what's wrong, or paste the computed background and the winning selector from devtools? I've already had two wrong swings at this one, so I'd rather not take a third blind. The current state is that `.cm-panels` and the vim panel inherit from `.cm-editor` (which is what monokai actually paints) rather than naming a colour, because every `--bks-text-editor-*` variable falls back to `$query-editor-bg`, which is white.

**Quick search focus.** Agreed it's not vim-specific — it reproduces with the default keymap, and the fix belongs wherever quick search closes rather than in here. Happy to take it as a separate PR.

---

rathboma:
Yuck, sorry claude did an auto-respond.

I fixed the dark mode vim bar.
I think this fixes race conditions with various keyboard shortcuts for vim

PR #4670, opened 2026-08-19
Enhancements to Table Properties tab

rathboma:
- filter box to search columns / indexes / etc
- copy button to copy as json/markdown/csv
- slight design update for the header buttons to accomodate

image

image

Copy respects what's visible in the table

### DIFF
--- a/apps/studio/src/assets/styles/app/tabs/table-properties.scss
+++ b/apps/studio/src/assets/styles/app/tabs/table-properties.scss
@@ -53,7 +53,14 @@ $cell-success:           $row-add;
     .table-title {
       display: flex;
       flex-grow: 1;
+      align-items: baseline;
+      gap: 10px;
       color: $text-dark;
+      .filter-match-label {
+        font-size: 12.5px;
+        color: $text-light;
+        white-space: nowrap;
+      }
     }
     h2 {
       margin: 0;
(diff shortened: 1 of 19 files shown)


azmy60:
Cool feature! I would reorganize the UI a little bit like this:
- I personally feel like the `.filter-match-label` position is a little bit awkward. So first, I would move the search bar to a new row with the `.filter-match-label`. This way, it makes more sense because the label is now one group with the search bar.
  * Also, I don't know which side is better, is it left (like the image) or right? whichever feels right. I'd pick left because we read text from left to right.
- For the new buttons, the order is; 1. icon button 2. flat button 3. primary button. I feel like this is cleaner and it doesn't need the divider.

Screenshot 2026-08-20 at 16 34 31

rathboma:
I agree the three action buttons look cleaner in your design, but having search on a different row just adds a bunch of clutter and wastes more vertical space

rathboma:
I have an idea, gimme a sec to mock it up
image

I think this will look neater, with search expanding on click

rathboma:
image

rathboma:
This has been updated with the new UI. Good suggestion to improve. I think it looks pretty sleek now

azmy60:
Okay that's good, I love the concept! I think it can be more discoverable though. Putting an icon beside the title like that, especially if it's a small icon, wouldn't be as obvious as placing it near the other icons (on the right).

image

I'm not saying we should increase the size or the weight. In this case, I think just applying the law of proximity would be fine.

I also like that this could become a strict standard design for our apps that there are two sections for header; left section is for title and static elements, right section is for actions, buttons or active elements.

Lastly, I just want to say that I'm a bit skeptical with the way that the match label is now visualized, but it's not a big deal, it's totally OK to leave it.

Screenshot_20260821_070247

azmy60:
@rathboma If you want, we could copy how Google did it for the match label.

image

Google doesn't look like this for me right now, but we could have the match label in between the title and the table  (not beside the title), showing "how many columns are there?". And if the user is actively searching, the label becomes "how many columns are there and how many columns are filtered?".

image
image

```css
font-size: 1rem;
color: var(--text-lighter); /** or var(--text-light) */
```

apps/studio/src/assets/styles/app/tabs/table-properties.scss
```diff
@@ -71,6 +78,204 @@ $cell-success:           $row-add;
...
+          align-items: center;
+          justify-content: center;
+          flex: none;
+          width: 16px;
+          height: 16px;
+          padding: 0;
+          border: 0;
+          border-radius: 999px;
+          background: transparent;
+          color: $text-light;
```
azmy60, line 189: There's a pure css version of this btw:

```css
color: var(--text-light);
```

apps/studio/src/components/tableinfo/TablePartitions.vue
```diff
@@ -94,18 +88,21 @@ import _ from 'lodash';
...
 
 export default Vue.extend({
 	components: {
     StatusBar,
-    ErrorAlert
+    ErrorAlert,
+    TableInfoToolbar
   },
-  mixins: [DataMutators],
+  mixins: [DataMutators, StructureCopyMixin],
```
azmy60: It would save us from forgetting `StructureCopyMixin` is needed in order to use `<table-info-toolbar>` if `copyStructure` is included in the `<table-info-toolbar>`.

azmy60 (approved):

rathboma:
Just to wrap up the thread. Final version removes the title entirely (brilliant!). We're just repeating ourselves anyway. In it's stead is the filter box, results look great. Shipping.

image

PR #4675, opened 2026-08-20
Phantom tabs

rathboma:
1. We interchanged saved connection and used connection when opening the CoreInterface
2. This caused an ID conflict. The backend always thought we were talking about a saved connection.
3. Requiring that core interface only accept a saved connection is the correct fix here

Ways users could experience this bug:
- Random tabs open when they connect to a database
- Pinned tables seemingly open on their own
- Deleted tabs when they went back to open a saved connection from previously.

So it didn't happen all the time, but when it did it was very frustrating.


rathboma:
fixing the accidentally bundled sql server code

apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -527,19 +526,34 @@ export default Vue.extend({
...
+      // a saved connection, or a new one built on this screen
+      if (_.isUndefined(config.connectionId)) return config
+
+      const init = _.omit(config, ['id', 'connectionId', 'createdAt', 'updatedAt', 'version'])
+      const unsaved = await this.$util.send('appdb/saved/new', { init })
+      unsaved.id = null
+      return unsaved
+    },
+    async edit(config) {
+      this.config = _.clone(await this.configFrom(config))
```
not-night-but: I don't think we need to clone here? `configFrom` already gives us a brand new object that doesn't have any references hanging about.

apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -571,15 +585,6 @@ export default Vue.extend({
...
         this.connecting = true
-        // If this is an existing used connection that doesn't have an associated saved connection
-        // we need to see if changes have been made to the config
-        if (this.config.connectionId === null && this.config.id) {
-          const oldConfig = this.usedConfigs.find((c) => c.id === this.config.id);
-          if (!_.isEqual(this.config, oldConfig)) {
-            this.config.id = null;
-          }
-        }
-
```
not-night-but, line 582: Do we just not want this behaviour anymore?

apps/studio/src/handlers/appDbHandlers.ts
```diff
@@ -64,6 +71,11 @@ function handlersFor<T extends Transport>(name: string, cls: any, transform: (ob
       const allCols = repo.metadata.columns
       .map((c: { propertyPath: string }) => `${alias}.${c.propertyPath}`);
       if (_.isArray(obj)) {
+          const dropped = obj.filter(unscoped);
+          if (dropped.length) {
+            log.warn(`Not saving ${dropped.length} ${name} row(s) with no connection`);
+            obj = obj.filter((e) => !unscoped(e));
+          }
```
not-night-but: This feels like it could just be done with a foreign key or some other constraint?

apps/studio/src/store/modules/data/used_connection/UtilityUsedConnectionModule.ts
```diff
@@ -22,38 +34,64 @@ export const UtilUsedConnectionModule: DataStore<IConnection, State> = {
...
+
+      const lastUsedConnection = context.state.items.find((c) => {
+        if (c.workspaceId !== config.workspaceId) return false
+        if (savedConnectionId && c.connectionId === savedConnectionId) return true
+        // A connection with nothing saved behind it has no id to match on, so
+        // match on where it points instead. That keeps repeat connects to the
+        // same unsaved connection on one recent-list entry, and lets a
+        // connection that has since been saved adopt the entry it grew out of
+        // rather than leaving a duplicate behind.
+        return _.isNil(c.connectionId) && sameServer(c, config)
```
not-night-but: This seems bad. I can see this 10000% selecting the wrong connection and overriding its details. Maybe you have two connections with different credentials (admin vs read-only creds or something), or it's a connection that has that stale data in it, but is now using az or aws or something to resolve, so it doesn't need a host, port, etc. We just shouldn't be doing something like this imo.

apps/studio/src/store/modules/data/used_connection/UtilityUsedConnectionModule.ts
```diff
@@ -22,38 +34,64 @@ export const UtilUsedConnectionModule: DataStore<IConnection, State> = {
...
+          id: null,
+          connectionId: savedConnectionId,
+        });
       }
+
+      // `config` is returned untouched: its id is a saved_connection id, or
+      // null for a connection that was never saved. A null id disables
+      // per-connection persistence for the session (see the usedConfig.id
+      // guards in TabModule and friends) instead of borrowing some other
+      // saved connection's key.
```
not-night-but: This does mean that the config \*could\* be out of sync with what is in the store and in the database. When save is called, we use `withProps` so that all of the setters for any fields get run (`connectionType` sets default ports and stuff, `socketPath` sets `_socketPath`. Not certain though

not-night-but (changes requested):
I'm not sure if the `isConnectionScope` stuff in `appDbHandlers` is the best option, ideally sqlite can just handle that for us.

I'm also concerned about matching when recording used connections.

apps/studio/src/components/ConnectionInterface.vue
rathboma: it's never an existing used connection anymore

apps/studio/src/components/ConnectionInterface.vue
rathboma: We do need the clone, but I moved it into that method

apps/studio/src/handlers/appDbHandlers.ts
rathboma: We can't use a foreign key because some connections are cloud connections. Unless we want to move pins, tabs, etc to the cloud

rathboma: I'm refactoring because this isn't very clear.

apps/studio/src/handlers/appDbHandlers.ts
```diff
@@ -168,15 +183,15 @@ async function transformConn(obj: SavedConnection, cls: any): Promise<IConnectio
 export const AppDbHandlers = {
   ...handlersFor<IConnection>('saved', SavedConnection, transformConn),
   ...handlersFor<IConnection>('used', UsedConnection, transformConn),
-  ...handlersFor<TransportPinnedConn>('pinconn', PinnedConnection),
-  ...handlersFor<TransportPinnedEntity>('pins', PinnedEntity),
+  ...handlersFor<TransportPinnedConn>('pinconn', PinnedConnection, defaultTransform, true),
```
rathboma: the `true` is `requiresConnectionId`

apps/studio/src/store/modules/data/used_connection/UtilityUsedConnectionModule.ts
rathboma: This was removed

not-night-but (approved):

PR #4679, opened 2026-08-21
SQLite and LibSQL flakiness fixes (hopefully)

rathboma:
Less flakes, more bran. This PR is the fix for SQLite's digestive issues.


rathboma:
This is slop coded for the finest breakfast

not-night-but (approved):

PR #4681, opened 2026-08-22
fix(ssh): treat agent errors as non-fatal so publickey auth can proceed

### LINKED ISSUE #4661: BUG: SSH Hostname in .ssh/config does not work in 6.0.1
**Describe the bug**
in previous versions I was able to enter a ssh hostname (found in ~/.ssh/config) with a SSH Auhtentication set to Automatic and the connection would go through. 
Since the latest update, this is broken. I get the error message Database Connection Error: SSH Tunnel Connection Error: Failed to retrieve identities from agent

**To Reproduce**
1. Go to connections
2. enter ssh hostname present in ~/.ssh/config
3. make sure ssh authentication is set to automatic
4. click on connect

**Expected behavior**
Connected to the database

 - OS: Windows
 - App Version 6.0.4 (also tested on 6.0.1 with same issue - previous version was working as expected)

stephaneminisini-fps:
Connecting on Windows with SSH auth set to Automatic fails with "SSH Tunnel Connection Error: Failed to retrieve identities from agent" when no agent is running, even though the key resolved from ~/.ssh/config would authenticate. Starting an empty Pageant works around it.

e8f7b775c made Windows always attempt agent auth (!!socketPath || isWindows) so Pageant users could reach their agent. With no Pageant running, ssh2 cannot read identities and emits an error with level 'agent'. ssh2 treats that as advisory and carries on with the next auth method, but connect() rejected on any error event, aborting a connection that publickey would have completed. Before 6.0.1 the agent was never queried on Windows without SSH_AUTH_SOCK, so this path was unreachable.

Ignore agent-level errors in the connect handler and log them instead. Terminal failures arrive as client-authentication and still reject, so a genuinely unauthenticated connection fails as before.

Fixes #4661

### DIFF
--- a/apps/studio/src/vendor/node-ssh-forward/index.ts
+++ b/apps/studio/src/vendor/node-ssh-forward/index.ts
@@ -232,6 +232,11 @@ class SSHConnection {
       })
 
       connection.on('error', (error) => {
+        // if pageant is not running on windows, we log the message and move on to the next auth method rather than rejecting the connection.
+        if ((error as any).level === 'agent') {
+          this.log.warn(`Ignoring non-fatal SSH agent error: ${error.message}`)
+          return
+        }
         reject(error)
       })
       try {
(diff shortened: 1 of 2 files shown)


not-night-but:
@stephaneminisini-fps Checked and confirmed that this fixes the issue. Thanks so much for the patch!

not-night-but (approved):

PR #4688, opened 2026-08-24
add ability to disable schema & keyword completion individually

not-night-but:
Add two new config keys, `disableKeywordCompletion` and `disableSchemaCompletion`. These do what they say on the tin. Disabling schema completion will prevent the building of completion levels which should help with some performance issues for bigger dbs.


apps/ui-kit/lib/components/sql-text-editor/extensions/vendor/@codemirror/lang-sql/src/sql.ts
```diff
@@ -49,10 +53,20 @@ export function sql(config: SQLConfig = {}) {
...
   // upperCaseKeywords predates keywordCasing (upstream option); keep honoring it.
   let casing = config.keywordCasing || (config.upperCaseKeywords ? "upper" : "preserve")
-  return new LanguageSupport(lang.language, [
-    schemaCompletion(config),
-    lang.language.data.of({
+  const extensions: Extension[] = [];
+
+  if (!config.disableSchemaCompletion) {
+    extensions.push(schemaCompletion(config));
+    console.log('enabled schema completion')
```
azmy60: Can we remove the `console.log`s?

azmy60 (approved):

rathboma:
LGTM, not sure this needs to go into rc-60, but fine to merge either way

not-night-but:
Yeah I just did it against rc-60 because it's easier to rebase to master than it is the other way around lol

PR #4693, opened 2026-08-26
extend bks type

azmy60:
- each client now overrides listTableColumnsRunner and selectTopRunner instead of listTableColumns and selectTop
- added TypeResolver to resolve bks type


apps/studio/src-commercial/backend/lib/db/clients/surrealdb/SurrealDBFieldResolver.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (column.dataType?.startsWith("record<") || column.columnName === "id") {
+      return "SURREALID";
+    }
+
+    // option<string> wraps the type we actually care about.
+    const declaration = column.dataType?.toLowerCase() ?? "";
+    const type = declaration
+      .replace(/^option<(.*)>$/, "$1")
+      .split("<")[0]
+      .trim();
```
not-night-but: I feel like this could just be included in the regex optionally, no?

apps/studio/src-commercial/backend/lib/db/clients/surrealdb/SurrealDBFieldResolver.ts
```diff
@@ -0,0 +1,71 @@
...
+  binary: /^bytes$/,
+  string: /^(?:string|uuid)$/,
+  number: /^(?:int|float|decimal|number)$/,
+  datetime: /^(?:datetime|duration)$/,
+  boolean: /^bool$/,
+};
+
+export class SurrealDBFieldResolver extends FieldResolver<SurrealDBQueryResult> {
+  // Surreal reports no types for a result, so the values decide.
+  // TODO (@day): may need to do some analysis here
```
not-night-but: ?

PR #4700, opened 2026-08-27
fix plugin shortcuts and type errors

not-night-but:
Currently in the app you can't open a plugin tab from a keyboard shortcut as we weren't properly passing the manifest to the function, this fixes that and several type issues

### DIFF
--- a/apps/studio/src/services/plugin/web/WebPluginLoader.ts
+++ b/apps/studio/src/services/plugin/web/WebPluginLoader.ts
@@ -31,7 +31,8 @@ type PluginResponseData = {
   };
 }[keyof RequestMap];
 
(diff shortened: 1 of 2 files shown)


apps/studio/src/services/plugin/web/WebPluginLoader.ts
```diff
@@ -31,7 +31,8 @@ type PluginResponseData = {
   };
 }[keyof RequestMap];
 
-type PluginNotificationData = {
+// This should probably be moved to the plugin packge
```
azmy60, line 34: Yep

azmy60 (approved):
Nice catch

PR #4707, opened 2026-09-01
Add tree for import from local workspace

not-night-but:
- Use a tree structure for import from local workspace: 
image
- Cleaned up some inconsistencies between data modules in the store
- Made local workspace function the same as the cloud workspace (progressive loading) and allow for searching in local workspace


apps/studio/src/handlers/appDbHandlers.ts
```diff
@@ -126,6 +152,33 @@ function handlersFor<T extends Transport>(name: string, cls: any, transform: (ob
...
+        }
+      }
+
+      const result = await builder
+        .orderBy(`c.${cls.searchableFields[0]}`, 'ASC')
+        .getMany();
+
+      return await Promise.all(result.map(async (value) => {
+        return await transform(value, cls);
+      }));
```
azmy60, line 164: I think the whole thing can be an ApplicationEntity method and then we can use it like this:

```suggestion
      const result = await cls.search(searchText);
      return await Promise.all(result.map((value) => transform(value, cls)));
```

Later, `ApplicationEntity.search()` can be tested as well.

apps/studio/src/handlers/appDbHandlers.ts
```diff
@@ -112,6 +135,9 @@ function handlersFor<T extends Transport>(name: string, cls: any, transform: (ob
       }
     },
     [`appdb/${name}/find`]: async function({ options }: { options?: FindManyOptions<any> }) {
+      if (options && !options.where && "params" in options) {
+        options.where = paramsToWhere(options.params);
```
azmy60, line 138: Same thing with this, it's probably better to be abstracted but I don't know if it's possible for this one.

apps/studio/src/components/sidebar/core/FavoriteList.vue
```diff
@@ -298,10 +269,9 @@ export default {
     ...mapState('data/queries/nodes', {'itemNodes': 'items'}),
     ...mapState('data/queryFolders/nodes', {'folderNodes': 'items'}),
     ...mapState('data/queries', {
-      'queries': 'items',
       'queriesError': 'error',
       'queriesPollError': 'pollError',
-      'savedQueryFilter': 'filter',
+      'queryFilter': 'filter',
```
azmy60, line 304: Love these small changes for consistency.

apps/studio/src/components/data/ImportQueriesModal.vue
```diff
@@ -89,39 +84,58 @@ export default Vue.extend({
...
+        await Promise.all(this.selectedQueries.map(async (id: number) => {
+          const query = await this.findLocal(id);
+
+          if (!query) {
+            log.error(`Could not find local query for ${id}`);
+            return;
+          }
+
+          const payload = { ...query, id: null, queryFolderId: null }
+          return this.$store.dispatch('data/queries/save', payload);
```
azmy60, line 107: Is it possible to bulk save this? Don't we have `/import` api in the cloud?

apps/studio/src/components/common/LocalItemTreePicker.vue
```diff
@@ -0,0 +1,403 @@
...
+            v-model="filterText"
+          >
+          <x-buttons class="filter-actions">
+            <x-button
+              @click="clearFilter"
+              v-if="filterText"
+            >
+              <i class="clear material-icons">cancel</i>
+            </x-button>
+          </x-buttons>
```
azmy60, line 19: At some point, we might want to make a generic component that wraps this.

apps/studio/src/components/common/LocalItemTreePicker.vue
```diff
@@ -0,0 +1,403 @@
...
+        newValue = [...this.value, id];
+      }
+      this.$emit('input', newValue)
+    },
+    clearFilter() {
+      this.filterText = null
+    },
+
+  },
+  beforeMount() {
```
azmy60, line 288: Hmm this looks like a red flag. Why do we need to register/unregister vue x modules? Why don't we keep the state and business logic in this file or the `ImportConnectionsModal.vue` and `ImportQueriesModal.vue`?

azmy60 (changes requested):

apps/studio/src/components/common/LocalItemTreePicker.vue
not-night-but: This is because we do so much heavy lifting in the vuex modules to generate the nodes, and we dynamically mount/unmount those when we switch workspaces. In this case we need all of the local data from several vuex stores, so rather than copying all of the loading and node building logic, we can just mount the store temporarily

apps/studio/src/components/data/ImportQueriesModal.vue
not-night-but: Hmmm it's possible, but we'd have to batch them. Might be worth just moving this logic to the backend so we can do that properly

apps/studio/src/components/common/LocalItemTreePicker.vue
not-night-but: Yeah agreed

apps/studio/src/handlers/appDbHandlers.ts
not-night-but: The problem with that would be accessing the current class statically I think to get the builder and whatnot

apps/studio/src/components/common/LocalItemTreePicker.vue
azmy60: That means we can just use the existing module, no? No mounting the store temporarily.

not-night-but: No because at this point we're in a workspace, so the cloud module is mounted. We need access to the local module temporarily

azmy60: I understand that the main problem here is that `expandedIds` is currently tied to the sidebar module. So in this case, it seems like there are two ways to solve this:

1. We make a new module for `ImportConnectionsModal` and then use it from the `poll()`:

```ts
// CloudConnectionModule.ts
export const CloudConnectionModule = {
  actions: {
    async poll(context) {
      // ...
      const expandedFolderIds = _.uniq([
        ...context.rootState.sidebar.connections.expandedIds,
        ...context.rootState.modal.importConnections.expandedIds,
      ]);
      const result = await context.dispatch('loadByParentIds', expandedFolderIds)
      // ...
    }
  },
};
```

This is similar to how we solved the sidebar, it's not great, but it works. Or...

2. We might want to support a proper watcher system to the CloudConnectionModule, similar to Vue's `this.$watch`:

```ts
// CloudConnectionModule.ts
export const CloudConnectionModule = {
  state: {
    watchedParentIds: [],
  },
  actions: {
    async poll(context) {
      const result = await context.dispatch('loadByParentIds', context.state.watchedParentIds)
    },
    watch(context, options: { parentIds: number[], immediate?: boolean }) {
      const unwatch = () => { /*...*/ };
      return unwatch;
    },
  },
};
```

And then we use it in the modals like:

```ts
state() {
  return {
    expandedIds: [],
    unwatchExpandedIds: null,
  };
},
watch: {
  expandedIds() {
    this.unwatchExpandedIds?.();
    this.unwatchExpandedIds = null;
    if (this.expandedIds.length > 0) {
      this.watchParentIds(this.expandedIds);
    }
  },
},
```

apps/studio/src/handlers/appDbHandlers.ts
azmy60: Hmm that's def a quirk. Maybe the closest we can get is this:

```ts
cls.search(cls, searchText);

class ApplicationEntity {
  static search(cls, term) {
    let builder = cls.createQueryBuilder("c");
    // ...
  }
}
```

apps/studio/src/components/common/LocalItemTreePicker.vue
azmy60: @not-night-but Oh I didn't see your last comment lol

azmy60: Ok I get your point. It still feels rough on the edges, but that's just by design.

azmy60 (changes requested):
For the UI, these are minor details:

1. In general, I think scroll bar should stick to the edge of the modal:

Image

I know this could be tricky to do, it'll require sticky positions, styling on the `<base-modal>`, paddings, hiding the elements above the title, etc.. I'll leave it up to you whether you'd work on this or not.

2. The checkboxes look a bit too large next to the folder icons. Might need to resize them a little bit.

Image

3. Too much space between the title and the text below it, and too little between the text and the filter field.

Image

Image

apps/studio/src/components/common/LocalItemTreePicker.vue
not-night-but: Gonna mark this as resolved as I'm not sure we can really get around this without copying a bunch of logic

apps/studio/src/components/data/ImportQueriesModal.vue
not-night-but: I've made some changes to the import workflow in another PR, and I think we can expand that to allow for handling this better, so I'll leave that for another day if that's okay with you?

apps/studio/src/handlers/appDbHandlers.ts
not-night-but: I've modified this in another PR, so if we want to make changes we'll have to do it there anyways

apps/studio/src/components/data/ImportQueriesModal.vue
azmy60: Sure!

apps/studio/src/handlers/appDbHandlers.ts
```diff
@@ -126,6 +151,17 @@ function handlersFor<T extends Transport>(name: string, cls: any, transform: (ob
       // Support both direct options or wrapped in { options: ... }
       const options = 'options' in args ? args.options : args;
       return await cls.count(options);
+    },
+    [`appdb/${name}/search`]: async function({ searchText }: { searchText: string }) {
+      if (!cls.searchableFields || cls.searchableFields.length === 0) {
```
azmy60, line 156: Claude says this could work?

```ts
class ApplicationEntity {
  static searchableFields = [];
  static search() {
    if (_.empty(this.searchableFields)) {
      throw new Error(`You need to configure the searchable fields for model ${name}`);
    }
    // do search
  }
}
```

apps/studio/src/store/modules/data/connection/UtilityConnectionModule.ts
```diff
@@ -31,15 +33,16 @@ export const UtilConnectionModule: DataStore<IConnection, State> = {
   actions: {
     ...utilActionsFor<IConnection>('saved', {}),
     ...accessGrantActions('connections'),
-    ...treeActions<IConnection>({ plural: 'connectionFolderIds', singular: 'connectionFolderId' }),
+    ...treeActions<IConnection>({ plural: 'connectionFolderIds', singular: 'connectionFolderId' }, true),
```
azmy60, line 36: I would do 

```
...treeActions<IConnection>({ plural: 'connectionFolderIds', singular: 'connectionFolderId', local: true }),
```

The bare boolean flag at the end just looks ambiguous to me.

not-night-but: I feel like having parentKeys together in an object makes sense, adding local to that object muddies the waters a bit

apps/studio/src/handlers/appDbHandlers.ts
not-night-but: I mean we can obviously move the error throw into the method, I just would prefer to have any errors that can be thrown to the client a bit more obvious (hence in the handler and not the method).

apps/studio/src/store/modules/data/connection/UtilityConnectionModule.ts
azmy60: Honestly, what I meant was if someone read this next month, would it be obvious what it means by just reading that line or do I need to go deeper? I'm just afraid I didn't get my point across. It's not a must, all good :)

azmy60 (approved):
The UI could've been better. There are issues like alignment and empty state that's not the same like the original. But those could be in a follow up PR.

PR #4708, opened 2026-09-01
Load saved connections that are tied to used connections

not-night-but:
Load saved connections that are tied to used connections show that their info can be used in the recent list, regardless of whether their parent has been expanded or not.

** PENDING CLOUD PR: https://github.com/beekeeper-studio/cloud/pull/184**

### DIFF
--- a/apps/studio/src/handlers/appDbHandlers.ts
+++ b/apps/studio/src/handlers/appDbHandlers.ts
@@ -36,23 +36,28 @@ const log = rawLog.scope('Appdb handlers');
 const pluralKeys = [
   'connectionFolderIds',
   'queryFolderIds',
-  'parentIds'
+  'parentIds',
+  'ids'
 ]
 
-function paramsToWhere(params: Record<string, any>): FindOptionsWhere<any> {
-  const where = {};
-  for (const key of pluralKeys) {
-    if (key in params) {
-      const singular = key.replace(/Ids$/, 'Id');
-      if (params[key] && params[key].length > 0) {
-        where[singular] = In(params[key]);
-      } else {
-        where[singular] = IsNull();
+function paramsToWhere(params: Record<string, any> | Array<Record<string, any>>): FindOptionsWhere<any>[] {
+  params = _.isArray(params) ? params : [params];
+
+  return params.map((p: Record<string, any>) => {
+    const where = {};
+    for (const key of pluralKeys) {
+      if (key in p) {
+        const singular = key.replace(/Ids$/, 'Id').replace(/ids$/, 'id');
+        if (p[key] && p[key].length > 0) {
+          where[singular] = In(p[key]);
+        } else {
+          where[singular] = IsNull();
+        }
       }
     }
-  }
 
-  return where;
+    return where;
+  })
 }
 
 async function defaultTransform<T extends Transport>(obj: T, cls: any) {
--- a/apps/studio/src/handlers/appDbHandlers.ts
+++ b/apps/studio/src/handlers/appDbHandlers.ts
@@ -38,7 +38,14 @@ const pluralKeys = [
   'queryFolderIds',
   'parentIds',
   'ids'
-]
+];
+
+const pluralToSingular = {
+  'connectionFolderIds': 'connectionFolderId',
+  'queryFolderIds': 'queryFolderId',
+  'parentIds': 'parentId',
+  'ids': 'id'
+};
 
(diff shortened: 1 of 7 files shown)


apps/studio/src/store/modules/data/tree/treeStore.ts
```diff
@@ -64,10 +64,16 @@ export type TreeState<T> = {
...
-  plural: "connectionFolderIds" | "queryFolderIds" | "parentIds",
-  singular: "connectionFolderId" | "queryFolderId" | "parentId"
-}, local: boolean = false): ActionTree<TreeState<T>, RootState> {
+export function treeActions<T extends HasId>(
+  parentKeys: {
+    plural: "connectionFolderIds" | "queryFolderIds" | "parentIds",
+    singular: "connectionFolderId" | "queryFolderId" | "parentId"
+  },
+  local: boolean = false,
+  extraIdsInfo: {
```
azmy60: Hmm I feel like this will break easily. It's fine for a hack, but there's a better way to do this.

azmy60 (changes requested):
Instead of reusing the same fetch that we use with `connectionFolderId`. I think a better way is that we make a new method for fetching the `linkedSavedConnectionIds`.

So in the `poll()`, we can do:

```
poll(context) {
  // ...
  context.dispatch("loadByParentIds", expandedFolderIds)
  context.dispatch("loadByIds", context.state.linkedSavedConnectionIds)
  // ...
}
```

I don't think you need to do much more than needed if you base the solution around `loadByIds` and the `poll()`, the whole thing will just work for you (hopefully). And you don't need changes from the cloud too.

apps/studio/src/store/modules/data/tree/treeStore.ts
not-night-but: Yeah I definitely don't love it either lmao. This is one issue with trying to create factories like this

apps/studio/src/handlers/appDbHandlers.ts
```diff
@@ -36,23 +36,28 @@ const log = rawLog.scope('Appdb handlers');
...
-      } else {
-        where[singular] = IsNull();
+function paramsToWhere(params: Record<string, any> | Array<Record<string, any>>): FindOptionsWhere<any>[] {
+  params = _.isArray(params) ? params : [params];
+
+  return params.map((p: Record<string, any>) => {
+    const where = {};
+    for (const key of pluralKeys) {
+      if (key in p) {
+        const singular = key.replace(/Ids$/, 'Id').replace(/ids$/, 'id');
```
azmy60: Instead of doing this, making a map would be nicer IMO, so we could do:

```ts
const singular = pluralToSingularMap[key];
if (singular) {
  // ...
}
```

apps/studio/src/store/modules/data/connection/CloudConnectionModule.ts
```diff
@@ -32,9 +38,8 @@ export const CloudConnectionModule: DataStore<ICloudSavedConnection, State> = {
   actions: {
     ...actionsFor<ICloudSavedConnection>('connections', {}),
     ...accessGrantActions('connections'),
-    ...treeActions<ICloudSavedConnection>({ plural: 'connectionFolderIds', singular: 'connectionFolderId' }),
-    async initialize() {
-      // noop
+    ...treeActions<ICloudSavedConnection>({ plural: 'connectionFolderIds', singular: 'connectionFolderId' }, false),
+    initialize() {
```
azmy60: Would be wise to not remove async. What happens if we do initialize.then or initialize.catch somewhere? or in the future? I feel like that's the same reason why all methods in utility connection are async methods?

azmy60: No actually, we use `context.dispatch` to call this so I don't know for sure.

azmy60 (approved):

apps/studio/src/store/modules/data/connection/CloudConnectionModule.ts
not-night-but: We are very inconsistent with whether or not we have this as async (see the corresponding Utility module, which apparently didn't have async, and I added it 🙃).

I'm not sure what would actually happen but I'll make them both async for consistency's sake

apps/studio/src/handlers/appDbHandlers.ts
not-night-but: Ugh yeah probably, this was me being lazy (probably about 15 characters difference)

PR #4710, opened 2026-09-01
Added cross button hover effect

Srijal-SS:
On Tabbar cross button I added hover effect on cross button. Previously the hover feedback was absent.

image
This is how cross button looks on hover now.


not-night-but:
Great! Thanks so much for the contribution!

not-night-but (approved):

PR #4713, opened 2026-09-02
added query index to multi-query errors

drannonymousxx:
Added shared annotateQueryError helper.
Added query indexing for SQLite, SQL Server, PostgreSQL, and MySQL.
Preserved existing error properties.
Added tests for single and multi-query failures.
MySQL statements are executed sequentially to identify failures safely.
Tests: 24 passed.
`git diff --check` passed.


rathboma:
Thanks for the PR! We'll look into it asap

apps/studio/tests/unit/lib/db/clients/mysql.spec.js
```diff
@@ -190,5 +190,90 @@ describe("MysqlChangeBuilder", () => {
       expect(result).not.toContain('COLLATE')
     })
   })
+
+  describe("Multi-statement query error annotation (issue #4698)", () => {
```
not-night-but, line 194: It would be nice if we could move this to the common tests in `apps/studio/tests/integration/lib/db/clients/all.js` Most clients just use those, and it would cover the 4 you've implemented here. Keep it in it's own describe block and just have it return if `getUtil().dbType` is not one of those four dialects for now

apps/studio/src/lib/db/clients/mysql.ts
```diff
@@ -1168,34 +1169,49 @@ export class MysqlClient extends BasicDatabaseClient<ResultType, mysql.PoolConne
...
-    });
-
-    if (!rows) {
-      return [];
-    }
-
     const commands = this.identifyCommands(queryText);
 
-    if (!isMultipleQuery(fields)) {
+    if (commands.length <= 1) {
```
not-night-but, line 1174: We already check for this in `annotateQueryError`, so it is probably easier to just do the loop even if there's only the one command? Previously we had to split it up because we were forcing everything through `driverExecuteSingle`

apps/studio/src/lib/db/clients/utils.ts
```diff
@@ -26,6 +26,25 @@ export class ClientError extends Error {
...
+  }
+
+  if (typeof error === 'object' && typeof error.message === 'string' && !error.message.includes('(@ query #')) {
+    error.message = `${error.message}${annotation}`;
+  }
+
+  return error;
+}
+
+
```
not-night-but, line 47: I think it might make more sense to put this on the `BasicDatabaseClient`, near `identifyCommands`

not-night-but (changes requested):
Looks good! I think we could move the tests into the common testing suite, as well as move that utility function into the base class, but the general idea is great! I think we could definitely expand on this system in the future.

Let me know if you're up to making those changes, and thanks for the contribution!

drannonymousxx:
I’ve moved annotateQueryError into BasicDatabaseClient, updated the database clients to use it, and relocated the related tests. All relevant tests are passing. Sorry for the delay. I was occupied with my semester exams.

not-night-but:
@drannonymousxx No worries! There were two other comments on the PR that still haven't been resolved

PR #4714, opened 2026-09-02
Send plugin events to renderer

azmy60:
- Plugin events are now camelCased, matching the convention.
- Plugin events can now be sent to renderer process via `NotificationModule`.
- We used `plugin/waitForInit` because we weren't able to send notifications from utility to renderer. Now this is removed, and replaced by `ReadyState`, which uses a notification.

Other than `ReadyState`, being able to send notifications to renderer allows us to get notified of processes that are running in the backend, for example, finished installing/updating a plugin.


azmy60:
I have something better. It's gonna be in the theme system PR.

PR #4722, opened 2026-09-03
Move to the renamed @trinodb/trino-js-client

mosabua:
The Trino client was republished under the `trinodb` scope. Versions up to 0.2.9 shipped as the unscoped `trino-client`; the scoped `@trinodb/trino-js-client` starts at 0.3.0. The unscoped package is no longer maintained and will not receive further updates.

I am a Trino maintainer, and we are working through the downstream consumers of the client as part of that move. Happy to open an issue first if you would rather discuss it, per CONTRIBUTING.md — this seemed small enough to just show.

## Risk

Low. The declaration files of `trino-client@0.2.9` and `@trinodb/trino-js-client@0.3.2` are identical byte for byte:

```
$ npm pack trino-client@0.2.9 && tar xzf trino-client-0.2.9.tgz
$ diff -u package/dist/index.d.ts node_modules/@trinodb/trino-js-client/dist/index.d.ts
$ echo $?
0
```

Same twenty exports, same shapes, so this is a rename of the dependency and its import specifiers with no API change.

## Changes

- `apps/studio/package.json` — the dependency
- `apps/studio/src-commercial/backend/lib/db/clients/trino.ts` — the import, and one comment that names the package
- `apps/studio/tests/unit/lib/db/clients/trino.spec.ts` — the import and the `jest.mock` specifier
- `yarn.lock`

## A note on the lockfile

The change also removes an inconsistency that was already there. The old entry read:

```
axios@1.13.2, axios@^1.15.0, axios@^1.15.1, axios@^1.18.0:
  version "1.19.0"
```

The old client pinned axios `1.13.2` exactly, but yarn deduplicated that request against the newer ranges and resolved the whole group to 1.19.0, so the lockfile claimed to satisfy an exact pin with a version that does not match it. 0.3.2 pins 1.19.0, so the range key and the resolved version now agree. No axios version actually changes.

## Verification

I did not run the Trino integration suite, which needs a live coordinator. Beyond the declaration comparison above, I verified 0.3.2 against a real Trino using a harness built from this driver's call patterns, including the `next()` iteration and the error handling your client comments describe. All checks pass. That harness now runs in the client's own CI (trinodb/trino-js-client#970).

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -139,7 +139,7 @@
     "surrealdb": "^2.0.3",
     "tabulator-tables": "^6.5.2",
     "tinyduration": "^3.2.4",
-    "trino-client": "^0.2.9",
+    "@trinodb/trino-js-client": "^0.3.2",
     "typeface-roboto": "^0.0.75",
     "typeface-source-code-pro": "^1.1.3",
     "typeorm": "^0.3.31",
--- a/apps/studio/src-commercial/backend/lib/db/clients/trino.ts
+++ b/apps/studio/src-commercial/backend/lib/db/clients/trino.ts
@@ -7,7 +7,7 @@ import {
   QueryResult,
   ConnectionOptions as TrinoConnectionOptions,
   SecureContextOptions
-} from 'trino-client'
+} from '@trinodb/trino-js-client'
 import {
   BaseQueryResult,
   BasicDatabaseClient,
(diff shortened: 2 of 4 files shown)


not-night-but:
@mosabua Thank you for the PR! I wasn't able to push an update to this PR to fix the conflicts, so I've opened another PR in #4758 with these changes

PR #4730, opened 2026-09-04
Connection import

not-night-but:
Adds a connection importer: 
image

Also refactored the query import system so that we could use it for connections and queries. Both now support single and recursive import, and recursive import isn't ult-only anymore.

Importing to a cloud workspace is also possible. Requires: https://github.com/beekeeper-studio/cloud/pull/187

### DIFF
--- a/apps/studio/src/backend/lib/objectimport/import.ts
+++ b/apps/studio/src/backend/lib/objectimport/import.ts
@@ -1,0 +1,174 @@
+import path from 'path';
+import { promises as fs } from 'fs';
+import { IObjectImportStats } from '@/common/interfaces/IObjectImportStats';
+import { IFolder } from '@/common/interfaces/IQueryFolder';
+import rawLog from '@bksLogger';
+import { CloudClient } from '@/lib/cloud/CloudClient';
+
+const log = rawLog.scope('ObjectImporter');
+
+export const MAX_SQL_FILE_LENGTH = 2_000_000;
+export const MAX_SQL_FILE_BYTES = 4 * MAX_SQL_FILE_LENGTH;
+export const SQL_FILE_EXTENSIONS = new Set(['.sql', '.txt']);
+
+export const SKIP_DIR_NAMES = new Set(['.git']);
+
+// I am just assuming that big batches might be an issue, so we have the ability to cap here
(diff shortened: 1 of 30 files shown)


apps/studio/src/backend/lib/objectimport/import.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    if (this.currentBatch && this.currentBatch.length > 0) {
+      await this.importCurrentBatch();
+    }
+
+    return this.stats;
+  }
+
+  async importDirectory(dir: string, parentId: number | null, importDir: boolean = true): Promise<IObjectImportStats> {
```
azmy60, line 76: Might be worth to unit test this.

apps/studio/src/backend/lib/objectimport/import.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  constructor(protected client?: CloudClient) {
+
+  }
+
+  abstract importFolders(folders: IFolder[]): Promise<IFolder[]>;
+  abstract importItems(items: T[]): Promise<T[]>;
+  abstract readObjectFromPath(dir: string, name: string, parentId?: number): Promise<ObjectDescriptor<T>>;
+
+  async importSelections(paths: string[], parentId: number): Promise<IObjectImportStats> {
```
azmy60, line 46: And this. I don't mind coverage honestly. A single test is better than none.

apps/studio/src/components/common/modals/ConnectionFilesImportModal.vue
```diff
@@ -0,0 +1,299 @@
...
+      </button>
+      <button
+        v-if="!importing"
+        class="btn btn-primary"
+        type="button"
+        :disabled="files.length === 0 || (isCloud && parentId === null)"
+        @click="submit"
+      >
+        Import
+      </button>
```
azmy60, line 111: Tip: you can skip adding `@click` by setting the button type to "submit"

```html
<base-modal @submit="handleSubmit">
  <button type="submit">Import</button>
</base-modal>
```

azmy60 (approved):

PR #4736, opened 2026-09-08
Make folders free

not-night-but:
Remove logic gating folders to ultimate


rathboma:
Couple of dangling code bits:

Upgrade panel still lists "Connection folders" as a paid feature
Leftover CSS for the deleted expired-folder-alert component
Unused isUltimate prop/getter left behind in the save-connection form and query editor

rathboma:
Will also need a follow up website change

rathboma (approved):

PR #4737, opened 2026-09-08
Make readonly mode free

not-night-but:


apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -191,16 +191,14 @@
                 <div class="form-group" v-if="!shouldUpsell">
                   <label class="checkbox-group" for="readOnlyMode">
                     <input
-                      :disabled="!isUltimate || editingDisabled"
```
rathboma, line 194: I think you have to also look at the connection model, it has an ultimate check I believe

rathboma: also appDbHandlers

rathboma (changes requested):
missing some bits

apps/studio/src/components/ConnectionInterface.vue
not-night-but: Good catch

rathboma (approved):

PR #4738, opened 2026-09-09
Theme Customization -  #1 Appearance modal

azmy60:
This PR will give a more flexibility way for the next PRs to implement color system with full vanilla css.

This also adds a modal that replaces `View > Theme` menu.

Screenshot 2026-09-11 at 18 00 08

### How does theming work?

A theme will be a css file that's most of it should only be for overriding pre-defined css variables.

```
// monokai.css

:root,
.light-theme {
  // all monokai colors in light mode here
  --gray-1: #002b36;
  --gray-2: #00303b;
  --gray-3: #15414c;
}


.dark-theme {
  // all monokai colors in dark mode here
  --gray-1: #fff6e3;
  --gray-2: #f8efdd;
  --gray-3: #ddd8c9;
}
```

We will use a color system [Radix Colors](https://www.radix-ui.com/colors), which this PR doesn't have yet, but I will explain more in the next PR when I implement this.

A color system should help us to become more organized and disciplined so it wouldn't be too hard for developers (or even non developers) to make custom themes because it makes the app more predictable.

### Full Plan

Each of these will be separated into PRs in order:

1. Add an Appearance modal **(we are here :wave:)** and the basic business logic 
2. Make colors in the first screen (connection card, titlebar, sidebar, etc.) customizable and make sure they are compatible with plugins + Radix Colors
3. Make text editor, table, json viewer components customizable + a new theme for testing
4. Make the rest components customizable (90% of this is the connected UI)   
5. Add one more theme, at the end we have 4 themes in total!


azmy60:
UPDATE:

OK, I think this will be my last update on this PR. Three key points that this PR introduce:

1. The appearance modal should look like this:

image

If you open it using this PR, it won't look like this. But it's fixed in the next PR, which will be linked.

2. There's no `data-theme-*` attributes anywhere added. In the `<body>` tag , we will add `.dark-theme` or `.light-theme` class, and `.theme-solarized` or `.theme-default` class depending on which one is active. For example, `<body class="theme-solarized dark-theme">`.

So a theme file is expected to be written like this (in pure css):

```css
.theme-solarized {
  --base-1: #002b36;
  --base-12: #fff6e3;
}

.theme-solarized.dark-theme {
  --gray-1: var(--base-12);
  --gray-12: var(--base-1);
}

.theme-solarized.light-theme {
  --gray-1: var(--base-1);
  --gray-12: var(--base-12);
}
```

This PR doesn't have solarized theme written in this format. But in the next PR, color scales, css vars, and stuff like that will be the main focus.

3. I added user_settings for theme, which is reflected by the Appearance modal. Those are:
  - `appearance` = "light" / "dark" / "auto" (default to auto)
  - `themeId` = "default" / "solarized"

apps/studio/src/background/WindowBuilder.ts
```diff
@@ -66,6 +69,10 @@ class BeekeeperWindow {
     // const appUrl = startUrl
     const queryObj: any = openOptions ? { ...openOptions } : {}
 
+    queryObj.themeId = this.getThemeId();
+    queryObj.appearance = this.getAppearance();
+    queryObj.systemDark = this.systemUsesDarkColors;
+
     if (platformInfo.isWayland) {
```
not-night-but, line 76: I think this means we need to change the check in the `will-navigate` handler below. In dev mode I am getting some issues where on some hot reloads and switching between branches in this stack is causing the navigation to go external: 

Image

not-night-but (changes requested):
Looks good! Kinda hard to review as I suspect a lot of this changes a bit up the stack? I guess we'll see lol. Did find an issue with hot reloading and our navigation blocking

General Notes:
- Looks like we are selecting light by default when opening the appearance modal, so we get this which looks a little strange: 
Image

- Looks like we've switched the icons here to be bold? Idk if that's intentional or not (or good or bad) just jumped out at me:
Image

azmy60:
@not-night-but Oh now I know why hot reloads acted weird :smile: !

I agree with all comments too. Will figure something out.

PR #4750, opened 2026-09-10
move to node 24 and electron 44 :)

### LINKED ISSUE #4734: FEAT: Update studio's electron version to 43 or 44
studio's electron version is currently 39, which is EOL

Bump Electron to stable 43 or 44 (42 is about to be EOL in a month-ish)

not-night-but:
fix #4734 

A couple notes:
- Electron removed the auto install in their post install script, the binary is now auto downloaded when you try to run the application, but just in case I have it running the install in our post install.
- Electron updated their clipboard API to be completely async, so we had to update how we do things
- node-sqlanywhere had to be updated
- bettersqlite3 bumped to v13
- msnodesqlv8 bumped
- node-abi bumped


apps/studio/src-commercial/entrypoints/main.ts
```diff
@@ -317,6 +316,27 @@ ipcMain.handle('closeWindow', () => {
...
+    const blob = new Blob([pngBuffer], { type: 'image/png' });
+    item = new ClipboardItem({
+      'image/png': blob
+    });
+  } else {
+    item = new ClipboardItem({
+      'text/plain': content
+    });
+  }
+  clipboard.write([item]);
```
rathboma: isn't this async? Don't we want to `await` in that case?

rathboma (approved):

PR #4758, opened 2026-09-14
Update trino package

not-night-but:
Supersedes #4722


mosabua (commented):
Looks good to me. Thx @not-night-but

PR #4761, opened 2026-09-14
Fix ssh tunnels when IdentityFile points to a public key only

### LINKED ISSUE #4347: BUG: SSH `Too many authentication failures` when having more than 6 ssh keys
**Describe the bug**
When connecting via SSH Agent Beekeper picks the hosts config from `~/.ssh/config`, but doesn't seem to pick up `Match Host` / `IdentityFile` config. This means that the SSH server's `MaxAuthTries` kicks in after some time and you can't connect. I have this via 1Password, but I think it is unrelated to this.

**To Reproduce**
1. In 1Password add a ssh key
2. Login to the host in a terminal
3. Use SSH bookmarks in 1Password to match the right key with the right host(s) (https://www.1password.dev/ssh/bookmarks#manage-ssh-bookmarks-in-1password-beta) (do this after SSHing so 1Password suggests the right host)
4. Create an item in Beekeeper using one that keys, and login successfully
5. In 1Password add more ssh keys, so you have at least 7
6. In Beekeeper, login again

**Expected behavior**
I can just login no matter the amount of SSH keys I have.

**Screenshots / Gifs**
Image

 - OS: Ubuntu 24.04.01
 - App Version: 5.7.3
 - Database type and version: Postgres

**Additional context**
Using 1Password's bookmark function a `~/.ssh/1Password/config` is created, this is included in the `~/.ssh/config` file. In this file it lists these kind of entries:

```
# 1Password Item: 1password-item-name
Match Host ssh-config-hostname
  IdentitiesOnly yes
  IdentityFile ~/.ssh/1Password/SHA256_ssh-key-public-hash.pub
```

### LINKED ISSUE #4643: BUG: SSH login fails with 'privateKey value does not contain a (valid) private key'
**Describe the bug**
Since Beekeeper 6.0.0 (also tried 6.0.1) I can't login using my SSH keys anymore. They are stored in 1Password and whenever I connect I get this error `privateKey value does not contain a (valid) private key`.

**To Reproduce**
1. Setup a `~/.ssh/config` for a tunnel with a custom hostname
2. Create a connection in Beekeeper with an SSH tunnel, specifying the custom hostname
3. Use `Automatic` for `SSH Authentication`, keep the other values empty
4. Connect & see error

**Expected behavior**
It should popup a 1Password confirmation dialog asking if I want to use this key and, after accepting, connect to the database.

**Screenshots / Gifs**
Image

 - OS: Ubuntu 24.04.4 LTS
 - App Version: 6.0.1
 - Database type and version: Postgres

**Additional context**
I tried 5.9.3 via AppImage and that works fine.

Let me know if I can provide more debug context.

lode:
Fixes #4643.

(Btw, this is also the first time I don't have the issue from #4347 anymore. Maybe that was actually fixed in 6.0.0, but without this PR I couldn't verify that at least. So maybe this also fixes #4347.)

This fixes the case where a ssh config `IdentityFile` points to a public key file. This is not the regular flow, but it is documented in https://manpages.ubuntu.com/manpages/noble/man5/ssh_config.5.html#IdentityFile (emphasis mine):

> IdentityFile
> Specifies a file from which the user's DSA, ECDSA, authenticator-hosted ECDSA, Ed25519, authenticator-hosted Ed25519 or RSA authentication identity is read. **You can also specify a public key file to use the corresponding private key that is loaded in [ssh-agent(1)](https://manpages.ubuntu.com/manpages/noble/man1/ssh-agent.1.html) when the private key file is not present locally.**

This is the case for 1Password's own ssh agent. The private key is stored inside 1Password, and 1Password allows to store a public key file on disk. The ssh config then references this public key. When connecting, BeekeeperStudio shows a `privateKey value does not contain a (valid) private key` error.

This was working up to and including 5.9.3 and regressed from 6.0.0 on.

I verified this PR locally with my setup via:

```
yarn vitest:unit
yarn lib:build
yarn bks:build --linux AppImage
```

For transparency; I understand the context and the patch, but I used an LLM to find the repro and how to fix it. See my chat for our line of thought, in case it might help with things we missed: https://gist.github.com/lode/349b137d4271f0479f64846bcaabd889. I'm sorry if I'm missing obvious things and did my best to verify things so I don't contribute with AI slop.

Btw, I tried to keep the patch minimal. I'm willing to help improve it in another direction, but wanted to start small.

For a local reproduction the important bits are:
- ssh config with `IdentitiesOnly yes`
- ssh config with `IdentityFile` pointing to a `.pub` file
- 1Password's agent keeping the private key (not per se needed, the private key should just not be on disk, but fullest way to test)
- no default ssh keys on disk, or look `authHandler` in the logs

See more details in https://gist.github.com/lode/349b137d4271f0479f64846bcaabd889#file-6-review-details-txt

I hope this helps to reproduce the issue. Let me know if there's something more I can help with.


apps/studio/src/lib/ssh/sshKeyUtils.ts
lode: The name `canParseKey` is not really true anymore with this change. Maybe `isUsablePrivateKey` is better? Or maybe a split into two functions?

not-night-but:
@lode Oh cool! Is that a specific way of setting up 1Password? Or did you just import the public key manually? I just want to ensure that I can reproduce your issue. I'm glad that we've found something that seems to work for you though!

not-night-but:
@lode Also, I do believe that #4347 may have been fixed because we do actually evaluate match rules now, so that should cut down on keys that are tried, and we also ensure that we don't needlessly try things that aren't keys or are encrypted

apps/studio/src/lib/ssh/sshKeyUtils.ts
not-night-but: Yeah I think renaming makes sense here

lode:
> Oh cool! Is that a specific way of setting up 1Password? Or did you just import the public key manually? I just want to ensure that I can reproduce your issue. I'm glad that we've found something that seems to work for you though!

I'm not really sure, but I could look into the details tomorrow?

I mainly just setup 1Password to manage my ssh keys, so I don't have any keys on disk anymore. See https://www.1password.dev/ssh. Also I think I needed to setup their ssh bookmarks (https://www.1password.dev/ssh/bookmarks). And I had to adjust a few things according to https://www.1password.dev/ssh/agent/compatibility.

lode:
> Also, I do believe that #4347 may have been fixed because we do actually evaluate match rules now, so that should cut down on keys that are tried, and we also ensure that we don't needlessly try things that aren't keys or are encrypted

Yeah great! That part probably needs the ssh bookmarks (https://www.1password.dev/ssh/bookmarks) in 1Password.

apps/studio/src/lib/ssh/sshKeyUtils.ts
```diff
@@ -34,6 +34,12 @@ export function canParseKey(keyPath: string): boolean {
       log.warn(`Could not parse ${keyPath}: ${parsed.message}`);
       return false;
     }
+    const key = Array.isArray(parsed) ? parsed[0] : parsed;
+    const hasPrivateKeyMaterial = key.getPrivatePEM() !== null;
```
not-night-but: We should us `key.isPrivateKey()` instead. They reference the same value, but it's a clearer api: check [here](https://github.com/mscdex/ssh2/blob/a3a39d795a8d7ff43d5f33aafadccfc03466d044/lib/protocol/keyParser.js#L429)

not-night-but (changes requested):
Great find! Glad you were able to figure it out. I think just rename the function so it makes more sense like you mentioned, and use `isPrivateKey` instead of `getPrivatePEM` and we'll be good to go

not-night-but:
@lode yeah I think my private key was still available on disk and that's why I wasn't able to repro. With the test and your confirmation that it fixes your issue, as well as the fact that it hasn't broken my test configs I think we'll be fine to merge once the two changes are made

apps/studio/src/lib/ssh/sshKeyUtils.ts
lode: Thanks for the hint!

lode:
Thanks for the help and review! Made the adjustments.

not-night-but:
Okay great! I think this is good to go! I'm gonna merge and also cherry pick to our rc branch so it can get into the next patch

not-night-but (approved):

PR #4763, opened 2026-09-15
Theme Customization -  #2 Make connection screen customizable

azmy60:
### Full Plan

Each of these will be separated into PRs in order:

1. Add an Appearance modal and the basic business logic 
2. **Make colors in the first screen (connection card, titlebar, sidebar, etc.) customizable and make sure they are compatible with plugins + Radix Colors (we are here :wave:)**
3. Make text editor, table, json viewer components customizable + a new theme for testing
4. Make the rest components customizable (90% of this is the connected UI)   
5. Add one more theme, at the end we have 4 themes in total!


apps/studio/src/assets/styles/app/_base.scss
```diff
@@ -1,5 +1,28 @@
...
+             "Cascadia Mono", "Segoe UI Mono",
+             "Roboto Mono",
+             "Oxygen Mono",
+             "Ubuntu Mono",
+             "Source Code Pro",
+             "Fira Mono",
+             "Droid Sans Mono",
+             "Consolas", "Courier New", monospace;
+$base-font-size:      0.875rem; // ~14px
+$form-control-font-size: 0.831rem; // ~13.3px
```
not-night-but, line 25: Did we not just remove this in the previous PR?

apps/studio/src/assets/styles/app/_layout.scss
```diff
@@ -411,18 +400,15 @@ select {
...
 }
 input[type="checkbox"] {
   $checkbox-size:     18px;
 
   position: relative;
   display: inline-flex;
 	-webkit-appearance: none;
-	background-color: transparent;
-  outline: 2px solid rgba($theme-base, 0.2);
+	background-color: var(--checkbox-bg);
```
not-night-but, line 410: I know the webkit part wasn't you but can we not mix tabs and spaces?

apps/studio/src/assets/styles/app/connection-labels.scss
```diff
@@ -1,12 +1,12 @@
...
-  blue:         $brand-info,
-  purple:       $brand-purple,
-  pink:         $brand-pink,
+  red:          var(--red-solid-bg-a),
+  orange:       var(--orange-solid-bg-a),
+  yellow:       var(--yellow-solid-bg-a),
+  green:        var(--green-solid-bg-a),
+  blue:         var(--blue-solid-bg-a),
+  purple:       var(--purple-solid-bg-a),
+  pink:         var(--pink-solid-bg-a),
```
not-night-but, line 9: Yeah these are def more saturated than before. I think the first 4 are the same? blue and purple might just be a bit darker, and pink is just very saturated3

apps/studio/src/assets/styles/app/titlebar.scss
```diff
@@ -136,8 +127,13 @@ $titlebar-actions-height:   30px;
 
 .titlebar-wrapper {
   padding-bottom: $gutter-h;
-  background-color: var(--bks-titleBar-activeBackground);
-  color: var(--bks-titleBar-activeForeground);
+  background-color: var(--titlebar-bg);
+  color: var(--titlebar-fg);
+  transition: background-color 0.15s ease-in-out, color 0.15s ease-in-out;
+  @at-root body.window-inactive .titlebar-wrapper {
+    background-color: var(--titlebar-inactive-bg);
```
not-night-but, line 134: Ah I see this was maybe intentional? It looks a little out of place tbh

apps/studio/src/assets/styles/themes/README.md
```diff
@@ -0,0 +1,51 @@
+# Themes
+
+A theme is a folder in here named after its theme id, with one file per color
+scale (`gray.css`, `yellow.css`, `blue.css`, ...) holding both its light and
+dark blocks.
+
```
not-night-but, line 6: So this is the format from Radix? To me it feels a bit overcomplicated but if it's needed then I'm cool with it. I have my own theme that is typically just like 4 colours and I end up having to copy paste them into multiple different categories lmao (`#ed1a3d` as an accent, `#000000` as the background for almost everything, `#4c4c59` for inactive colours, and then `#ffffff` for a text colour if you're curious :))

apps/studio/src/services/plugin/web/cssVars.ts
```diff
@@ -1,42 +1,247 @@
...
+  { name: "--placeholder", value: "var(--gray-a9)" },
+  { name: "--selection", value: "var(--selection-bg)" },
+  { name: "--input-highlight", value: "var(--primary-solid-bg)" },
 
-  "--query-editor-bg",
-  "--menu-bg",
-  "--menu-shadow",
+  { name: "--query-editor-bg", value: "var(--bg-base)" },
+  { name: "--menu-bg", value: "var(--panel-bg)" },
+  { name: "--menu-shadow", value: "var(--panel-shadow)" },
```
not-night-but, line 37: Unless I'm misunderstanding, this is a system that only we use. We set the value to `var(--some-var)` and then later have to manually strip out the `var()` part of that to get the variable name, so why not just store `--some-var` and save having to run all that regex?

apps/studio/src/services/plugin/web/PluginStoreService.ts
```diff
@@ -69,13 +69,22 @@ export default class PluginStoreService {
...
-      palette[camelKey] = styles.getPropertyValue(name).trim();
+    for (const entry of cssVars) {
+      if (typeof entry === "string") {
+        palette[_.camelCase(entry)] = styles.getPropertyValue(entry).trim();
+        continue;
+      }
+      palette[_.camelCase(entry.name)] = entry.value.replace(
+        /var\((--[\w-]+)\)/g,
+        (_match, name) => styles.getPropertyValue(name).trim()
+      );
```
not-night-but, line 80: Yeah see here we should just be able to do `palette[_.camelCase(entry.name)] = styles.getPropertyValue(entry.value)`, no?

not-night-but (changes requested):
Some of my notes may just be personal preference, so please let me know if you disagree :)

Notes:
- Seems like some of the default colours got a whole lot more saturated? I feel like I liked the more pastel look we had originally: 
Image
- Something feels weird with the font but I could be going insane
- Icons have again gotten bolder, and I think brighter?
- When the window loses focus, the titlebar changes colour:

Image
- I feel like we're using the wrong background colour for the ssh tunnel here?

Image

apps/studio/src/assets/styles/app/titlebar.scss
azmy60: Yes. I'm just copying other apps in general. The css selector could be better too honestly.

apps/studio/src/assets/styles/themes/README.md
azmy60: Cool! 

What you have there are probably apps that have theme generator built in. It simplifies the process of curating these colors.

I'm probably more of a developer than a UI designer, so it's definitely complicated if I pick these colors by hand.

I will think about how to make this more approachable to devs, and let me know if you got ideas.

Btw, I'm just using a color system that already exists so that I don't reinvent the wheel. I hope we don't do something like `rgba($theme-base, 0.5)` anymore because we literally have 24 colors for each family. I hope this makes it easier for theme devs not to worry about **accessibillity** (texts are harder to read cause it's not contrast, etc..) You'll probably find this helpful https://www.radix-ui.com/colors/docs/palette-composition/understanding-the-scale. And this is prob not so much compared [primer](https://primer.style/primitives/storybook/?path=/story/color-base-display-scales--all-scales) for example :)

PR #4769, opened 2026-09-17
fix: read only mode is not saved

andreasarf:
Read Only Mode can't be turned on any more: the checkbox doesn't stick, and connections that already had it on lose it the next time they're saved.

`transformConn` in `appDbHandlers.ts` sets `obj.readOnlyMode = false` unconditionally. It's the transform for both the `saved` and `used` handler sets, so it runs on `find`, `findOne`, `findOneBy`, `search`, `new` and `save` — every connection handed to the renderer comes back with the flag off, and the next save writes that back to the DB.

This is a leftover from making read-only mode free (b9956e31 / e78e055c). That assignment used to be guarded:

```ts
const status = await LicenseKey.getLicenseStatus();
const canBeReadOnly = status.isUltimate || platformInfo.testMode;
if (!canBeReadOnly) {
  obj.readOnlyMode = false;
}
```

When the gate was removed the side effect stayed behind, so it now applies to everyone, Ultimate included. Removing the line is the whole fix.

Added a Vitest spec covering save/read/round-trip of the flag — 3 of its 4 cases fail on master and pass here.


not-night-but:
@andreasarf Good catch, annoyed I missed this lol

not-night-but (approved):

PR #4771, opened 2026-09-18
Theme Customization - #3 Themable text editor components, status bar, tabs, sidebar, and a new Dracula theme :vampire: 

azmy60:
The main focus here is any components that we see when we are connected to a database for the first time.
- Query Editor / Text Editor 
- Tab headers
- Tab empty state
- Query result empty state
- Sidebar
- Tree view
- Tab icons
- Status bar
- and to make it easier to test and feel how things are, a new theme is added, Dracula!

image

Coverage: **~50%** of our styles/components (including ui-kit) use the color system.


apps/studio/src/assets/styles/app/_layout.scss
```diff
@@ -560,48 +565,48 @@ input[type=file] {
...
-      color: $brand-pink;
+      color: var(--pink-solid-bg);
     }
     &.routine-function-icon {
-      color: $brand-success;
+      color: var(--green-solid-bg);
     }
     &.routine-aggregate-icon {
-      color: $brand-warning;
+      color: var(--orange-solid-bg);
```
rathboma, line 584: this looks good

not-night-but (changes requested):
NOTES:
- Okay it seems like some of the icons were fixed in this branch? But these ones are still bold: 

Image

- For some background colours we need to switch the text colour, ie in the status bar when it is orange: 
Image

In the latest release we do this: 
Image

- Looks like we have some overlapping colours between the background and the connection form background: 

Image

apps/studio/src/assets/styles/themes/dracula/blue.css
```diff
@@ -0,0 +1,55 @@
+.theme-dracula {
```
rathboma, line 1: these color files I find really confusing and hard to understand -- why do we need them? We don't even seem to use these color values at all?

If I were a normal joe user of Beekeeper Studio looking to make a theme and I saw this I'd think there's no way I could create a theme. How am I going to define like a million blues?

Is this file auto-generated from the 'main' colors in some way?

apps/studio/src/assets/styles/themes/dracula/index.css
```diff
@@ -0,0 +1,98 @@
...
+@import "./green.css";
+@import "./orange.css";
+@import "./red.css";
+@import "./yellow.css";
+@import "./pink.css";
+@import "./purple.css";
+@import "./editor.css";
+
+/* https://github.com/dracula/cursor/blob/main/dev/src/classic.yml */
+.theme-dracula {
```
rathboma, line 14: Here's kinda what I'd expect from a theme system:

### Part 1 - Set `theme colors`
A single place where all the colors of the theme are defined as variables. Not like every blue that ever existed, but like these here 'dracula-red', or 'dracula-background'

### Part 2 - Apply `theme colors` to `app variables`
A separate section where we apply theme colors to beekeeper variables, we also do this here, like `--titlebar-bg`. This makes total sense to me.

### What I find confusing
1. Why is `editor.scss` allowed to hard-code colors and not take them from the theme colors?
2. Why do blue/grey/red/etc.scss exist at all & what purpose do they serve? We don't seem to use them anywhere at all
3. If these color.css files really define our color profiles, why don't we use their variables at all in `index.css`?

### Takeaway
I guess if you'd showed me a theme system that included `index.css` and `editor.css` (with some changes to that file), I'd be sold, but with these other color files, this feels overwhelming to me - imagine how this would feel to some 18yo kid who's using Beekeeper Studio at school for the first time.

I'm fine with complexity if it's `needed`, but I don't see why it's needed here?

Assuming you're going to say something like `radix standardizes colors and uses them in set specific places and that's why we should use them`, but Radix needs to standardize colors because **it's built for meta** which has 10,000 software engineers. Likewise the thrift RPC library is built by meta. I would never use it for our REST APIs because It's complexity is driven not by technical need, but by the need to manage a lot of people working on the same thing without breaking it -- is that true of Radix also?

Just as a counter example, here is the VSCode Dracula theme which follows the part1/part2 pattern we discussed: colors at the top, applying to selectors below, sure there are a lot of selectors, but that's Microsoft, which also has 10,000 developers 🤷 .
https://github.com/dracula/visual-studio-code/blob/main/src/dracula.yml

rathboma (commented):
Love everything except the `blue/gray/etc` css files - the rest seems solid, but don't get them at all.

apps/studio/src/assets/styles/themes/dracula/index.css
rathboma: Just as an example of how we might simplify the theme system, could users define `red`, then have the app convert that into red-1,2,3,4,5,6,.....? Like does the them builder need to specify everything like that?

rathboma: Some other examples:
- tmTheme format - which is from Textmate and is supported by Sublime text, [here's an example](https://github.com/rainglow/sublime/blob/master/rainglow/Azure%20Contrast%20(rainglow).tmTheme), supporting this format wouldn't have been the worst idea tbh
- Obsidian minimal theme with many color variants [here]. This one is actually pretty complicated.

Maybe this is just missing docs? Do you have docs for making a theme you could add? They maybe we could try it and get a better sense

apps/studio/src/assets/styles/themes/dracula/blue.css
azmy60: Yes, auto generated.

apps/studio/src/assets/styles/themes/dracula/index.css
azmy60: For context, we use to do `rgba($theme-base, 0.2)`, `color.adjust($theme-base, $lightness: 30%)` , `color.mix($query-editor-bg, $brand-danger, 70%)`, and so on. We had **endless** options to define which color to use and even make new colors on every new feature we introduce, but that's not my concern. The main concern is; how do we make these colors predictable? How do we make sure that theme devs do not make too much effort into thinking if `$theme-base` is readable or not? Not just for theme devs, for us internal devs as well. Are we sure `color.mix($query-editor-bg, $brand-danger, 70%)` is easy to read?

With these endless options, I cap them into 24 colors, (well not me but the color system does), that way it's more predictable.

Have you seen `apps/studio/src/assets/styles/themes/utilities.scss`? This is where all "raw colors" (like --gray-1, --gray-2, etc.) are mapped to aliases. 


image

image

And no you don't have to pick these colors by your own. I used the theme generator all the time (with some manual tweaking ofc, it's not perfect).

To be fair, I don't know about the company behind Radix and how it's actually meant to be used, but technically, I chose this because it's well documented, especially this [page](https://www.radix-ui.com/colors/docs/palette-composition/understanding-the-scale) tells us which colors to use in what cases very specifically, and they have an open source theme generator that we can fork, and if we are consistent with the rules, would that be easier for theme devs to make themes? In a sense that, you might just find yourself making 100 lines of code, instead of ~1100 lines (the vscode theme link you added).  I don't know :shrug:

I think you are right that this is missing clear docs. This also tells me that the theme generator is a must, and that we can't do this with "pure css". We can but the theme generator is written in js.

azmy60: For us, I hope we are not using anymore scss utilities, because we literally have 24 different colors to pick lol.

These colors are aliased as well:

```
"bg-base","bg-subtle", "bg", "bg-hover", "bg-active", "border-subtle", "border", "border-hover", "solid-bg", "solid-bg-hover", "text", "text-contrast",
   "bg-base-a", "bg-subtle-a", "bg-a", "bg-hover-a", "bg-active-a", "border-subtle-a", "border-a",
   "border-hover-a", "solid-bg-a", "solid-bg-hover-a", "text-a", "text-contrast-a",
```

azmy60: And those are not strict aliases. Meaning, we can still use `--gray-1` or `--gray-2` whenever it makes sense, and be assured that other themes will automatically follow, as long as they follow the docs (or use the generator).

Good feedback though. I'll make a new PR in the stack that simplifies all of this, this really lacks the theme generator.

(With those files removed too; `blue.css`/`gray.css`, etc.)

PR #4772, opened 2026-09-18
Theme Customization - #4 Standardize generic components colors

azmy60:
Generic components now use the color system. E.g. buttons, alerts, menus, etc..

You can see how many components we have by going to `Dev > Theme Playground`.

Recording 2026-09-18 at 16 24 48

Coverage: **57%**


apps/studio/src/assets/styles/themes/dracula/index.css
```diff
@@ -30,6 +30,37 @@
...
+  --primary-a5: var(--purple-a5);
+  --primary-a6: var(--purple-a6);
+  --primary-a7: var(--purple-a7);
+  --primary-a8: var(--purple-a8);
+  --primary-a9: var(--purple-a9);
+  --primary-a10: var(--purple-a10);
+  --primary-a11: var(--purple-a11);
+  --primary-a12: var(--purple-a12);
+  --primary-solid-fg: var(--purple-solid-fg);
+  --primary-solid-fg-hover: var(--purple-solid-fg);
```
not-night-but, line 59: Do we need 24 + different primary variables?

apps/studio/src/components/common/modals/ThemePlaygroundModal.vue
```diff
@@ -0,0 +1,552 @@
+<template>
+  <base-modal :name="modalName" max-width="960px" height="85vh">
```
not-night-but, line 2: This is not wide enough for me to see the entire palette: 

Image

apps/studio/src/components/common/modals/ThemePlaygroundModal.vue
```diff
@@ -0,0 +1,552 @@
...
+            <p>Some content inside a card.</p>
+          </div>
+          <div class="card-flat padding">
+            <h4 class="card-title">
+              Flat card
+            </h4>
+            <p>Some content inside a flat card.</p>
+          </div>
+        </div>
+      </section>
```
not-night-but, line 301: Do we actually use cards in the app?

apps/studio/src/components/common/modals/ThemePlaygroundModal.vue
```diff
@@ -0,0 +1,552 @@
...
+        <div class="form-group">
+          <label>Multi select</label>
+          <multi-select
+            v-model="multiSelectQuery"
+            placeholder="Add a member"
+            :suggestions="multiSelectSuggestions"
+            :selected-options="multiSelectSelected"
+            @item-add="multiSelectSelected.push($event)"
+            @item-remove="multiSelectSelected.splice(multiSelectSelected.indexOf($event), 1)"
+          />
```
not-night-but, line 281: The styling for this seems to be broken, but not sure if that's fixed later: 

Image

apps/studio/src/components/ConnectionInterface.vue
```diff
@@ -281,6 +281,18 @@
...
+        <div style="display:flex;">
+          <div v-for="i in 12" :key="i" :style="`width: 1em; height: 1em; background: var(--gray-${i})`" />
+        </div>
+      </div>
+      <div class="dracula-theme">
+        <div style="display:flex;">
+          <div v-for="i in 12" :key="i" :style="`width: 1em; height: 1em; background: var(--gray-a${i})`" />
+        </div>
+      </div>
+    </div>
```
not-night-but, line 295: This doesn't seem like it should have been committed?

not-night-but (changes requested):
NOTES: 
- When opening a menu from the menu bar, something appears between the icon and file: 

Image

- I believe IAM Auth should be the darker background colour, and the optional text next to custom endpoint here is too dark to read:

Image

apps/studio/src/components/common/modals/ThemePlaygroundModal.vue
azmy60: Yes

apps/studio/src/components/ConnectionInterface.vue
azmy60: Nope

apps/studio/src/assets/styles/themes/dracula/index.css
azmy60: Yes. What if you use `--primary-6` in the future and this theme doesn't cover it? Please read [this](https://github.com/beekeeper-studio/beekeeper-studio/pull/4763#discussion_r4068895296) comment.

